PR #44, opened 2017-11-18
 private/model/api: Move service specific config to client

jasdel:
Moves service specific configuration fields from `aws.Config` to the specific clients.  Cleans up aws.Config removing non-global configuration flags.

Moves:
* `S3ForcePathStyle`, `S3Disable100Continue`, `S3UseAccelerate` to S3 service client.
* `DisableComputeChecksums` to SQS and DynamoDB service clients.

Fix #51


crystal-lee:
I find the UseAccelerate UsePathStyle in s3.Options. But I can not find param S3Disable100Continue int the s3 service ,where is it ?

PR #72, opened 2017-12-22
Improve SDK's service client API interface (iface)s

jasdel:
Updates the SDK to generate an interface per API call. This design idea allows the user of the SDK to selectively use build interfaces for the SDK's client that use a specific subset of the APIs. The goal of generating the API interfaces individually is also to prevent breaking code when the service adds a new API operation. In v1 the interface would "break" each time a new API was added.

* Is this a better approach than a single interface per service?
* How would this manifest in the s3manager Upload/Downloader, and Crypto client? Would using these subset of interfaces prevent the s3manager from adding new S3 functionality in the future?
   * Optional interface support is an option. Not enforced from code signatures.

### DIFF
--- a/service/acm/acmiface/interface.go
+++ b/service/acm/acmiface/interface.go
@@ -13,78 +13,52 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/acm"
 )
 
-// ACMAPI provides an interface to enable mocking the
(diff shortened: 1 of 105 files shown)


jasdel:
Closing this PR in favor of the V2 SDK' code generation strategy being done in the feature/SmithyAPIClient branch.

PR #76, opened 2017-12-26
private/protocol: adding generated marshalers

xibz:


jasdel:
Could you expand this PR description to describe the change being made and its impact?  Also which protocols are being impacted by this change. Similar idea to aws/aws-sdk-go#1409.  It would be helpful to include up-to-date performance metric impact as well.

PR #86, opened 2018-01-09
Update v2 SDKs example and documentation

### LINKED ISSUE #85: Old v1 documentation needs to be cleaned up and removed for v2.
There are several places where the v1 SDK's documentation is still exists in v2 repo. These need to be cleaned up and removed.

Related to the confusion caused in #83

jasdel:
Updates the SDK's examples and documentation to use the v2 configuration pattern.

Fix #85

### DIFF
--- a/service/s3/doc_custom.go
+++ b/service/s3/doc_custom.go
@@ -6,11 +6,11 @@
 // for optimizations if the Body satisfies that type. Once the Uploader instance
 // is created you can call Upload concurrently from multiple goroutines safely.
 //
-//   // The session the S3 Uploader will use
-//   sess := session.Must(session.NewSession())
+//   // The config the S3 Uploader will use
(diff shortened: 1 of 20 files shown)


service/s3/doc_custom.go
```diff
@@ -38,11 +38,11 @@
 // Once the Downloader instance is created you can call Upload concurrently from
 // multiple goroutines safely.
 //
-//   // The session the S3 Downloader will use
-//   sess := session.Must(session.NewSession())
+//   // The config the S3 Downloader will use
+//   cfg, err := external.LoadDefaultAWSConfig()
```
xibz: Not going to handle error?

jasdel: Yeah left it out for brevity.

PR #94, opened 2018-01-15
aws: Fix Pagination handling of empty string NextToken

### LINKED ISSUE #84: Inconsistent pagination
### Version of AWS SDK for Go?

What's the command to determine this?
Else, going to assume https://github.com/aws/aws-sdk-go-v2/commit/493e3ced184b77eff33393092e91fd3e1d0dbfd8

go version go1.9.2 linux/amd64

### What issue did you see?

With medialive the way to test an empty token is `*next == ""`. on mediapackage it's `next == nil`. I'm puzzled why.

* https://godoc.org/github.com/aws/aws-sdk-go-v2/service/medialive#ListChannelsOutput
* https://godoc.org/github.com/aws/aws-sdk-go-v2/service/mediapackage#ListOriginEndpointsOutput

### Steps to reproduce

Compare

* https://s.natalian.org/2018-01-09/live.go
* https://s.natalian.org/2018-01-09/pkgs.go

Notice if live.go uses:

        if next == nil {
            break
        }

It just goes in an infinite loop.

jasdel:
Fixes the SDK's handling of a Pagination NextToken's value being an
empty string compared to a nil value. The SDK was expecting NextToken's
to always be unset (nil) and treating any non-nil value as a valid
value. This was not the case in MediaLive's List APIs. As those APIs
return a empty string value instead of null or not setting the field at
all.

This issue exists in both the v1 and v2 SDKs.

Fix #84

### DIFF
--- a/aws/request_pagination.go
+++ b/aws/request_pagination.go
@@ -121,26 +121,18 @@ func (r *Request) nextPageTokens() []interface{} {
 		return nil
 	}
 	if r.Operation.TruncationToken != "" {
-		vs, _ := awsutil.ValuesAtPath(r.Data, r.Operation.TruncationToken)
-		if len(vs) == 0 {
+		tr, _ := awsutil.ValuesAtPath(r.Data, r.Operation.TruncationToken)
+		if len(tr) == 0 {
 			return nil
 		}
 
-		switch v := vs[0].(type) {
+		switch v := tr[0].(type) {
 		case *bool:
 			if !BoolValue(v) {
 				return nil
 			}
 		case bool:
 			if v == false {
-				return nil
-			}
-		case *string:
-			if len(StringValue(v)) == 0 {
-				return nil
-			}
-		case string:
-			if len(v) == 0 {
 				return nil
 			}
 		}


aws/request_pagination_test.go
```diff
@@ -586,6 +586,82 @@ func TestPagination_Standalone(t *testing.T) {
...
+		Value     *string
+		NextToken *string
+	}
+
+	expect := []struct {
+		Value, PrevToken, NextToken *string
+	}{
+		{aws.String("FirstValue"), aws.String("InitalToken"), aws.String("FirstToken")},
+		{aws.String("SecondValue"), aws.String("FirstToken"), aws.String("SecondToken")},
+		{aws.String("ThirdValue"), aws.String("SecondToken"), nil},
```
xibz: Do we not have a test to test for `""`?

aws/request_pagination.go
```diff
@@ -141,13 +141,29 @@ func (r *Request) nextPageTokens() []interface{} {
...
+				continue
+			}
+		case string:
+			if len(tv) == 0 {
+				tokens = append(tokens, nil)
+				continue
+			}
 		}
+
+		tokenAdded = true
```
xibz: if they are bunch of empy strings, wouldn't `tokenAdded` not be set to `true`? Meaning this would never return the `tokens`

jasdel: `tokenAdded` is present if a valid token is added. The `tokens` slice is a sparse array. Its index directly correspond to indexes in another slice.

If all of the values at the `outTokens` paths are empty strings there will be no valid tokens in the `tokens` slice. In that case returning `nil` would be the correct action.  If there is at least one valid value in `outTokens` paths the `tokenAdded` will be set to true.

aws/request_pagination_test.go
jasdel: Yeah, [TestPagination_Standalone](https://github.com/aws/aws-sdk-go-v2/pull/94/files#diff-c0f28140336767027cefe425e291979cR528) is that testcase. The SDK already had this testcase but the way it [worked](https://github.com/aws/aws-sdk-go-v2/pull/94/files#diff-c0f28140336767027cefe425e291979cR561) prevented the bug from manifesting.

PR #97, opened 2018-01-15
aws: Refactor pagination test based on v1 port

jasdel:
Refactors the SDK's request pagination test based on the work done for v1 aws/aws-sdk-go#1733

### DIFF
--- a/aws/request_pagination_test.go
+++ b/aws/request_pagination_test.go
@@ -559,7 +559,7 @@ func TestPagination_Standalone(t *testing.T) {
 					// Setup handlers for testing
 					r.Handlers.Clear()
 					r.Handlers.Build.PushBack(func(req *aws.Request) {
-						if e, a := len(c), i+1; a > e {
+						if e, a := len(c), i; a >= e {
 							t.Fatalf("expect no more than %d requests, got %d", e, a)
 						}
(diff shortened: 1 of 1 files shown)


aws/request_pagination_test.go
```diff
@@ -513,152 +513,93 @@ func TestPaginationWithContextNilInput(t *testing.T) {
...
+								InputTokens:  []string{"NextToken"},
+								OutputTokens: []string{"NextToken"},
+							},
+						},
+						&input, &testPageOutput{},
+					)
+					// Setup handlers for testing
+					r.Handlers.Clear()
+					r.Handlers.Build.PushBack(func(req *aws.Request) {
+						if e, a := len(c), i+1; a > e {
```
xibz: this check seems weird. Wouldn't e always be greater than a? Obviously missing something as this test passes.

The first test case e would be three. Calling the first `Next` would result in `i` being `0`. So would this check fail??

jasdel: This check is to identify if too many requests have been made and fail the build, because `i` now is greater than the number of expected tests.

This could be simplified as just `e, a := len(c), i; a == e`  Using equal instead of greater than would remove the need for the `+1`

PR #119, opened 2018-02-07
private/mode/api: refactoring paginators

xibz:
Pagination will be typed now and users will be able to grab a `aws.Pager` from the API requests.

```go
svc := s3.New(cfg)
req := svc.ListObjectsRequest(input)
p := req.Paginate()

for p.Next() {
    page := p.CurrentPage()
    fmt.Println("List Objects' contents", len(page.Contents))
}

if err := p.Err(); err != nil {
    fmt.Println("unexpected error occurred", err.Error())
}
```

### DIFF
--- a/service/s3/s3manager/batch.go
+++ b/service/s3/s3manager/batch.go
@@ -127,25 +127,16 @@ type BatchDeleteIterator interface {
 //	}
 type DeleteListIterator struct {
 	Bucket    *string
-	Paginator aws.Pager
+	Paginator s3.ListObjectsPager
 	objects   []s3.Object
 }
 
 // NewDeleteListIterator will return a new DeleteListIterator.
 func NewDeleteListIterator(svc s3iface.S3API, input *s3.ListObjectsInput, opts ...func(*DeleteListIterator)) BatchDeleteIterator {
+	req := svc.ListObjectsRequest(input)
 	iter := &DeleteListIterator{
-		Bucket: input.Bucket,
-		Paginator: aws.Pager{
-			NewRequest: func() (*aws.Request, error) {
-				var inCpy *s3.ListObjectsInput
-				if input != nil {
-					tmp := *input
-					inCpy = &tmp
-				}
-				req := svc.ListObjectsRequest(inCpy)
-				return req.Request, nil
-			},
-		},
+		Bucket:    input.Bucket,
+		Paginator: req.Paginate(),
 	}
 
 	for _, opt := range opts {
@@ -161,7 +152,7 @@ func (iter *DeleteListIterator) Next() bool {
 	}
 
 	if len(iter.objects) == 0 && iter.Paginator.Next() {
-		iter.objects = iter.Paginator.CurrentPage().(s3.ListObjectsOutput).Contents
+		iter.objects = iter.Paginator.CurrentPage().Contents
 	}
 
 	return len(iter.objects) > 0
--- a/service/s3/s3manager/batch_test.go
+++ b/service/s3/s3manager/batch_test.go
@@ -314,12 +314,16 @@ type mockS3Client struct {
 }
 
 func (client *mockS3Client) ListObjectsRequest(input *s3.ListObjectsInput) s3.ListObjectsRequest {
-	object := client.objects[client.index]
-	client.index++
 	req := client.S3.ListObjectsRequest(input)
-	req.Handlers.Send.PushBack(func(r *aws.Request) {
-		r.Data = object
-	})
+	req.Copy = func(v *s3.ListObjectsInput) s3.ListObjectsRequest {
+		r := client.S3.ListObjectsRequest(v)
+		r.Handlers.Send.PushBack(func(r *aws.Request) {
+			object := client.objects[client.index]
+			client.index++
+			r.Data = &object
+		})
+		return r
+	}
 
 	return req
 }
@@ -371,20 +375,11 @@ func TestBatchDeleteList(t *testing.T) {
 	input := &s3.ListObjectsInput{
 		Bucket: aws.String("bucket"),
 	}
-	iter := &DeleteListIterator{
-		Bucket: input.Bucket,
-		Paginator: aws.Pager{
-			NewRequest: func() (*aws.Request, error) {
-				var inCpy *s3.ListObjectsInput
-				if input != nil {
-					tmp := *input
-					inCpy = &tmp
-				}
 
-				req := svc.ListObjectsRequest(inCpy)
-				return req.Request, nil
-			},
-		},
+	req := svc.ListObjectsRequest(input)
+	iter := &DeleteListIterator{
+		Bucket:    input.Bucket,
+		Paginator: req.Paginate(),
 	}
 
 	if err := batcher.Delete(aws.BackgroundContext(), iter); err != nil {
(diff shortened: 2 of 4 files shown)


aws/request_pagination.go
```diff
@@ -42,24 +41,24 @@ type Pagination struct {
...
 }
 
-// HasNextPage will return true if Pagination is able to determine that the API
+// hasNextPage will return true if Pagination is able to determine that the API
 // operation has additional pages. False will be returned if there are no more
 // pages remaining.
 //
 // Will always return true if Next has not been called yet.
-func (p *Pagination) HasNextPage() bool {
+func (p *Pager) hasNextPage() bool {
```
jasdel: out of curiosity why not export this?

aws/request_pagination_test.go
```diff
@@ -5,182 +5,89 @@ import (
 	"testing"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
+	"github.com/aws/aws-sdk-go-v2/aws/defaults"
```
jasdel: so much red! :)

aws/request_pagination_test.go
```diff
@@ -5,182 +5,89 @@ import (
...
-		t.Errorf("expected nil value, but received %v", params.ExclusiveStartTableName)
+
+	if e, a := expected, results; !reflect.DeepEqual(e, a) {
+		t.Error("expected %v, but received %v", e, a)
 	}
 }
 
-// Use DynamoDB methods for simplicity
-func TestPaginationEachPage(t *testing.T) {
+/*// Use DynamoDB methods for simplicity
```
jasdel: large block of commented code here.

aws/request_pagination_test.go
```diff
@@ -675,4 +507,4 @@ func BenchmarkEachPageIterator(b *testing.B) {
 			return true
 		})
 	}
-}
+}*/
```
jasdel: can this block be uncommented, or tested directly?

private/model/api/operation.go
```diff
@@ -70,6 +70,7 @@ var tplOperation = template.Must(template.New("operation").Funcs(template.FuncMa
 	"GetCrosslinkURL": GetCrosslinkURL,
 }).Parse(`
 {{ $reqType := printf "%sRequest" .ExportedName -}}
+{{ $pagerType := printf "%sPager" .ExportedName -}}
```
jasdel: We want to make sure this `__Pager` type won't collide with modeled shapes.

private/model/api/operation.go
```diff
@@ -162,43 +163,35 @@ func (c *{{ .API.StructName }}) {{ $reqType }}(input {{ .InputRef.GoType }}) ({{
...
+	return {{ $pagerType }}{
+			aws.Pager {NewRequest: func() (*aws.Request, error) {
+				var inCpy {{ .InputRef.GoType }}
+				if p.Input != nil  {
+					tmp := *p.Input
+					inCpy = &tmp
+				}
+
+				var output {{ .OutputRef.GoTypeElem }}
+				req := aws.New(p.Request.Config, p.Request.Metadata, p.Request.Handlers.Copy(), p.Request.Retryer, p.Request.Operation, inCpy, &output)
```
jasdel: Not directly related to this PR, but we should rename the `aws.New` method to `aws.NewRequest`

jasdel: If there are in custom code generated into the API's __Request method this will bypass that code. e.g. Signing logic, ect.  

For example its possible for the __Request logic to have a closure referencing some instance value such as a output shape. This code probably should just use the base __Request instead of bypasing it.

private/protocol/ec2query/build_bench_test.go
```diff
@@ -1,85 +0,0 @@
-// +build bench
```
jasdel: Think this is a merge issue, these benchmark files have been removed.

private/protocol/restjson/build_bench_test.go
```diff
@@ -9,9 +9,7 @@ import (
 	"testing"
```
jasdel: Shouldn't these files of been updated in the generated marshaler change?

private/model/api/operation.go
xibz: Ah! Good point. Will fix.

private/model/api/operation.go
xibz: Okay, that makes sense.

private/protocol/restjson/build_bench_test.go
xibz: Yea, this is a rebase issue.

aws/request_pagination.go
xibz: With the new design, I couldn't think of a good reason to export it.

service/s3/s3manager/batch.go
```diff
@@ -162,7 +161,7 @@ func (iter *DeleteListIterator) Next() bool {
 	}
 
 	if len(iter.objects) == 0 && iter.Paginator.Next() {
-		iter.objects = iter.Paginator.Page().(s3.ListObjectsOutput).Contents
+		iter.objects = iter.Paginator.CurrentPage().(s3.ListObjectsOutput).Contents
```
jasdel: The s3/s3manager/DeleteBatch can be updated to use the Pager created by S3.ListObjectsRequest.Paginate() method. 

This would also get rid of the closure here. If there are other places that have the aws.Pager{NewRequest:...} like this they probably can be updated as well.

```go
func NewDeleteListIterator(...
    pager := svc.ListObjectsRequest(input).Paginate()
    iter := &DeleteListIterator{
        Bucket: input.Bucket,
        Paginator: pager.Pager,
    }

//...
}

File: example/service/s3/listObjects/listObjects.go [hunk 10]
```diff
@@ -31,19 +31,16 @@ func main() {
...
-	i := 0
-	err = svc.ListObjectsPages(&s3.ListObjectsInput{
-		Bucket: &os.Args[1],
-	}, func(p *s3.ListObjectsOutput, last bool) (shouldContinue bool) {
-		fmt.Println("Page,", i)
-		i++
-
-		for _, obj := range p.Contents {
-			fmt.Println("Object:", *obj.Key)
+	req := svc.ListObjectsRequest(&s3.ListObjectsInput{Bucket: &os.Args[1]})
```
jasdel: Noticed the `listObjectsConcurrently` example isn't using the paginator like it should be.  Only the first page is used in that example. If you don't mind would be good to update that one as well.

service/s3/s3manager/batch_test.go
```diff
@@ -439,8 +438,8 @@ func TestBatchDeleteList_EmptyListObjects(t *testing.T) {
 	// called on it
 	iter := &DeleteListIterator{
 		Bucket: input.Bucket,
-		Paginator: request.Pagination{
-			NewRequest: func() (*request.Request, error) {
+		Paginator: aws.Pager{
+			NewRequest: func() (*aws.Request, error) {
```
jasdel: These tests can be updated to use the ListObjectsRequest(input).Paginate() returned pager value.

PR #166, opened 2018-04-13
Fix XML unmarshaler not correctly unmarshaling list of timestamp values

### LINKED ISSUE #1892: GetBucketLocation fail if s3 bucket is locate in us-east-1
### Describe the bug

We use aws-sdk-go-v2 to list all buckets and then we will do so put action for some buckets.
There is  a bucket locate on us-east-1, and we found when we perform `s3.GetBucketLocation` for this bucket, if fail.
and if we use aws cli to list this bucket region, it success.
the reason we found is file `service/s3/types/enums.go` it does not support `us-east-1` region.

### Expected Behavior

we can perform `s3.GetBucketLocation` successfully and get bucket region back if bucket locate on us-east-1.

### Current Behavior

if bucket locate on us-east-1, perform `s3.GetBucketLocation` can not get the bucket region.

### Reproduction Steps

create bucket on `us-east-1`, then perform `s3.GetBucketLocation(ctx, &s3.GetBucketLocationInput{Bucket: aws.String(name)})`

### Possible Solution

_No response_

_No response_

v1.16.16

### Compiler and Version used

1.18.3

RedHat 8.5

jasdel:
Fixes a bug in the XML unmarshaler that would incorrectly try to
unmarshal "time.Time" parameters that did not have the struct tag type
on them. This would occur for nested lists like CloudWatch's
GetMetricDataResponse MetricDataResults timestamp parameters.

Adds protocol tests for XML unmarshaling with list of timestamps.

Related to aws/aws-sdk-go#1894

Fix #1892

### DIFF
--- a/models/protocol_tests/output/rest-xml.json
+++ b/models/protocol_tests/output/rest-xml.json
@@ -48,6 +48,9 @@
           },
           "Timestamps": {
             "shape": "Timestamps"
+          },
+          "Float64s": {
+            "shape": "Float64s"
           }
         }
       },
@@ -93,6 +96,10 @@
       "Timestamps":{
         "type":"list",
         "member":{"shape":"TimestampType"}
+      },
+      "Float64s": {
+        "type":"list",
+        "member":{"shape":"FloatType"}
       }
     },
     "cases": [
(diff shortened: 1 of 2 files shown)


private/protocol/xml/xmlutil/unmarshal.go
```diff
@@ -52,14 +54,26 @@ func parse(r reflect.Value, node *XMLNode, tag reflect.StructTag) error {
 	if t == "" {
 		switch rtype.Kind() {
 		case reflect.Struct:
-			t = "structure"
+			// also it can't be a time object
+			_, tok := r.Interface().(*time.Time)
```
xibz: May be cleaner to do something like this
```go
if _, ok := reflect.Indirect(v).Interface().(time.Time); ok {
}
```

jasdel: Wouldn't be able to use `Indirect` because that would produce a nil value for fields, which `Interface` will panic on.

jasdel:
Merge pending until we're able to regression test this with the other REST XML protocol services.

opavader:
Can this be expedited ?
Seems like major bug to me.

models/protocol_tests/output/rest-xml.json
```diff
@@ -42,6 +42,15 @@
...
             "shape": "TimestampType"
+          },
+          "Blobs": {
+            "shape": "Blobs"
+          },
+          "Timestamps": {
+            "shape": "Timestamps"
+          },
+          "Float64s": {
+            "shape": "Float64s"
```
xibz: Can we add `JSONValue` to this test?

Issue #74, opened 2017-12-22
Refactor SDK's error types


jasdel:
The v1 SDK exposes a single interface that most errors in the SDK try to conform to. `awserr.Error`. This error doesn't make sense for a lot of the errors the SDK core and utilities expose. Specifically the `code` and `message` values only make sense for API response errors.

The v2 SDK should refactor `awserr.Error` into an `APIError` interface type that basically is the `awserr.RequestFailure` type.  I think the generic `awserr.Error` can be replaced in favor with errors that support the `github.com/pkg/errors` package's `Wrap` and `Cause` pattern.

Refactoring these error types should also satisfy a goal of error checking in user applications should be based more on type assertions instead of string comparison.

rajender:
Looks like Go1.13 may have more error inspection capabilities. We should consider it part of this.

https://github.com/golang/go/issues/29934
https://godoc.org/golang.org/x/xerrors

jasdel:
Thanks for the suggestion @rajender. I agree taking advantage of `xerrors` package with 1.13 would help to standardize the SDK's error chaining. Right now this chaining is an odd mix of hand rolled chain, and `github.com/pkg/errors`.

This would mean the SDK's minimum supported version would raise go Go 1.13. Since the SDK is still in developer preview, this is may not be a deal breaker, but 1.13 most likely won't release until around August. The SDK may need to consider back filling support for `xerrors` for older version of Go if 1.13 ends up being too recent of a change when the SDK does GA release.

jasdel:
APIErrors returned from an API operation call should probably include the metadata for the operation request/response that failed. This context would help will logging, and identifying context of a failed operation call.

Does it make sense to expose this behavior to all SDK errors that are returned from an operation call?  One way this could happen is the operation's Send potentially could wrap the underlying error to ensure context is maintained.

jasdel:
The SDK's error handling refactor should be built with Unwrap in mind so that layering of errors can easily be peeled apart.

jasdel:
#487 made significant changes to the SDK's awserr.Error and internal error wrapping. The SDK now supports Unwrap in numerous places in the SDK.

There are still several instances of SDK logic errors that are using the awserr.Error Code wrapper. Ideally the SDK doesn't use awserr.Error for any non-API error response. All SDK logic errors are typed, or generic errors. The outstanding SDK logic awserr.Error usage are a good target for further burn down.

PR #167, opened 2018-04-14
service/s3: Add s3.RequestFailure.Extra()

matope:
Hi, folks.

For now, aws-sdk-go(-v2) users are not able to access all error information which S3 responded in the xml (For example, we can't access ArgumentName and ArgumentValue elements of InvlalidArgument xml). If we'd like to access those elements in the XML, we need to write some custom UnmarshalError handler.

ref: https://github.com/aws/aws-sdk-go/issues/542#issuecomment-180879375

But I think writing those codes in each apps are not so convenient. So, I'd like to add an method called `Extra()` to access extra (optional) error information to s3.RequestFailure.

This method returns an `map[string]string` which represents all the extra XML elements other than Code and Message (which are already accessible through Code() and Message()) in the S3 error xml. As far as I know, there are no S3 error xml that have special nested elements so I believe we can map them into simple `map[string]string` type generically.

With this method, users don't have to write custom XML unmarshallers specified for their apps.

Could you please review this PR?

Thanks.

### DIFF
--- a/service/s3/unmarshal_error.go
+++ b/service/s3/unmarshal_error.go
@@ -13,9 +13,24 @@ import (
 )
 
 type xmlErrorResponse struct {
-	XMLName xml.Name `xml:"Error"`
-	Code    string   `xml:"Code"`
-	Message string   `xml:"Message"`
+	XMLName xml.Name      `xml:"Error"`
+	Code    string        `xml:"Code"`
+	Message string        `xml:"Message"`
+	Extra   extraElements `xml:",any"`
+}
+
+type extraElements map[string]string
+
+func (e *extraElements) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
+	var data string
+	if err := d.DecodeElement(&data, &start); err != nil {
+		return err
+	}
+	if *e == nil {
+		*e = make(extraElements)
+	}
+	(*e)[start.Name.Local] = data
+	return nil
 }
 
 func unmarshalError(r *request.Request) {
@@ -69,6 +84,7 @@ func unmarshalError(r *request.Request) {
 			r.RequestID,
 		),
 		hostID: hostID,
+		extra:  resp.Extra,
 	}
 }
 
@@ -81,12 +97,17 @@ type RequestFailure interface {
 
 	// Host ID is the S3 Host ID needed for debug, and contacting support
 	HostID() string
+
+	// Extra returns a map which represents the S3 error XML elements
+	// other than Code and Message.
+	Extra() map[string]string
 }
 
 type requestFailure struct {
 	awserr.RequestFailure
 
 	hostID string
+	extra  map[string]string
 }
 
 func (r requestFailure) Error() string {
@@ -99,4 +120,10 @@ func (r requestFailure) String() string {
 }
 func (r requestFailure) HostID() string {
 	return r.hostID
+}
+func (r requestFailure) Extra() map[string]string {
+	if r.extra == nil {
+		return map[string]string{}
+	}
+	return r.extra
 }
--- a/service/s3/unmarshal_error_test.go
+++ b/service/s3/unmarshal_error_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -19,6 +20,7 @@ type testErrorCase struct {
 	RespFn           func() *http.Response
 	ReqID, HostID    string
 	Code, Msg        string
+	Extra            map[string]string
 	WithoutStatusMsg bool
 }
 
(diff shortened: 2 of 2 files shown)


jasdel:
Thanks for taking the time to create this PR @matope. With the V2 SDK we have the opportunity to take a much more radical and wide reaching improvement of how errors are exposed by the SDK, #74 captures this basic idea of how this could be done.  For the V2 SDK we'd like to use more strongly typed error messages and rely less on metadata and loosely typed attributes.

One step is to add generated unmarshaler to the SDK. These unmarshaler would allow the SDK to also have typed errors from the services that would provide all metadata the service exception is modeled with. There are a few services which add additional metadata on their error messages, that generated unmarshaling would help solve the pain of working with.

Do you have an example of some of the additional metadata fields that are being added to S3 exceptions responses?

matope:
Thank you for your replying, @jasdel.

Generating individual error types sounds great if we're interested in some specific error codes.
But in my case, as we just want to do something like below, strong error types sounds a little overkill for us.

```go
fmt.Errorf("Something wrong happened! Code: %s, Message: %s, AdditionalInfo: %+v", e.Code(), e.Message(), e.Extra() )
```

In this case, we're not interested in what the error was, but I'd like to output enough information for our users to handle it (We want more info than Error() or String() provides here).

If I understand your plan correctly, I think we need a ton of type switches on error handling logics with the strongly typed errors and it might not fit for our purpose.
 
```go
out, err := client.GetObjectRequest(...)
if err != nil {
  switch ee := err.(type) {
    case s3err.NoSuchBucket:
      return fmt.Errorf("bucket was not found. Bucket: %s", ee.Bucket)
    case s3err.NoSuchKey:
      return fmt.Errorf("bucket was found but key was not: Key: %s", ee.Key)
    case s3err.RequestTimeTooSkewed:
      return fmt.Errorf("your server time is wrong! ServerTime: %v, RequestTime:%v", ee.ServerTime, ee.RequestTime )
    // ...
    default:
      return fmt.Errorf("You got an error: %s", ee.Error() )
  }
}
```

Thank you.

t2y:
I have same issue getting extra error attributes. For example, I got a `NoSuchBucket` error and want to get `BucketName` attributes.

```xml
<Error>
  <Code>NoSuchBucket</Code>
  <Message>The specified bucket does not exist</Message>
  <BucketName>not-exist-bucket</BucketName>
  <RequestId>edfa12ee-aa3b-40fb-a147-4f55e5920cd0</RequestId>
  <HostId>myhostname</HostId>
</Error>
```

Also, I got a `NoSuchKey`, then want to get `Key` attributes.

```xml
<Error>
  <Code>NoSuchKey</Code>
  <Message>The specified key does not exist.</Message>
  <Key>not-exist-key</Key>
  <RequestId>92eac590-0afb-4d82-8cee-995bde73adce</RequestId>
  <HostId>myhostname</HostId>
</Error>
```

Another example, `SignatureDoesNotMatch` is like this.

```xml
<Error>
  <Code>SignatureDoesNotMatch</Code>
  <Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
  <AWSAccessKeyId>myAccessKeyId</AWSAccessKeyId>
  <StringToSign>GET Wed, 25 Apr 2018 11:11:46 +0000 ... </StringToSign>
  <SignatureProvided>...</SignatureProvided>
  <StringToSignBytes>...</StringToSignBytes>
  <RequestId>eb3600a0-372a-4a4c-b009-8087d46f6cde</RequestId>
  <HostId>myhostname</HostId>
</Error>
```

> But I think writing those codes in each apps are not so convenient. So, I'd like to add an method called Extra() to access extra (optional) error information to s3.RequestFailure.

I'm not sure that [models/apis/s3/2006-03-01/*.json](https://github.com/aws/aws-sdk-go-v2/tree/master/models/apis/s3/2006-03-01) can handle such attributes of the error. So, I think `Extra()` method is simple and reasonable.

jasdel:
Thanks for the feedback @matope. The idea of the typed errors would only need broken out switch statements if the individual fields of the typed error are meaningful to the application. If the fields are not meaningful the standard types satisfy the `error` interface and calling the `Error()` method will return a formatted string including all extra metadata that was modeled in the error message.

You'd still be able to use a generic `awserr.APIError` interface that would provide a `Code` and `Message` methods in addition to the formatted Error method.

jasdel:
In working on an item for the V1 SDK it reminded me about this outstanding PR. I think another approach the v2 SDK could take instead of parsing the `,any` fields is to provide a way for the user have access to the raw response message it self. Services like S3 can return (un-modeled) arbitrarily nested error messages that I think would be difficult to work with using the `,any` catch all.

I think the V2 SDK could have generated types for the errors with support for unmarshaling modeled fields. This is apart of our plan for the V2 SDK. The V2 SDK also needs to support the case of arbitrary unmodeled fields. the `,any` catch all is one way to do that for a single depth (nicely). An alternative would be for the SDK's APIError type should provide a method to get the raw error response message.

This does push the unmarshaling of the error message onto users if they wish to have access to the unmodeled error fields, but I think it also ensures that the value the SDK provides back to the user won't lose any information.

jasdel:
Thanks for reporting this issue and providing an implementation. The V2 SDK has shifted its code generation utilities to use one that will generate typed errors with the metadata that you're looking for in the `feature/SmithyAPIClient` branch. Our next major change to the V2 repo will be to merge that feature branch into master. We're currently working on the XML deserialization flow that will include serializing errors for APIs like S3, and other APIs that have modeled error values.

PR #168, opened 2018-04-15
Added STS endpoint configuration for eu-west-1.

GroovyCarrot:
Quick fix to allow calling STS in `endpoints.EuWest1RegionID`. I have an edge-case where this is necessary

### DIFF
--- a/aws/endpoints/defaults.go
+++ b/aws/endpoints/defaults.go
@@ -2182,7 +2182,12 @@ var awsPartition = partition{
 				"aws-global":     endpoint{},
 				"ca-central-1":   endpoint{},
 				"eu-central-1":   endpoint{},
-				"eu-west-1":      endpoint{},
+				"eu-west-1":      endpoint{
+					Hostname: "sts.eu-west-1.amazonaws.com",
+					CredentialScope: credentialScope{
+						Region: "eu-west-1",
+					},
+				},
 				"eu-west-2":      endpoint{},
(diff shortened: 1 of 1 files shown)


jasdel:
Thanks for posting this PR @GroovyCarrot Could you describe more about the issue your are experiencing with STS and the endpoint `eu-west-1` not being modeled within the SDK?

jasdel:
STS currently models its endpoints for the AWS SDKs to consume as only a single global endpoint for all regions. The AWS SDKs do not yet have a clear way to specify an alternative endpoint via a region identifier. This is a feature that the SDK teams are working with internally with the STS team to provide this functionality.

jasdel:
The `endpoints.json` is a model file that is used to generate the `aws/endpoints/defaults.go` file. Because of this, we won't be able to modify either of these files directly.

As a workaround you can manually specify the endpoint the SDK will use via the [EndpointResolver](https://godoc.org/github.com/aws/aws-sdk-go-v2/aws#Config.EndpointResolver) and [ResolveWithEndpointURL](https://godoc.org/github.com/aws/aws-sdk-go-v2/aws#ResolveWithEndpointURL)

```go
cfg.Region = "eu-west-1"
cfg.EndpointResolver = aws.ResolveWithEndpointURL("https://sts.eu-west-1.amazonaws.com")

svc := sts.New(cfg)
```

PR #169, opened 2018-04-20
handle special case of 0 value float64 in parseXML

opavader:

`parseScalar` `in private/protocol/xml/xmlutil/unmarshal.go` is unable to parse `0` as `*float64` value

To reproduce call `GetMetricData` from `service/cloudwatch` package where returned metric values contain 0

### DIFF
--- a/private/protocol/xml/xmlutil/unmarshal.go
+++ b/private/protocol/xml/xmlutil/unmarshal.go
@@ -251,6 +251,12 @@ func parseScalar(r reflect.Value, node *XMLNode, tag reflect.StructTag) error {
 			return err
 		}
 		r.Set(reflect.ValueOf(&v))
+	case float64:
+		v, err := strconv.ParseFloat(node.Text, 64)
+		if err != nil {
+			return err
+		}
+		r.Set(reflect.ValueOf(v))
 	case *time.Time:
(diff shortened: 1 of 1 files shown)


jasdel:
Thanks for posting this PR @opavader. I think the issue you're seeing is related to #166. This PR found an issue with the SDK's handling of [value vs pointer items in lists/maps](https://github.com/aws/aws-sdk-go-v2/pull/166/files#diff-57397599b3bd870740e1955ae082bcdcR238). I think we should add a protocol test similar to what was added for #166 that highlight this issue with float64.

opavader:
@jasdel can you please merge #166 on a emergency basis.
having it fixed it in my own fork is a lot of effort as the xml-unmarshler is being used almost everywhere
(so a lot of import path to be changed, times when you miss golang's lack of relative imports)

jasdel:
Thanks for creating this PR @opavader I'm going to be merging in #166 shortly which includes a fix for the issue you identified. Pr #166 also includes tests for this use case ensuring it works with the SDK.

jasdel:
@opavader #166 has been merged into master, and is now available. Thanks for your assistance and discovery of the issue with regular lists of scalars not working. These should all be working now. Let us know if you run into an additional issues, or have feedback.

PR #173, opened 2018-05-11
Fix typos for migrated S3 specific config options

t2y:
refer to #44


jasdel:
Thanks for taking the time to create this PR @t2y the changes look good.

PR #181, opened 2018-05-25
aws/endpoints: Fix SDK endpoint signing name resolution

### LINKED ISSUE #153: Ensure endpoint resolution falls back to C2J modeled SigningName
The SDK needs to ensure that the C2J modeled SigningName is used if the modeled endpoint does not contain the SigningName field.

Related to aws/aws-sdk-go#1854

### LINKED ISSUE #163: InvalidSignatureException: Credential should be scoped to correct service: 'application-autoscaling'.
Please fill out the sections below to help us address your issue.

2.0.0-preview.3

go version go1.10 darwin/amd64

### What issue did you see?

InvalidSignatureException: Credential should be scoped to correct service: 'application-autoscaling'. 

### Steps to reproduce

If you have have an runnable example, please include it.

```go
	if err != nil {
		panic(err)
	}
	cfg.Region = endpoints.UsWest2RegionID

	scale := applicationautoscaling.New(cfg)
	req := scale.RegisterScalableTargetRequest(&applicationautoscaling.RegisterScalableTargetInput{
		MaxCapacity:       aws.Int64(200),
		MinCapacity:       aws.Int64(90),
		ResourceId:        aws.String("service/foo/bar"),
		ScalableDimension: applicationautoscaling.ScalableDimensionEcsServiceDesiredCount,
		ServiceNamespace:  applicationautoscaling.ServiceNamespaceEcs,
	})
	_, err = req.Send()
	if err != nil {
		panic(err)
	}
```

jasdel:
Fixes how the SDK derives service signing names. If the signing name is
not modeled in the endpoints package the service will fallback to the
signing name modeled in the service model.

Fix #163
Fix #153
V2 Port of aws/aws-sdk-go#1854

### DIFF
--- a/internal/awstesting/integration/smoke/autoscalingplans/autoscalingplans.feature
+++ b/internal/awstesting/integration/smoke/autoscalingplans/autoscalingplans.feature
@@ -1,0 +1,7 @@
+# language: en
+@autoscalingplans @client
+Feature: AWS Auto Scaling Plans
+
+  Scenario: Making a request
+    When I call the "DescribeScalingPlans" API
+    Then the request should be successful
--- a/internal/awstesting/integration/smoke/autoscalingplans/client.go
+++ b/internal/awstesting/integration/smoke/autoscalingplans/client.go
@@ -1,0 +1,17 @@
+// +build integration
+
+//Package autoscalingplans provides gucumber integration tests support.
+package autoscalingplans
+
+import (
+	"github.com/aws/aws-sdk-go-v2/internal/awstesting/integration"
+	_ "github.com/aws/aws-sdk-go-v2/internal/awstesting/integration/smoke"
+	"github.com/aws/aws-sdk-go-v2/service/autoscalingplans"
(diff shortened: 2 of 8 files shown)


aws/request.go
```diff
@@ -103,7 +103,7 @@ func New(cfg Config, metadata Metadata, handlers Handlers,
 	if err == nil {
 		// TODO so ugly
 		metadata.Endpoint = endpoint.URL
-		if len(endpoint.SigningName) > 0 {
+		if len(endpoint.SigningName) > 0 && !endpoint.SigningNameDerived {
```
xibz: Can we just have the endpoint have a get signing name method that returns the correct signing name? I think that'll at least make it cleaner when dealing with cases where it isn't derived.

jasdel: because the signing name are modeled in both endpoint and service models, the SDK has to rely on both sources. In some cases the signing name is not included in the endpoint model,which is the default source.

PR #188, opened 2018-06-07
aws/signer/v4: Fix X-Amz-Content-Sha256 being in to query for presign

jasdel:
Fixes the bug which would allow the X-Amz-Content-Sha256 header to be
promoted to the query string when presigning a S3 request.

This bug also was preventing users from setting their own sha256 value
for a presigned URL. Presigned requests generated with the custom sha256
would of always failed with invalid signature.

S3 presign requests without a user specified X-Amz-Content-Sha256 will
no longer include the X-Amz-Content-Sha256 in the header nor query. The
X-Amz-Content-Sha256 header only needs to be set if it contains a non
UNSIGNED-PAYLOAD value.

Related to aws/aws-sdk-go#1974

### DIFF
--- a/aws/signer/v4/v4_test.go
+++ b/aws/signer/v4/v4_test.go
@@ -197,13 +197,33 @@ func TestSignBodyGlacier(t *testing.T) {
 	}
 }
 
-func TestPresignEmptyBodyS3(t *testing.T) {
-	req, body := buildRequest("s3", "us-east-1", "hello")
+func TestPresign_SignedPayload(t *testing.T) {
+	req, body := buildRequest("glacier", "us-east-1", "hello")
 	signer := buildSigner()
-	signer.Presign(req, body, "s3", "us-east-1", 5*time.Minute, time.Now())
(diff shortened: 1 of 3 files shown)


aws/signer/v4/v4_test.go
```diff
@@ -201,9 +201,8 @@ func TestPresignEmptyBodyS3(t *testing.T) {
 	req, body := buildRequest("s3", "us-east-1", "hello")
 	signer := buildSigner()
 	signer.Presign(req, body, "s3", "us-east-1", 5*time.Minute, time.Now())
-	hash := req.Header.Get("X-Amz-Content-Sha256")
-	if e, a := "UNSIGNED-PAYLOAD", hash; e != a {
-		t.Errorf("expect %v, got %v", e, a)
+	if a := req.Header.Get("X-Amz-Content-Sha256"); len(a) != 0 {
```
xibz: Do we have a test where the service isn't `"s3"`? To ensure it is included during presigning

jasdel: Added additional tests to cover these cases.

PR #189, opened 2018-06-08
Update models synced with v1 sdk

jasdel:
Update SDK's service models to be in sync with the v1 SDK.


jaredgisin:
Looks fine to me.

PR #191, opened 2018-06-10
Update readme to use "go get -u" instead of "go get"

artemnikitin:
It's better to use `go get -u` instead of `go get` for libraries, because `go get` will not update library if it's already existed in GOPATH.
Same is specified for v1 https://github.com/aws/aws-sdk-go


jasdel:
Thanks for submitting the update @artemnikitin. The change to the `go get` style looks good.

PR #192, opened 2018-06-11
Fix a broken README link and minor wording issue

euank:
The link to GitHub Issues used both '[GitHub Issues]' and 'GitHub
[Issues]' in different places, but only had links for one.

I think linking just 'Issues' is congruous with how 'GitHub Projects' is
linked, so I standardized them to that. I also fixed a sentence
fragment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,9 @@ func main() {
 
 ## Feedback and contributing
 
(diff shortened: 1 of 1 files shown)


jasdel:
Thanks for fixing the README links and wording @euank.

PR #211, opened 2018-09-03
fix: hashbang typo in cleanup_models.sh

rixtox:


jasdel:
Thanks for fixing this script's hashbang @rixtox.

PR #239, opened 2018-11-26
fix sqs batch send panic on failure

ccbrown:
Previously, if a message failed to send, `verifySendMessageBatch` would panic trying to access `*e.MessageId` where `e` is zero-initialized because it didn't exist in `out.Successful`. This fixes that panic and adds a unit test.


jasdel:
Thanks for creating this PR @ccbrown, and providing a unit test for the change.

jasdel:
Thanks for creating this PR @ccbrown, and providing a unit test for the change.

PR #265, opened 2019-01-25
Update SDK API operation request send to required Context

### LINKED ISSUE #264: Update client operation request send method to require Context
The v2 SDK should more clearly encourage the best practices of cancellation and tracing when using request operations. The SDK currently provides a optional way to specify Context, but this optional way is hidden and easily missed. The SDK should be updated so all operation request `Send` methods take a Context as their first parameter.

Standardizing on this pattern will also help reduce code paths which accidentally do not have the Context causing the cancellation and tracing chain to be lost. Leading to difficult to trace down losses of cancellation and tracing within an application.

```go
func (r *AnAPIOpRequest) Send(ctx context.Context) {...}
```

The SDK should consider if it should still provide a way to call `Send`, and `Pagination` without a `Context`. There are usecases where Context isn't very important, or desired, such as simple script programs. The `context` package does provide a `context.Background()` function for this purpose, but is that sufficient, or should the SDK provide a `SendWithoutContext` like method? 

* [ ] extra context with Pagination?
* [ ] Waiter Context?
* [x] should have SendWithoutContext?

jasdel:
Updates the SDK's API operation request Send method to require a
`context.Context` value when called. This is done to encourage best
applications to use Context for cancellation and request tracing.
Standardizing on this pattern will also help reduce code paths which
accidentally do not have the Context causing the cancellation and
tracing chain to be lost. Leading to difficult to trace down losses of
cancellation and tracing within an application.

If a context is not desired using `context.Background`, or
`context.TODO` instead of specific context, like cancel or trace.

Fix #264

**TODO**:
* [x] Documentation
* [x] Examples
* [x] Pagination with context
* [x] Waiters with context

### DIFF
--- a/example/service/s3/listObjects/listObjects.go
+++ b/example/service/s3/listObjects/listObjects.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 
@@ -33,7 +34,7 @@ func main() {
 
 	req := svc.ListObjectsRequest(&s3.ListObjectsInput{Bucket: &os.Args[1]})
 	p := req.Paginate()
-	for p.Next() {
+	for p.Next(context.TODO()) {
 		page := p.CurrentPage()
 		for _, obj := range page.Contents {
 			fmt.Println("Object: ", *obj.Key)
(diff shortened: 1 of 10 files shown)


cracraafiam:
How do I access my data

Issue #282, opened 2019-03-29
aws: Should "Paginate" method of API operation request be renamed?


jasdel:
All API operations that are paginatable have the "Paginate" method on the operation's request type. This name may imply that the method performs the pagination, when pagination doesn't begin until the `Next` method of the returned pager value is called.

Consider renaming `Paginate` to something that does not imply or suggest pagination request will start then, (e.g. `GetPager`, `GetPaginator`, `BuildPaginator`, `Paginator`, etc)

Issue #275, opened 2019-03-04
codegen: Remove unused shapes before processing model


jasdel:
The v2 SDK should trim unused shapes prior to processing the model, ot prevent generated name classing with unused shapes.

V1 Issue: aws/aws-sdk-go#2358

jasdel:
Fixed in #309

PR #298, opened 2019-04-30
Refactoring SDK's code generation patterns

### LINKED ISSUE #294: Add ServiceID client metadata to V2 SDK
The V2 SDK should generate its client package and struct identifiers from the service's modeled ServiceID not other metadata names.

jasdel:
Updates the SDK's code generation pattern with several improvements to improve discoverability, and consistency in client naming.

* Standardizing of the service model's ServiceID for the name of the
service client's package.

* Adds an operation `Response` type that is in addition from the
operation's `Output` parameters. This prevents collisions between SDK
response metadata, marshaling methods, and modeled output parameters.

* Refactor service client package's client name to be named Client
instead of the service's short name. Removes the stuttering of client
package and type, (e.g. s3.S3 becomes s3.Client)

* Fix service endpoint lookups use the the service's modeled EndpointID
instead of inconsistent service name.

* Generate API operations into their own file instead of all operations
and parameter types being generated into a single file. This improves
readability of individual source files for documentation reference.
Non-input/output types are still generated into a single file. This change also fixes several occurrences of incorrectly generated API operation input/output type names.

* Removes aws/endpoints Service identifiers from endpoint list. These
were not actually service identifiers, but EndpointsID. In addition the
values modeled by the services were unstable, and could change causing
breaking changes in the SDK's aws/endpoints package. The service
client package's `EndpointsID` const should be used instead.

* Move Paginate method from the initialized Request type into its own
construct via the New<RequestType>Paginator function to create the
request paginator. (e.g. req.Paginate becomes NewListObjectsPaginator(req))

Fix #294 #282 #275


jasdel:
Notes:
* `api_parameter_types.go` should probably renamed to just `api_types.go`
* Paginate's `CurrentPage` should return the `Response` type instead of the `Output` type. This allows access to the request metadata.
* Verify Paginator example `Next` takes context. And add package names to example.

stealthycoin:
Looks good.

Issue #113, opened 2018-01-31
service/s3/s3manager: Add support for resuming multipart upload between applications sessions.


jocot:
It seems that there is no facility to resume multipart uploads via the aws-sdk-go-v2 api once a session has ended or been disconnected.

jasdel:
Hi @jpeddicord Thanks for reaching out to use. The SDK's S3 Upload manager does not currently support resuming multipart uploads between application sessions. We did not have any direct plans to support this, but I can add this as a feature request so it can be considered added the the SDK's backlog.

As a workaround If you're application uses the S3 APIs directly it will have the capabilities to perform multipart upload resuming between application sessions.

jviney:
I have implemented a rough version of this feature if anyone is interested.

https://github.com/jviney/aws-sdk-go-v2/commit/925a73cc85d50382a2ddf456b1a3e762ab185797

microyahoo:
We also need it, do you have plans to support it? @jasdel

vbmithr:
I need this too.

nickajacks1:
This would be relatively straightforward to implement on your own if `Upload` somehow exposed the UploadId returned from `CreateMultipartUpload` before the actual upload starts.  Maybe add a callback?
A workaround is to use a middleware, e.g.:
```go
	var uploadId string
	mw := middleware.InitializeMiddlewareFunc("extractUploadID", func(ctx context.Context, ii middleware.InitializeInput, ih middleware.InitializeHandler) (middleware.InitializeOutput, middleware.Metadata, error) {
		switch v := ii.Parameters.(type) {
		case *s3.UploadPartInput:
			if uploadId == "" {
				uploadId = *v.UploadId
			}
		}
		return ih.HandleInitialize(ctx, ii)
	})
```

PR #300, opened 2019-05-03
Add support for resuming S3 uploads.

jviney:
Issue #113.


jasdel:
Thanks for taking the time to create this PR @jviney. It looks like the CI test failed because.

```
service/s3/s3manager/upload.go:162:2: struct field UploadIdCallback should be UploadIDCallback 
Makefile:148: recipe for target 'lint' failed 
```

To assist will testing you can run `make unit` or `make vet lint` locally to verify changes. I'll look into if there is a way to make the CodeBuild results public.

jviney:
Great - I've updated the patch with the linting fixes, and extended the tests to properly cover the UploadIDCallback behaviour.

service/s3/s3manager/upload.go
```diff
@@ -504,20 +520,67 @@ func (a completedParts) Len() int           { return len(a) }
...
+		req.ApplyOptions(u.cfg.RequestOptions...)
+		resp, err := req.Send(u.ctx)
+		if err != nil {
+			return nil, err
+		}
+		u.uploadID = *resp.UploadId
+	}
+
+	if u.cfg.UploadIDCallback != nil {
+		u.cfg.UploadIDCallback(u.uploadID)
```
jasdel: What is the use case that the UploadIDCallback would address?

service/s3/s3manager/upload.go
```diff
@@ -638,6 +706,43 @@ func (u *multiuploader) send(c chunk) error {
 	return nil
 }
 
+// isChunkUploaded is called before uploading each chunk to determine
+// if this chunk can be skipped. Chunks can be skipped when resuming an upload
+// if the ETag matches for the previously uploaded chunk.
+func (u *multiuploader) isChunkUploaded(c chunk) bool {
```
jasdel: I'm not sure we can use the `ETag` to check the upload part is the same as the local content. Specifically because the `ETag` is not required to be the MD5 of the part. By convention this the `ETag` probably is the MD5, but there is not guarantee that this is so. We'd probably not want to rely on unofficial behavior of the ETag.

service/s3/s3manager/upload.go
```diff
@@ -504,20 +520,67 @@ func (a completedParts) Len() int           { return len(a) }
...
-	req := u.cfg.S3.CreateMultipartUploadRequest(params)
-	req.ApplyOptions(u.cfg.RequestOptions...)
-	resp, err := req.Send(u.ctx)
-	if err != nil {
-		return nil, err
+	// Fetch the completed parts when resuming an upload.
+	// The ResumeUploadID is only used if the completed parts could be fetched.
+	if len(u.uploader.cfg.ResumeUploadID) > 0 {
+		if err := u.requestCompletedParts(); err == nil {
+			u.uploadID = u.uploader.cfg.ResumeUploadID
```
jasdel: If the user provided a UploadID to resume, and that ID either doesn't exist, or an error is encountered while getting the parts, the Uploader should return an error instead of creating a new upload. This is to prevent unexpected behavior, because the upload id didn't exist already.

service/s3/s3manager/upload.go
```diff
@@ -504,20 +520,67 @@ func (a completedParts) Len() int           { return len(a) }
...
+			partNumberMarker = output.NextPartNumberMarker
+		}
+	}
+
+	return nil
+}
+
 // upload will perform a multipart upload using the firstBuf buffer containing
 // the first chunk of data.
 func (u *multiuploader) upload(firstBuf io.ReadSeeker) (*UploadOutput, error) {
```
jasdel: When a `ResumeUploadID` is provided, the uploader probably should not attempt to upload a single part object, i.e. PutObject. A `ResumeUploadID` implies that a multi part upload is being performed, and it may be unexpected to upload a single part object instead.

service/s3/s3manager/upload.go
```diff
@@ -638,6 +706,43 @@ func (u *multiuploader) send(c chunk) error {
...
+// if the ETag matches for the previously uploaded chunk.
+func (u *multiuploader) isChunkUploaded(c chunk) bool {
+	if !u.resumed {
+		return false
+	}
+
+	var expectedETag string
+
+	u.m.Lock()
+	for _, part := range u.parts {
```
jasdel: Is the lock needed here? If `u.parts` is only written to once before the uploads begin it should be safe to iterate through the parts concurrently. Since the parts list may be a sparse slice, a map may be a better data structure to use for these. Where the part number is the map key.

service/s3/s3manager/upload.go
```diff
@@ -504,20 +520,67 @@ func (a completedParts) Len() int           { return len(a) }
...
+			UploadId:         &u.uploader.cfg.ResumeUploadID,
+			PartNumberMarker: partNumberMarker,
+		})
+
+		output, err := req.Send(u.ctx)
+		if err != nil {
+			return err
+		}
+
+		for _, part := range output.Parts {
```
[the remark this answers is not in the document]
jviney: Replaced with the paginator.

service/s3/s3manager/upload.go
```diff
@@ -177,6 +190,7 @@ func NewUploader(cfg aws.Config, options ...func(*Uploader)) *Uploader {
 		Concurrency:       DefaultUploadConcurrency,
 		LeavePartsOnError: false,
 		MaxUploadParts:    MaxUploadParts,
+		ResumeUploadID:    "",
```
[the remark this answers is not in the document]
jviney: Yep.

service/s3/s3manager/upload.go
```diff
@@ -504,20 +520,67 @@ func (a completedParts) Len() int           { return len(a) }
...
+			PartNumberMarker: partNumberMarker,
+		})
+
+		output, err := req.Send(u.ctx)
+		if err != nil {
+			return err
+		}
+
+		for _, part := range output.Parts {
+			completed := s3.CompletedPart{ETag: part.ETag, PartNumber: part.PartNumber}
```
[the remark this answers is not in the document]
jviney: Added a local `completedPart` type since `s3.CompletedPart` does not contain the part size. Is it actually necessary to validate the part sizes?

When resuming, the part size is now taken from the first previously uploaded part.

service/s3/s3manager/upload.go
jviney: This was to be able to store the UploadID for resuming. I've removed it for now, as our use case can be handled with the MultiUploadFailure interface.

service/s3/s3manager/upload.go
jviney: I can't find good AWS documentation defining the `ETag` values for multipart uploads. Using the MD5 seems pretty standard though. Is there a better option than using the MD5?

service/s3/s3manager/upload.go
jviney: ResumeUpload() will return an error now in this situation.

service/s3/s3manager/upload.go
jviney: Agreed.

service/s3/s3manager/upload.go
jviney: `u.parts` is written and read concurrently by `multiuploader.send()`, so I think the lock is necessary.

jviney:
Updated with `ResumeUpload` and `ResumeUploadWithContext` methods.

harshavardhana:
You should read why it may not be safe to use ListParts to resume uploads https://github.com/aws/aws-sdk-go/issues/1518

harshavardhana:
@jasdel is your comment here changed in any way 

https://github.com/aws/aws-sdk-go/issues/1518#issuecomment-329014393

> I would not recommend using `ListParts` when performing a `CompleteMultipartUpload`. While today it may work it is documented that the part numbers returned should not be used for completing a multipart.
> 
> The recommended way is for the application to maintain the list of parts that were uploaded and use that list for completing the multi-part upload.

rkrymer85:
What is the status of this change? I need the Resume Upload feature in my current project, and it looks like this PR is close completion, once the merge conflict is resolved. When can I expect this to be included in a release?

jviney:
The comments above seem to indicate that this PR might not be accepted because it uses `ListParts` to determine the previously uploaded parts. @jasdel can you confirm this either way?

If it won't be merged, we might as well close this PR.

I'm unlikely to provide another PR in the short term that implements a different approach, as this patch is currently sufficient for my needs - despite its potential shortcomings.

rkrymer85:
If that is the case, I will likely reference your patch to come up with a solution that meets my own needs. Thanks for the work anyway, @jviney !

PR #325, opened 2019-06-22
Correction in the use exemple in the README.md and doc.go files

pedropinheiro75:
* Add the `context` in the send method in the README.md and doc.go files.

The `req.Send` method used in the initial example of SDK needs to pass the `context` parameter or the following error is displayed:

```shell
not enough arguments in call to req.Send
        have ()
        want (context.Context)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This example shows how you can use the v2 SDK to make an API request using the S
 package main
 
 import (
+	"context"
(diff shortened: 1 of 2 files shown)


jasdel:
Thanks for fixing this typo in the SDK's README example @pedropinheiro75 . We'll get this change merged in.

PR #326, opened 2019-06-23
private/protocol: Fix benchmarks

Yimu-Yang:
Fix the existing benchmarks in private/protocol/restjson, private/protocol/restxml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This example shows how you can use the v2 SDK to make an API request using the S
 package main
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -7,3 +7,5 @@ Gemfile.lock
 /internal/awstesting/integration/smoke/_test/
 /vendor
 /private/model/cli/gen-api/gen-api
+.idea/
+*.DS_Store
```
jasdel: Would prefer not to include these in the SDK's `.gitignore` file you. For Git you can configure your local environment to ignore these platform/IDE files via git's global exclude configuration options.

https://help.github.com/en/articles/ignoring-files

Yimu-Yang:
.gitignore has been updated. Please check.

PR #350, opened 2019-08-07
aws: Add example for custom HTTP client idle connection options

jasdel:
Adds example to the SDK for configuring custom HTTP client idle
connection keep alive options.

### DIFF
--- a/aws/http_client_test.go
+++ b/aws/http_client_test.go
@@ -1,11 +1,9 @@
 package aws_test
 
 import (
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"testing"
-	"time"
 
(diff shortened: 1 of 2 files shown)


aws/http_client_test.go
```diff
@@ -29,33 +27,3 @@ func TestBuildableHTTPClient_NoFollowRedirect(t *testing.T) {
 		t.Errorf("expect %v code, got %v", e, a)
```
skotambkar: Should this be fatal too? Since it would be a network error.

jasdel: I think in this case since the assertion failing wouldn't prevent further assertions being performed the `Errorf` is good. `Fatalf` is great further assertions may not be testable.

PR #352, opened 2019-08-08
private/model/api: Create rest/xml/payload generated unmarshaler templates

Yimu-Yang:
Add shape_unmarshal.go inside private/model/api, which contains three templates for xml/rest/payload generated unmarshaler, respectively.

The templates are already wired into the generator. Some top level templates have also been modified.

### DIFF
--- a/private/protocol/restxml/unmarshal_test.go
+++ b/private/protocol/restxml/unmarshal_test.go
@@ -378,7 +378,7 @@ func (s *OutputService1TestShapeOutputService1TestCaseOperation1Output) unmarsha
 					return fmt.Errorf("fail to UnmarshalAWSXML OutputService1TestShapeOutputService1TestCaseOperation1Output.%s, %s", name, err)
 				}
 				v, _ := tok.(xml.CharData)
-				value, _ := time.Parse(protocol.ISO8601TimeFormat, string(v))
+				value, _ := protocol.ParseTime(protocol.ISO8601TimeFormatName, string(v))
 				s.Timestamp = &value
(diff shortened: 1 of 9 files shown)


service/cloudfront/api_op_GetDistributionConfig.go
```diff
@@ -86,6 +91,51 @@ func (s GetDistributionConfigOutput) MarshalFields(e protocol.FieldEncoder) erro
...
+			if s.DistributionConfig == nil {
+				s.DistributionConfig = &DistributionConfig{}
+			}
+			err = s.DistributionConfig.unmarshalAWSXML(d, start)
+			if err != nil {
+				return fmt.Errorf("fail to UnmarshalAWSXML GetDistributionConfigOutput, %s", err)
+			}
+			return nil
+		}
+	}
```
jasdel:
```go
	for {
		tok, err := d.Token()
		if tok == nil || err != nil {
			if err == io.EOF {
				return nil
			}
			return fmt.Errorf("fail to UnmarshalAWSXML GetDistributionConfigOutput, %s", err)
		}
		start, ok := tok.(xml.StartElement)
		if !ok {
		        continue
		}

		if s.DistributionConfig == nil {
			s.DistributionConfig = &DistributionConfig{}
		}
		err = s.DistributionConfig.unmarshalAWSXML(d, start)
		if err != nil {
			return fmt.Errorf("fail to UnmarshalAWSXML GetDistributionConfigOutput, %s", err)
		}
		return nil
	}
```

PR #359, opened 2019-08-19
private/model/api: Fix API doc being generated with wrong value

jasdel:
Fixes the SDK's generated API documentation for structure member being generated with the wrong documentation value when the member was included multiple times in the model doc-2.json file, but under different types.

The SDK was not considering the member's type when associating documentation with the member. Only parent struct name and member names was being used.

V2 Port of aws/aws-sdk-go#2748


CHANGELOG_PENDING.md
```diff
@@ -3,6 +3,5 @@
 ### SDK Enhancements
 
 ### SDK Bugs
-* `private/model`: Handles empty map vs unset map behavior in send request ([#337](https://github.com/aws/aws-sdk-go-v2/pull/337))
-  * Updates shape marshal model to handle the empty map vs nil map behavior. Also adds a test case to assert behavior when a user sends an empty map vs unset map.
-  * Fixes [#332](https://github.com/aws/aws-sdk-go-v2/issues/332)
+* `private/model/api`: Fix API doc being generated with wrong value ([#359](https://github.com/aws/aws-sdk-go-v2/pull/359))
```
skotambkar: Should we add reference to the v1 git issue or create and reference v2 git issue?

PR #375, opened 2019-08-29
aws : Adds configurations to the default retryer

### LINKED ISSUE #370: service/ec2: Fix int overflow in minTime on 386 and arm
Port V1 SDK's aws/aws-sdk-go#2787 for aws/aws-sdk-go#2786 to the v2 SDK fixing a potential int overflow when determining retry delay.

skotambkar:
Provides more retryer customization options by adding a constructor for default Retryer which accepts functional options. Adds NoOpRetryer to support no retry behavior. Exposes members of default retryer. 

Updates the underlying logic used by the default retryer to calculate jittered delay for retry. Also updates the custom retry logic for service/ec2. Handles int overflow for retry delay.

Includes changes for PR [#373 ](https://github.com/aws/aws-sdk-go-v2/pull/373). The changes will be squashed once the PR #373 is merged in.  

Fixes #370

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -14,8 +14,6 @@
   * Provides more customization options for retryer by adding a constructor for default Retryer which accepts functional options. Adds NoOpRetryer to support no retry behavior. Exposes members of default retryer.
   * Updates the underlying logic used by the default retryer to calculate jittered delay for retry. Handles int overflow for retry delay. 
   * Fixes [#370](https://github.com/aws/aws-sdk-go-v2/issues/370)
-* `service/ec2`: Adds custom retryer implementation for service/ec2.
-  * Adds test case to test custom retryer. 
   
 ### SDK Bugs
 * `aws`: Fixes bug in calculating throttled retry delay ([#373](https://github.com/aws/aws-sdk-go-v2/pull/373))


aws/client.go
```diff
@@ -64,7 +64,7 @@ func NewClient(cfg Config, metadata Metadata) *Client {
 	retryer := cfg.Retryer
 	if retryer == nil {
 		// TODO need better way of specifing default num retries
-		retryer = DefaultRetryer{NumMaxRetries: 3}
+		retryer = DefaultRetryer{3, 0}
```
jasdel: This should use named struct field initialization. e.g.

```
retryer = DefaultRetryer{
    NumMaxRetries: 3,
}
```

also don't need to initialize the `MinTime` member since it `0` is its default [zero value](https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful).

aws/client.go
```diff
@@ -76,6 +76,7 @@ func NewClient(cfg Config, metadata Metadata) *Client {
 // NewRequest returns a new Request pointer for the service API
 // operation and parameters.
 func (c *Client) NewRequest(operation *Operation, params interface{}, data interface{}) *Request {
+
```
jasdel: new line needed?

aws/config.go
```diff
@@ -1,5 +1,10 @@
 package aws
 
+// UseServiceDefaultRetries instructs the config to use the service's own
+// default number of retries. This will be the default action if
+// Config.MaxRetries is nil also.
+const UseServiceDefaultRetries = -1
```
jasdel: This const isn't referenced anywhere, probably don't needed it. The V1 SDK used this value but the v2 doesn't since default retryer is created by the client.

aws/default_retryer.go
```diff
@@ -21,6 +21,7 @@ import (
 //    func (d retryer) MaxRetries() int { return 100 }
 type DefaultRetryer struct {
 	NumMaxRetries int
+	MinTime       int64
```
jasdel: Probably want to include a member for min throttle time as well, e.g. (`minThrottleTime`)  Since this is separate value than non-throttle retry delay minimum.

In addition need to document what the default behavior is if the members are zero.

This is also a good opportunity to consider changing the minimum times from a `int64` to `time.Duration`  time.Duration is much nicer to use than an integer representing seconds.

```go
type DefaultRetryer struct {
    NumMaxRetries        int
    MinimumRetryDelay    time.Duration
    MinimumThrottleDelay time.Duration
}
```

aws/default_retryer.go
```diff
@@ -34,25 +35,39 @@ var seededRand = rand.New(&lockedSource{src: rand.NewSource(time.Now().UnixNano(
 // RetryRules returns the delay duration before retrying this request again
 func (d DefaultRetryer) RetryRules(r *Request) time.Duration {
 	// Set the upper limit of delay in retrying at ~five minutes
-	minTime := 30
+	var minTime int64
+	if d.MinTime != 0 {
+		minTime = d.MinTime
+	} else {
+		minTime = 30
```
jasdel: Since the min retry and throttle delay will be configuration should create constants for both that document the default minimum delay.

```go
const (
    DefaultRetryerRetryDelay    = 30 * time.Milliseconds 
    DefaultRetryerThrottleDelay = 30 * time.Milliseconds
    DefaultRetryerMaxNumRetries = 3
)
```

aws/default_retryer.go
```diff
@@ -34,25 +35,39 @@ var seededRand = rand.New(&lockedSource{src: rand.NewSource(time.Now().UnixNano(
...
+	var initialDelay time.Duration
 	throttle := d.shouldThrottle(r)
 	if throttle {
-		if delay, ok := getRetryDelay(r); ok {
-			return delay
+		if delay, ok := getRetryAfterDelay(r); ok {
+			initialDelay = delay
 		}
 
 		minTime = 500
```
jasdel: This should use a const type now that the value will be configurable, e.g. `DefaultRetryerThrottleDelay`

aws/default_retryer.go
```diff
@@ -34,25 +35,39 @@ var seededRand = rand.New(&lockedSource{src: rand.NewSource(time.Now().UnixNano(
...
 	}
 
 	retryCount := r.RetryCount
-	if retryCount > 13 {
-		retryCount = 13
-	} else if throttle && retryCount > 8 {
+	if throttle && retryCount > 8 {
 		retryCount = 8
+	} else if retryCount > 12 {
+		retryCount = 12
```
jasdel: This is a good time to consider adding a configuration for `MaxiumRetryDelay` instead of capping the `retryCount`.

aws/default_retryer.go
```diff
@@ -34,25 +35,39 @@ var seededRand = rand.New(&lockedSource{src: rand.NewSource(time.Now().UnixNano(
...
-	delay := (1 << uint(retryCount)) * (seededRand.Intn(minTime) + minTime)
-	return time.Duration(delay) * time.Millisecond
+	maxRetriesAllowed := d.MaxRetries()
+	if maxRetriesAllowed != 0 {
+		if retryCount > maxRetriesAllowed-1 {
+			retryCount = maxRetriesAllowed - 1
+		}
+	}
+
+	delay := (1 << uint(retryCount)) * (seededRand.Int63n(minTime) + minTime)
```
jasdel: If we update the above logic to use a `MaxiumRetryDelay` instead of capping based on `retryCount` this logic will need to be updated as `retryCount` greater than 30 will have an integer overflow. (63, if the exponent is wrapped in as int64)

private/model/api/shape_marshal.go
```diff
@@ -79,7 +79,7 @@ func getContentType(s *Shape) string {
 	if s.API.Metadata.JSONVersion != "" && s.API.Metadata.Protocol == "json" {
 		return fmt.Sprintf("application/x-amz-json-%s", s.API.Metadata.JSONVersion)
 	}
-	if  s.API.Metadata.Protocol == "json" || s.API.Metadata.Protocol == "rest-json" {
+	if s.API.Metadata.Protocol == "json" || s.API.Metadata.Protocol == "rest-json" {
```
jasdel: Assuming these are gofmt changes?

service/ec2/customizations.go
```diff
@@ -8,7 +8,43 @@ import (
...
+	case opAssignPrivateIpAddresses:
+		return d.customRetryRule(r)
+	default:
+		return d.DefaultRetryer.RetryRules(r)
+	}
+}
+
+func (d retryer) customRetryRule(r *request.Request) time.Duration {
+	d.MinTime = 1000
+	return d.DefaultRetryer.RetryRules(r)
```
jasdel: Specifying the min time should be done when the custom retrier is created, instead of in the function call.

service/ec2/customizations.go
```diff
@@ -8,7 +8,43 @@ import (
...
+		},
+	}
+}
+
 func init() {
+	initClient = func(c *Client) {
+		if c.Config.Retryer == nil {
+			// Only override the retryer with a custom one if the config
+			// does not already contain a retryer
+			setCustomRetryer(c)
```
jasdel: I think it would be better for `setCustomRetryer` return a `retryer` instead of setting the `Retryer` field within its own function.  (Also would be good example of using the constant suggested above for `DefaultRetryerMaxNumRetries`)

```go
if c.Config.Retryer == nil {
    c.Retryer = retryer{
         DefaultRetryer: request.DefaultRetryer{
              NumMaxRetries: 3,
              MinTime: 30,
         },
    }
}
```

service/ec2/customizations.go
```diff
@@ -8,7 +8,43 @@ import (
...
+}
+
+func (d retryer) RetryRules(r *request.Request) time.Duration {
+	switch r.Operation.Name {
+	case opModifyNetworkInterfaceAttribute:
+		fallthrough
+	case opAssignPrivateIpAddresses:
+		return d.customRetryRule(r)
+	default:
+		return d.DefaultRetryer.RetryRules(r)
```
jasdel: We could remove the need of having two code paths usages of the same utility by only adding the custom retrier for the API operations that it is needed for. This would mean the custom retrier would be added in `initRequest` instead of `initClient`. And would use this switch above to determine when to add the custom retrier to a request.

aws/default_retryer.go
```diff
@@ -31,28 +53,75 @@ func (d DefaultRetryer) MaxRetries() int {
...
+	} else {
+		delay = getDelaySeed(maxDelay / 2)
 	}
+	return delay + initialDelay
+}
 
-	delay := (1 << uint(retryCount)) * (seededRand.Intn(minTime) + minTime)
-	return time.Duration(delay) * time.Millisecond
+func getDelaySeed(duration time.Duration) time.Duration {
+	return time.Duration(seededRand.Int63n(int64(duration)) + int64(duration))
```
jasdel: Nit, this isn't really returning a delay `seed`, but returning the delay.  I suggest renaming this function to something more similar to `getJitterDelay`.

aws/default_retryer_test.go
```diff
@@ -152,7 +154,7 @@ func TestGetRetryDelay(t *testing.T) {
 	}
 
 	for i, c := range cases {
-		a, ok := getRetryDelay(&c.r)
+		a, ok := getRetryAfterDelay(&c.r)
```
jasdel: The test function for this test case should be renamed from `TestCanUseRetryAfter` to `TestGetRetryAfterDelay`

aws/request_test.go
```diff
@@ -207,7 +213,9 @@ func TestRequestExhaustRetries(t *testing.T) {
 		{StatusCode: 500, Body: body(`{"__type":"UnknownError","message":"An error occurred."}`)},
 	}
 
-	s := awstesting.NewClient(unit.Config())
+	cfg := unit.Config()
+	cfg.Retryer = aws.NewDefaultRetryer()
+	s := awstesting.NewClient(cfg)
```
jasdel: Did `unit.Config` not already include a retryer? Its not clear from this why it needed to be added. I would of expected the retryer to of been either already set in `unit.Config` or when the below `awstesting.NewClient` is created.

aws/request_test.go
```diff
@@ -572,7 +588,9 @@ func TestRequest_NoBody(t *testing.T) {
 
 			cfg := unit.Config()
 			cfg.Region = "mock-region"
-			cfg.Retryer = aws.DefaultRetryer{NumMaxRetries: 0}
+			cfg.Retryer = aws.NewDefaultRetryer(func(d *aws.DefaultRetryer) {
+				d.NumMaxRetries = 0
```
jasdel: It would be nice to have a no op retryer that always disabled retries, e.g `NopRetryer`

internal/awstesting/client.go
```diff
@@ -7,7 +7,7 @@ import (
 // NewClient creates and initializes a generic service client for testing.
 func NewClient(cfg aws.Config) *aws.Client {
 	if cfg.Retryer == nil {
-		cfg.Retryer = aws.DefaultRetryer{NumMaxRetries: 3}
+		cfg.Retryer = aws.NewDefaultRetryer()
```
jasdel: This should prevent the need for adding the `cfg.Retryer = aws.NewDefaultRetryer()` in the above tests when no special configuration is being specified.

internal/awstesting/unit/unit.go
```diff
@@ -16,6 +16,7 @@ func init() {
 			Source: "unit test credentials",
 		},
 	}
+	config.Retryer = aws.NewDefaultRetryer()
```
jasdel: Probably want to leave this unset, so that the behavior doesn't change between a config not specifying a retryer an the client automatically creating the default behavior.

service/ec2/customizations.go
```diff
@@ -4,19 +4,41 @@ import (
...
+
+	// customRetryerMinRetryDelay sets min retry delay
+	customRetryerMinRetryDelay = 1 * time.Second
+
+	// customRetryerMaxRetryDelay sets max retry delay
+	customRetryerMaxRetryDelay = 8 * time.Second
+)
+
+// setCustomRetryer overides the default Retryer values
+func setCustomRetryer(d *aws.DefaultRetryer) {
```
jasdel: minor nit, this is configuring the default retryer's options not setting the custom retryer. Maybe changed to `setRetryerConfig`?

service/ec2/customizations.go
```diff
@@ -4,19 +4,41 @@ import (
...
+}
+
 func init() {
-	initRequest = func(c *Client, r *request.Request) {
+	initRequest = func(c *Client, r *aws.Request) {
 		if r.Operation.Name == opCopySnapshot { // fill the PresignedURL parameter
 			r.Handlers.Build.PushFront(fillPresignedURL)
 		}
+		if c.Config.Retryer == nil && (r.Operation.Name == opModifyNetworkInterfaceAttribute || r.Operation.Name == opAssignPrivateIpAddresses) {
+			r.Retryer = aws.Retryer(aws.NewDefaultRetryer(setCustomRetryer))
```
jasdel: Shouldn't need to case the returned value of `aws.NewDefaultRetryer` to `aws.Retryer`

aws/request_test.go
skotambkar: unit.Config didn't include the retryer previously. I added a retryer in unit.config with this PR, but didn't remove the redundant changes which I made earlier.

internal/awstesting/client.go
skotambkar: Removing retryer initialization from awstesting as aws/client.go has NewClient method to initialize retryer with default retryer if not specified

CHANGELOG_PENDING.md
```diff
@@ -1,6 +1,12 @@
 ### SDK Features
 
 ### SDK Enhancements
-
+* `aws`: Provides more customization options for retryer by adding a constructor for default Retryer which accepts functional options. Adds NoOpRetryer to support no retry behavior. Exposes members of default retryer. [#375](https://github.com/aws/aws-sdk-go-v2/pull/375)
+    * Updates the underlying logic used by the default retryer to calculate jittered delay for retry. 
+    * Handles int overflow for retry delay. Fixes [#370](https://github.com/aws/aws-sdk-go-v2/issues/370)
```
jasdel: Suggest updating this so that their is a short title for the entry, with a paragraph for the description.

PR #378, opened 2019-09-01
service/s3/s3manager: Fix index out of range when a io.Reader returns -1

dgrr:
Fixes the S3 Upload Manager's handling of an unbounded streaming reader that returns negative bytes read.


dgrr:
Hello @jasdel 

My error is happening because I am using a custom reader I made myself to read the files from S3 and process the file in between. When my custom reader detects an error returns -1 and the error. I can change that to return 0 and an error, but I think it's not the expected behavior in an io.Reader when an error happens.

jasdel:
Thanks for the update @dgrr. In the case of an error returned by a `Read` call the `n` parameter "should" be ignored by the caller. With that said, it looks like the `nextReader` function's usage of the `Read`'s returned `n` should be updated to only return the error and not create the buffer if there was an error.

We probably don't want to check for a value of `n`, but should update the code to return an `err` if there was an error, and otherwise return the sized buffer.

```go
n, err := readFillBuf(r, part)
if err != nil {
    return nil, n, err
}

u.readerPos += int64(n)

return bytes.NewReader(part[0:n]), n, nil
```

There is no recovery when the `Read` method returns an error, so don't need to worry about updating `u.readerPos` in the case of an error.

dgrr:
What's next? @jasdel

service/s3/s3manager/upload.go
```diff
@@ -440,14 +440,17 @@ func (u *uploader) nextReader() (io.ReadSeeker, int, error) {
 	default:
 		part := make([]byte, u.cfg.PartSize)
 		n, err := readFillBuf(r, part)
+		if n < 0 {
+			return nil, n, err
```
jasdel: Probably want a check here in case `err` is nil that fills error.

jasdel: e.g.:

```go
		n, err := readFillBuf(r, part)
		if n < 0 {
			if err == nil {
				err = fmt.Errorf("unable to read part, got negative bytes read(%d) without error, %v",
					n, err)
			}
			return nil, n, err
		}
```

dgrr:
@jasdel updated. I also resolved the conflicts in CHANGELOG_PENDING.md

jasdel:
Thanks for making those updates. The change looks good, we'll get this merged in.

PR #382, opened 2019-09-05
Be more explicit in the README about the state of the project

TopherGopher:
We performed a fairly significant upgrade from `aws-sdk-go` to `aws-sdk-go-v2` thinking that it was the community recommended version. It's very gloss over the mention that -v2 is in beta.

As a direct result of this misunderstanding, we now will need to perform a significant downgrade back to the original library, as each minor version update contains large breaking changes.

This MR calls out more explicitly that this is a beta project and should not be consumed for production code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # AWS SDK for Go v2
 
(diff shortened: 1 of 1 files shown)


jasdel:
Thanks for clarifying the README @TopherGopher. The v2 SDK is still experiencing breaking changes. We hope to more clearly communicate our road map and path to GA for the v2 SDK in the future.

PR #384, opened 2019-09-12
aws: Refactors request retry behavior path logic 

### LINKED ISSUE #145: Remove 501, NotImplemented, from being retried.
https://github.com/aws/aws-sdk-go/pull/1826

### LINKED ISSUE #372: Port V1 Request retry behavior path logic to v2
Port v1 SDK's aws/aws-sdk-go#2744 fix related to aws/aws-sdk-go#1376 to the v2 SDK. 

Include porting of related fix aws/aws-sdk-go#2774 for aws/aws-sdk-go#2773

skotambkar:
Ports v1 SDK's retry utilities to follow a consistent code path. `aws.IsErrorRetryable` is the primary entry point to determine if a request is retryable. Also ports support for `service/kinesis` to retry service specific API errors . Corrects sdk's behavior by not retrying errors with status code 501.

Related to https://github.com/aws/aws-sdk-go/pull/2744, https://github.com/aws/aws-sdk-go/pull/2774, https://github.com/aws/aws-sdk-go/pull/2751 and https://github.com/aws/aws-sdk-go/pull/1826

Fixes #372; Fixes #145

### DIFF
--- a/aws/default_retryer.go
+++ b/aws/default_retryer.go
@@ -109,9 +109,6 @@ func (d DefaultRetryer) ShouldRetry(r *Request) bool {
 		return *r.Retryable
 	}
 
-	if r.HTTPResponse.StatusCode >= 500 && r.HTTPResponse.StatusCode != 501 {
-		return true
-	}
 	return r.IsErrorRetryable() || r.IsErrorThrottle()
 }
 
--- a/aws/retryer.go
+++ b/aws/retryer.go
@@ -207,6 +207,18 @@ func (r *Request) IsErrorRetryable() bool {
 		return true
 	}
 
+	// HTTP response status code 501 should not be retried.
+	// 501 represents Not Implemented which means the request method is not
+	// supported by the server and cannot be handled.
+	if r.HTTPResponse != nil {
+		// HTTP response status code 500 represents internal server error and
+		// should be retried without any throttle.
+		if r.HTTPResponse.StatusCode == 500 {
+			return true
+		}
+
+	}
+
 	return IsErrorRetryable(r.Error)
 }
 
@@ -221,7 +233,11 @@ func (r *Request) IsErrorThrottle() bool {
 
 	if r.HTTPResponse != nil {
 		switch r.HTTPResponse.StatusCode {
-		case 429, 502, 503, 504:
+		case
+			429, // error caused due to too many requests, thus retry should be throttled
+			502, // Bad Gateway error should be throttled
+			503, // caused when service is unavailable, thus retry should be throttled
+			504: // error occurred due to gateway timeout, thus retry should be throttled
 			return true
 		}
 	}


aws/default_retryer.go
```diff
@@ -79,7 +79,7 @@ func (d DefaultRetryer) RetryRules(r *Request) time.Duration {
 	retryCount := r.RetryCount
 
 	maxDelay := d.MaxRetryDelay
-	if throttle {
+	if isThrottle {
 		maxDelay = d.MaxThrottleDelay
```
skmcgrail: Maybe move this logic up into the isThrottle check above?

aws/defaults/handlers.go
```diff
@@ -158,9 +158,10 @@ func handleSendError(r *aws.Request, err error) {
 			Body:       ioutil.NopCloser(bytes.NewReader([]byte{})),
 		}
 	}
-	// Catch all other request errors.
+
+	// Catch all request errors, and let the default retrier determine
```
skmcgrail: np: reference as the interface type name "Catch all request errors, and let the retryer determine if error is retryable)

aws/http_request_retry_test.go
```diff
@@ -37,15 +32,13 @@ func TestRequestCancelRetry(t *testing.T) {
 	s.Handlers.UnmarshalError.Clear()
 	s.Handlers.Send.PushFront(func(r *aws.Request) {
 		reqNum++
-		r.Error = errors.New("net/http: request canceled")
 	})
 	out := &testData{}
 	r := s.NewRequest(&aws.Operation{Name: "Operation"}, nil, out)
 	r.HTTPRequest.Cancel = c
```
skmcgrail: Since this is v2 should we change this particular test to not use the deprecated cancel behavior while we are making modifications?

aws/request_retry_test.go
```diff
@@ -0,0 +1,202 @@
...
+			Code: "UnknownCode",
+			RetryErrorCodes: []string{
+				"RetryMePlease",
+				"SomeOtherError",
+			},
+			Retryable: false,
+		},
+	}
+
+	for name, c := range cases {
```
skmcgrail: We should use `t.Run(name, func(t *testing.T){ // code })` to make these cases visible

aws/retryer.go
```diff
@@ -106,32 +114,90 @@ func isNestedErrorRetryable(parentErr awserr.Error) bool {
...
+		// `awserr.Error.OrigErr()` can be nil, meaning there was an error but
+		// because we don't know the cause, it is marked as retryable. See
+		// TestRequest4xxUnretryable for an example.
+		return true
+
+	default:
+		switch err.Error() {
+		case "net/http: request canceled",
+			"net/http: request canceled while waiting for connection":
+			// known 1.5 error case when an http request is cancelled
```
skmcgrail: Should we remove this 1.5 specific case?

service/kinesis/customizations_test.go
```diff
@@ -1,14 +1,17 @@
 package kinesis
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws/defaults"
```
skmcgrail: goimports

service/sts/customizations_test.go
```diff
@@ -1,6 +1,12 @@
 package sts_test
 
 import (
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws/defaults"
```
skmcgrail: goimports

aws/defaults/handlers.go
skotambkar: I didn't get what you meant by this, can you elaborate?

aws/retryer.go
skotambkar: Need to ascertain that this is 1.5 specific case

aws/request.go
```diff
@@ -33,13 +32,18 @@ const (
 	ErrCodeRequestCanceled = "RequestCanceled"
 )
 
+// NoBody is a http.NoBody reader instructing Go HTTP client to not include
+// and body in the HTTP request.
+var NoBody = http.NoBody
```
jasdel: This type alias is no longer needed. It was a compatability with Go 1.8. The SDK's code can refer to `http.NoBody` instead now.

aws/request_retry_test.go
```diff
@@ -0,0 +1,206 @@
+// +build go1.7
```
jasdel: Don't really need this build tag in v2 SDK.

aws/request_1_8_test.go
```diff
@@ -1,84 +0,0 @@
...
-	"net/http/httptest"
-	"strings"
-	"testing"
-
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
-	"github.com/aws/aws-sdk-go-v2/internal/awstesting/unit"
-)
-
-func TestResetBody_WithEmptyBody(t *testing.T) {
```
jasdel: Want to make sure these tests are not deleted. I think they are still valid for the v2 SDK.

aws/request_1_8_test.go
```diff
@@ -1,84 +0,0 @@
...
-
-	r.ResetBody()
-
-	if a, e := r.HTTPRequest.Body, http.NoBody; a != e {
-		t.Errorf("expected request body to be set to reader, got %#v",
-			r.HTTPRequest.Body)
-	}
-}
-
-func TestRequest_FollowPUTRedirects(t *testing.T) {
```
jasdel: Want to make sure these tests are not deleted. I think they are still valid for the v2 SDK.

aws/default_retryer.go
```diff
@@ -111,26 +109,10 @@ func (d DefaultRetryer) ShouldRetry(r *Request) bool {
...
-	if r.HTTPResponse != nil {
-		switch r.HTTPResponse.StatusCode {
-		case 429:
-		case 502:
-		case 503:
-		case 504:
-		default:
-			return r.IsErrorThrottle()
-		}
+	if r.HTTPResponse.StatusCode >= 500 && r.HTTPResponse.StatusCode != 501 {
```
jasdel: The status code determination for retry probably should happen in `r.IsErrorThrottle` since the `Request.IsErrorThrottle` method already checks status codes. would be better not to duplicate this here and `r.IsErrorThrottle`

skotambkar: Retryable is true for all the status code >= 500 except 501. while r.isErrorThrottle is true only when status code is either 429, 502, 503,or 504.  Is this the correct behavior?

jasdel: Good point, I was incorrectly thinking `500` should be included in throttle. The `500` statuscode check should be a part of `r.IsErrorRetryable`  A 500 should be a regular retry, not throttle.

PR #385, opened 2019-09-16
awstesting: Fixes AssertXML to correctly assert on differences

skmcgrail:
Ports change from V1 to V2 to correct the AssertXML behavior.


skmcgrail:
Need to regenerate model

skmcgrail:
Closing this for the time being, this issue has uncovered issues within the current design and implementation of our generated marshalers. We will follow-up with this change once we have completed the redesign and implementation of the marshaler code.

PR #393, opened 2019-09-19
sdk: Remove deprecated go versions

skotambkar:
Removes deprecated go version specific code from the SDK. Fixes and updates broken test on code generation. Replaces `Retryer` embedded field on `Request` with a named field.

Raises the minimum supported version to Go 1.11 for the SDK. Older versions may work, but are not actively supported

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,7 @@ Deprecations
   * Removing plugin credential provider from the v2 SDK developer preview. This feature may be made available as a separate module.
 * Removes support for deprecated Go versions ([#393](https://github.com/aws/aws-sdk-go-v2/pull/393))
   * Removes support for Go version specific files from the SDK. Also removes irrelevant build tags, and updates the README.md file. 
+  * Raises the minimum supported version to Go 1.11 for the SDK. Older versions may work, but are not actively supported
   
 SDK Features
 ---
--- a/internal/ini/walker_test.go
+++ b/internal/ini/walker_test.go
@@ -11,7 +11,7 @@ import (
 
 func TestValidDataFiles(t *testing.T) {
 	const expectedFileSuffix = "_expected"
-	err := filepath.Walk("./testdata/valid", func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk(filepath.Join("testdata", "valid"), func(path string, info os.FileInfo, err error) error {
 		if strings.HasSuffix(path, expectedFileSuffix) {
 			return nil
 		}


CHANGELOG_PENDING.md
```diff
@@ -1,11 +1,18 @@
 Services
 ---
 
+Deprecates
+---
+* Removes deprecated go versions ([#393](https://github.com/aws/aws-sdk-go-v2/pull/393))
```
jasdel: suggest, `Remove support for deprecated Go versions`

README.md
```diff
@@ -4,7 +4,7 @@
 
 `aws-sdk-go-v2` is the **Developer Preview** for the v2 AWS SDK for the Go programming language. While in Developer Preview (aka **beta**) the SDK may release minor version changes that break backwards compatability. The release notes for the breaking change will include information about the breaking change, and how you can migrate to the latest version.
 
-Check out the [Issues] and [Projects] for design and updates being made to the SDK. The v2 SDK requires a minimum version of Go 1.9.
+Check out the [Issues] and [Projects] for design and updates being made to the SDK. The v2 SDK requires a minimum version of Go 1.11.
```
jasdel: Might be helpful to bold `Go 1.11` to be easier to notice.

aws/defaults/handlers.go
```diff
@@ -189,12 +189,12 @@ var AfterRetryHandler = aws.NamedHandler{
 	Fn: func(r *aws.Request) {
 		// If one of the other handlers already set the retry state
 		// we don't want to override it based on the service's state
-		if r.Retryable == nil || r.Config.EnforceShouldRetryCheck {
-			r.Retryable = aws.Bool(r.ShouldRetry(r))
+		if r.Retryable == nil {
+			r.Retryable = aws.Bool(r.Retryer.ShouldRetry(r))
```
jasdel: I think the `Retryer.ShouldRetry` should be called regardless of the state of the `r.Retryable` flag. The SDK's `DefaultRetryer` should inspect this flag and if set, the DefaultRetryer should return that value, true|false.

This would remove the need for the `nil` check here in this function.

aws/request.go
```diff
@@ -389,19 +383,6 @@ func (r *Request) getNextRequestBody() (body io.ReadCloser, err error) {
 			"failed to get request body error", err)
 	}
 
-	// Go 1.8 tightened and clarified the rules code needs to use when building
```
jasdel: I think we want to keep this doc comment. The SDK still needs to not set the `http.Request.Body` value when there has not been a body specified by the SDK's request. WE should have a github issue tracking this fix.

aws/retryer.go
```diff
@@ -170,9 +171,7 @@ func shouldRetryError(origErr error) bool {
 
 	default:
 		switch err.Error() {
-		case "net/http: request canceled",
-			"net/http: request canceled while waiting for connection":
-			// known 1.5 error case when an http request is cancelled
+		case "net/http: request canceled":
```
jasdel: Can this use a strings has prefix, or strings contains instead of exact match.

aws/retryer.go
```diff
@@ -152,6 +152,7 @@ func shouldRetryError(origErr error) bool {
 		}
 		// *url.Error only implements Temporary after golang 1.6 but since
 		// url.Error only wraps the error:
+		// Todo: Investigate use case for this error case for +Go1.11
```
jasdel: This todo should have a matching Github issue so we don't loose track of the task.

jasdel: We should have a Github issue that tracks refactoring the SDK's retry behavior to take advantage of the supported Go version's behavior for request failures and cancellation.

internal/ini/walker_test.go
```diff
@@ -93,6 +91,9 @@ func TestValidDataFiles(t *testing.T) {
 
 		return nil
 	})
+	if err != nil {
+		t.Errorf("Error while walking the file tree rooted at root, %d", err)
```
jasdel: That really should fatalf

internal/sdkmath/floor.go
```diff
@@ -1,5 +1,3 @@
-// +build go1.10
-
 package sdkmath
 
 import "math"
```
jasdel: I think we can  remove this file as well, and use the stdlib version directly in the SDK's code.

aws/defaults/handlers.go
skotambkar: Default Retryer already checks for r.Retryable flag in ShouldRetry fn. If it is set, we return the *r.Retryable.

CHANGELOG_PENDING.md
```diff
@@ -5,12 +5,16 @@ Deprecations
 ---
 * `aws`: Removes plugin credential provider ([#391](https://github.com/aws/aws-sdk-go-v2/pull/391))
   * Removing plugin credential provider from the v2 SDK developer preview. This feature may be made available as a separate module.
+* Removes support for deprecated Go versions ([#393](https://github.com/aws/aws-sdk-go-v2/pull/393))
+  * Removes support for Go version specific files from the SDK. Also removes irrelevant build tags, and updates the README.md file.
```
jasdel: probably want to call out the minimum supported Go version is now 1.11 as well.

internal/ini/walker_test.go
```diff
@@ -13,7 +11,7 @@ import (
 
 func TestValidDataFiles(t *testing.T) {
 	const expectedFileSuffix = "_expected"
-	filepath.Walk("./testdata/valid", func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk("./testdata/valid", func(path string, info os.FileInfo, err error) error {
```
jasdel: Was already existing before, but this should use `filepath.Join("testdata", "valid")` instead of the literal string. `filepath.Join` will correctly join the path for the OS.

PR #394, opened 2019-09-20
service/s3/s3crypto: Deprecates the crypto client from the SDK

skotambkar:
s3crypto client is deprecated and may be removed from the future versions of the SDK.


service/s3/s3crypto/doc.go
```diff
@@ -1,67 +1,4 @@
 /*
-Package s3crypto provides encryption to S3 using KMS and AES GCM.
```
jasdel: We should probably keep the rest of this package's doc for now, until the client is removed.

PR #401, opened 2019-09-26
service/dynamodb/dynamodbattribute: New Encoder and Decoder Behavior for Empty Collections

skmcgrail:
This PR pulls over the changes from the V1 SDK for supporting the (un)marshaling of empty collections. The change has been altered here to default to this behavior rather then having to opt-in. This PR does not consolidate/condense the unit-tests now that there is no feature flag, and has been left for another point in time as part of an overall cleanup.

V1: https://github.com/aws/aws-sdk-go/pull/2834

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,6 +13,9 @@ Deprecations
   
 SDK Features
 ---
+* `service/dynamodb/dynamodbattribute`: New Encoder and Decoder Behavior for Empty Collections ([#401](https://github.com/aws/aws-sdk-go-v2/pull/401))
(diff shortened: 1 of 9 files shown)


CHANGELOG_PENDING.md
```diff
@@ -13,6 +13,8 @@ Deprecations
 
 SDK Features
 ---
+* `service/dynamodb/dynamodbattribute`: New Encoder and Decoder Behavior for Empty Collections
```
skotambkar: Nit: Add PR reference

jasdel: Also would be great to reference the V1 associated change

skmcgrail:
Updated changelog with requested changes.

PR #404, opened 2019-09-26
Ports S3 Manager Upload and Download Buffering Strategies

### LINKED ISSUE #402: services/s3/s3manager: Uploader does not utilize buffer pool
The S3 Upload Manager does not utilize a pool of buffers when performing a streaming upload where the size of the `io.Reader` is not known.

### LINKED ISSUE #403: service/s3/s3manager: Uploader does not support configuring a MaxUploadParts
V1 SDK allows users to configure the max number of upload parts. This allows for configuring a smaller number then the default of 10,000.

V1 Documentation:
```go
// MaxUploadParts is the max number of parts which will be uploaded to S3.
// Will be used to calculate the partsize of the object to be uploaded.
// E.g: 5GB file, with MaxUploadParts set to 100, will upload the file
// as 100, 50MB parts. With a limited of s3.MaxUploadParts (10,000 parts).
```

skmcgrail:
Ports Upload and Download Buffering Strategies from V1 SDK and in addition fixes #402  which is a v1-sync issue.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,7 +10,7 @@ Deprecations
 * Removes support for deprecated Go versions ([#393](https://github.com/aws/aws-sdk-go-v2/pull/393))
   * Removes support for Go version specific files from the SDK. Also removes irrelevant build tags, and updates the README.md file. 
   * Raises the minimum supported version to Go 1.11 for the SDK. Older versions may work, but are not actively supported
-
+  
 SDK Features
 ---
 * `service/s3/s3manager`: Add Upload Buffer Provider ([#404](https://github.com/aws/aws-sdk-go-v2/pull/404))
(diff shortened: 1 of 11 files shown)


jasdel:
Does it make sense to include #403 in this PR as well?

skmcgrail:
> Does it make sense to include #403 in this PR as well?

Yes I can pull this over as well in this PR. I will push a new commit to include this.

service/s3/s3manager/download_test.go
```diff
@@ -644,6 +646,80 @@ func TestDownload_WithFailure(t *testing.T) {
 	}
 }
 
+func TestDownloadBufferStrategy(t *testing.T) {
```
jasdel: It would be helpful to pull in the s3manager v1 test updates made in https://github.com/aws/aws-sdk-go/pull/2841 to this PR as well. The `fillRandom` function took enough time it could cause Go tests to sometimes timeout with 10m when running in Travis.

skmcgrail: Added

PR #417, opened 2019-10-17
internal/sdkio: add RingBuffer data structure to the sdk

skotambkar:
The PR adds in an implementation of RingBuffer data structure which acts as a revolving buffer of a predefined length. The RingBuffer implements io.ReadWriter interface.


internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,100 @@
...
+func New(capacity int) *RingBuffer {
+	slice := make([]byte, capacity, capacity)
+	ringBuf := RingBuffer{
+		slice: slice,
+		start: 0,
+		end:   0,
+		size:  0,
+	}
+	return &ringBuf
+}
```
jasdel: Instead of using of `New` creating the byte slice, if the byte slice is passed in it allows the usage of RingBuffer that doesn't allocate the backing array onto the heap.

```go
var buf [1024]byte
ringBuf := sdkio.NewRingBuffer(buf[:])

reader := io.TeeReader(body, ringBuf)

decoder := xml.Decoder(reader)
```

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,100 @@
...
+type RingBuffer struct {
+	slice []byte
+	start int
+	end   int
+	size  int
+}
+
+// New method takes in the capacity as an int
+// and returns a RingBuffer.
+func New(capacity int) *RingBuffer {
```
jasdel: This should be named `NewRingBuffer`

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,100 @@
...
+		size:  0,
+	}
+	return &ringBuf
+}
+
+// Write method inserts the elements
+// in the byte slice, and return
+// the number of bytes written and an error
+//
+// This satisfies io.Writer interface.
```
jasdel: Nit don't have to wrap comment so signfiicantly at least 80 characters is fine :)

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return r.size, nil
+}
+
+// Read method on RingBuffer returns the read count
+// along with Error encountered while reading.
+//
+// Read copies the data on the ring buffer into
+// the byte slice provided to the method.
+func (r *RingBuffer) Read(p []byte) (int, error) {
```
jasdel: This method probably should decrement the `r.size` count since it is reading bytes out of the buffer. Incrementing  the start cursor

internal/sdkio/ringbuffer_test.go
```diff
@@ -0,0 +1,213 @@
...
+				t.Errorf("expect written bytes to be %v , got %v", e, a)
+			}
+		})
+	}
+
+}
+
+func TestRingBuffer_Read(t *testing.T) {
+	cases := []struct {
+		name                  string
```
jasdel: Instead of a `name` field a `map[string]struct{}` would be preferred. Makes it easier to identify the case as its a separate column and prevents the potential error of not setting a name.

internal/sdkio/ringbuffer_test.go
```diff
@@ -0,0 +1,213 @@
...
+			expectedReadByteSlice: []byte(""),
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			ringBuffer := New(c.capacity)
+			readSlice := make([]byte, c.sizeofReadBytes)
+
+			ringBuffer.Write(c.input)
+			_, err := ringBuffer.Read(readSlice)
```
jasdel: We'll want to add a couple additional test cases here. 

* Read larger than buffer
* Read smaller than buffer
* Multiple reads, `e.g. io.Copy`
* Validation the io.EOF is returned only once the ring buffer has been fully read

jasdel: Probably also want a test case that does both write and read interwoven

internal/sdkio/ringbuffer.go
skotambkar: [TeeReader](https://golang.org/pkg/io/#TeeReader) takes a reader as first param, and writer as second. So here we can use io.TeeReader(ringBuf, body) as ringBuf satisfies io.ReadWriter; But we can also use a MultiReader in this case. Is that correct?

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,100 @@
...
+
+		if s == r.end {
+			break
+		}
+		if s == len(r.slice) {
+			s = 0
+		}
+	}
+
+	return readCount, io.EOF
```
skotambkar: Should this also be changed to return nil, instead of io.EOF

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,81 @@
...
+		r.slice[r.end] = b
+		r.end++
+		r.size++
+	}
+	return r.size, nil
+}
+
+// Read method on RingBuffer returns the read count along with Error encountered while reading.
+//
+// Read copies the data on the ring buffer into the byte slice provided to the method.
```
jasdel: minor nit for the documentation here. I think it could be simplified as:

```
// Read copies the data on the ring buffer into the byte slice provided to the method.
// Returns the read count along with Error encountered while reading
```

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,81 @@
...
+func (r *RingBuffer) Read(p []byte) (int, error) {
+	// readCount keeps track of the number of
+	// bytes read.
+	readCount := 0
```
jasdel: nit, `var readCount int`

internal/sdkio/ringbuffer.go
```diff
@@ -0,0 +1,81 @@
...
+		// decrement the size of ring buffer
+		r.size--
+
+		if r.start == len(r.slice) {
+			r.start = 0
+		}
+
+		if s == len(r.slice) {
+			s = 0
+		}
```
jasdel: is the extra `s` var needed?  It looks like this loop could just use `r.start` to keep track of the start, and wrapping back around to the beginning of the slice's array?

internal/sdkio/ringbuffer_test.go
```diff
@@ -0,0 +1,369 @@
...
+	"bytes"
+	"io"
+	"strings"
+	"testing"
+)
+
+func TestNew(t *testing.T) {
+	byteSlice := make([]byte, 10)
+	cases := map[string]struct {
+		expectedStart int
```
jasdel: Is this extra test case needed? It seems like the behavior of `New` would be captured by the `Write` and `Read` cases.

jasdel:
Go 1.13 Codebuild job is broken :( we can bypass this until we fix the CodeBuild job.

PR #420, opened 2019-10-25
aws/endpoints: Fix SDK resolving endpoint without region

jasdel:
Fixes the SDK's endpoint resolve incorrectly resolving endpoints for a service when the region is empty. Also fixes the SDK attempting to resolve a service when the service value is empty.

Related to: aws/aws-sdk-go#2911

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,5 +10,5 @@ SDK Enhancements
 SDK Bugs
 ---
 * `aws/endpoints`: aws/endpoints: Fix SDK resolving endpoint without region ([#420](https://github.com/aws/aws-sdk-go-v2/pull/420))
-  * Fixes the SDK's endpoint resolve incorrectly resolving endpoints for a service when the region is empty. Also fixes the SDK attempting to resolve a service when the service value is empty..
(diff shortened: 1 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -9,4 +9,6 @@ SDK Enhancements
 
 SDK Bugs
 ---
-
+* `aws/endpoints`: aws/endpoints: Fix SDK resolving endpoint without region ([#420](https://github.com/aws/aws-sdk-go-v2/pull/420))
+  * Fixes the SDK's endpoint resolve incorrectly resolving endpoints for a service when the region is empty. Also fixes the SDK attempting to resolve a service when the service value is empty..
```
skotambkar: Nit `..` at end

aws/endpoints/v3model.go
```diff
@@ -74,16 +74,29 @@ func (p partition) canResolveEndpoint(service, region string, strictMatch bool)
 	return p.RegionRegex.MatchString(region)
 }
 
+func allowLegacyEmptyRegion(service string) bool {
+	legacy := map[string]struct{}{
+		"ec2metadata": {},
+	}
```
skotambkar: Does it help to keep the legacy map outside this function to avoid creating it every time this function is called?

jasdel: updated to represent this better. moved map into a package variable.

PR #441, opened 2019-11-27
sdk: Add support for types/enums pkg as sub-packages of each service pkg

skotambkar:
**When merging this PR, we should remove api_types, api_enums from all service pkgs.**
* Modifies existing sdk service pkg layout. `types` and `enums` sub-packages has been added to `service/*` pkg. 
* Input, output types along with other api types of a service are now generated in a `types` sub-package. Every operation has it's own types file which consist of input, output types for each operation. 
* api-enums would now be generated in a `enums` sub-package.
* Few additional customizations, code cleanup are introduced for review in this PR.

Related to #445


private/model/api/customization_passes.go
```diff
@@ -17,28 +17,16 @@ type service struct {
...
 
 	// Enable generated marshalers
-	switch a.Metadata.Protocol {
-	case "rest-xml", "rest-json":
-		a.NoGenMarshalers = false
-	}
+	// switch a.Metadata.Protocol {
+	// case "rest-xml", "rest-json":
+	// 	a.NoGenMarshalers = false
+	// }
```
skotambkar: uncommented this in later commits.

private/model/api/smoke.go
```diff
@@ -59,10 +59,12 @@ func (a *API) APISmokeTestsGoCode() string {
 	a.AddImport("context")
 	a.AddImport("testing")
 	a.AddImport("time")
+
 	a.AddSDKImport("aws")
 	a.AddSDKImport("aws/awserr")
 	a.AddSDKImport("internal/awstesting/integration")
-	a.AddImport(a.ImportPath())
+	a.AddSDKImport("service", a.PackageName())
+	a.AddSDKImport("service", a.PackageName(), TypesPkgName)
```
jasdel: I think changing the `a.AddImport(a.ImportPath())`, to `a.AddImport(a.ImportPath())` will prevent providing a base import path that doesn't start with `service`.

private/protocol/restxml/build_bench_test.go
```diff
@@ -14,7 +14,11 @@ import (
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting/unit"
 	"github.com/aws/aws-sdk-go-v2/private/protocol/restxml"
 	"github.com/aws/aws-sdk-go-v2/service/cloudfront"
+	"github.com/aws/aws-sdk-go-v2/service/cloudfront/enums"
+	"github.com/aws/aws-sdk-go-v2/service/cloudfront/types"
```
jasdel: Nit suggest usign `cfEnums` and `cfTypes`. and us camel casing instead of underscores.

private/model/api/passes.go
```diff
@@ -158,6 +160,42 @@ func (a *API) fixStutterNames() {
...
+
+					fmt.Printf("Renamed shape %v to %v for package %v \n", s.ShapeName, name, a.PackageName())
+					s.Rename(name)
+				}
+			}
+		}
+
+		// Throw an error if shape is struct or enum and is still starting with non alphabetical characters
+		if (s.Type == "structure" || s.IsEnum()) && !unicode.IsLetter(rune(s.ShapeName[0])) {
+			log.Fatalf("Shape starting with non alphabetical character found: %v", s.ShapeName)
```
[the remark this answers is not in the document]
skotambkar: debuglogger doesn't serve the purpose here. I think this should be a fatal error.

aws/example_test.go
```diff
@@ -8,7 +8,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws/defaults"
 	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	types2 "github.com/aws/aws-sdk-go-v2/service/sqs/types"
```
jasdel: Nit `types2` would be better as `sqsTypes` or similar

example/aws/endpoints/customEndpoint/customEndpoint.go
```diff
@@ -9,8 +9,11 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	dynamodb_types "github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	s3_types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/aws/aws-sdk-go-v2/service/sqs"
+	sqs_types "github.com/aws/aws-sdk-go-v2/service/sqs/types"
```
jasdel: Would be better for the aliases not to include underscores(`_`)

models/protocol_tests/generate.go
```diff
@@ -592,6 +592,8 @@ func getType(t string) uint {
 func main() {
 	fmt.Println("Generating test suite", os.Args[1:])
 	out := generateTestSuite(os.Args[1])
+
+	// Todo: fix test suites
 	if len(os.Args) == 3 {
```
jasdel: probably don't need this `todo` A tracking Github issue would be a better way to track the protocol tests no longer can be generated.

private/model/api/api.go
```diff
@@ -302,7 +305,7 @@ func (a *API) APIGoCode() string {
 		a.AddSDKImport("private/protocol")
 	}
 	if !a.NoGenMarshalers || !a.NoGenUnmarshalers {
-		a.AddSDKImport("private/protocol")
+		a.imports[path.Join(SDKImportRoot, "private/protocol")] = true
```
jasdel: This should use the `AddSDKImport` utility instead of the `a.imports` direclty.

```go
a.AddSDKImport("private/protocol")
```

private/model/api/api.go
```diff
@@ -319,23 +322,33 @@ func (a *API) APIGoCode() string {
 // string.
 func (a *API) APIOperationGoCode(op *Operation) string {
 	a.resetImports()
-	a.AddSDKImport("aws")
-	a.AddSDKImport("internal/awsutil")
 	a.AddImport("context")
+	a.AddSDKImport("aws")
+	a.AddSDKImport("service", a.PackageName(), ServiceTypesPkgName)
```
jasdel: This should be 

```go
a.AddImport(a.ImportPath(), ServiceTypesPkgName)
```

private/model/api/api.go
```diff
@@ -496,6 +501,26 @@ func (a *API) ServiceID() string {
...
+	case "rest-xml":
+		return "aws_restxml"
+	case "rest-json":
+		return "aws_restjson"
+	case "ec2":
+		return "aws_ec2query"
+	case "query":
+		return "aws_query"
+	case "xml":
+		return "aws_xml"
```
jasdel: XML is not a protocol on its own.  We probably want to remove this case from the switch.

private/model/api/api.go
```diff
@@ -803,6 +819,7 @@ func (a *API) InterfaceGoCode() string {
 		a.AddSDKImport("aws")
 	}
 	a.AddImport(a.ImportPath())
+	a.AddSDKImport("service", a.PackageName(), ServiceTypesPkgName)
```
jasdel: This should use `a.AddImport(a.ImportPath(), ServiceTypesPkgName)`

Or even better would be to create a utility that does this directly

.e.g 
```go
a.AddSDKServiceTypesImport()
a.AddSDKServiceEnumsImport()
```

private/model/api/example.go
```diff
@@ -288,11 +275,19 @@ type exHeader struct {
...
+	a.AddImport("fmt")
+	a.AddImport("context")
+	a.AddImport("strings")
+	a.AddImport("time")
+
+	a.AddSDKImport("aws")
+	a.AddSDKImport("aws/awserr")
+	a.AddSDKImport("aws/external")
+	a.AddSDKImport("service", a.PackageName())
+	a.AddSDKImport("service", a.PackageName(), "types")
```
jasdel: would perferable to use utility instead of adding `service` pkg direclty

private/model/api/passes.go
```diff
@@ -158,6 +160,38 @@ func (a *API) fixStutterNames() {
 	}
 }
 
+func (a *API) validateShapeNames() {
+	for _, s := range a.Shapes {
+		// name should start with an alphabet, that is ascii character 65 to 90, and 97 to 122;
+		// else skip that character
+		name := s.ShapeName
```
jasdel: It would be helpful to have unit test for this validation behavior.

In addition to leading underscores(`_`), what should the behavior be for other underscores in the shape's name. both within the shape's name and suffix?

private/model/api/s3manager_input.go
```diff
@@ -21,7 +21,7 @@ func S3ManagerUploadInputGoCode(a *API) string {
 	a.resetImports()
 	a.AddImport("io")
 	a.AddImport("time")
-	a.AddSDKImport("service/s3")
+	a.AddSDKImport("service", "s3", ServiceEnumsPkgName)
```
jasdel: This shouldn't use the `service` package direclty but the helper instead.

private/model/api/shape.go
```diff
@@ -287,6 +287,12 @@ func (s *Shape) GoStructType(name string, ref *ShapeRef) string {
 		}
 	}
 
+	// check if enum needs to be imported
+	if ref.Shape.IsEnum() || (ref.Shape.MemberRef.Shape != nil && ref.Shape.MemberRef.Shape.IsEnum()) ||
+		(ref.Shape.ValueRef.Shape != nil && ref.Shape.ValueRef.Shape.IsEnum()) {
+		s.API.AddSDKImport("service", s.API.PackageName(), ServiceEnumsPkgName)
```
jasdel: prefer not to use `service` directly.

private/model/api/shape.go
```diff
@@ -330,25 +349,18 @@ func getPkgName(s *Shape) string {
...
-		return name
-	}
-
 	prefix := ""
 	if pointer {
 		prefix = "*"
 	}
 
+	var pkgName string
+	pkgName = getPkgName(s)
```
jasdel: Nit
```go
pkgName := getPkgName(s)
```

private/model/api/shape.go
```diff
@@ -360,6 +372,13 @@ func goType(s *Shape, withPkgName, pointer bool) string {
 	case "boolean":
 		return prefix + "bool"
 	case "string", "character":
+		if s.IsEnum() {
+			s.API.AddSDKImport("service", s.API.PackageName(), ServiceEnumsPkgName)
```
jasdel: Nit should use helper

private/model/api/shape_value_builder.go
```diff
@@ -252,12 +265,11 @@ func (b ShapeValueBuilder) GoType(ref *ShapeRef, elem bool) string {
 }
 
 func getEnumName(ref *ShapeRef, t, name string) string {
-	pkg := getPkgName(ref.Shape)
+	ref.API.AddSDKImport("service", ref.API.PackageName(), ServiceEnumsPkgName)
```
jasdel: nit utility

private/model/api/smoke.go
```diff
@@ -59,10 +59,12 @@ func (a *API) APISmokeTestsGoCode() string {
 	a.AddImport("context")
 	a.AddImport("testing")
 	a.AddImport("time")
+
 	a.AddSDKImport("aws")
 	a.AddSDKImport("aws/awserr")
 	a.AddSDKImport("internal/awstesting/integration")
 	a.AddImport(a.ImportPath())
+	a.AddSDKImport("service", a.PackageName(), ServiceTypesPkgName)
```
jasdel: nit utility

private/model/api/waiters.go
```diff
@@ -45,16 +45,17 @@ type Waiter struct {
...
 func (a *API) WaitersGoCode() string {
 	var buf bytes.Buffer
-	fmt.Fprintf(&buf, "import (\n%q\n%q\n\n%q\n)",
-		"time",
-		"context",
-		SDKImportRoot+"/aws",
-	)
+
+	a.resetImports()
+	a.AddSDKImport("service", a.PackageName(), "types")
```
jasdel: nit utility

service/dynamodb/dynamodbattribute/decode_test.go
```diff
@@ -1,5 +1,7 @@
 package dynamodbattribute
 
+// code generation? No
```
jasdel: Still needed?

private/model/api/passes.go
skotambkar: We validate only leading characters of a shape name.  

ex.
1. shape name: `Application_ShapeName` is considered a valid name and not renamed  
2. shape name: `__Application_ShapeName` will be renamed to `Application_ShapeName`. 
3. shape name : `&Application_ShapeName` will fail validation and we will exit with appropriate log message.

private/model/api/passes.go
```diff
@@ -160,36 +159,58 @@ func (a *API) fixStutterNames() {
...
-			} else {
-				break
+		if s.Type == "structure" || s.IsEnum() {
+			name := s.ShapeName
+
+			// The regexp used allows underscores(_) at the beginning of the shape name
+			// There may be 0 or more underscores(_). The next character which would be the leading character
+			// in the renamed shape must be an alphabetic character.
+			// The regex allows alphanumeric characters along with underscores(_) in rest of the string.
+			compiledRegExp, err := regexp.Compile("^[_]*[a-zA-Z][a-zA-Z0-9_]*$")
```
jasdel: Probably should move this `regexp.Compile` to be a package variable instead of compiled per shape. Compiling per shape probably will add a non-trivial amount of time to the SDK's code generation.

Can use [regexp.MustCompile](https://golang.org/pkg/regexp/#MustCompile) instead of `Compile` so don't have to check error for the package variable.

private/model/api/passes.go
```diff
@@ -160,36 +159,58 @@ func (a *API) fixStutterNames() {
...
+			if err != nil {
+				return fmt.Errorf("failed to compile regex while validating shape name")
+			}
+
+			// check if the shape name satisfies the regex
+			if b := compiledRegExp.Match([]byte(name)); !b {
+				return fmt.Errorf("invalid shape name found: %v", s.ShapeName)
```
jasdel: can use [regexp.Regexp.MatchString](https://golang.org/pkg/regexp/#Regexp.MatchString) instead of `Match` here so you don't have to cast the `name` to a `[]byte`

private/model/api/passes.go
```diff
@@ -160,36 +159,58 @@ func (a *API) fixStutterNames() {
...
-		if s.ShapeName != name {
-			debugLogger.Logf("Renamed shape %v to %v for package %v \n", s.ShapeName, name, a.PackageName())
-			if a.Shapes[name] != nil {
-				// throw an error if shape with a new shape name already exists
-				log.Fatalf("Tried to rename shape %v to %v, but the new name results in shape name collision",
-					s.ShapeName, name)
+			// replace all _ with whitespace
+			name = strings.Replace(name, "_", " ", -1)
+			// remove whitespace and get a slice of strings
+			slice := strings.Fields(name)
```
jasdel: Could combine the `Replace` and `Fields` statement together with [strings.FieldsFunc](https://golang.org/pkg/strings/#FieldsFunc). This utility will swash multiple delimiters together when separating fields.

https://play.golang.org/p/9NDH5VonVp9

PR #447, opened 2019-12-03
private/model: Templates for code generated marshaling stubs

skotambkar:
This PR adds code generation templates for generated marshaling stubs. The generated stubs use the reflection based marshalers. This PR uses types, enums pkg; and thus requires merging of PR #441 before this is used.


jasdel:
Closing as no longer needed, XML protocol (de)serializers are being generated in `feature/SmithyAPIClient` branch.

Issue #445, opened 2019-12-02
Refactoring API client package layout to group enums, types in their own respective packages.


skotambkar:
This issue details how the v2 AWS SDK for Go's API client packages can be updated as a step towards improving generated serializers performance. The proposed changes in this design seek to improve the v2 SDK package layout to group service enums, types in their own respective packages.

We'd like to hear your feedback on this proposed refactor to the v2 SDK package layout.

## Motivation

We aim to improve the generated serializers performance for the AWS SDK for Go. A step towards this goal is to create stand-alone operation serializer functions and reduce use of reflection/ run-time assertions in the SDK. This would require us to refactor the Service client package and group enums, types in their own respective packages to avoid cyclic dependencies in the Go SDK. 

Currently, the Go SDK's API Client Package is cluttered with client operations, types, enums, errors and client specific customizations. The proposed change of moving the types, enums in their own respective packages would also help improve their discoverability. 

## Proposed Changes

We propose a refactor to the SDK's API Client package layout. The changes proposes splitting the API Client package in three primary packages with canonical dependency. 

For example for s3 service, the three packages would be : 
* Package s3 
* Package types 
* Package enums

The dependency flow would be as follows:
```
Given a set of objects S and a transitive relation R ⊆ S × S with ( a , b ) ∈ R modeling a dependency "a depends on b"
 S = { s3 pkg , types pkg , enums pkg}
 R = {(s3 pkg , types pkg), (s3 pkg , enums pkg), (types pkg , enums pkg)}

```

* s3 package will depend on both types and enums packages
* types package might depend on enums packages. 

The enums package would not depend on any of the above two packages. This packaging layout helps avoid circular reference 

The proposed refactor of the SDK's API Client package would create a package layout as follows :

```
|-service/
	|- s3/
		|- api_client.go
		|- ap_op_opname.go	
		|- enums
			|- api_enums.go
		|- types
			|- api_op_opname.go
				- operation input
				- operation output
			|- api_types.go 			
		|- <Others>
```

## Usage example 

The following example invokes the Amazon S3 API Client's Create Bucket Request using the new types, enums package. 

```go
// To create a bucket in a specific region
//
// The following example creates a bucket. The request specifies an AWS region
// where to create the bucket.
package main 

import (
	"context"
	"fmt"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/aws/awserr"
[...]

PR #449, opened 2019-12-04
aws/protocol/rest: V2 REST Encoder Implementation

skmcgrail:
V2 implementation of the HTTP bindings for use in implementing AWSREST and AWSXML protocols.

### DIFF
--- a/aws/protocol/rest/encode.go
+++ b/aws/protocol/rest/encode.go
@@ -1,0 +1,79 @@
+package rest
+
+import (
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// An Encoder provides encoding of REST URI path, query, and header components
+// of an HTTP request. Can also encode a stream as the payload.
+//
+// Does not support SetFields.
+type Encoder struct {
+	req *http.Request
+
+	path, rawPath, pathBuffer []byte
+
+	query  url.Values
+	header http.Header
+}
+
+// NewEncoder creates a new encoder from the passed in request. All query and
+// header values will be added on top of the request's existing values. Overwriting
+// duplicate values.
+func NewEncoder(req *http.Request) *Encoder {
+	e := &Encoder{
+		req: req,
+
+		path:    []byte(req.URL.Path),
+		rawPath: []byte(req.URL.Path),
+		query:   req.URL.Query(),
+		header:  req.Header,
+	}
+
+	return e
+}
+
+// Encode returns a REST protocol encoder for encoding HTTP bindings
+// Returns any error if one occurred during encoding.
+func (e *Encoder) Encode() error {
+	e.req.URL.Path, e.req.URL.RawPath = string(e.path), string(e.rawPath)
+	e.req.URL.RawQuery = e.query.Encode()
+	e.req.Header = e.header
+
+	return nil
+}
+
+// AddHeader returns a HeaderValue for appending to the given header name
+func (e *Encoder) AddHeader(key string) HeaderValue {
+	return newHeaderValue(e.header, key, true)
+}
+
+// SetHeader returns a HeaderValue for setting the given header name
+func (e *Encoder) SetHeader(key string) HeaderValue {
+	return newHeaderValue(e.header, key, false)
+}
+
+// Headers returns a Header used encoding headers with the given prefix
+func (e *Encoder) Headers(prefix string) Headers {
+	return Headers{
+		header: e.header,
+		prefix: strings.TrimSpace(prefix),
+	}
+}
+
+// SetURI returns a URIValue used for setting the given path key
+func (e *Encoder) SetURI(key string) URIValue {
+	return newURIValue(&e.path, &e.rawPath, &e.pathBuffer, key)
+}
+
+// SetQuery returns a QueryValue used for setting the given query key
+func (e *Encoder) SetQuery(key string) QueryValue {
+	return newQueryValue(e.query, key, false)
+}
+
+// AddQuery returns a QueryValue used for appending the given query key
+func (e *Encoder) AddQuery(key string) QueryValue {
+	return newQueryValue(e.query, key, true)
+}
--- a/aws/protocol/rest/header.go
+++ b/aws/protocol/rest/header.go
@@ -1,0 +1,103 @@
+package rest
+
+import (
+	"encoding/base64"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/private/protocol"
+)
+
+// Headers is used to encode header keys using a provided prefix
+type Headers struct {
+	header http.Header
+	prefix string
+}
+
+// AddHeader returns a HeaderValue used to append values to prefix+key
+func (h Headers) AddHeader(key string) HeaderValue {
+	return h.newHeaderValue(key, true)
+}
+
+// SetHeader returns a HeaderValue used to set the value of prefix+key
+func (h Headers) SetHeader(key string) HeaderValue {
+	return h.newHeaderValue(key, false)
+}
+
+func (h Headers) newHeaderValue(key string, append bool) HeaderValue {
+	return newHeaderValue(h.header, h.prefix+strings.TrimSpace(key), append)
+}
+
+// HeaderValue is used to encode values to an HTTP header
+type HeaderValue struct {
+	header http.Header
+	key    string
+	append bool
+}
+
+func newHeaderValue(header http.Header, key string, append bool) HeaderValue {
+	return HeaderValue{header: header, key: strings.TrimSpace(key), append: append}
+}
+
+func (h HeaderValue) modifyHeader(value string) {
+	lk := strings.ToLower(h.key)
+
+	val := h.header[lk]
+
+	if h.append {
+		val = append(val, value)
+	} else {
+		val = append(val[:0], value)
+	}
+
+	h.header[lk] = val
+}
+
+// String encodes the value v as the header string value
+func (h HeaderValue) String(v string) {
+	h.modifyHeader(v)
+}
+
+// Integer encodes the value v as the header string value
+func (h HeaderValue) Integer(v int64) {
+	h.modifyHeader(strconv.FormatInt(v, 10))
+}
+
+// Boolean encodes the value v as a header string value
+func (h HeaderValue) Boolean(v bool) {
+	h.modifyHeader(strconv.FormatBool(v))
+}
+
+// Float encodes the value v as a header string value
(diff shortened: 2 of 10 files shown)


private/protocol/rest/v2/query.go
```diff
@@ -0,0 +1,71 @@
...
+	return &QueryValue{
+		query:  query,
+		key:    key,
+		append: append,
+	}
+}
+
+func (qv *QueryValue) updateKey(value string) {
+	if qv.append {
+		qv.query.Add(qv.key, value)
```
skotambkar: Should `Add` check if the key already exists?

private/protocol/rest/v2/encode.go
```diff
@@ -0,0 +1,75 @@
...
+// Encode will return the request and body if one was set. If the body
+// payload was not set the io.ReadSeeker will be nil.
+//
+// returns any error if one occured while encoding the API's parameters.
+func (e *Encoder) Encode() *http.Request {
+	e.req.URL.Path, e.req.URL.RawPath = string(e.path), string(e.rawPath)
+	e.req.URL.RawQuery = e.query.Encode()
+	e.req.Header = e.header
+
+	return e.req
```
skotambkar: Why not set request here, instead of returning the `*http.Request`?

private/protocol/rest/v2/encode.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (e *Encoder) Headers(prefix string) *Headers {
+	return &Headers{
+		header: e.header,
+		prefix: strings.TrimSpace(prefix),
+	}
+}
+
+func (e *Encoder) AddURI(key string) *URIValue {
```
skmcgrail: This should be `SetURI`, will update in the next revision.

private/protocol/rest/v2/encode.go
```diff
@@ -0,0 +1,75 @@
...
+
+// NewEncoder creates a new encoder from the passed in request. All query and
+// header values will be added on top of the request's existing values. Overwriting
+// duplicate values.
+func NewEncoder(req *http.Request) *Encoder {
+	e := &Encoder{
+		req: req,
+
+		path:    []byte(req.URL.Path),
+		rawPath: []byte(req.URL.Path),
```
jasdel: Should `rawPath` start off with the value from `req.URL.RawPath`? This could of been set.

private/protocol/rest/v2/encode.go
```diff
@@ -0,0 +1,75 @@
...
+func (e *Encoder) SetHeader(key string) *HeaderValue {
+	return newHeaderValue(e.header, key, false)
+}
+
+func (e *Encoder) Headers(prefix string) *Headers {
+	return &Headers{
+		header: e.header,
+		prefix: strings.TrimSpace(prefix),
+	}
+}
```
jasdel: Can this builder be done without returning a new pointer for the `Headers` structure?  Limiting the pointer use to only where needed will help reduce allocations. This will increase the struct copies but the copies shouldn't need to be put onto the heap, should only exist on the stack.

Since the underlying `e.header` is a pointer we probably can get away with the value setter types not being pointers.

private/protocol/rest/v2/query.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (qv *QueryValue) JSONValue(v aws.JSONValue) error {
+	encodeJSONValue, err := protocol.EncodeJSONValue(v, protocol.NoEscape)
+	if err != nil {
+		return err
+	}
+	qv.updateKey(encodeJSONValue)
+	return nil
+}
```
jasdel: Is JSONValue supported for REST's Querystring?

private/protocol/rest/v2/uri.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (u *URIValue) JSONValue(v aws.JSONValue) error {
+	encodeJSONValue, err := protocol.EncodeJSONValue(v, protocol.NoEscape)
+	if err != nil {
+		return err
+	}
+
+	return u.modifyURI(encodeJSONValue)
+}
```
jasdel: Probably don't need to support JSONValue for URI

private/protocol/rest/v2/utils_test.go
```diff
@@ -0,0 +1,36 @@
+package rest
```
jasdel: What do you think about naming this `shared_test.go`. would follow similar testing behavior/logic that is shared between multiple test files.

private/protocol/rest/v2/query.go
skmcgrail: Today the reflection based rest-encoder supports JSON value (at least in the case where it supports adding it as a `query.Set(...)` call. This would not be encoded as base64, but would technically get encoded using the URI path encoding.

private/protocol/rest/v2/uri.go
skmcgrail: Similar case to Query, URI would have supported this in the reflection code-base, we can remove it here if we believe it to be an invalid use-case.

private/protocol/rest/v2/encode.go
skmcgrail: Good point.

aws/protocol/rest/header_test.go
```diff
@@ -0,0 +1,213 @@
...
+				t.Errorf("expected %v, got %v", e, a)
+			}
+		})
+	}
+}
+
+func TestHeaders(t *testing.T) {
+	const prefix = "x-amzn-meta-"
+	cases := map[string]struct {
+		headers  http.Header
```
jasdel: would be good to have a test case that ensures that adding a `headers` doesn't mutate the map key's casing.

e.g. 

```
map[string]string{
     "foo": "fooValue",
}
```
becomes
```
X-Amzn-Meta-foo: fooValue
```

Would address the S3 Metadata casing issue where the SDK drops the original casing of the header causing unexpected behavior with S3 Metadata for an object.

This probably means the `Encoder` needs to set the `http.Header` map/values  directly and not go through the `Set` and `Add` Methods. https://github.com/aws/aws-sdk-go/issues/445

skmcgrail: For issue #445 this is actually an issue when retrieving objects and metadata from S3. When Go sends the PutObject request the header would be in the form of: `X-Amzn-Meta-Foo`. S3 normalizes this as `x-amzn-meta-foo` which can be confirmed using the S3 console. But when a GetObject is performed, Go is converting the headers received into canonical mime format, which results in the map key for the returned metadata to become titled. This can be seen in this example:

```go
package main

import (
	"bytes"
	"crypto/rand"
	"flag"
	"fmt"
	"io"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/s3"
	"github.com/davecgh/go-spew/spew"
)

var bucket, region string

func init() {
	flag.StringVar(&bucket, "bucket", "", "bucket name")
	flag.StringVar(&region, "region", "", "region name")
}

func main() {
	sess := session.Must(session.NewSession(&aws.Config{Region: &region}))

	uid := uniqueID()

	fmt.Printf("uid: %v\n", uid)

	svc := s3.New(sess)
	_, err := svc.PutObject(&s3.PutObjectInput{
		Bucket: &bucket,
		Key:    &uid,
		Metadata: map[string]*string{
			"key-one": aws.String("value-one"),
			"key-two": aws.String("value-two"),
		},
		Body: bytes.NewReader(make([]byte, 1024)),
	})
	if err != nil {
		panic(fmt.Errorf("service error: %v", err))
	}

	obj, err := svc.GetObject(&s3.GetObjectInput{
		Bucket: &bucket,
		Key:    &uid,
	})
	if err != nil {
		panic(fmt.Errorf("service error: %v", err))
	}

	spew.Dump(obj.Metadata)
}

func uniqueID() string {
	uuid := make([]byte, 16)
	io.ReadFull(rand.Reader, uuid)
	return fmt.Sprintf("%x", uuid)
}
```

This outputs:
```
uid: de8a11a34eccedcc5fcd4b7ec24d6ae0
(map[string]*string) (len=2) {
 (string) (len=7) "Key-One": (*string)(0xc00018ed60)((len=9) "value-one"),
 (string) (len=7) "Key-Two": (*string)(0xc00018ed70)((len=9) "value-two")
}
```

Also viewable if you capture the HTTPS requests:
Screen Shot 2019-12-11 at 1 26 27 PM
Screen Shot 2019-12-11 at 1 26 44 PM

jasdel: Should the SDK rely on the behavior of S3? or set the key in the casing the user provided it in? I think this is the key question we should use with regard to S3's Metadata. If another service were to add support for `headers` would it have the same behavior?

aws/protocol/rest/encode.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	return e
+}
+
+// Encode will return the request and body if one was set. If the body
+// payload was not set the io.ReadSeeker will be nil.
+//
+// returns any error if one occured while encoding the API's parameters.
+func (e *Encoder) Encode() *http.Request {
```
skotambkar: Just so we don't miss this. As per our discussion this should return error.

aws/protocol/rest/header.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (h HeaderValue) Time(t time.Time, format string) (err error) {
+	value, err := protocol.FormatTime(format, t)
+	if err != nil {
+		return err
+	}
+	h.modifyHeader(value)
+	return nil
+}
```
skotambkar: Nit: Do we need to name the return parameter? 
Can it be just 
```go 
func (h HeaderValue) Time(t time.Time, format string) error {
	...
}
```

aws/protocol/rest/header_test.go
skmcgrail: I've modified the encoder to normalize headers as case-insensitive.

aws/protocol/rest/encode.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Encode returns a REST protocol encoder for encoding HTTP bindings
+// Returns any error if one occurred during encoding.
+func (e *Encoder) Encode() error {
+	e.req.URL.Path, e.req.URL.RawPath = string(e.path), string(e.rawPath)
+	e.req.URL.RawQuery = e.query.Encode()
+	e.req.Header = e.header
+
+	return nil
+}
```
skotambkar: How will this return an error, if it occurred during encoding?

jasdel: Ok to return error in this case since, we may want to add validation in the future

PR #452, opened 2019-12-06
`service`: Adds prototyped operations from APIGateWay and S3 service that uses rest encoder utility 

skotambkar:
* Adds prototype for `getApiKey` request for apigateway service and `putObject` request for s3 service that use rest encoder utility from PR [aws/aws-sdk-go-v2#449](https://github.com/aws/aws-sdk-go-v2/pull/449)
* This PR adds changes on top of changes added for PR [aws/aws-sdk-go-v2#441](https://github.com/aws/aws-sdk-go-v2/pull/441), also takes in changes from PR [aws/aws-sdk-go-v2#449](https://github.com/aws/aws-sdk-go-v2/pull/449). 
* We add Marshal.go for service/apigateway & service/s3. It contains prototyped marshaler operations and functions for respective operations.
* We also add test to compare the marshaled request using rest encoder utility with the old rest marshaler.

### DIFF
--- a/service/apigateway/internal/awsrestjson/marshal.go
+++ b/service/apigateway/internal/awsrestjson/marshal.go
@@ -1,7 +1,7 @@
 package awsrestjson
 
 import (
-	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/private/protocol/jsonrpc"
@@ -43,10 +43,16 @@ func (m ProtoGetAPIKeyMarshaler) MarshalOperation(r *aws.Request) {
 // This method uses the rest encoder utility
 func MarshalGetAPIKeyInputShapeAWSREST(input *types.GetApiKeyInput, encoder *restV2.Encoder) error {
 	// encode using rest encoder utility
-	if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
-		return fmt.Errorf("failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
+	if input.ApiKey != nil {
+		if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
+			return awserr.New(aws.ErrCodeSerialization, "failed to marshal API KEY to URI using REST encoder:\n \t %v", err)
+		}
 	}
-	encoder.AddQuery("includeValue").Boolean(*input.IncludeValue)
+
+	if input.IncludeValue != nil {
+		encoder.AddQuery("includeValue").Boolean(*input.IncludeValue)
+	}
+
 	return nil
 }
 
--- a/service/s3/internal/awsrestxml/marshal.go
+++ b/service/s3/internal/awsrestxml/marshal.go
@@ -41,41 +41,98 @@ func MarshalPutObjectInputShapeAWSREST(input *types.PutObjectInput, encoder *res
 	// Encoding shapes with location `headers`
 	marshalShapeMapForHeaders(encoder, "x-amz-meta-", input.Metadata)
 	//  Encoding shapes with location `header`
-	encoder.AddHeader("Cache-Control").String(*input.CacheControl)
-	encoder.AddHeader("Content-Disposition").String(*input.ContentDisposition)
-	encoder.AddHeader("Content-Language").String(*input.ContentLanguage)
-	encoder.AddHeader("Content-Length").Integer(*input.ContentLength)
-	encoder.AddHeader("Content-Md5").String(*input.ContentMD5)
-	encoder.AddHeader("Content-Type").String(*input.ContentType)
-	encoder.AddHeader("x-amz-acl").String(string(input.ACL))
-	encoder.AddHeader("x-amz-grant-full-control").String(*input.GrantFullControl)
-	encoder.AddHeader("x-amz-grant-read").String(*input.GrantRead)
-	encoder.AddHeader("x-amz-grant-read-acp").String(*input.GrantReadACP)
-	encoder.AddHeader("x-amz-grant-write-acp").String(*input.GrantWriteACP)
-	encoder.AddHeader("x-amz-object-lock-legal-hold").String(string(input.ObjectLockLegalHoldStatus))
-	encoder.AddHeader("x-amz-object-lock-mode").String(string(input.ObjectLockMode))
-	encoder.AddHeader("x-amz-tagging").String(*input.Tagging)
-	encoder.AddHeader("x-amz-request-payer").String(string(input.RequestPayer))
-	encoder.AddHeader("x-amz-server-side-encryption-context").String(*input.SSEKMSEncryptionContext)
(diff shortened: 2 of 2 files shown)


service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -0,0 +1,65 @@
...
+	err = MarshalGetAPIKeyInputShapeAWSREST(m.Input, encoder)
+	if err != nil {
+		r.Error = err
+	}
+
+	err = MarshalGetAPIKeyInputShapeAWSJSON(m.Input, r)
+	if err != nil {
+		r.Error = err
+	}
+}
```
jasdel: If an error is encountered while Marshaling within `MarshalOperation` need to return and not continue.

service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -0,0 +1,65 @@
...
+	encoder := v2.NewEncoder(r.HTTPRequest)
+	// adds content-type header
+	encoder.AddHeader("Content-Type").String("application/json")
+
+	err = MarshalGetAPIKeyInputShapeAWSREST(m.Input, encoder)
+	if err != nil {
+		r.Error = err
+	}
+
+	err = MarshalGetAPIKeyInputShapeAWSJSON(m.Input, r)
```
jasdel: It would be helpful here to add a TODO that this needs to be updated not to pass in `r`  directly but a payload

service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -0,0 +1,65 @@
...
+	if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
+		log.Fatalf("Failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
+	}
+	encoder.AddQuery("includeValue").Boolean(*input.IncludeValue)
+	encoder.Encode()
+	return nil
+}
+
+// MarshalGetAPIKeyInputShapeAWSJSON is a stand alone function used to marshal the json body
+func MarshalGetAPIKeyInputShapeAWSJSON(v *types.GetApiKeyInput, r *aws.Request) error {
```
jasdel: It would be helpful here to add a TODO that this needs to be updated not to pass in `r`  directly but a payload

service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -19,40 +19,44 @@ type ProtoGetAPIKeyMarshaler struct {
...
+	encoder := restV2.NewEncoder(r.HTTPRequest)
 	// adds content-type header
 	encoder.AddHeader("Content-Type").String("application/json")
 
 	err = MarshalGetAPIKeyInputShapeAWSREST(m.Input, encoder)
 	if err != nil {
 		r.Error = err
+		return
 	}
+	encoder.Encode()
```
jasdel: If `Encode` returns an `error` need to check and return it if so.

service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -19,40 +19,44 @@ type ProtoGetAPIKeyMarshaler struct {
...
 // MarshalGetAPIKeyInputShapeAWSREST is a stand alone function used to marshal the HTTP bindings a input shape.
 // This method uses the rest encoder utility
-func MarshalGetAPIKeyInputShapeAWSREST(input *types.GetApiKeyInput, encoder *v2.Encoder) error {
+func MarshalGetAPIKeyInputShapeAWSREST(input *types.GetApiKeyInput, encoder *restV2.Encoder) error {
 	// encode using rest encoder utility
 	if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
-		log.Fatalf("Failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
+		return fmt.Errorf("failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
 	}
 	encoder.AddQuery("includeValue").Boolean(*input.IncludeValue)
```
jasdel: This check should first check that `input.ApiKey` is not nil, before setting the value into the encoder.

```go
if input.ApiKey != nil {
    if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
         return fmt.Errorf("failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
    }
}
```

service/apigateway/internal/awsrestjson/marshal.go
```diff
@@ -19,40 +19,44 @@ type ProtoGetAPIKeyMarshaler struct {
...
 }
 
 // MarshalGetAPIKeyInputShapeAWSREST is a stand alone function used to marshal the HTTP bindings a input shape.
 // This method uses the rest encoder utility
-func MarshalGetAPIKeyInputShapeAWSREST(input *types.GetApiKeyInput, encoder *v2.Encoder) error {
+func MarshalGetAPIKeyInputShapeAWSREST(input *types.GetApiKeyInput, encoder *restV2.Encoder) error {
 	// encode using rest encoder utility
 	if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
-		log.Fatalf("Failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
+		return fmt.Errorf("failed to marshal API KEY to URI using REST encoder:\n \t %v", err.Error())
```
jasdel: Not too important ATM, but these errors should be Serialization errors so we don't have to add the formatting in place here.  Wanted to call this out, but not something we need to address direclty here, but should be thinking about it as we move forward with the prototypes.

service/s3/internal/awsrestxml/marshal.go
```diff
@@ -60,35 +64,33 @@ func MarshalPutObjectInputShapeAWSREST(input *types.PutObjectInput, encoder *v2.
 	encoder.AddHeader("x-amz-storage-class").String(string(input.StorageClass))
 	encoder.AddHeader("x-amz-server-side-encryption").String(string(input.ServerSideEncryption))
 	if err := encoder.AddHeader("Expires").Time(*input.Expires, protocol.RFC822TimeFormatName); err != nil {
```
jasdel: Need to make sure to check for nil before setting any pointer based value.

For enums need to see if the enum value is not empty string.

service/apigateway/internal/awsrestjson/marshal.go
skotambkar: I was thinking code generation should only add in code statements that have a value that is not nil; 

For eg: if `input.APIKey` is nil, the code generation should not create the statement `encoder.SetURI("api_Key").String(*input.ApiKey)` altogether. 

Would that make sense?

service/apigateway/internal/awsrestjson/marshal.go
skotambkar: Sure, `Encode` doesn't directly return an error, but returns a pointer to http.request. I will check if r.Error != nil and return if not nil

jasdel:
For this PR what do you think about moving the `apigateway/internal/awsrestjson` package content into the API client's package, `apigateway`?.  This would allow us to continue making progress on marshaling prototypes and merge this PR before we finalize on the client's package structure.

PR #459, opened 2019-12-16
s3/proto_api_op_GetObject: Adds prototyped GetObject operation 

skotambkar:
This PR adds a prototyped getobject operation for S3. The PR is an initial step for generated unmarshaling for restxml protocol. 

Here we delegate unmarshaling to reflection based unmarshalers.


jasdel:
Looks like `go vet` catching bug in

```
# github.com/aws/aws-sdk-go-v2/service/s3/internal/awsrestxml 
service/s3/internal/awsrestxml/unmarshal.go:95:17: call of Decode passes non-pointer 
```

jasdel:
Pending the discussion on API client package layout, what do you think about rebasing this so that the `s3/internal/awsrestxml` package content are put in the `s3` package as unexported types/funcs.  This will allow us to merge this functionality now and continue prototyping "generated" protocol marshaling.

PR #461, opened 2019-12-19
`aws`: Add context on credential providers, EC2Metadata client operations

skotambkar:
This PR makes a breaking change to the aws sdk for go v2, by adding context as a required parameter to each operation on EC2Metadata client. We also modified the Retrieve method on Credential Provider to take in a user provided context. 

SDK operations that relied on credential providers have been updated to accommodate this change. Tests have also been modified.

### DIFF
--- a/example/service/sts/assumeRole/assumeRole.go
+++ b/example/service/sts/assumeRole/assumeRole.go
@@ -47,8 +47,7 @@ type CredentialsProvider struct {
 	*sts.Credentials
 }
 
-func (s CredentialsProvider) Retrieve() (aws.Credentials, error) {
-
+func (s CredentialsProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
 	if s.Credentials == nil {
 		return aws.Credentials{}, errors.New("sts credentials are nil")
 	}


aws/endpointcreds/provider.go
```diff
@@ -85,7 +86,7 @@ func New(cfg aws.Config) *Provider {
 
 // Retrieve will attempt to request the credentials from the endpoint the Provider
 // was configured for. And error will be returned if the retrieval fails.
-func (p *Provider) retrieveFn() (aws.Credentials, error) {
+func (p *Provider) retrieveFn(ctx context.Context) (aws.Credentials, error) {
 	resp, err := p.getCredentials()
```
skotambkar: This should take context. set context to request in get credentials method

skotambkar: This should take context. set context to request in get credentials method

aws/signer/v2/v2_test.go
```diff
@@ -75,7 +75,7 @@ func TestSignRequestWithAndWithoutSession(t *testing.T) {
 
 	signer := builder.BuildSigner()
 
-	err := signer.Sign()
+	err := signer.Sign(nil)
```
skotambkar: should not take nil for context. 

Investigate why this didn't panic.

aws/stscreds/provider.go
```diff
@@ -248,7 +248,10 @@ func (p *AssumeRoleProvider) retrieveFn() (aws.Credentials, error) {
 	}
 
 	req := p.Client.AssumeRoleRequest(input)
-	resp, err := req.Send(context.Background())
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
skotambkar: shouldn't check for nil

aws/external/provider.go
```diff
@@ -320,6 +321,6 @@ type WithEC2MetadataRegion struct {
 }
 
 // GetRegion attempts to retreive the region from EC2 Metadata service.
-func (p WithEC2MetadataRegion) GetRegion() (string, error) {
-	return p.Client.Region()
+func (p WithEC2MetadataRegion) GetRegion(ctx context.Context) (string, error) {
+	return p.Client.Region(ctx)
 }
```
jasdel: This change will prevent the `WithEC2MetadataRegion` provider to be used as a `RegionProvider` since the `RegionProvider.GetRegion` interface method doesn't take a context.  The `RegionProvider` interface will be used by the `ResolveRegion` function in `aws/external/resolve.go:82`

This is probably one of those cases where `ctx` should be `context.Background` here for the `p.Client.Region` call.

jasdel:
Are #461 and #462 duplicates of each other? They seem very similar.

skotambkar:
#462  is based of #461

aws/credentials_test.go
```diff
@@ -51,7 +52,7 @@ func TestSafeCredentialsProvider_Cache(t *testing.T) {
 	}
 
 	for i := 0; i < 2; i++ {
-		creds, err := p.Retrieve()
+		creds, err := p.Retrieve(context.Background())
```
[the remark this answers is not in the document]
skotambkar: TestSafeCredentialsProvider_Cache uses SafeCredentialsProvider and not stubProvider.

The stubProvider probably should be moved to chain_provider_test where it is used.

aws/static_provider.go
```diff
@@ -31,7 +33,13 @@ func NewStaticCredentialsProvider(key, secret, session string) StaticCredentials
 }
 
 // Retrieve returns the credentials or error if the credentials are invalid.
-func (s StaticCredentialsProvider) Retrieve() (Credentials, error) {
+func (s StaticCredentialsProvider) Retrieve(ctx context.Context) (Credentials, error) {
+	select {
+	case <-ctx.Done():
+		return Credentials{}, awserr.New(ErrCodeRequestCanceled, "context canceled", ctx.Err())
+	default: // do nothing
+	}
```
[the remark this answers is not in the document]
skotambkar: Static credential provider won't make an actual API call, so probably won't need to check for context cancellation here.

aws/external/provider.go
skotambkar: Added the change proposed in #462

aws/credentials_test.go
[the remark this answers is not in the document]
jasdel: Sorry this comment was on mis read of the changed code. this line is fine.

service/rds/rdsutils/connect.go
```diff
@@ -33,7 +34,7 @@ import (
 //
 // See http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html
 // for more information on using IAM database authentication with RDS.
-func BuildAuthToken(endpoint, region, dbUser string, credProvider aws.CredentialsProvider) (string, error) {
+func BuildAuthToken(ctx context.Context, endpoint, region, dbUser string, credProvider aws.CredentialsProvider) (string, error) {
```
jasdel: suggest instead of taking credential provide this utility takes a signer instead.

```go
HTTPV4Signer interface {
    Presign(context.Context, *http.Request, ...)
}
```

example/service/rds/rdsutils/authentication/iam_authentication.go
```diff
@@ -36,7 +37,7 @@ func main() {
 
 	credProvider := stscreds.NewAssumeRoleProvider(sts.New(cfg), os.Args[5])
 
-	authToken, err := rdsutils.BuildAuthToken(dbEndpoint, awsRegion, dbUser, credProvider)
+	authToken, err := rdsutils.BuildAuthToken(context.Background(), dbEndpoint, awsRegion, dbUser, credProvider)
```
jasdel: If we're updateing this `BuildAuthToken` and similar utilities to take a signer instead of credential provider, we'll still want to updates these examples on how to create a signer for the credProvider.

skotambkar: will update BuildAuthToken to take a pointer to the signer.

PR #462, opened 2019-12-20
`aws/ec2metadata`: modifies ec2metadata client to use appropriate request timeouts

skotambkar:
The PR modifies EC2Metadata client to use request context within its operations.
Reduces the dialer timeout and response header timeout on the EC2Metadata client to help reduce latency for known issues with EC2Metadata client running inside a container. 

Note: This PR accommodates the changes included by PR #461 , and should be merged after it has been merged.

### DIFF
--- a/aws/static_provider.go
+++ b/aws/static_provider.go
@@ -34,12 +34,6 @@ func NewStaticCredentialsProvider(key, secret, session string) StaticCredentials
 
 // Retrieve returns the credentials or error if the credentials are invalid.
 func (s StaticCredentialsProvider) Retrieve(ctx context.Context) (Credentials, error) {
-	select {
-	case <-ctx.Done():
-		return Credentials{}, awserr.New(ErrCodeRequestCanceled, "context canceled", ctx.Err())
-	default: // do nothing
-	}
-
 	v := s.Value
 	if v.AccessKeyID == "" || v.SecretAccessKey == "" {
 		return Credentials{Source: StaticCredentialsProviderName}, ErrStaticCredentialsEmpty
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -12,10 +12,6 @@ Breaking Change
 Services
 ---
 
-SDK Breaking Changes
----
-  * `aws`: Context has been added to EC2Metadata operations and credential provider operations that use EC2Metadata.([#461](https://github.com/aws/aws-sdk-go-v2/pull/461)) 
-
 SDK Features
 ---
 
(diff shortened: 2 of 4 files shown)


aws/external/provider.go
```diff
@@ -320,6 +321,6 @@ type WithEC2MetadataRegion struct {
 }
 
 // GetRegion attempts to retreive the region from EC2 Metadata service.
-func (p WithEC2MetadataRegion) GetRegion() (string, error) {
-	return p.Client.Region()
+func (p WithEC2MetadataRegion) GetRegion(ctx context.Context) (string, error) {
+	return p.Client.Region(ctx)
```
jasdel: Adding `context` to this helper will prevent the `WithEC2MetadataRegion` provider from beign used because it will no longer satisfy the `RegionProvider` interface that is used by the config to generically get the region.

I suggest instead, refactor the `WithEC2MetadataRegion` to have a constructor function that takes a context, and client. Making the `Client` member of the struct unexported.

```go
type WithEC2MetadataRegion struct {
     ctx context.Context
     client *ec2metadata.Client
}
func NewWithEC2MetadataRegion(ctx context.Context, client *ec2metadata.Client) *WithEC2MetadataRegion {
//...
}
```

Usage
```go
ec2metaClient := ec2metadata.New(defaults.Config())

cfg, err := external.LoadDefaultAWSConfig(
    external.NewWithEC2MetadataRegion(ctx, ec2metaClient),
)
```

aws/signer/v4/v4.go
```diff
@@ -324,7 +325,7 @@ func (v4 Signer) signWithBody(r *http.Request, body io.ReadSeeker, service, regi
 	}
 
 	var err error
-	ctx.credValues, err = v4.Credentials.Retrieve()
+	ctx.credValues, err = v4.Credentials.Retrieve(context.Background())
```
jasdel: The `signWithBody` method, (and its callers) will need to be updated to pass in the context like was done with v2's signer. 

```go
func (v4 Signer) Sign(ctx, context.Context, r *http.Request, body io.ReadSeeker, service, region string, signTime time.Time) (http.Header, error) {
}
```

aws/static_provider.go
```diff
@@ -31,7 +33,13 @@ func NewStaticCredentialsProvider(key, secret, session string) StaticCredentials
...
 
 // Retrieve returns the credentials or error if the credentials are invalid.
-func (s StaticCredentialsProvider) Retrieve() (Credentials, error) {
+func (s StaticCredentialsProvider) Retrieve(ctx context.Context) (Credentials, error) {
+	select {
+	case <-ctx.Done():
+		return Credentials{}, awserr.New(ErrCodeRequestCanceled, "context canceled", ctx.Err())
+	default: // do nothing
+	}
+
```
jasdel: Probably don't need this check, the context can probably be ignored in this case. If the context is canceled the request will be terminated before it is invoked.

jasdel:
Are #461 and #462 duplicates of each other? They seem very similar.

skotambkar:
#462 is based of #461 . The last commit should we reviewed for this PR. https://github.com/aws/aws-sdk-go-v2/pull/462/commits/a5e44c06ee2509fcbcc18c0d24e12de981abec77

aws/static_provider.go
skotambkar: Updated in #461

aws/external/provider.go
skotambkar: updated in #461

aws/signer/v4/v4.go
skotambkar: Updated in #461

PR #464, opened 2019-12-24
service/dynamodb/dynamodbattribute: Adds clarifying docs on dynamodbattribute.UnixTime

edwardsmatt:
* Updates the `dynamodbattribute.UnixTime` documentation describing the
default encoding of `time.Time` encoding from `time.RFC3339` to state `time.RFC3339Nano`;
as described in:

```go
func (e *Encoder) encodeStruct(av *dynamodb.AttributeValue, v reflect.Value, fieldTag tag) error
```

* Adds clarification around `dynamodbattribute.UnixTime` documentation
regarding the suitability of use as a range/sort key if the attribute value type is defined as a string

* Updates `time.Time` marshalling documentation to `time.RFC3339Nano` for the `dynamodbattribute` package

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,7 @@ SDK Enhancements
 ---
 * `aws/ec2metadata`: Adds support for EC2Metadata client to use secure tokens provided by the IMDS ([#453](https://github.com/aws/aws-sdk-go-v2/pull/453)) 
   * Modifies and adds tests to verify the behavior of the EC2Metadata client.
+* `service/dynamodb/dynamodbattribute`: Adds clarifying docs on dynamodbattribute.UnixTime ([#464](https://github.com/aws/aws-sdk-go-v2/pull/464))
 
 SDK Bugs
 --


edwardsmatt:
I wasn’t sure whether to raise an issue for this because it seems by design; however, I thought an update to the documentation was a good start.

Whilst I realise this is likely an edge case, we got tripped up by the internals of the `time.RFC3339Nano` nano truncation and thought it worth clarifying (obviously because we didn’t read the documentation thoroughly enough). 

I have another repo demonstrating this issue and a `time.Time` alias that we use to ensure sort-ability if you think it’s worth a look. 🙂

jasdel:
In your doc update it included mention that the time could not be used as a string sort key. Would you mind creating a GitHub issue describing this issue. I think this is something we want to track to see if another way marshaling the UnixTime should be supported as well. As a number makes sense for the default case, but if there are usecases were time is being stored as a number string we should track that request.

edwardsmatt:
No problem. I will put together a use case to demonstrate.

Cheers,
Matt

PR #468, opened 2020-01-07
REST marshaler prototype using REST Encoder utility

skotambkar:
**Note**: Tests fail, due to pending change required to restrict lower casing header's to unmarshaling. 

* Adds prototype for `getApiKey` operation for apigateway service and `putObject` operation for s3 service. The prototypes use rest encoder utility from PR [aws/aws-sdk-go-v2#449](https://github.com/aws/aws-sdk-go-v2/pull/449)
* We add Marshal.go for service/apigateway & service/s3. It contains prototyped marshaler operations and functions for respective operations.
* We also add test to compare the marshaled request using rest encoder utility with the old rest marshaler.

### DIFF
--- a/service/apigateway/marshal.go
+++ b/service/apigateway/marshal.go
@@ -8,19 +8,27 @@ import (
 	restlegacy "github.com/aws/aws-sdk-go-v2/private/protocol/rest"
 )
 
-// protoGetApiKeyMarshaler defines a marshaler for GetApiKey operation
-type protoGetApiKeyMarshaler struct {
+// protoGetAPIKeyMarshaler defines a marshaler for GetApiKey operation
+type protoGetAPIKeyMarshaler struct {
 	input *GetApiKeyInput
 }
 
 // marshalOperation is the top level method used within a handler stack to marshal an operation
 // This method calls appropriate marshal shape functions as per the input shape and protocol used by the service.
-func (m protoGetApiKeyMarshaler) marshalOperation(r *aws.Request) {
+func (m protoGetAPIKeyMarshaler) marshalOperation(r *aws.Request) {
 	var err error
 	encoder := rest.NewEncoder(r.HTTPRequest)
-	// adds content-type header
+
+	// We add Content-Type Header if input shape's is not of type streaming payload
+	// The value of Content-Type is decided by following:
+	// a. if shape's metadata has JSONVersion and protocol is json.
+	//    - application/x-amz-json-%s where %s is ths JSONVersion.
+	// b. else if protocol is either json or rest-json
+	//     - application/json
+	// Here protocol is rest-json, and shape is not of type streaming payload,
+	// thus content-type header with value application/json is added.
 	encoder.AddHeader("Content-Type").String("application/json")
-	err = marshalGetApiKeyInputShapeAWSREST(m.input, encoder)
+	err = marshalGetAPIKeyInputShapeAWSREST(m.input, encoder)
 	if err != nil {
 		r.Error = err
 		return
@@ -32,18 +40,18 @@ func (m protoGetApiKeyMarshaler) marshalOperation(r *aws.Request) {
 		return
 	}
 
-	err = marshalGetApiKeyInputShapeAWSJSON(m.input, r)
+	err = marshalGetAPIKeyInputShapeAWSJSON(m.input, r)
 	if err != nil {
 		r.Error = err
 	}
 }
 
-// marshalGetApiKeyInputShapeAWSREST is a stand alone function used to marshal the HTTP bindings a input shape.
+// marshalGetAPIKeyInputShapeAWSREST is a stand alone function used to marshal the HTTP bindings a input shape.
 // This method uses the rest encoder utility
-func marshalGetApiKeyInputShapeAWSREST(input *GetApiKeyInput, encoder *rest.Encoder) error {
+func marshalGetAPIKeyInputShapeAWSREST(input *GetApiKeyInput, encoder *rest.Encoder) error {
 	if input.ApiKey != nil {
 		if err := encoder.SetURI("api_Key").String(*input.ApiKey); err != nil {
-			return awserr.New(aws.ErrCodeSerialization, "failed to marshal API KEY to URI using REST encoder:\n \t %v", err)
+			return awserr.New(aws.ErrCodeSerialization, "failed to marshal API KEY to URI using REST encoder", err)
 		}
 	}
 	if input.IncludeValue != nil {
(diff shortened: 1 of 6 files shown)


service/apigateway/marshal.go
```diff
@@ -0,0 +1,71 @@
...
+	input *GetApiKeyInput
+}
+
+// marshalOperation is the top level method used within a handler stack to marshal an operation
+// This method calls appropriate marshal shape functions as per the input shape and protocol used by the service.
+func (m protoGetApiKeyMarshaler) marshalOperation(r *aws.Request) {
+	var err error
+	encoder := rest.NewEncoder(r.HTTPRequest)
+	// adds content-type header
+	encoder.AddHeader("Content-Type").String("application/json")
```
jasdel: For the purpose of this prototype it would be helpful to call out what drives this header and when it should be set.

service/apigateway/marshal.go
```diff
@@ -0,0 +1,71 @@
...
+func marshalGetApiKeyInputShapeAWSJSON(v *GetApiKeyInput, r *aws.Request) error {
+	// delegate to reflection based marshaling
+	if t := restlegacy.PayloadType(r.Params); t == "structure" || t == "" {
+		jsonrpc.Build(r)
+	}
+	return nil
+}
+
+// getNamedBuildHandler returns a Named Build Handler for an operation marshal function
+func (m protoGetApiKeyMarshaler) getNamedBuildHandler() aws.NamedHandler {
```
jasdel: minor nit, this method could be named `NamedHandler` for brevity.

service/apigateway/proto_api_op_GetApiKey_test.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	if diff := cmp.Diff(request.HTTPRequest.Header, prototypeRequest.HTTPRequest.Header); diff != "" {
+		t.Errorf("Found diff: %v", diff)
+	}
+	if diff := cmp.Diff(request.HTTPRequest.URL, prototypeRequest.HTTPRequest.URL); diff != "" {
+		t.Errorf("Found diff: %v", diff)
+	}
```
jasdel: While not directly updated in the prototype it would be helpful for the test to include comparing the marshaled JSON document.

service/s3/marshal.go
```diff
@@ -0,0 +1,162 @@
...
+	if input.StorageClass != "" {
+		encoder.AddHeader("x-amz-storage-class").String(string(input.StorageClass))
+	}
+	if input.ServerSideEncryption != "" {
+		encoder.AddHeader("x-amz-server-side-encryption").String(string(input.ServerSideEncryption))
+	}
+	if input.Expires != nil {
+		if err := encoder.AddHeader("Expires").Time(*input.Expires, protocol.RFC822TimeFormatName); err != nil {
+			return awserr.New(aws.ErrCodeSerialization, "failed to marshal header for shape Expires using REST encoder:\n \t %v", err)
+		}
```
jasdel: these error messages have (probably) unintentional formatting in them

service/s3/marshal.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// marshalShapeMapForHeaders is marshal function that takes in a map[string]string as an input along with an encoder
+// and location Name which should be used to marshal the shape with location headers.
+func marshalShapeMapForHeaders(encoder *rest.Encoder, locationName string, input map[string]string) {
+	headerObject := encoder.Headers(locationName)
+	for k, v := range input {
+		headerObject.AddHeader(k).String(v)
+	}
+}
```
jasdel: Not as important for headers but, when the SDK starts generating XML and JSON objects from `map` and `slices`. Specifically the handling around empty vs nil maps/slices.  A empty, but not nil map/slice would be serialized as an empty JSON object/list. Whereas if the map/slice were nil nothing would be marshaled.

service/apigateway/marshal.go
skotambkar: Why exported `NamedHandler` here? Can rename to `namedhandler` instead

skotambkar: If we export the `NamedHandler`, then we don't need an exported const name for the BuildHandler. However, I think exporting a const name might be better than exporting NamedHandler method as there might be a case where we have multiple Handler's

service/s3/marshal.go
```diff
@@ -112,23 +112,23 @@ func marshalPutObjectInputShapeAWSREST(input *PutObjectInput, encoder *rest.Enco
...
 	if input.Expires != nil {
 		if err := encoder.AddHeader("Expires").Time(*input.Expires, protocol.RFC822TimeFormatName); err != nil {
-			return awserr.New(aws.ErrCodeSerialization, "failed to marshal header for shape Expires using REST encoder:\n \t %v", err)
+			return awserr.New(aws.ErrCodeSerialization, "failed to marshal header for shape Expires using REST encoder: %v", err)
 		}
 	}
 	if input.ObjectLockRetainUntilDate != nil {
 		if err := encoder.AddHeader("x-amz-object-lock-retain-until-date").Time(*input.ObjectLockRetainUntilDate, protocol.ISO8601TimeFormatName); err != nil {
-			return awserr.New(aws.ErrCodeSerialization, "failed to marshal header for shape ObjectLockRetainUntilDate using REST encoder:\n \t %v", err)
+			return awserr.New(aws.ErrCodeSerialization, "failed to marshal header for shape ObjectLockRetainUntilDate using REST encoder: %v", err)
```
jasdel: This error message shouldn't include the trailing `%v` this component will be added automatically by the SDK's `awserr.Error` type.

jasdel:
Should be able to rebase this PR with the feature branch, pulling in the updated header encoding.

service/s3/marshal.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// marshalPutObjectInputShapeAWSXML is a stand alone function used to marshal the xml payload
+// This should be generated according to the payload type for rest-xml protocol
+func marshalPutObjectInputShapeAWSXML(r *aws.Request, input *PutObjectInput) error {
+	if input.Body != nil {
+		r.SetReaderBody(input.Body)
+	}
+	return nil
+}
```
jasdel: We'll want to consider how  the request body is used across multiple request attempts. This is handled in an awkward way within the SDK's `Request` type, but leads to the SDK requiring the `input.Body` to be an `io.ReadSeeker`. Ideally the SDK should always except an `io.Reader`, and fail if the request requires a hash to be computed of the body, or disable retries if the body cannot be seeked.

jasdel:
smithy branch includes generated rest (de)serializers.

PR #470, opened 2020-01-07
S3: REST-XML prototype unmarshaler deligates to Legacy Unmarshaler (GetObject operation)

skotambkar:
* PR serves as a first step towards our new unmarshaler design. 
* unmarshal.go outlines our unmarshaler type and associated unmarshaling methods that currently delegate to legacy unmarshaler of the SDK.

### DIFF
--- a/service/s3/unmarshal.go
+++ b/service/s3/unmarshal.go
@@ -36,22 +36,22 @@ func (u protoGetObjectUnmarshaler) unmarshalOperation(r *aws.Request) {
 		ringBuff := sdkio.NewRingBuffer(buff)
 		body := io.TeeReader(r.HTTPResponse.Body, ringBuff)
 		decoder := xml.NewDecoder(body)
-		startToken, err := decoder.Token()
-		if err != nil {
-			ringBuff.Read(readBuff)
-			r.Error = awserr.New(aws.ErrCodeSerialization, fmt.Sprintf("Failed to decode response body with invalid XML. Here's a snapshot : %v", readBuff), err)
-			return
-		}
 
-		if start, ok := startToken.(xml.StartElement); ok {
-			r.Error = unmarshalErrorPrototype(r, decoder, start, ringBuff)
-		} else {
-			ringBuff.Read(readBuff)
-			r.Error = awserr.New(aws.ErrCodeSerialization, fmt.Sprintf(
-				"Failed to decode response body with invalid XML. Here's a snapshot : %v", readBuff),
-				err)
+		// recurse thru the xml body till we get the startElement,
+		// we ignore the xml preamble.
+		for {
+			startToken, err := decoder.Token()
+			if err != nil {
+				ringBuff.Read(readBuff)
+				r.Error = awserr.New(aws.ErrCodeSerialization, fmt.Sprintf("Failed to decode response body with invalid XML. Here's a snapshot : %v", readBuff), err)
+				return
+			}
+			// deligate to error unmarshaler if startElement is retrieved
+			if start, ok := startToken.(xml.StartElement); ok {
+				r.Error = unmarshalErrorPrototype(r, decoder, start, ringBuff)
+				return
+			}
 		}
-		return
 	}
 
 	// delegate to reflection based rest unmarshaler
@@ -77,9 +77,11 @@ type protoXMLErrorResponse struct {
 	HostID    string `xml:"HostId"`
 }
 
-// unmarshalError unmarshal's the error response
+// unmarshalErrorPrototype unmarshal's the error response.
+// The function takes in a startElement; this is to support 200 Errors
+// where we will be reading the startElement.
 // Note: Services may have customizations that require custom error unmarshaling.
-// In this prototype, we do not prototype modeled error unmarshaling.
+//  also, in this prototype, we do not prototype modeled error unmarshaling.
 func unmarshalErrorPrototype(r *aws.Request, d *xml.Decoder, start xml.StartElement, buffer *sdkio.RingBuffer) error {
 	// protoXMLErrorResponse is error response struct for xml errors.
 	var respErr = protoXMLErrorResponse{}


service/s3/unmarshal.go
```diff
@@ -0,0 +1,111 @@
...
+// unmarshalOperation is the top level method used with a handler stack to unmarshal an operation response
+// This method calls appropriate unmarshal shape functions as per the output shape and protocol used by the
+// service.
+func (u protoGetObjectUnmarshaler) unmarshalOperation(r *aws.Request) {
+	var buff bytes.Buffer
+	body := io.TeeReader(r.HTTPResponse.Body, &buff)
+	// an xml decoder should be used to decode an xml document
+	d := xml.NewDecoder(body)
+	startToken, err := d.Token()
+	if err != nil && err != io.EOF {
```
jasdel: Attempting to get the token from a decoder if there is not response body will result in an error. This can happen if the request fails for connection error, or there just is no XML .

Specifically for this operation, the successful response body probably won't contain a token. The prototype should first check if it thinks the operation was successful or not based on status code. If the operation was not successful, then the unmarshal error could be called with the first token.

We probably don't want to optimize for the 200 error case. But use that as a specific customization.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := unmarshalGetObjectOutputShapeAWSPayload(u.output, r); err != nil {
+		r.Error = err
+		return
+	}
+}
+
+// isRequestError would check if a request response was an error
+// It takes in *aws.Request, & a startTag to check for error.
+func isRequestError(r *aws.Request) bool {
+	if r.HTTPResponse.StatusCode != http.StatusOK {
```
jasdel: Probably don't want to check only for 200, there are other 200 status codes that could be returned like 204. For example the SDK's response validation handlers, `ValidateResponseHandler` checks of the status code is `0`, or 
>= `300` to determine if the response status code as an error.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,111 @@
...
+	type xmlResponseError struct {
+		xmlErrorResponse
+	}
+
+	var respErr *xmlResponseError
+	// delegate to reflection based error decoder
+	err := d.Decode(respErr)
+	if err != nil && err != io.EOF {
+		return awserr.New(aws.ErrCodeSerialization, "Serialization error: Failed to unmarshal error", err)
+	}
```
jasdel: I don't think this will have the intended behavior. Because the caller of `unmarshalErrorPrototype` calls `d.Token()` the decoder's position in the XML document has been incremented, subsequent calls to `d.Decode` will be from the nested within the tag returned by the original `d.Token`.

service/s3/unmarshal.go
skotambkar:
```go 
startToken, err:= d.Token()
if err!= nil && err != io.EOF {
	r.Error = err
	return
}

// isRequestError checks if operation response returned a error response
if isRequestError(r) {
	r.Error = unmarshalErrorPrototype(r, d, startToken)
	return
}
```

When a response body doesn't have xml body, we should get an err of io.EOF, and proceed to check if isRequestError. 

My intial thought was that if operation is one of the few operations where we need to check for status 200 error, we should generate a different isRequestError definition, where we should also have the start tag as an param. Thus we try to retrieve the startToken before checking for error.

```go 
func isRequestError(r *aws.Request, startToken xml.Token)
``` 

The current implementation won't address the case where HTTPResponse.Body is nil. We should probably add a check for if response body is nil ans skip trying to retrieve the start tag.

service/s3/unmarshal.go
skotambkar: True, deflecting to reflection based error decoder won't have the same intended behavior here. Having a peek method on decoder would have really helped to not mess up the original structure.

An alternative but a bad workaround, is to re-init the xml.Decoder with the original io.Reader

service/s3/proto_api_op_GetObject_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestProtoGetObjectRequest_Send(t *testing.T) {
+	svc := s3.New(mock.Config())
+	input := s3.GetObjectInput{
+		Bucket:                     aws.String("mock bucket"),
+		IfMatch:                    nil,
+		IfModifiedSince:            nil,
+		IfNoneMatch:                nil,
+		IfUnmodifiedSince:          nil,
+		Key:                        aws.String("mock key"),
+		PartNumber:                 nil,
```
jasdel: Would be helpful to fill out these values for this test, and future iterations of this protototype.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if r.HTTPResponse.Body != nil {
+		body = io.TeeReader(r.HTTPResponse.Body, &buff)
+		decoder = xml.NewDecoder(body)
+		startToken, err = decoder.Token()
+		// we will encounter io.EOF if body does not have an xml
+		if err != nil && err != io.EOF {
+			r.Error = awserr.New(aws.ErrCodeSerialization, "Failed to decode response body with invalid XML", err)
+			return
+		}
```
jasdel: Don't want to attempt to parse the response body as XML unless the `isRequestError` returns true. Attempting to parse the body too soon like here will incorrectly result in fail to unmarshal XML errors.


Creating the decoder and getting first token should be moved within the `isREquestError` check below. This will also remove the need for optional nil pointers accessible at the top level of this method.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,109 @@
...
+
+// unmarshalOperation is the top level method used with a handler stack to unmarshal an operation response
+// This method calls appropriate unmarshal shape functions as per the output shape and protocol used by the
+// service.
+func (u protoGetObjectUnmarshaler) unmarshalOperation(r *aws.Request) {
+	// isRequestError checks if operation response returned a error response
+	if isRequestError(r) {
+		// if startToken is nil, it would mean there is no xml response body
+		if r.HTTPResponse.Body == nil {
+			// fall back to status code error message
```
jasdel: Probably want to ensure that `r.HTTPResponse.Body.Close` is called before this method returns.

jasdel: It would be helpful to move the content of the error case block into its own function so won't have be be duplicated as much.

skotambkar: Modified and used `getStatusCodeErrorMessage()` instead . We can consider putting this method as an aws util as it is pretty standard.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// unmarshalOperation is the top level method used with a handler stack to unmarshal an operation response
+// This method calls appropriate unmarshal shape functions as per the output shape and protocol used by the service.
+func (u protoGetObjectUnmarshaler) unmarshalOperation(r *aws.Request) {
+	defer r.HTTPResponse.Body.Close()
+	// isRequestError checks if operation response returned a error response
+	if isRequestError(r) {
+		// if startToken is nil, it would mean there is no xml response body
+		if r.HTTPResponse.Body == nil {
```
jasdel: I think `Body` here will always not be nil.  Can use `EOF` error when `decoder.Token()` is called to determine there is no XML error message.

jasdel: Can investigate what to do if Body is nil. This would only happen if the request was not sent at all, but unmarshal was called for some reason. This probably isn't a case we want to handle.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,114 @@
...
+
+// unmarshalError unmarshal's the error response
+func unmarshalErrorPrototype(r *aws.Request, d *xml.Decoder, startToken xml.Token) error {
+	// protoXMLErrorResponse is error response struct for xml errors.
+	var respErr = protoXMLErrorResponse{}
+
+	// Delegate to reflection based decoding utils
+	if start, ok := startToken.(xml.StartElement); ok {
+		err := d.DecodeElement(&respErr, &start)
+		if err != nil && err != io.EOF {
```
jasdel: would be great for this to take a `StartElement` type instead of `Token`. The caller should be responsible for getting the first StartElement. This is helpful because the first Token might be something like the XML preamble.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	} else {
+		return awserr.New(aws.ErrCodeSerialization, "Serialization error: Failed to unmarshal invalid xml", nil)
+	}
+
+	reqID := respErr.RequestID
+	if len(reqID) == 0 {
+		reqID = r.RequestID
+	}
+	return awserr.NewRequestFailure(awserr.New(respErr.Code, respErr.Message, nil), r.HTTPResponse.StatusCode, reqID)
```
jasdel: Nit would be helpful to add a comment about this not extracting the modeled errors, and that will handled in another prototype for generated error unmarshaling.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,114 @@
...
+type protoXMLErrorResponse struct {
+	Code      string `xml:"Code"`
+	Message   string `xml:"Message"`
+	RequestID string `xml:"RequestId"`
+}
+
+// unmarshalError unmarshal's the error response
+func unmarshalErrorPrototype(r *aws.Request, d *xml.Decoder, startToken xml.Token) error {
+	// protoXMLErrorResponse is error response struct for xml errors.
+	var respErr = protoXMLErrorResponse{}
```
jasdel: Would be good to callout via comment that services can have customizations like S3, that require custom error unmarshaling.

service/s3/unmarshal.go
```diff
@@ -0,0 +1,114 @@
...
+		ringBuff := sdkio.NewRingBuffer(buff)
+		body := io.TeeReader(r.HTTPResponse.Body, ringBuff)
+		decoder := xml.NewDecoder(body)
+		startToken, err := decoder.Token()
+		if err != nil {
+			r.Error = awserr.New(aws.ErrCodeSerialization, "Failed to decode response body with invalid XML", err)
+			return
+		}
+
+		r.Error = unmarshalErrorPrototype(r, decoder, startToken)
```
jasdel: Serialization errors encountered here, will not include the ring buffer, but should.

service/s3/unmarshal.go
jasdel: Nit this should use the S3 specific `RequestFailure` wrapper so the `HostID` value in the error response can be captured.

PR #471, opened 2020-01-07
`private/model`: Add shape name validation for structs and enums

skotambkar:
* Adds `validateShapeName` utility to validate that the shape names for structs and enums adhere to a defined format. 
* Fixes bug which allowed service package structs, enums to start with non alphabetic character 
* Fixes the incorrect enum types in mediapackage service package, changing enum types __AdTriggersElement, __PeriodTriggersElement to AdTriggersElement, PeriodTriggersElement respectively.
* Adds unit tests to test the validateShapeName utility.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,15 +1,16 @@
 Breaking Change
 ---
-* Add generated service for wafregional and dynamodbstreams ([#463](https://github.com/aws/aws-sdk-go-v2/pull/463))
+* `service`: Add generated service for wafregional and dynamodbstreams #463
   * Updates the wafregional and dynamodbstreams API clients to include all API operations, and types that were previously shared between waf and dynamodb API clients respectively. This update ensures that all API clients include all operations and types needed for that client, and shares no types with another client package.
   * To migrate your applications to use the updated wafregional and dynamodbstreams you'll need to update the package the impacted type is imported from to match the client the type is being used with.
-* Add utility for validating shape names for structs and enums for the service packages ([#471](https://github.com/aws/aws-sdk-go-v2/pull/471))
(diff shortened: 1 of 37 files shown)


CHANGELOG_PENDING.md
```diff
@@ -4,6 +4,10 @@ Breaking Change
   * Updates the wafregional and dynamodbstreams API clients to include all API operations, and types that were previously shared between waf and dynamodb API clients respectively. This update ensures that all API clients include all operations and types needed for that client, and shares no types with another client package.
   * To migrate your applications to use the updated wafregional and dynamodbstreams you'll need to update the package the impacted type is imported from to match the client the type is being used with.
 
+* Add utility for validating shape names for structs and enums for the service packages.
+  * Fixes bug which allowed service package structs, enums to start with non alphabetic character 
+  * Fixes the incorrect enum types in mediapackage service package, changing enum types __AdTriggersElement, __PeriodTriggersElement to AdTriggersElement, PeriodTriggersElement respectively.
```
jasdel: nit extra new line, and would be good to refer to the PR

private/model/api/load.go
```diff
@@ -172,6 +173,9 @@ func (a *API) Setup() {
 	}
 
 	a.fixStutterNames()
+	if err := a.validateShapeNames(); err != nil {
+		log.Fatalf(err.Error())
+	}
```
jasdel: It would be helpful for this `log.Fatalf` to include the name of the API that failing.

skotambkar: The error was already wrapped with shape name, added package name to error. 
```go
return fmt.Errorf("attempted to rename shape %v to %v for package %v, but the new name results in shape name collision",s.ShapeName, name, a.PackageName())
```

PR #472, opened 2020-01-13
Modifies REST V2 encoder to not lower-case header names

skmcgrail:
Modifies the REST v2 encoder to not lower-case values, but use the provided Set/Add headers which will canonicalize the headers.

### DIFF
--- a/aws/protocol/rest/encode_test.go
+++ b/aws/protocol/rest/encode_test.go
@@ -21,8 +21,8 @@ func TestEncoder(t *testing.T) {
 	expected := http.Request{
 		Header: map[string][]string{
 			"custom-user-header": {"someValue"},
-			"x-amzn-header-foo":  {"someValue"},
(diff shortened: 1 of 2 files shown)


jasdel:
Tests failing due to not being in sync with casing change.

skmcgrail:
> Tests failing due to not being in sync with casing change.

Fixed

aws/protocol/rest/header.go
```diff
@@ -43,17 +43,11 @@ func newHeaderValue(header http.Header, key string, append bool) HeaderValue {
 }
```
skotambkar: Do we need `newHeaderValue` constructor? It isn't exported and we can just use `HeaderValue` directly on line 31.

skmcgrail: This was just to keep construction of the header consistent as the structure is created in a number of spots.

PR #473, opened 2020-01-15
aws.Endpoints Usage Refactor for aws.Request and aws.Metadata

skmcgrail:
This PR is the refactor work to the aws.Request structure to store and utilize the resolved Endpoint structure from the EndpointResolver. This included updates to the usage of the aws.Metadata and the SigningName, SigningRegion, and Endpoint which were originally stored there.

### DIFF
--- a/aws/request_test.go
+++ b/aws/request_test.go
@@ -1157,7 +1157,7 @@ func TestRequestEndpointConstruction(t *testing.T) {
 		ExpectedEndpoint aws.Endpoint
 	}{
 		"resolved modeled endpoint": {
-			EndpointResolver: func(_, _ string) (endpoint aws.Endpoint, err error) {
+			EndpointResolver: func(_, _ string) (aws.Endpoint, error) {
 				return aws.Endpoint{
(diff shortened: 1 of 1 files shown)


aws/request.go
```diff
@@ -111,12 +112,8 @@ func New(cfg Config, metadata Metadata, handlers Handlers,
 
 	if err == nil {
 		// TODO so ugly
-		metadata.Endpoint = endpoint.URL
-		if len(endpoint.SigningName) > 0 && !endpoint.SigningNameDerived {
-			metadata.SigningName = endpoint.SigningName
-		}
-		if len(endpoint.SigningRegion) > 0 {
-			metadata.SigningRegion = endpoint.SigningRegion
```
skmcgrail: Looking at this again, with this change, it should probably have a similar check as the new code I wrote below which checks to see if the returned endpoint had an empty signing region, and fallback to the metadata signing region. Metadata signing region would be initialized with the aws.Config.Region as part of client construction.

PR #474, opened 2020-01-16
Support for Loading Service Specific Configuration from External Sources During Client Creation

skmcgrail:
Adds support for loading service specific configuration options from external sources during client creation.

Remaining Items:
[X] Unit-test generation of generated service configuration resolver

### DIFF
--- a/service/s3/api_client.go
+++ b/service/s3/api_client.go
@@ -98,7 +98,7 @@ func New(config aws.Config) *Client {
 	}
 
 	if config.AdditionalConfig != nil {
-		if err := config.AdditionalConfig.ResolveConfig(resolveServiceConfig(svc)); err != nil {
+		if err := config.AdditionalConfig.ResolveConfig(resolveClientConfig(svc)); err != nil {
 			panic(fmt.Errorf("failed to resolve service configuration: %v", err))
 		}
 	}
--- a/service/s3/api_client_config.go
+++ b/service/s3/api_client_config.go
@@ -6,7 +6,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3/internal/external"
 )
 
-func resolveServiceConfig(svc *Client) func(configs []interface{}) error {
+func resolveClientConfig(svc *Client) func(configs []interface{}) error {
 	return func(configs []interface{}) error {
 		if value, ok, err := external.ResolveUseARNRegion(configs); err != nil {
 			return err
--- a/service/s3/api_client_config_test.go
+++ b/service/s3/api_client_config_test.go
@@ -18,7 +18,7 @@ func TestExternalConfigResolver(t *testing.T) {
 	t.Run("UseARNRegion", func(t *testing.T) {
 		t.Run("value not found", func(t *testing.T) {
 			svc := &Client{}
-			err := resolveServiceConfig(svc)([]interface{}{
+			err := resolveClientConfig(svc)([]interface{}{
 				mockUseARNRegion(func() (value bool, ok bool, err error) {
 					return value, ok, err
 				}),
@@ -32,7 +32,7 @@ func TestExternalConfigResolver(t *testing.T) {
 		})
 		t.Run("resolve error", func(t *testing.T) {
 			svc := &Client{}
-			err := resolveServiceConfig(svc)([]interface{}{
+			err := resolveClientConfig(svc)([]interface{}{
 				mockUseARNRegion(func() (value bool, ok bool, err error) {
 					err = fmt.Errorf("resolve error")
 					return value, ok, err
@@ -47,9 +47,12 @@ func TestExternalConfigResolver(t *testing.T) {
 		})
 		t.Run("value found", func(t *testing.T) {
 			svc := &Client{}
-			err := resolveServiceConfig(svc)([]interface{}{
+			err := resolveClientConfig(svc)([]interface{}{
 				mockUseARNRegion(func() (value bool, ok bool, err error) {
 					value = true
+					return value, true, err
+				}),
+				mockUseARNRegion(func() (value bool, ok bool, err error) {
 					return value, true, err
 				}),
 			})


private/model/api/api.go
```diff
@@ -694,6 +708,10 @@ func (a *API) ServiceGoCode() string {
 	}
 	a.AddSDKImport("private/protocol", a.ProtocolPackage())
 
+	if a.HasServiceClientConfig() {
+		a.AddSDKImport("service", a.PackageName(), a.ConfigPackageName())
+	}
+
```
skotambkar: Nit: Conditional AddSDKImport probably can be a part of the template.

skotambkar:
Were generated config.go files intentionally not added in this PR?

private/model/api/api.go
skmcgrail: This is included in the most recent commit.

skmcgrail: It could be, but I don't agree with adding unnecessarily adding complexity to the cognitive load of understanding the template for a single import.

aws/config.go
```diff
@@ -78,6 +78,17 @@ type Config struct {
...
+	// clients.
+	ExternalConfig ExternalConfig
+}
+
+// ExternalConfig is an interface that encapsulates the behavior of loading
+// additional configuration from an external source.
+type ExternalConfig interface {
+	// ResolveConfig calls the provide function passing slice of configuration sources
+	ResolveConfig(func(configs []interface{}) error) error
 }
```
jasdel: Does it make sense to rename this field to be more generic to additional config, instead of specific to `External`?   Even though that is what we plan to use it for.

aws/external/config.go
```diff
@@ -100,6 +100,15 @@ func (cs Configs) ResolveAWSConfig(resolvers []AWSConfigResolver) (aws.Config, e
 	return cfg, nil
 }
 
+// ResolveConfig calls the provide function passing slice of configuration sources
+func (cs Configs) ResolveConfig(f func(configs []interface{}) error) error {
```
jasdel: Might be helpful to document the interface this intends to implement.

jasdel: Does `ResolveConfig` need to be a separate method from `ResolveAWSConfig?  They basically do the same thing, with the exception, `ResolveAWSConfig` takes multiple resolvers where `ResolveConfig` only takes one. Does it make sense for `ResolveAWSConfigs` to be updated to take variadic arguments, instead of adding a new method?

aws/external/provider.go
```diff
@@ -343,3 +343,13 @@ func NewWithEC2MetadataRegion(ctx context.Context, client *ec2metadata.Client) W
...
+
+// WithS3UseARNRegion is used to specify whether to enable usage of the region specified
+// in an ARN passed as an S3 bucket value.
+type WithS3UseARNRegion bool
+
+// GetS3UseARNRegion returns whether to enable usage of the region specified
+// in an ARN passed as an S3 bucket value.
+func (w WithS3UseARNRegion) GetS3UseARNRegion() (bool, error) {
+	return bool(w), nil
+}
```
jasdel: Does this need to be exported in the `external` package or unexported within the s3 client package? This provider probably can be generated.

private/model/api/api.go
```diff
@@ -631,6 +667,12 @@ func New(config aws.Config) *{{ .StructName }} {
     	),
     }
 
+	{{ if .HasExternalServiceConfigFields }}
+		if config.ExternalConfig != nil {
+			config.ExternalConfig.ResolveConfig(resolveServiceConfig(svc))
```
jasdel: even though not best place to use it, should capture error here, for when we add client builders.

private/model/api/api.go
```diff
@@ -666,13 +737,29 @@ func (c *{{ .StructName }}) newRequest(op *aws.Operation, params, data interface
...
+}
+{{- end }}
 `))
 
 // ServicePackageDoc generates the contents of the doc file for the service.
 //
 // Will also read in the custom doc templates for the service if found.
 func (a *API) ServicePackageDoc() string {
-	a.imports = map[string]bool{}
+	a.imports = map[packageImport]bool{}
```
jasdel: This could be `a.resetImports()` instead

private/model/api/api.go
```diff
@@ -704,6 +796,72 @@ func (a *API) ServiceGoCode() string {
 	return code
 }
 
+// ServiceTestGoCode generates service client specific unit-tests
+func (a *API) ServiceTestGoCode() string {
+	// At this time we only need to generate client tests for external config resolution
+	if !a.HasExternalServiceConfigFields() {
```
jasdel: Nit should be `Client`TestGoCode :) or maybe `APIClientTestGoCode` if we're using that naming.

what do you think about putting all of these templates, and behavior into its own go file. e.g. `client_config.go` similar to `smoke.go` and `examples.go`.

service/s3/external/external.go
```diff
@@ -0,0 +1,9 @@
+// Code generated by private/model/cli/gen-api/main.go. DO NOT EDIT.
+
+package external
+
+type WithUseARNRegion bool
+
+func (v WithUseARNRegion) GetS3UseARNRegion() (value bool, ok bool, err error) {
+	return bool(v), true, nil
+}
```
jasdel: It might be better for the SDK not to generate this file at all. On second look at a sub package with just this file seems excessive.

private/model/api/api.go
```diff
@@ -983,3 +1121,112 @@ func (a *API) hasNonIOShapes() bool {
 	}
 	return false
 }
+
+var tplClientConfig = template.Must(template.New("tplClientConfig").Funcs(map[string]interface{}{
+	"ExternalConfigFields": externalConfigFields,
+}).Parse(`
+func resolveServiceConfig(svc *{{ .StructName }}) func(configs []interface{}) error {
```
jasdel: Nit, `Service` -> `Client`

private/model/api/service_specific_config.go
```diff
@@ -4,21 +4,53 @@ package api
...
 type serviceConfigFields []serviceConfigField
 
-func (fs serviceConfigFields) GoCode() string {
+// ServiceClientGoCode returns rendered service configuration fields suitable for a structure definition
+func (fs serviceConfigFields) ServiceClientGoCode() string {
 	w := bytes.NewBuffer(nil)
 
-	if err := svcConfigTmpl.Execute(w, fs); err != nil {
+	if err := tplServiceConfigFields.Execute(w, fs); err != nil {
 		panic("failed to render serviceConfigFields " + err.Error())
```
jasdel: Nit `Service` -> `Client`

PR #476, opened 2020-01-24
service/s3: Adds Support for S3 AccessPoints

skmcgrail:
This ports the V1 SDK's support for S3 AccessPoints.

### DIFF
--- a/private/model/api/api.go
+++ b/private/model/api/api.go
@@ -324,13 +324,13 @@ var tplAPI = template.Must(template.New("api").Parse(`
 
 // AddImport adds the import path to the generated file's import.
 func (a *API) AddImport(v ...string) error {
-	a.imports[packageImport{Path: strings.Join(v, "/")}] = true
+	a.imports[packageImport{Path: path.Join(v...)}] = true
 	return nil
 }
 
 // AddImportWithAlias adds the import path with the given alias to the generated file's import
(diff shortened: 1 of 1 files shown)


private/model/api/api.go
```diff
@@ -395,6 +395,11 @@ func (a *API) APIOperationGoCode(op *Operation) string {
 		a.AddSDKImport("private/protocol")
 	}
 
+	if op.HasEndpointARN {
+		a.AddImport("fmt")
+		a.AddSDKImport("service", a.PackageName(), "internal", "arn")
```
jasdel: this should use the `a.ImportPath()` instead of `service, a.PackageName` with `AddImport`. The `service` component can be passed into the `gen-api` utility as a cli command.

private/model/api/passes.go
```diff
@@ -36,9 +36,22 @@ func (a *API) updateTopLevelShapeReferences() {
 // writeShapeNames sets each shape's API and shape name values. Binding the
 // shape to its parent API.
 func (a *API) writeShapeNames() {
+	writeOrigShapeName := func(s *ShapeRef) {
+		if len(s.ShapeName) > 0 {
+			s.OrigShapeName = s.ShapeName
+		}
+	}
```
jasdel: Can this be a static function? Doesn't look like its use a closure.

PR #477, opened 2020-01-24
Add timestamp support for SDK's generated API smoke tests

jasdel:
Adds support for `timestamp` API parameters  in generated smoke tests. Enables the integration test model for ``.

Update also removes unneeded dependencies from generated  API `examples_test.go`, and refactors how Time is represented within the examples.

Fixes SDK's example generation of enum values to use the generated enum value constant instead of literal string.

### DIFF
--- a/private/model/api/example.go
+++ b/private/model/api/example.go
@@ -24,7 +24,7 @@ func (a *API) ExamplesGoCode() string {
 	a.AddSDKImport("aws")
 	a.AddSDKImport("aws/awserr")
 	a.AddSDKImport("aws/external")
-	a.AddSDKClientImport()
+	a.AddImport(a.ImportPath())
 
 	var extra = `
 var _ aws.Config
--- a/private/model/api/api.go
+++ b/private/model/api/api.go
@@ -356,17 +356,6 @@ func (a *API) AddSDKImportWithAlias(alias string, sdkPath ...string) error {
 	return nil
 }
 
-// AddSDKClientImport adds the import path for the generate API client's path,
-// and any sub package specified.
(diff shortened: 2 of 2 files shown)


private/model/api/eventstream.go
```diff
@@ -2,27 +2,43 @@
...
+		inputRef := getEventStreamMember(op.InputRef.Shape)
+		outputRef := getEventStreamMember(op.OutputRef.Shape)
 
-		if !(outbound || inbound) {
+		if inputRef == nil && outputRef == nil {
 			continue
 		}
 
-		a.removeOperation(name)
+		err := fmt.Errorf("eventstream support not implemented, %s, %s",
```
skotambkar: Is'%s' extra here?

private/model/api/eventstream.go
```diff
@@ -2,27 +2,43 @@
...
+			op.ExportedName)
```
skotambkar: Nit: one extra `%s` here ?

private/model/api/example.go
```diff
@@ -14,7 +14,41 @@ import (
...
+// examples.json file.
+func (a *API) ExamplesGoCode() string {
+	a.resetImports()
+	a.AddImport("context")
+	a.AddImport("fmt")
+
+	a.AddSDKImport("aws")
+	a.AddSDKImport("aws/awserr")
+	a.AddSDKImport("aws/external")
+	a.AddSDKClientImport()
```
skotambkar: We could have used `a.AddImport(a.ImportPath())` here instead of a.AddSDKClientImport()

This change looks good, but feels like we have lot of ways to Import and it may get confusing and result in different usages in the SDK for similar outcome.

PR #478, opened 2020-01-27
Simplify Service Client Configuration Resolving

skmcgrail:
Further simplifies the resolution of service specific client configuration from external sources.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -79,16 +79,9 @@ type Config struct {
 	// for testing that do not support the modeled host prefix pattern.
 	DisableEndpointHostPrefix bool
 
-	// ConfigSources are the sources that were used to the aws.Config.
+	// ConfigSources are the sources that were used to construct the Config.
 	// Allows for additional configuration can be loaded by clients.
 	ConfigSources []interface{}
-}
-
-// ConfigResolver is an interface that encapsulates the behavior of loading
-// additional configuration.
-type ConfigResolver interface {
-	// ResolveConfig calls the provide function passing a slice of configuration sources
-	ResolveConfig(func(configs []interface{}) error) error
 }
 
 // NewConfig returns a new Config pointer that can be chained with builder


aws/external/config_test.go
```diff
@@ -101,4 +103,13 @@ func TestConfigs_ResolveAWSConfig(t *testing.T) {
...
 		t.Errorf("expect %v provider, got %v", e, a)
 	}
+
+	var expectedSources []interface{}
+	for _, s := range cfg.ConfigSources {
+		expectedSources = append(expectedSources, s)
+	}
+
+	if e, a := expectedSources, cfg.ConfigSources; !reflect.DeepEqual(e, a) {
+		t.Errorf("expected %v, got %v", e, a)
```
skotambkar: QQ: When to use `reflect.DeepEqual` vs `cmp.Diff`?

I thought cmp.Diff was preferred, but for []interface not sure

private/model/api/api.go
```diff
@@ -699,10 +699,8 @@ func New(config aws.Config) *{{ .StructName }} {
     }
 
 	{{ if .HasExternalClientConfigFields }}
-		if config.AdditionalConfig != nil {
-			if err := config.AdditionalConfig.ResolveConfig(resolveClientConfig(svc)); err != nil {
-				panic(fmt.Errorf("failed to resolve service configuration: %v", err))
-			}
+		if err := resolveClientConfig(svc, config.ConfigSources); err != nil {
+			panic(fmt.Errorf("failed to resolve service configuration: %v", err))
```
skotambkar: Should we use log.Fatalf here instead of panic?

skmcgrail: This solution is only temporary until we implement the client builder pattern. Today the client.New(...) operation does not have the ability to return an error. So we opt for panic in this case.

aws/external/config_test.go
skmcgrail: I think this is a much bigger discussion we can take offline, but in essence cmp.Diff has much more granular comparisons, for example know that two timestamps are equal across timezones. Additionally it has the ability to set options like ignoring/not-ignoring unexported structure fields. In this instance of comparison the need for using cmp.Diff is unwarranted.

PR #483, opened 2020-02-02
s3/s3manager: Reduce of memory allocations & GC runs for S3 downloader

### LINKED ISSUE #482: service/s3/s3manager: Reduce downloader memory allocations & GC runs
### Version of AWS SDK for Go?
v0.19.0

go version go1.13.4 darwin/amd64

### What issue did you see?
in case the user of the library doesn't create a buffer with a particular size and pass it to the S3 `downloader`, it downloads file part-by-part from 1st byte up to file length. Thus `WriteAtBuffer` increase buffer size based on the current max downloaded part, but we know file size at the start after this code:

https://github.com/aws/aws-sdk-go-v2/blob/master/service/s3/s3manager/download.go#L314
```go
d.getChunk()
```

so I propose download file backward => from the end of the file, so we will allocate the right memory at the start as the last part of the file will be downloaded at the beginning of the process and then buffer never be reallocated. It gives good optimization and GC doesn't work during the whole process, at the beginning only.

Before pull request:
image

After pull request:
image

[trace_2_bad_gc.txt](https://github.com/aws/aws-sdk-go-v2/files/4146577/trace_2_bad_gc.txt)
[trace_2_good_gc.txt](https://github.com/aws/aws-sdk-go-v2/files/4146578/trace_2_good_gc.txt)

### Steps to reproduce
My sample code:

```go
package main

import (
	"context"
	"fmt"

	"github.com/aws/aws-sdk-go-v2/service/s3/s3manager"
)

func HandleDemoRecord(bucket, region, item string) error {
	fmt.Printf("Bucket = %s, Key = %s\n", bucket, item)

	config, err := external.LoadDefaultAWSConfig()
	if err != nil {
		return err
	}
	config.Region = region

	headObject, err := getHeadObject(bucket, item, config)
	if err != nil {
		return err
	}

	fmt.Printf("obj: %v\nbytes:%+v\n", headObject, *headObject.ContentLength)

	downloader := s3manager.NewDownloader(config, func(d *s3manager.Downloader) {
		d.Concurrency = 5
	})

	var buf []byte
	// allocate before
	//buf := make([]byte, int(*headObject.ContentLength))

	fmt.Println("Starting download of file")
	buffer := aws.NewWriteAtBuffer(buf)
	numBytesDownloaded, err := downloader.Download(buffer, &s3.GetObjectInput{
		Bucket: aws.String(bucket),
		Key:    aws.String(item),
	})

	if err != nil {
		return err
	}

	fmt.Println("Downloaded ", numBytesDownloaded, " bytes")

	return nil
}

func getHeadObject(bucket string, key string, config aws.Config) (*s3.HeadObjectResponse, error) {
	client := s3.New(config)

	input := &s3.HeadObjectInput{
		Bucket: aws.String(bucket),
		Key:    aws.String(key),
	}

	request := client.HeadObjectRequest(input)

	headObjectResponse, err := request.Send(context.TODO())
	if err != nil {
		return nil, err
	}

	fmt.Printf("Downloaded: %v\n", request)

	return headObjectResponse, nil
}

func main() {
	err := HandleDemoRecord("vvelikodny-s3-upload", "ca-central-1", "mysql-workbench-community-8.0.18-macos-x86_64.dmg")
	if err != nil {
		panic(err)
	}
}
```

here I investigated also memory issues with huge allocations on start, but think it should be in the next issue/pull request.

vvelikodny:
Download parts of the file from the end. Reduce the count of mem allocation & GC runs.

fix #482 

see aws/aws-sdk-go#3085

### DIFF
--- a/service/s3/s3manager/download.go
+++ b/service/s3/s3manager/download.go
@@ -42,7 +42,7 @@ func (e *errReadingBody) Unwrap() error {
 // Mutating the Downloader's properties is not safe to be done concurrently.
 type Downloader struct {
 	// The size (in bytes) to request from S3 for each part.
-	// The minimum allowed part size is 5MB, and  if this value is set to zero,
+	// The minimum allowed part size is 5MB, and if this value is set to zero,
 	// the DefaultDownloadPartSize value will be used.
 	//
 	// PartSize is ignored if the Range input parameter is provided.
@@ -322,15 +322,20 @@ func (d *downloader) download() (n int64, err error) {
 			go d.downloadPart(ch)
 		}
 
+		tailLen := d.cfg.PartSize
+		if tail := total % d.cfg.PartSize; tail > 0 {
+			tailLen = tail
+		}
+		d.pos = total - tailLen
+
 		// Assign work
 		for d.getErr() == nil {
-			if d.pos >= total {
+			if d.pos < d.cfg.PartSize {
 				break // We're finished queuing chunks
 			}
-
 			// Queue the next range of bytes to read.
 			ch <- dlchunk{w: d.w, start: d.pos, size: d.cfg.PartSize}
-			d.pos += d.cfg.PartSize
+			d.pos -= d.cfg.PartSize
 		}
 
 		// Wait for completion
--- a/service/s3/s3manager/download_test.go
+++ b/service/s3/s3manager/download_test.go
@@ -220,7 +220,7 @@ func TestDownloadOrder(t *testing.T) {
 		t.Errorf("expect %v API calls, got %v", e, a)
 	}
 
-	expectRngs := []string{"bytes=0-5242879", "bytes=5242880-10485759", "bytes=10485760-15728639"}
+	expectRngs := []string{"bytes=0-5242879", "bytes=10485760-15728639", "bytes=5242880-10485759"}
 	if e, a := expectRngs, *ranges; !reflect.DeepEqual(e, a) {
 		t.Errorf("expect %v ranges, got %v", e, a)
 	}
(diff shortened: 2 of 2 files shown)


skmcgrail:
Hello @vvelikodny, thank you for reaching out to us and putting together a pull request for improving the S3 transfer manager that is included with the SDK. We are looking to make some significant changes to the S3 Upload and Download manager offered prior to the general availability of the SDK, and as such I think it might be best to hold off this particular change and potentially include it as part of our refactor work. 

In the interim, to decrease the overall total allocations that you are experiencing, we recommend that you preallocate that the `WriteAtBuffer` that you are passing to the download manager by passing in an already allocated byte slice using the `NewWriteAtBuffer(buf []byte)` function provided in the `aws` package.

vvelikodny:
> Hello @vvelikodny, thank you for reaching out to us and putting together a pull request for improving the S3 transfer manager that is included with the SDK. We are looking to make some significant changes to the S3 Upload and Download manager offered prior to the general availability of the SDK, and as such I think it might be best to hold off this particular change and potentially include it as part of our refactor work.
> 

ok, thanks!

> In the interim, to decrease the overall total allocations that you are experiencing, we recommend that you preallocate that the `WriteAtBuffer` that you are passing to the download manager by passing in an already allocated byte slice using the `NewWriteAtBuffer(buf []byte)` function provided in the `aws` package.

right, in the case when we allocate it at the start we don't have such issue. thank you for the answer! @skmcgrail

vvelikodny:
Hello, @skmcgrail! Just noticed still getting points on that issue on StackOverflow recently and the downloader works well only with a pre-allocated buffer, although changes could fix it and weren't implemented in new SDK versions. Lambda function with v.1.2.1 still eat well memory if the client doesn't pass proper buffer.

vvelikodny:
Hello, @skmcgrail! An additional interesting thing I discovered during the tests:
`Billed Duration: 24524 ms` without changes vs `Billed Duration: 510 ms` with changes

it seems could affect the costs of Lambda

PR #484, opened 2020-02-03
[V1-Sync] Port Endpoint Discovery

skmcgrail:
Ports V1 SDK's support for performing service endpoint discovery.


aws/crr/sync_map.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"sync"
+)
+
+type syncMap sync.Map
+
+func newSyncMap() syncMap {
+	return syncMap{}
+}
+
```
jasdel: Is this file/type needed anymore?  Since SDK's min version includes `sync.Map` it may not be.

aws/crr/sync_map_test.go
```diff
@@ -0,0 +1,110 @@
+package crr
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestRangeDelete(t *testing.T) {
```
jasdel: Probably not needed if `sync_map.go` isn't needed.

aws/external/shared_config.go
```diff
@@ -105,6 +108,21 @@ type SharedConfig struct {
...
+
+	// EnableEndpointDiscovery can be enabled in the shared config by setting
+	// endpoint_discovery_enabled to true
+	//
+	//	endpoint_discovery_enabled = true
+	EnableEndpointDiscovery *bool
+}
+
+// GetEnableEndpointDiscovery returns whether to enable service endpoint discovery
+func (c *SharedConfig) GetEnableEndpointDiscovery() (value, ok bool, err error) {
```
jasdel: Will be helper to ensure the `SharedConfig` and `EnvConfig` can satisfy the `EnableEndpointDiscoveryProvider` interface via test type assertion.

Issue #272, opened 2019-03-04
SDK should be able to correct for clock skew


jasdel:
The SDK should be able to correct for clock skew between client's time and server.

V1 Issue: aws/aws-sdk-go#2012

PR #487, opened 2020-02-06
Refactor SDK's request retry, and error handling behavior

jasdel:
This update refactors the SDK's request retry behavior by updating the SDK's implementation to a model that will be the standard across the SDKs. This model simplifies retry behavior, and will retry all connection errors. Not just write connection reset errors.

This update also refactors the SDK's error wrapping and handling. The SDK's error types such as awserr.Error implement the Unwrap method allowing the nested errors to be more easily extracted using the Go 1.13 errors.As, Is, and Unwrap functions.

This update also provides more information about why a request failed, and was retried, but exhausted attempts. The aws.MaxAttemptsError will wrap the connection or service API error.

TODO:
* [x] Update SDK's retry and error tests
* [x] Add changelog
* [x] Add missing/commented out tests to retry package

### DIFF
--- a/aws/request.go
+++ b/aws/request.go
@@ -516,10 +516,10 @@ func (r *Request) Send() error {
 		return err
 	}
 
-	// TODO (jasdel) Refactoring request error handling needs to consider the
-	// difference between API, connection, and SDK behavior errors. This
-	// consideration must include wrapping of underlying error when any SDK
-	// error occurs.
+	// TODO (jasdel), Issue #74 - Refactoring request error handling needs to
+	// consider the difference between API, connection, and SDK behavior
+	// errors. This consideration must include wrapping of underlying error
+	// when any SDK error occurs.
 
 	relRetryToken := r.Retryer.GetInitialToken()
 	for {


jasdel:
Two commints, first includes the hand written code changes, second commit is all code generation.

jasdel:
Expect Go 1.11 and Go 1.12 builds to fail due to `errors.As` and other functions added in Go 1.13

aws/defaults/handlers.go
```diff
@@ -182,6 +183,35 @@ var ValidateResponseHandler = aws.NamedHandler{
 		}
 	}}
 
+// RequestInvocationIDHeaderHandler sets the invocation id header for request
+// tracking across attempts.
+var RequestInvocationIDHeaderHandler = aws.NamedHandler{
+	Name: "core.RequestInvocationIDHeaderHandler",
+	Fn: func(r *aws.Request) {
+		const invocationIDHeader = "amz-sdk-invocation-id"
+		r.HTTPRequest.Header.Set(invocationIDHeader, r.InvocationID)
```
skotambkar: scope of `const invocationIDHeader` will be limited to the named Handler func; Is that intended/needed?

jasdel: Yeah, the value didn't really need to be a constant, and could of been in-lined. Didn't want it to be exported.

aws/defaults/handlers.go
```diff
@@ -182,6 +183,35 @@ var ValidateResponseHandler = aws.NamedHandler{
 		}
 	}}
 
+// RequestInvocationIDHeaderHandler sets the invocation id header for request
+// tracking across attempts.
+var RequestInvocationIDHeaderHandler = aws.NamedHandler{
+	Name: "core.RequestInvocationIDHeaderHandler",
```
skotambkar: Should `Name` for a NamedHandler be an exported constant? In this case, should the `core.RequestInvocationIDHeaderHandler` be exported constant of defaults package?

This would help swap the handlers in case we need to.

aws/defaults/handlers.go
```diff
@@ -182,6 +183,35 @@ var ValidateResponseHandler = aws.NamedHandler{
...
+	Fn: func(r *aws.Request) {
+		const invocationIDHeader = "amz-sdk-invocation-id"
+		r.HTTPRequest.Header.Set(invocationIDHeader, r.InvocationID)
+	}}
+
+// RetryMetricHeaderHandler sets an additional header to the API request that
+// includes retry details for the service to consider.
+var RetryMetricHeaderHandler = aws.NamedHandler{
+	Name: "core.RetryMetricHeaderHandler",
```
skotambkar: Should `Name be exported constant`?

aws/defaults/handlers.go
jasdel: Since the variable is already exported the general practice has been doing `RequestInvocationIDHeaderHandler.Name` to get the named handler. Having a separate constant duplicates this, and doesn't add a lot of value.

aws/defaults/handlers.go
```diff
@@ -161,55 +162,84 @@ func handleSendError(r *aws.Request, err error) {
...
-				return
-			}
+		parts = append(parts, fmt.Sprintf("attempt=%d", r.AttemptNum))
+		if max := r.Retryer.MaxAttempts(); max != 0 {
+			parts = append(parts, fmt.Sprintf("max=%d", max))
+		}
+		// TODO "ttl=YYYYmmddTHHMMSSZ"
+		//    ttl = current_time + socket_read_timeout + estimated_skew
+		// SDK doesn't implement clock skew yet and not obvious how to obtain
+		// read deadlines.
```
skmcgrail: https://github.com/aws/aws-sdk-go-v2/issues/272 is the relevent issue to track this work, in case you want to throw that issue number in the PR

aws/ec2metadata/token_provider.go
```diff
@@ -48,19 +49,23 @@ func (t *tokenProvider) fetchTokenHandler(r *aws.Request) {
...
 			// Check if request timed out while waiting for response
-			if e, ok := requestFailureError.OrigErr().(awserr.Error); ok {
-				if e.Code() == "RequestError" {
-					atomic.StoreUint32(&t.disabled, 1)
-				}
+			var re *aws.RequestSendError
+			var ce *aws.RequestCanceledError
+			if errors.As(rf, &re) || errors.As(rf, &ce) {
+				atomic.StoreUint32(&t.disabled, 1)
 			}
```
skmcgrail: Just from a styling perspective, whats your thoughts on using As vs. Is in this case? Seems As is more for if you need to assign the error to that type and use it, versus Is being more of a check.

aws/ratelimit/token_bucket.go
```diff
@@ -0,0 +1,53 @@
...
+// specified.
+func NewTokenBucket(i uint) *TokenBucket {
+	return &TokenBucket{
+		capacity:    i,
+		maxCapacity: i,
+	}
+}
+
+// Retrieve attempts to reduce the available tokens by the amount requested. If
+// available tokens are already negative, or the retrieve would make available
```
skmcgrail: np: Wording that available tokens are already negative, technically they can't be negative since the type is uint.

aws/ratelimit/token_rate_limit.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func isTimeoutError(error) bool {
+	return false
+}
+
+// GetToken may cause a available pool of retry quota to be
+// decremented. Will return an error if the decremented value can not be
+// reduced from the retry quota.
+func (l *TokenRateLimit) GetToken(ctx context.Context, cost uint) (func() error, error) {
```
skmcgrail: Should we check context here to make sure its not canaceled?

aws/ec2metadata/token_provider.go
jasdel: `Is` works great for equality check, but they must be exactly equal. Where as `As` works good for an error that is like, or of type of.  `Is` is more helpful for error value comparisons, not so much for type/interfaces.

aws/retry/retry.go
```diff
@@ -0,0 +1,99 @@
...
+	// request. Will return error if the if the delay could not be determined.
+	RetryDelay(attempt int, opErr error) (time.Duration, error)
+
+	// GetRetryToken attempts to deduct the retry cost from the retry token pool.
+	// Returning the token release function, or error.
+	GetRetryToken(ctx context.Context, opErr error) (releaseToken func(error) error, err error)
+
+	// GetInitalToken returns the initial request token that can increment the
+	// retry token pool if the request is successful.
+	GetInitialToken() func(error) error
```
skmcgrail: Might be useful to also name the return paramaters here for documentation similar to GetRetryToken

aws/retry/retryable_error.go
```diff
@@ -0,0 +1,168 @@
...
+
+// IsErrorRetryable returns if the error is retryable if it satisfies the
+// Retryable interface, and returns if the attempt should be retried.
+func (RetryableOperationError) IsErrorRetryable(err error) aws.Ternary {
+	var v interface{ RetryableError() bool }
+
+	if !errors.As(err, &v) {
+		return aws.UnknownTernary
+	}
+
```
skmcgrail: As versus Is?

aws/retry/retryable_error.go
```diff
@@ -0,0 +1,168 @@
...
+	case errors.As(err, &tempErr):
+		// url.Error and net.OpError implement Temporary but have different
+		// expectations for when an error is or is not temporary.
+		conErrVal = tempErr.Temporary()
+
+	default:
+		if strings.Contains(err.Error(), "connection reset") {
+			return aws.TrueTernary
+		}
+		return aws.UnknownTernary
```
skmcgrail: Is there cases where this happens and it is not a net.OpErrror?

aws/retry/retryable_error_test.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func TestRetryHTTPStatusCodes(t *testing.T) {
+}
+
+//func TestIsSerializationErrorRetryable(t *testing.T) {
+//	testCases := []struct {
+//		err      error
+//		expected bool
+//	}{
```
skmcgrail: Did you mean to remove these commented test cases, or were you going to follow-up with additional ones?

jasdel: these are cases copied from the old retry behavior. Commented our at while i port them over to this new logic.

aws/retry/standard.go
```diff
@@ -0,0 +1,210 @@
...
+	return releaseToken(s.incrementTokens).release
+}
+
+func (s *Standard) incrementTokens() error {
+	return s.options.RateLimiter.AddTokens(s.options.NoRetryIncrement)
+}
+
+// GetRetryToken attempts to deduct the retry cost from the retry token pool.
+// Returning the token release function, or error.
+func (s *Standard) GetRetryToken(ctx context.Context, err error) (func(error) error, error) {
```
skmcgrail: Should context be checked here?

aws/retry/retryable_error.go
jasdel: yeah this is more to cover the bases, where anything that contains connection reset is considered a connection error. This is what the old logic did, but sprinkled in a few places through the checks. With this update the logic is centralized here. I have a few updates to this RetryConnectionError that I'll publish shortly.

aws/signer/v4/functional_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
-	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
```
skmcgrail: I've noticed this happening recently, is this a Go formater change? Is it noticing that we have a package that is called v4, but also might confusingly match import path versioning?

jasdel: :( `goimports` is doing this. I'm not positive why, but this is an artifact of that utility.

jasdel: I can revert this file though since there are no other changes here.

aws/retry/standard.go
jasdel: yeah, will add that check.

aws/ratelimit/token_rate_limit.go
jasdel: Added context check to the TokenRateLimit the Standard GetToken call will defer to this check.

aws/ratelimit/token_rate_limit_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+func TestTokenRateLimit(t *testing.T) {
+	type usage struct {
+		Cost      uint
+		Release   bool
+		Err       string
+		AddTokens uint
+	}
+
+	cases := map[string]struct {
```
skmcgrail: Might be good to add a test case here around checking if the context was canceled, just so it does not get removed in the future

aws/request.go
```diff
@@ -445,29 +516,57 @@ func (r *Request) Send() error {
 		return err
 	}
 
+	// TODO (jasdel) Refactoring request error handling needs to consider the
+	// difference between API, connection, and SDK behavior errors. This
+	// consideration must include wrapping of underlying error when any SDK
+	// error occurs.
+
```
skmcgrail: Consider making a Github issue to track this, and put the issue in the comment

jasdel: Thanks I can actually remove this comment. since the Unwrap makes the embedded errors available.

jasdel: Though #74 does cover refactoring the errors. Will update the GitHub issue to ensure the redesign includes that expectation.

PR #488, opened 2020-02-07
Port Shared Config Improvements & Credential Providers

skmcgrail:
Criteria for Success:
* [X] Shared Config Improvements from V1
* [X] Process Credential Provider
* [X] Web Identity Token Provider
* [X] Credential Chaining and Loading from Environment and Shared Config
* [X] Separate Credential Provider Configurable Options, Provide Functional Arguments 
* [x] Support Configuring Credential Providers created in the Credential Chain via functional arguments

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -6,6 +6,14 @@ Services
 
 SDK Features
 ---
+* `aws`: Add Support for additional credential providers and credential configuration chaining ([#488](https://github.com/aws/aws-sdk-go-v2/pull/488))
+  * `aws/processcreds`: Adds Support for the Process Credential Provider
+    * Fixes [#249](https://github.com/aws/aws-sdk-go-v2/issues/249)
+  * `aws/stscreds`: Adds Support for the Web Identity Credential Provider
+    * Fixes [#475](https://github.com/aws/aws-sdk-go-v2/issues/475)
+    * Fixes [#338](https://github.com/aws/aws-sdk-go-v2/issues/338)
+  * Adds Support for `credential_source`
+    * Fixes [#274](https://github.com/aws/aws-sdk-go-v2/issues/274)
 
 SDK Enhancements
 ---


aws/external/resolve_credentials.go
```diff
@@ -0,0 +1,404 @@
...
+var (
+	ecsContainerEndpoint = "http://169.254.170.2" // not constant to allow for swapping during unit-testing
+
+	// WebIdentityEmptyRoleARNErr will occur if 'AWS_WEB_IDENTITY_TOKEN_FILE' was set but
+	// 'AWS_ROLE_ARN' was not set.
+	WebIdentityEmptyRoleARNErr = awserr.New(stscreds.ErrCodeWebIdentity, "role ARN is not set", nil)
+
+	// WebIdentityEmptyTokenFilePathErr will occur if 'AWS_ROLE_ARN' was set but
+	// 'AWS_WEB_IDENTITY_TOKEN_FILE' was not set.
+	WebIdentityEmptyTokenFilePathErr = awserr.New(stscreds.ErrCodeWebIdentity, "token file path is not set", nil)
```
jasdel: Probably don't want to expose error values because we'll never be able to change them to errors that can be includes metadata about the filename that failed, or other attributes about the error that may be useful to return.

aws/external/resolve_credentials.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		return err
+	}
+
+	envConfig, sharedConfig, other := getAWSConfigSources(configs)
+
+	switch {
+	case sharedProfileSet:
+		err = resolveCredsFromProfile(cfg, envConfig, sharedConfig, other)
+	case envConfig.Credentials.HasKeys():
```
jasdel: Nit for readability it would be nice to have a trailing newline after each case block to make little easier to read the case sections.

aws/external/shared_config.go
```diff
@@ -18,27 +19,49 @@ const (
...
+// ErrSharedConfigSourceCollision will be returned if a section contains both
+// source_profile and credential_source
+var ErrSharedConfigSourceCollision = awserr.New(ErrCodeSharedConfig, "only source profile or credential source can be specified, not both", nil)
+
+// ErrSharedConfigECSContainerEnvVarEmpty will be returned if the environment
+// variables are empty and Environment was set as the credential source
+var ErrSharedConfigECSContainerEnvVarEmpty = awserr.New(ErrCodeSharedConfig, "EcsContainer was specified as the credential_source, but 'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI' was not set", nil)
+
+// ErrSharedConfigInvalidCredSource will be returned if an invalid credential source was provided
+var ErrSharedConfigInvalidCredSource = awserr.New(ErrCodeSharedConfig, "credential source values must be EcsContainer, Ec2InstanceMetadata, or Environment", nil)
```
jasdel: probably want to avoid defining package error values. We won't be able to add metadata to error values in the future.  Such as profile name, file, etc.

aws/processcreds/provider.go
```diff
@@ -0,0 +1,350 @@
...
+// executeCredentialProcess starts the credential process on the OS and
+// returns the results or an error.
+func (p *Provider) executeCredentialProcess(ctx context.Context) ([]byte, error) {
+	ctx, cancelFunc, err := p.prepareCommand(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer cancelFunc()
+
+	output := bytes.NewBuffer(make([]byte, 0, 1*sdkio.MebiByte))
```
jasdel: Would something closer to (5-10) KB be more appropriate? A response as large as 1MB seems fairly excessive for a default.

aws/stscreds/provider.go
```diff
@@ -199,43 +204,48 @@ type AssumeRoleProvider struct {
...
-		} else if p.TokenProvider != nil {
-			input.SerialNumber = p.SerialNumber
-			code, err := p.TokenProvider()
+	if p.options.SerialNumber != nil {
+		if p.options.TokenCode != nil {
+			input.SerialNumber = p.options.SerialNumber
+			input.TokenCode = p.options.TokenCode
+		} else if p.options.TokenProvider != nil {
+			input.SerialNumber = p.options.SerialNumber
+			code, err := p.options.TokenProvider()
```
jasdel: What do you think about simplifying this, dropping `TokenCode`, and only keeping `TokenProvider`? The extra parameter duplicates the behavior of `TokenProvider` and `TokenProvider` can satisfy the role of `TokenCode`.

aws/stscreds/web_identity_provider.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// WebIdentityRoleProviderOptions is a structure of configurable options for WebIdentityRoleProvider
+type WebIdentityRoleProviderOptions struct {
+	ExpiryWindow time.Duration
+}
+
+// NewWebIdentityRoleProvider will return a new WebIdentityRoleProvider with the
+// provided stsiface.STSAPI
+func NewWebIdentityRoleProvider(svc stsiface.ClientAPI, roleARN, roleSessionName, path string, options ...func(*WebIdentityRoleProviderOptions)) *WebIdentityRoleProvider {
```
jasdel: What do you think about changing the `path` parameter to a interface that has a method to get the JWT bytes. 
Minor difference, but it decouples where the JWT bytes come from outside of the web identity role provider. This also has a small benefit for tests, that a actual JWT file isn't needed for unit testing this provider.

```go
type JWTBytes interface {
    GetJWTBytes() ([]byte, error)
}

type JWTFilename string
func (f *FileBytes) GetJWTBytes() ([]byte, error) { /* ... */ }
```

Usage:
```go
stscreds.NewWebIdentityRoleProvider(client, myARN, "sessionName", stscreds.JWTFilename("myfile.jwt"))
```

aws/external/resolve_credentials.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	sts := sts.New(*cfg)
+
+	opts := []func(*stscreds.AssumeRoleProviderOptions){
+		func(options *stscreds.AssumeRoleProviderOptions) {
+			options.RoleSessionName = sharedCfg.RoleSessionName
+			if sharedCfg.RoleDurationSeconds != nil {
+				if *sharedCfg.RoleDurationSeconds/time.Second < 900 {
+					options.Duration = time.Second * 900
```
jasdel: Minor nit but using `time.Minute*15` would be a bit easier to read then 900. (Don't have to think about how many minutes 900 seconds are)

aws/processcreds/provider.go
skmcgrail: Yes this was a miss on my part, I couldn't remember what the default had been recently changed to in V1. Forgot to put in a TODO to come back and correct this!

aws/processcreds/provider.go
```diff
@@ -0,0 +1,342 @@
...
+// executeCredentialProcess starts the credential process on the OS and
+// returns the results or an error.
+func (p *Provider) executeCredentialProcess(ctx context.Context) ([]byte, error) {
+	ctx, cancelFunc, err := p.prepareCommand(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer cancelFunc()
+
+	output := bytes.NewBuffer(make([]byte, 0, DefaultBufSize))
```
jasdel: Do we want to export `DefaultBufSize` the value cannot be set? Since we're not using the limit reader anymore probably don't need to export this value.

Issue #493, opened 2020-02-22
[feature request] Add Empty() API to expression builders


alextbok:
Please fill out the sections below to help us address your issue.

### Version of AWS SDK for Go?
v0.19.0

### Version of Go (`go version`)?
go version go1.13.7 darwin/amd64

### What issue did you see?
I think it'd be useful to have a function on `service/dynamodb/expression/condition.go#ConditionBuilder`
that enables checks to see if the ConditionBuilder has any conditions added. As it currently stands, the `expression.Builder` that the `ConditionBuilder` is pass into will fail with a "buildTree" error if no expressions have been set. Open to anything, but proposed APIs include `func (ConditionBuilder) IsSet() bool` or `func (ConditionBuilder) IsEmpty() bool`

### Steps to reproduce
Suppose we had the following function where a ConditionBuilder is provided to us and we want to add it to the builder if and only if it is set. Note that this will fail in the unset case.
```
func buildWithFilter(filter expression.ConditionBuilder) (Expression, error) {
  builder := expression.NewBuilder().WithKeyCondition(...)
  builder = builder.WithFilter(filter)
  return builder.Build()
}
```

With the proposed API, we could make a check to know whether to add it or not:
```
func buildWithFilter(filter expression.ConditionBuilder) (Expression, error) {
  builder := expression.NewBuilder().WithKeyCondition(...)
  if filter.IsSet() {
    builder = builder.WithFilter(filter)
  }
  return builder.Build()
}
```
This is a wasteful and hacky workaround with the current limitations:
```
func buildWithFilter(filter expression.ConditionBuilder) (Expression, error) {
  builder := expression.NewBuilder().WithKeyCondition(...)

  if _, err := expression.NewBuilder().WithFilter(filter).Build(); err != nil {
    if err != nil {
      switch err.(type) {
        case expression.UnsetParameterError:
          // Do nothing.
        default:
          return expression.Expression{}, err
       }
     }
  } else {
    builder = builder.WithFilter(filter)
  }

  return builder.Build()
}
```

jasdel:
Thanks for reaching out and taking the time to to create this feature request. It makes sense to provide the ability to check if the condition builder has any conditions configured.

We'd be happy to review a pull request if you're interested in [submitting one](https://github.com/aws/aws-sdk-go-v2/blob/master/CONTRIBUTING.md).

alextbok:
@jasdel awesome, just submitted a PR.

[...]

PR #494, opened 2020-02-27
service/dynamodb/expression: Add IsSet helper for ConditionBuilder and KeyConditionBuilder

alextbok:
Adds `IsSet` method to the `ConditionBuilder` and `KeyConditionBuilder` types. This methods makes it easier to discover if the condition builders have any conditions added to them.

Implements #493.


alextbok:
Thanks @jasdel. I just made the requested changes. You mentioned the signature as `HasConditions` but the changelog specified the original `IsSet`, so I just kept `IsSet`, which I feel is more consistent with the internal package terminology (as well as the unset error message that is exposed), but let me know if you'd prefer `HasConditions` and I'll update the code and changelog

PR #495, opened 2020-02-28
internal/ini: Normalize Section keys to lowercase

gaqzi:
This is to match the behavior of the parsing in Python and the boto
based AWS CLI, where the keys are normalized to lowercase. Per the [config parser documentation](https://docs.python.org/3.8/library/configparser.html#quick-start) in Python:

> Note also that keys in sections are case-insensitive and stored in lowercase.

Because this indicates that only keys are case-insensitive in Python, I didn't touch the sections themselves (and I had a lot more failing tests if I did, so I opted for the smallest possible change that required changing existing tests).

We had a case where things were working using the CLI but when we tried
to do the same action, list S3 bucket through Terraform, which relies on the Go SDK, it didn't work because we had UPPERCASED the access_key_id.

Note: This issue is also present in the v1 SDK as the INI parser seems to be copied from there, but I figured I would start by making this PR here and if it's accepted handle the backport. :)

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,6 +21,8 @@ SDK Features
   
 SDK Enhancements
 ---
+* `internal/ini`: Normalize Section keys to lowercase ([#495](https://github.com/aws/aws-sdk-go-v2/pull/495))
+  * Update's SDK's ini utility to store all keys as lowercase. This brings the SDK inline with the AWS CLI's behavior.
 
 SDK Bugs
 ---


internal/ini/testdata/valid/utf_8_profile_expected
```diff
@@ -1,6 +1,6 @@
 {
 	"ʃʉʍΡιξ": {
-		"ϰϪϧ": "Ϯϴϖ",
+		"ϰϫϧ": "Ϯϴϖ",
```
gaqzi: wish I didn't have to change this, but, it has a capitalized UTF-8 character.

PR #500, opened 2020-03-15
aws/defaults: Add recording clock skew metadata per request attempt

jasdel:
Updates the SDK's default request handlers to record metadata about the clock skew from a request attempt. This metadata will be used by the SDK's request retry metadata header. This header may be used by AWS services in the future to better determine how to handle request load balancing.

This update does NOT add clock skew correction to the SDK's request signatures.

### DIFF
--- a/aws/defaults/handlers.go
+++ b/aws/defaults/handlers.go
@@ -301,7 +301,7 @@ var AttemptClockSkewHandler = aws.NamedHandler{
 		}
 
 		r.AttemptClockSkews = append(r.AttemptClockSkews,
-			respDate.Add(-1*time.Second).Sub(r.ResponseAt.Add(-1*time.Second)),
+			respDate.Sub(r.ResponseAt),
 		)
 	},
 }


aws/defaults/handlers.go
```diff
@@ -204,10 +206,25 @@ var RetryMetricHeaderHandler = aws.NamedHandler{
...
-		// read deadlines.
+
+		type timeoutGetter interface {
+			GetTimeout() time.Duration
+		}
+
+		// Attempt extract the TTL from context deadline, or timeout on the client.
+		var ttl time.Time
+		if v, ok := r.Config.HTTPClient.(timeoutGetter); ok && v.GetTimeout() > 0 {
+			ttl = sdk.NowTime().Add(v.GetTimeout())
```
skotambkar: `GetTimeOut()` should be called just once.

aws/defaults/handlers.go
```diff
@@ -204,10 +206,25 @@ var RetryMetricHeaderHandler = aws.NamedHandler{
...
+		} else if deadline, ok := r.Context().Deadline(); ok {
+			ttl = deadline
+		}
+
+		// Only append the ttl if it can be determined.
+		if !ttl.IsZero() && len(r.AttemptClockSkews) > 0 {
+			const unixTimeFormat = "20060102T150405Z"
+			ttl = ttl.Add(r.AttemptClockSkews[len(r.AttemptClockSkews)-1])
+			parts = append(parts, fmt.Sprintf("ttl=%s", ttl.Format(unixTimeFormat)))
```
skotambkar: Will this always be Unix time format for every service?

aws/defaults/handlers.go
```diff
@@ -204,10 +206,25 @@ var RetryMetricHeaderHandler = aws.NamedHandler{
...
+		}
 
 		r.HTTPRequest.Header.Set("amz-sdk-request", strings.Join(parts, "; "))
```
skotambkar: can use `retryMetricHeader` const instead of literal for `amz-sdk-request`

aws/defaults/handlers.go
```diff
@@ -252,3 +269,31 @@ var ValidateEndpointHandler = aws.NamedHandler{Name: "core.ValidateEndpointHandl
...
+		}
+
+		respDateHeader := r.HTTPResponse.Header.Get("Date")
+		if len(respDateHeader) == 0 {
+			return
+		}
+
+		respDate, err := http.ParseTime(respDateHeader)
+		if err != nil {
+			// unable to parse response error for some reason.
```
skotambkar: err can be logged, if we don't want to put it on request.

aws/defaults/handlers.go
jasdel: Yeah, per design, this header fields are independent of any API protocols. Generic across all APIs over HTTP.

aws/defaults/handlers.go
jasdel: In this case since the else condition needs to be invoked if there is no timeout, the else has to be duplicated, or the if becomes more complicated if the `GetTimeout` call is only called once.

aws/defaults/handlers.go
jasdel: can add logging here

aws/defaults/handlers.go
```diff
@@ -252,3 +269,34 @@ var ValidateEndpointHandler = aws.NamedHandler{Name: "core.ValidateEndpointHandl
...
+			if r.Config.Logger != nil {
+				r.Config.Logger.Log(fmt.Sprintf("ERROR: unable to determine clock skew for %s/%s API response, invalid Date header value, %v",
+					r.Metadata.ServiceName, r.Operation.Name, respDateHeader))
+			}
+			return
+		}
+
+		r.AttemptClockSkews = append(r.AttemptClockSkews,
+			respDate.Add(-1*time.Second).Sub(r.ResponseAt.Add(-1*time.Second)),
+		)
```
skmcgrail: What's the purpose of subtracting the one second from each value?

jasdel: I have an outstanding question about that as well from the design doc this is implemented from. I cannot see a reason for it, but wanted to confirm before removing it.

jasdel: Going to remove this for now. Can add it back in if needed.  Best guess is that the design doc used this as some kind of error correction if the Date header wasn't available.

PR #501, opened 2020-03-16
JSON marshaler prototype using JSON encoder utility

skotambkar:
This prototype uses json encoder utility.

### DIFF
--- a/service/apigateway/proto_api_op_CreateApiKey.go
+++ b/service/apigateway/proto_api_op_CreateApiKey.go
@@ -2,9 +2,9 @@ package apigateway
 
 import (
 	"context"
-	"github.com/aws/aws-sdk-go-v2/private/protocol/restjson"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/private/protocol/restjson"
 )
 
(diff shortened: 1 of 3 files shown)


service/apigateway/marshal.go
```diff
@@ -0,0 +1,88 @@
...
+		jsonVersion := r.Metadata.JSONVersion
+		r.HTTPRequest.Header.Add("Content-Type", "application/x-amz-json-"+jsonVersion)
+	}
+
+	obj := enc.Object()
+	obj.Key("customerId").String(*v.CustomerId)
+	obj.Key("description").String(*v.Description)
+	obj.Key("enabled").Boolean(*v.Enabled)
+	obj.Key("generateDistinctId").Boolean(*v.GenerateDistinctId)
+	obj.Key("name").String(*v.Name)
```
skmcgrail: dereferencing values should be guarded with a `if v.<Field> != nil`

skotambkar: Will add null checks!

skotambkar: For marshaling input into JSON, guarding with v.<Field>!=nil might not be enough; If there exists cases where `not setting a key in json` is different than `setting null value for that key`. Those cases, would need customizations.

PR #502, opened 2020-03-16
JSON unmarshaler prototype: walking the JSON body 

skotambkar:
The PR adds prototyped implementation of unmarshaler. 

In this PR, we walk the json body. The aim is to reduce reflection when unmarshaling.


service/apigateway/unmarshal.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	return nil
+}
+
+func unmarshalTagsMapShapeAWSJSON(output *map[string]string, decoder *json.Decoder, ringBuffer *sdkio.RingBuffer) error {
+	startToken, err := decoder.Token()
+	if err == io.EOF {
+		// "Empty List"
+		return nil
+	}
```
skmcgrail: Does this actually indicate the end of the map in this case? Or does this actually indicate the end of the entire IO stream?

PR #504, opened 2020-03-16
aws: Fix BuildableHTTPClient datarace bug

jasdel:
Fixes the aws.BuildableHTTPClient's copy behavior to not encounter dataraces when being used concurrently.

Also fixes a broken unit test missed when implementing #487.

### DIFF
--- a/aws/http_client.go
+++ b/aws/http_client.go
@@ -41,7 +41,7 @@ type BuildableHTTPClient struct {
 	transport *http.Transport
 	dialer    *net.Dialer
 
-	initOnce *sync.Once
+	initOnce sync.Once
 
 	clientTimeout time.Duration
 	client        *http.Client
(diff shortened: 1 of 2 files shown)


aws/http_client.go
```diff
@@ -110,51 +91,49 @@ func (b BuildableHTTPClient) initReset() BuildableHTTPClient {
 // If a non (*http.Transport) was set as the round tripper, the round tripper
 // will be replaced with a default Transport value before invoking the option
 // functions.
-func (b BuildableHTTPClient) WithTransportOptions(opts ...func(*http.Transport)) HTTPClient {
-	b = b.initReset()
+func (b *BuildableHTTPClient) WithTransportOptions(opts ...func(*http.Transport)) HTTPClient {
+	cpy := b.clone()
 
-	tr := b.GetTransport()
+	tr := cpy.GetTransport()
```
skmcgrail: Should this still b `b`? Wouldn't we want to copy the existing transport that might have been set?

PR #510, opened 2020-03-18
aws: Cleanup exported concrete error types, only export error codes

skmcgrail:
Cleans up cases in the `aws` package where we were exporting concrete awserr.Error types, instead we will only export error code constants which can be compared and will be immutable.


skmcgrail:
A good point for discussion is we have a mix of naming schemes for error codes. `ErrCode<Name>` and `<Name>ErrCode` seem to be common patterns, but we should likely standardize on one for consistency.

jasdel:
So for this PR, what do you think about taking the step further and getting rid of the error code string, and instead use typed errors similar to the work that was done with the retry + error handling PR, #487.

PR #512, opened 2020-03-18
aws/endpoints: Remove types used for exposing iteration types over partitions, regions, and service models.

skmcgrail:
This PR further cleans up the endpoints package to only expose the default generated resolver based on the endpoint model. This removes supporting types that were used to iterate over the model, and that were inherently tied to and were exposing the underlying data model.


flyinprogrammer:
You might want to fix your README:

https://github.com/aws/aws-sdk-go-v2/blame/master/README.md#L68

jcarter3:
I know I'm late to the party since this is already merged, but upon updating versions of the SDK I had a lot of code that broke due to this. I have some lambdas that iterate over all regions and do some work in each one, and it was implemented by iterating over all the regions in all the partitions. But now those are no longer exposed. Is there an alternative way to do this? I supposed I can just build and maintain a hardcoded list, but using the exports on the endpoint package seemed harmless enough at the time.

bobheadxi:
What was the reason for removing this? https://github.com/aws/aws-sdk-go-v2/pull/512#issuecomment-621442799 and #557 both point out use cases for this functionality

kaihendry:
Breaking changes are nasty `undefined: endpoints.UsWest2RegionID`, though there doesn't appear to any documentation how to fix code that depended on setting say `cfg.Region = endpoints.UsWest2RegionID` to send email, or am I missing something?

Thank you

dustin-decker:
Documentation still mentions things removed by this, and there is no clear equivalent functionality: https://docs.aws.amazon.com/sdk-for-go/v2/api/aws/endpoints/

PR #514, opened 2020-03-23
`aws`: add de/serialization typed error to the SDK

skotambkar:
### Rev 1
* Adds HTTP based de/serialization typed error 
* This PR doesn't replace the old usages of Err codes for de/serialization error.

### Rev 2
* Changes the serialization/ deserialization error types to be generic, and not transport specific. This change conforms with layered error design discussed.
* Simplified tests 

### Rev 3
* Modifications to use %w verb in serialization/deserilization error types.
* Removes the `not-so-needed` field: Reason string from the error types.

### DIFF
--- a/aws/request.go
+++ b/aws/request.go
@@ -26,13 +26,12 @@ const (
 
 // SerializationError provides a generic request serialization error
 type SerializationError struct {
-	Reason string // provides more context of error caused
-	Err    error  // original error
+	Err error // original error
 }
 
 // Error returns a formatted error for SerializationError
 func (e *SerializationError) Error() string {
-	return fmt.Sprintf("serialization failed: %v %v", e.Reason, e.Err)
+	return fmt.Sprintf("serialization failed: %v", e.Err)
 }
 
 // Unwrap returns the underlying Error in DeserializationError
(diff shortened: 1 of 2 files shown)


aws/request.go
```diff
@@ -24,6 +24,47 @@ const (
...
+	Err    error  // original error
+}
+
+// Error returns a formatted error for HTTPSerializationError
+func (e *HTTPSerializationError) Error() string {
+	return fmt.Sprintf("serialization failed: %v %v", e.Reason, e.Err)
+}
+
+// Unwrap returns the underlying Error in HTTPDeserializationError
+func (e *HTTPSerializationError) Unwrap() error { return e.Err }
```
jasdel, line 39: Since `HTTPSerializationError` is generic, regardless of transport should this just be `SerializationError`?

aws/request.go
```diff
@@ -24,6 +24,47 @@ const (
...
+// ErrorStatusCode returns the Response status code
+func (e *HTTPDeserializationError) ErrorStatusCode() int {
+	if e.Response != nil {
+		return e.Response.StatusCode
+	}
+	return 0
+}
+
+// ErrorRequestID returns the Request ID for error response request
+func (e *HTTPDeserializationError) ErrorRequestID() string { return e.RequestID }
```
jasdel, line 63: These methods should just be `StatusCode`, and `RequestID` to conform to the current `awserr.RequestFailure` interface. Will have to address the member name collision though. Maybe use the v1 sdk's idea of a `protocol.ResponseMetadata` type as a member instead of the two individual members here. Poterntially adding `http.Response` to that type as well.

aws/typed_error_test.go
```diff
@@ -0,0 +1,167 @@
...
+			Response:  r.HTTPResponse,
+			RequestID: r.RequestID,
+			Reason:    fmt.Sprintf("Here's a snapshot of response being deserialized: %s", snapshot), // Additional context
+			Err:       err,
+		}
+	}
+}
+
+// Call unmarshal output shape of JSON doc
+func unmarshalJSON(dec *json.Decoder, output *mockOP) error {
```
jasdel, line 121: For simplicity and maintainability fo this test might be better just to use the standard json.Unmarshal instead of needing to hand write a deserializer.

PR #517, opened 2020-03-27
Smithy SigV4 Middlwares and Retry Middleware Changes

skmcgrail:
Includes two middlwares compatible with the proposed Smithy middleware designs that handle SigV4 authentication.

### DIFF
--- a/aws/signer/v4/middleware.go
+++ b/aws/signer/v4/middleware.go
@@ -81,7 +81,7 @@ func (m *ComputePayloadSHA256Middleware) HandleFinalize(ctx context.Context, in 
 		return out, metadata, &HashComputationError{Err: fmt.Errorf("failed to compute payload hash, %w", err)}
 	}
 
-	if err := req.Rewind(); err != nil {
+	if err := req.RewindStream(); err != nil {
 		return out, metadata, &HashComputationError{Err: fmt.Errorf("failed to seek body to start, %w", err)}
 	}
 
--- a/aws/retry/middleware.go
+++ b/aws/retry/middleware.go
@@ -67,6 +67,14 @@ func (r AttemptMiddleware) HandleFinalize(ctx context.Context, in smithymiddle.F
 			AttemptClockSkew: attemptClockSkew,
 		})
 
+		if attemptNum > 1 {
+			if rewindable, ok := in.Request.(interface{ RewindStream() error }); ok {
+				if err := rewindable.RewindStream(); err != nil {
+					return out, metadata, fmt.Errorf("failed to rewind transport stream for retry, %w", err)
+				}
+			}
+		}
+
 		out, metadata, reqErr := next.HandleFinalize(attemptCtx, attemptInput)
 
 		relRetryToken(reqErr)


aws/signer/v4/v4.go
```diff
@@ -46,7 +46,7 @@
 // through 1.7.4 you should use the URL.RawPath as the pre-escaped form of the
 // request URL. https://github.com/golang/go/issues/16847 points to a bug in
 // Go pre 1.8 that fails to make HTTP2 requests using absolute URL in the HTTP
-// message. URL.Opaque generally will force Go to make requests with absolute URL.
+// cause. URL.Opaque generally will force Go to make requests with absolute URL.
```
skmcgrail: This change is unintentional and will be removed.

aws/client/client.go
```diff
@@ -0,0 +1,34 @@
...
+var _ middleware.MiddlewareHandler
+
+type signingMetadataKey struct{}
+
+type SigningMetadata struct {
+	SigningName   string
+	SigningRegion string
+}
+
+func GetSigningMetadata(ctx context.Context) SigningMetadata {
```
jasdel: Should this signing metadata be apart of a base client package? This seems more specific to a signer package, that a generated client would compose it self with, or have members for.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
+package v4
```
jasdel: Nit, probably want to rename this file to middleware, or signing_middleware

aws/signer/v4/handler_test.go
```diff
@@ -0,0 +1,172 @@
+package v4
```
jasdel: nit probably want to rename this file to middleware_test or signing_middleware_test

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Error is the error message
+func (e *HashComputationError) Error() string {
+	msg := "failed to compute payload hash"
+	if len(e.cause) == 0 {
+		return msg
+	} else {
+		return fmt.Sprintf("%s: %s", msg, e.cause)
+	}
```
jasdel: nit, don't really need the else case since returning within the if.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+	err   error
+}
+
+func (e *SigningError) Error() string {
+	msg := "failed to sign request"
+	if len(e.cause) == 0 {
+		return msg
+	} else {
+		return fmt.Sprintf("%s: %s", msg, e.cause)
+	}
```
jasdel: nit, don't really need the else case since returning within the if.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+
+// UnsignedPayloadMiddleware sets the SigV4 request payload hash to unsigned
+type UnsignedPayloadMiddleware struct{}
+
+func (m *UnsignedPayloadMiddleware) Name() string {
+	return "SigV4 unsigned payload middleware"
+}
+
+func (m *UnsignedPayloadMiddleware) HandleFinalize(ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler) (out middleware.FinalizeOutput, err error) {
+	ctx = context.WithValue(ctx, payloadHashKey{}, v4Internal.UnsignedPayload)
```
jasdel: What do you think about putting these WithValue, GetValue calls in helper utility functions? Their public interface would be fore the expected type of the payload.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+type ComputePayloadHashMiddleware struct{}
+
+// Name is the middleware name
+func (m *ComputePayloadHashMiddleware) Name() string {
+	return "SignV4 Payload Hash Middleware"
+}
+
+// HandleFinalize compute the payload has fof the request payload for SigV4 authentication
+func (m *ComputePayloadHashMiddleware) HandleFinalize(ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler) (out middleware.FinalizeOutput, err error) {
+	req := in.Request.(*smithyHTTP.Request)
```
jasdel: Probably want to return error if Request is not the correct type.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+type SignHTTPRequestMiddleware struct {
+	Signer interface {
+		SignHTTP(ctx context.Context, r *http.Request, payloadHash string, service string, region string, signingTime time.Time)
+	}
+}
+
+func NewSignHTTPRequestMiddleware(signer interface {
+	SignHTTP(ctx context.Context, r *http.Request, payloadHash string, service string, region string, signingTime time.Time)
+}) *SignHTTPRequestMiddleware {
+	return &SignHTTPRequestMiddleware{Signer: signer}
```
jasdel: This pkg could define its own `HTTPSigner` interface type for the `SignHTTP` interface needed. Would help with being able to document this parameter type potentially.

aws/signer/v4/handler.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// Name is the middlware name
+func (s *SignHTTPRequestMiddleware) Name() string {
```
jasdel: Smithy PR for handlers were updated renaming `Name` to `ID`

aws/client/client.go
skmcgrail: Agreed, this should probably be moved to the v4 package.

skmcgrail:
Builds are broken as we don't yet depend on the smithy-go package in our go.mod file.

aws/signer/v4/middleware.go
```diff
@@ -0,0 +1,191 @@
...
+
+// SignHTTPRequestMiddleware is a `FinalizeMiddleware` implementation for SigV4 HTTP Signing
+type SignHTTPRequestMiddleware struct {
+	Signer HTTPSigner
+}
+
+// NewSignHTTPRequestMiddleware constructs a SignHTTPRequestMiddleware using the given Signer for signing requests
+func NewSignHTTPRequestMiddleware(signer HTTPSigner) *SignHTTPRequestMiddleware {
+	return &SignHTTPRequestMiddleware{Signer: signer}
+}
```
jasdel, line 120: should `Signer` member be unexported so using constructor is required?

aws/signer/v4/middleware_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (*semiSeekable) Read(p []byte) (n int, err error) {
+	return 0, io.EOF
+}
+
+type finalizeHandlerFunc func(ctx context.Context, in middleware.FinalizeInput) (middleware.FinalizeOutput, middleware.Metadata, error)
+
+func (f finalizeHandlerFunc) HandleFinalize(ctx context.Context, in middleware.FinalizeInput) (middleware.FinalizeOutput, middleware.Metadata, error) {
+	return f(ctx, in)
+}
```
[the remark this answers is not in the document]
skmcgrail: Will follow up with a separate request to add these there, and then can follow up with a separate pr to update this middleware.

aws/signer/v4/middleware.go
```diff
@@ -0,0 +1,176 @@
...
+
+	body, ok := req.Stream.(io.ReadSeeker)
+	if !ok {
+		return middleware.FinalizeOutput{}, metadata, &HashComputationError{Err: fmt.Errorf("stream does not implement io.Seeker, got %T", req.Stream)}
+	}
+
+	_, err = body.Seek(0, io.SeekStart)
+	if err != nil {
+		return middleware.FinalizeOutput{}, metadata, &HashComputationError{Err: fmt.Errorf("failed to seek body to start, %w", err)}
+	}
```
jasdel, line 86: So we want to be careful about seeking the payload. Seeking to 0 vs the current position the stream was received at could cause unexpected behavior. Its an odd situation but should the SDK attempt to reseek back to the begin of a reader if the user gave it to the input parameter at some other offset. Good example of this is I want to upload only the trailing portion of a file.

this is the reason the current seeking behavior in the SDK keeps track of initial offset. Might make sense to put all of the seeking behavior into a helper since it will be reused by retry. Maybe as apart of smithyhttp.Request?

aws/signer/v4/middleware.go
```diff
@@ -0,0 +1,176 @@
...
+
+	signingMetadata := GetSigningMetadata(ctx)
+	payloadHash := GetPayloadHash(ctx)
+	if len(payloadHash) == 0 {
+		return middleware.FinalizeOutput{}, metadata, &SigningError{Err: fmt.Errorf("computed payload hash missing from context")}
+	}
+
+	err = s.signer.SignHTTP(ctx, req.Request, payloadHash, signingMetadata.SigningName, signingMetadata.SigningRegion, sdk.NowTime())
+	if err != nil {
+		return middleware.FinalizeOutput{}, metadata, &SigningError{Err: fmt.Errorf("failed to sign http request, %w", err)}
```
jasdel, line 136: could just use `out` for the return instead of middleware.FinalizeOutput{} similar to metadata

skmcgrail:
Latest changes utilize the new updates to the smithy http stream body: https://github.com/awslabs/smithy-go/pull/15

PR #520, opened 2020-04-02
Retry Middleware Implementation

skmcgrail:
There are some TODO comments I've made as points to discuss and come back to. Will follow up with some unit tests as well.

### DIFF
--- a/aws/middleware/middleware.go
+++ b/aws/middleware/middleware.go
@@ -17,11 +17,13 @@ type RequestInvocationIDMiddleware struct{}
 
 // ID the identifier for the RequestInvocationIDMiddleware
 func (r RequestInvocationIDMiddleware) ID() string {
-	return "Request Invocation ID Middleware"
+	return "RequestInvocationIDMiddleware"
 }
 
 // HandleBuild attaches a unique operation invocation id for the operation to the request
-func (r RequestInvocationIDMiddleware) HandleBuild(ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (out middleware.BuildOutput, metadata middleware.Metadata, err error) {
+func (r RequestInvocationIDMiddleware) HandleBuild(ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (
+	out middleware.BuildOutput, metadata middleware.Metadata, err error,
+) {
 	const invocationIDHeader = "amz-sdk-invocation-id"
 
 	invocationID, err := sdk.UUIDVersion4()
@@ -43,12 +45,15 @@ func (r RequestInvocationIDMiddleware) HandleBuild(ctx context.Context, in middl
 // TODO: Could be a better name, since this calculates more then skew
 type AttemptClockSkewMiddleware struct{}
 
+// ID is the middleware identifier
 func (a AttemptClockSkewMiddleware) ID() string {
-	return "Attempt Clock Skew Middlware"
+	return "AttemptClockSkewMiddlware"
 }
 
 // HandleDeserialize calculates response metadata and clock skew
-func (a AttemptClockSkewMiddleware) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (out middleware.DeserializeOutput, metadata middleware.Metadata, err error) {
+func (a AttemptClockSkewMiddleware) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (
+	out middleware.DeserializeOutput, metadata middleware.Metadata, err error,
+) {
 	respMeta := ResponseMetadata{}
 
 	deserialize, metadata, err := next.HandleDeserialize(ctx, in)
@@ -86,13 +91,9 @@ type ResponseMetadata struct {
 }
 
 // GetResponseMetadata retrieves response metadata from the context, if nil returns an empty value
-func GetResponseMetadata(metadata middleware.Metadata) ResponseMetadata {
-	switch v := metadata.Get(responseMetadataKey{}).(type) {
-	case ResponseMetadata:
-		return v
-	default:
-		return ResponseMetadata{}
-	}
+func GetResponseMetadata(metadata middleware.Metadata) (v ResponseMetadata) {
+	v, _ = metadata.Get(responseMetadataKey{}).(ResponseMetadata)
+	return v
 }
 
 // SetResponseMetadata sets the ResponseMetadata on the given context
--- a/aws/retry/middleware.go
+++ b/aws/retry/middleware.go
@@ -7,9 +7,9 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	middleware2 "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	awsmiddle "github.com/aws/aws-sdk-go-v2/aws/middleware"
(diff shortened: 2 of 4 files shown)


aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func (r RetryMiddleware) HandleFinalize(ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler) (out middleware.FinalizeOutput, err error) {
+	const invocationIDHeader = "amz-sdk-invocation-id"
+
+	req := in.Request.(*http.Request)
+
+	// TODO: Don't see a compelling reason to make this a separate middleware that makes this available via context
+	invocationID, err := sdk.UUIDVersion4()
+	if err != nil {
```
jasdel, line 43: This might be helpful via logging, would make sense to be in a `Build` handler, separate from retry middleware implementation. Helpful if the retry middleware is replaced.

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+
+	relRetryToken := r.Retryer.GetInitialToken()
+	for {
+		attemptNum++
+
+		ctx = context.WithValue(ctx, retryMetadataKey{}, retryMetadata{
+			AttemptNum:       attemptNum,
+			AttemptTime:      sdk.NowTime(),
+			MaxAttempts:      maxAttempts,
+			AttemptClockSkew: attemptClockSkew,
```
jasdel, line 62: Clock skew probably should be a separate middleware that determines this value?

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+
+		retryDelay, err := r.Retryer.RetryDelay(attemptNum, reqErr)
+		if err != nil {
+			return out, err
+		}
+
+		relRetryToken, err = r.Retryer.GetRetryToken(ctx, reqErr)
+		if err != nil {
+			return out, err
+		}
```
jasdel, line 93: not issue, just curious for reason for swapping the order of these two,

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+
+		respContainer, ok := out.Result.(responseMeta)
+		if ok {
+			metadata := respContainer.GetResponseMetadata()
+			responseAt := metadata.GetResponseAt()
+			serverTime := metadata.GetServerTime()
+
+			if !(responseAt.IsZero() || serverTime.IsZero()) {
+				attemptClockSkew = responseAt.Sub(serverTime)
+			}
```
jasdel, line 116: it would be helpful to determine clockskew in the deserializer, and return that upstream the response metadata for any consumer to use. This will reduce the amount of not-directly related logic  that needs to be in the retry middleware.

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+		if !retryable {
+			return out, err
+		}
+
+		if maxAttempts > 0 && attemptNum >= maxAttempts {
+			err = &aws.MaxAttemptsError{
+				Attempt: attemptNum,
+				Err:     err,
+			}
+			return out, err
```
jasdel, line 82: don't need to solve it in this PR, but this is a good example of a place where we should think about how to include response metadata upstream.

Especially since this is returning both `out` response and `err` generally this is an anti-pattern that we should avoid. It probably would be better to have a way of decorating the error with response and retry metadata. Potentially a wrapper retry always wraps any error returned with, (e.g. via a defer)

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+
+	parts = append(parts, fmt.Sprintf("attempt=%d", metadata.AttemptNum))
+	if metadata.MaxAttempts != 0 {
+		parts = append(parts, fmt.Sprintf("max=%d", metadata.MaxAttempts))
+	}
+
+	// TODO: Is there a method that we could know the client timeout? Maybe this is passed via middleware constructor?
+	//type timeoutGetter interface {
+	//	GetTimeout() time.Duration
+	//}
```
jasdel, line 150: passing this in via the context is probably our only option.  Either that or the concept of a handler is changed to make use of a client that is provided via the context. Not sure about the ladder approach, though.

aws/retry/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (r RetryMetricsHeaderMiddleware) ID() string {
+	return "Retry Metrics Header Middleware"
+}
+
+func (r RetryMetricsHeaderMiddleware) HandleFinalize(ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler) (out middleware.FinalizeOutput, err error) {
+	metadata, ok := ctx.Value(retryMetadataKey{}).(retryMetadata)
+	if !ok {
+		return out, fmt.Errorf("retry metadata value not found on context")
+	}
```
jasdel, line 137: nit probably will be helpful to move context getter setters to helper functions. even if they are unexported. 

also in case of no retry metadata, this middleware should just forward to next FinalizeHandler, not error.

jasdel:
Biggest thing we need to workthrough is our various upstream metadata values go. We have Context conveniently for downstream metadata, but not an equivalent for the upstream.

One way to handle this would be for the smithy middleware stack to change the expected contract between steps.

Though at the end of the data, the generated API response type, and all API errors will still need some metadata.  Maybe having an interface the middleware expects Result and error values to implement is the best appraoch.

aws/retry/middleware.go
skmcgrail: No reason, believe this was unintentional code movement during migration.

aws/retry/middleware.go
skmcgrail: Agreed on calculating this as part of de-serialization and having someone to access it in the response.

aws/middleware/middleware.go
```diff
@@ -0,0 +1,101 @@
...
+type AttemptClockSkewMiddleware struct{}
+
+func (a AttemptClockSkewMiddleware) ID() string {
+	return "Attempt Clock Skew Middlware"
+}
+
+// HandleDeserialize calculates response metadata and clock skew
+func (a AttemptClockSkewMiddleware) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (out middleware.DeserializeOutput, metadata middleware.Metadata, err error) {
+	respMeta := ResponseMetadata{}
+
```
jasdel, line 53: what do you think about putting return parameters on another line similar to the smithy middleware pkg?

aws/middleware/middleware.go
```diff
@@ -0,0 +1,101 @@
+package middleware
+
+import (
```
jasdel, line 3: We'll probably want to take a look at how request/retry middleware are grouped in various packages.

aws/retry/middleware.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			return out, metadata, err
+		}
+
+		if err = sdk.SleepWithContext(ctx, retryDelay); err != nil {
+			err = &aws.RequestCanceledError{Err: err}
+			return out, metadata, err
+		}
+
+		responseMetadata := middleware2.GetResponseMetadata(metadata)
```
jasdel, line 107: maybe `smithymiddle` and `awmiddle` for aliases of the two middleware packages?

skmcgrail: Yeah I forgot to fix the Intellij auto-name

PR #523, opened 2020-04-03
aws/external: Add Support for setting a default fallback region and resolving region from EC2 IMDS

### LINKED ISSUE #244: LoadDefaultAWSConfig: Automatically get REGION from EC2 metadata
More info can be found from the same issue on v1: https://github.com/aws/aws-sdk-go/issues/1103

### LINKED ISSUE #515: Unable to assume web identity inside EC2 when no region is set in the environment.
  
**Describe the bug**
All requests fail when executed on an **EC2 (EKS) instance trying to use AssumeWebIdentity** and `AWS_REGION` wasn't set.

v0.20.0

go1.14

The following code snippet exhibits the problem:

```go
package main

import (
)

func () main() {
    awsCfg, err := external.LoadDefaultAWSConfig()

    if err != nil {
        panic(fmt.Errorf("unable to load AWS SDK config: %s", err))
    }

    if awsCfg.Region == "" {
        awsCfg.Region = "us-east-1"
    }

    stsClient := sts.New(awsCfg)
    _, err := stsClient.GetCallerIdentityRequest(&sts.GetCallerIdentityInput{}).Send(ctx)

    // This call fails even though the region was set manually. If set before hand via `AWS_REGION`, it works as expected.
    if err != nil {
        panic(fmt.Errorf("failed to fetch AWS STS caller identity: %s", err))
    }
}
```

**Expected behavior**
The code should work but doesn't when running on an EC2 instance assuming a web identity.

**Additional context**
I debugged the situation a bit, and it seems the internal AssumeRole call is performed using a default AWS config. Since the default config on EC2 doesn't fetch the EC2 metadata region (yet? https://github.com/aws/aws-sdk-go-v2/issues/244), the call fails with `UnknownEndpointError`, even though a region was set on the configuration. From my understanding, the underlying provider in charge of performing the assume-role call doesn't get the manually set region, and uses an empty one instead.

I could find no way of setting the region on the provider itself.

If I set `AWS_REGION` from inside the process and reload the default configuration once more, the code starts working again. However, this seems overcomplicated and I assume it is not intentional.

### LINKED ISSUE #522: Credentials Setup
Confirm by changing [ ] to [x] below:
- [ x] I've gone though the [API reference](https://github.com/aws/aws-sdk-go-v2/issues)
- [ x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers

Does anyone know how i could go about using IAM credentials to login to an external application I am developing?  I have it properly configured to read from the ~./aws/credentials file, but is there a way to load it based on credentials passed through?  Not sure if there is another AWS service for this or if using sessions would be the right way to go.

Thanks.

skmcgrail:
This PR adds support for users to configure a default region that will be fallen back to if a region is not found in the shared config or within the process environment. This option will take less precedence from the normal region resolution. Additionally support for resolving a region from EC2 IMDS (if available) has been added which will be queried after attempting to resolve a region from shared config or the process environment, but before falling back to the default fallback region if configured.

Fixes #515 
Fixes #244

### DIFF
--- a/aws/external/resolve.go
+++ b/aws/external/resolve.go
@@ -148,7 +148,7 @@ func ResolveDefaultRegion(cfg *aws.Config, configs Configs) error {
 
 	region, found, err := GetDefaultRegion(configs)
 	if err != nil {
-		return nil
+		return err
 	}
 	if !found {
 		return nil
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,7 +18,13 @@ SDK Features
   * `SignHTTP` replaces `Sign`, and usage of `Sign` should be migrated before it's removal at a later date
   * `PresignHTTP` replaces `Presign`, and usage of `Presign` should be migrated before it's removal at a later date
   * `DisableRequestBodyOverwrite` and `UnsignedPayload` are now deprecated options and have no effect on `SignHTTP` or `PresignHTTP`. These options will be removed at a later date.
-  
+* `aws/external`: Add Support for setting a default fallback region and resolving region from EC2 IMDS ([#523](https://github.com/aws/aws-sdk-go-v2/pull/523))
+  * `WithDefaultRegion` helper has been added which can be passed to `LoadDefaultAWSConfig`
+    * This helper can be used to configure a default fallback region in the event a region fails to be resolved from other sources
+  * Support has been added to resolve region using EC2 IMDS when available
+    * The IMDS region will be used if region as not found configured in either the shared config or the process environment.
+  * Fixes [#244](https://github.com/aws/aws-sdk-go-v2/issues/244)
+  * Fixes [#515](https://github.com/aws/aws-sdk-go-v2/issues/515)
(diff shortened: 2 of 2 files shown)


aws/external/resolve.go
```diff
@@ -136,3 +138,52 @@ func ResolveEndpointResolverFunc(cfg *aws.Config, configs Configs) error {
...
+// region if region had not been resolved from other sources.
+func ResolveDefaultRegion(cfg *aws.Config, configs Configs) error {
+	if len(cfg.Region) > 0 {
+		return nil
+	}
+
+	region, found, err := GetDefaultRegion(configs)
+	if err != nil {
+		return nil
+	}
```
jasdel, line 152: should this resolver allow the GetDefaultRegion error to be retruned?

skmcgrail: Ah yes it should good catch!

PR #529, opened 2020-04-14
Prototype serialize middleware for lex runtime

skmcgrail:
Prototyped serializer middleware for the smithy based Lex runtime service prototype.


service/smithyprototype/lexruntimeservice/api_op_DeleteSession.go
```diff
@@ -201,3 +205,62 @@ type DeleteSessionResponse struct {
...
+		return out, metadata, fmt.Errorf("unknown input parameters type %T", in.Parameters)
+	}
+
+	restEncoder := rest.NewEncoder(request.Request)
+
+	restEncoder.AddHeader("Content-Type").String("application/json")
+
+	if err := serializeDeleteSessionInputAWSREST(input, restEncoder); err != nil {
+		return middleware.SerializeOutput{}, metadata, err
+	}
```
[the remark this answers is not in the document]
skmcgrail: IDE strikes again!

service/smithyprototype/lexruntimeservice/api_types.go
```diff
@@ -487,3 +488,123 @@ func (s SentimentResponse) MarshalFields(e protocol.FieldEncoder) error {
 	}
 	return nil
 }
+
+func serializeIntentSummaryListAWSJSON(v []IntentSummary, value json.Value) error {
+	if v == nil {
+		value.Null()
+		return nil
+	}
+
```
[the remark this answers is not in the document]
skmcgrail: Agreed, though in this context serializeIntentSummaryListAWSJSON would not have been called if the list is nil. So this was just more of a fallback case if the intent was to actually encode a null. Otherwise this would have to return an error to signal it can't populate a null value.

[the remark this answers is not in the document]
jasdel: What do you think about using error as the return behavior in this case?  Since we're not actively supporting encoding null? Would prevent unexpected code paths silently encoding null when its not supposed to be supported.

service/smithyprototype/lexruntimeservice/api_op_GetSession.go
```diff
@@ -238,3 +242,66 @@ type GetSessionResponse struct {
...
+	}
+
+	input, ok := in.Parameters.(*GetSessionInput)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown input parameters type %T", in.Parameters)
+	}
+
+	restEncoder := rest.NewEncoder(request.Request)
+
+	restEncoder.AddHeader("Content-Type").String("application/json")
```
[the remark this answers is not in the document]
skmcgrail: GetSession API sets the `content-type` as `application/json` today both in the current generated marshaler as well as the reflection based one.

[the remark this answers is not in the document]
skmcgrail: Though looking at the smithy protocol test it seems like the cases will be far more nuanced, as from what I've seen mediaType traits can indicate that a member marked as a payload should drive what the content-type is set to for a header.

service/smithyprototype/lexruntimeservice/api_op_PostText.go
```diff
@@ -461,3 +467,95 @@ type PostTextResponse struct {
...
+	object := value.Object()
+	defer object.Close()
+
+	if v == nil {
+		return nil
+	}
+
+	if v.InputText != nil {
+		object.Key("inputText").String(*v.InputText)
+	}
```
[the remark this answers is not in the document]
skmcgrail: The string method already takes care of by the encoder.

service/smithyprototype/lexruntimeservice/api_op_PutSession.go
```diff
@@ -389,3 +394,103 @@ type PutSessionResponse struct {
...
+	input, ok := in.Parameters.(*PutSessionInput)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown input parameters type %T", in.Parameters)
+	}
+
+	restEncoder := rest.NewEncoder(request.Request)
+
+	restEncoder.AddHeader("Content-Type").String("application/json")
+
+	if err := serializePutSessionInputAWSREST(input, restEncoder); err != nil {
```
skotambkar, line 422: Should serializing functions be in a separate go file rater than being in operation file? (That was our initial design for serializers)

service/smithyprototype/lexruntimeservice/api_op_PutSession.go
```diff
@@ -389,3 +394,103 @@ type PutSessionResponse struct {
...
+		return out, metadata, err
+	}
```
skotambkar, line 424: Is no call to `restEncoder.Encode()` intentional?

service/smithyprototype/lexruntimeservice/serde.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	restEncoder := rest.NewEncoder(request.Request)
+
+	if err := awsRestJson1_serializeRestDeleteSessionInput(input, restEncoder); err != nil {
+		return out, metadata, err
+	}
+
+	if err := restEncoder.Encode(); err != nil {
+		return out, metadata, err
```
skotambkar, line 43: Should these errors at the top level middleware be wrapped as `SerializationError`?

service/smithyprototype/lexruntimeservice/serde.go
```diff
@@ -0,0 +1,438 @@
...
+
+	input, ok := in.Parameters.(*DeleteSessionInput)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown input parameters type %T", in.Parameters)
+	}
+
+	restEncoder := rest.NewEncoder(request.Request)
+
+	if err := awsRestJson1_serializeRestDeleteSessionInput(input, restEncoder); err != nil {
+		return out, metadata, err
```
skotambkar, line 39: Should these errors at the top level middleware be wrapped as SerializationError?

service/smithyprototype/lexruntimeservice/api_op_PutSession.go
```diff
@@ -390,3 +395,103 @@ type PutSessionResponse struct {
...
+}
+
+func serializePutSessionInputAWSJSON(v *PutSessionInput, encoder json.Value) error {
+	object := encoder.Object()
+	defer object.Close()
+
+	if v == nil {
+		return nil
+	}
+
```
skotambkar, line 444: I believe this was commented before, but wanted to reiterate. 

if v == nil is even possible, there shouldn't be a serializing function generated for that shape.

service/smithyprototype/lexruntimeservice/api_op_PutSession.go
```diff
@@ -390,3 +395,103 @@ type PutSessionResponse struct {
 func (r *PutSessionResponse) SDKResponseMetdata() *aws.Response {
 	return r.response
 }
+
+type putSessionSerializeMiddleware struct{}
+
+// ID is the middleware identifier
+func (p putSessionSerializeMiddleware) ID() string {
+	return "PutSessionSerializeMiddleware"
+}
```
skotambkar, line 404: Would these still be the sane with our new naming scheme?

service/smithyprototype/lexruntimeservice/api_op_PutSession.go
skmcgrail: Will add error to these ones, the previous one was around whether encoding actual json null vs error for other types.

service/smithyprototype/lexruntimeservice/serde.go
skmcgrail: Will update

service/smithyprototype/lexruntimeservice/api_op_PostContent.go
```diff
@@ -7,7 +7,8 @@ import (
 	"io"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	v4 "github.com/aws/aws-sdk-go-v2/aws/protocol/rest"
+	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
```
skotambkar, line 11: Nit

service/smithyprototype/lexruntimeservice/serde.go
```diff
@@ -0,0 +1,542 @@
...
+
+	jsonEncoder := json.NewEncoder()
+	if err := awsRestJson1_serializeJsonPostTextInput(input, jsonEncoder.Value); err != nil {
+		return out, metadata, &aws.SerializationError{Err: err}
+	}
+
+	if request, err = request.SetStream(bytes.NewReader(jsonEncoder.Bytes())); err != nil {
+		return out, metadata, &aws.SerializationError{Err: err}
+	}
+	in.Request = request
```
jasdel, line 161: Does it make sense for teh `json.NewEncoder` to take a buffer to write to?  That would save the need to create another buffer later when calling `SetStream`

jasdel: Then wouldn't need a `Bytes` method on the encoder? it would write to the underlying buffer, similar to std json.Encoder, and similar utilities.

skmcgrail:
> Would need to rebase with master as types has been moved, and that would introduce some nit changes.

This was already rebased with master.

service/smithyprototype/lexruntimeservice/serde.go
skmcgrail: Yeah agreed let's follow that up in a separate PR as that requires changes out side of the scope and intentions here.

Issue #439, opened 2019-11-23
Refactoring API client Paginators for better refactored API Client


jasdel:
This issue details how the v2 AWS SDK for Go's API client Paginators can be updated based on the API client refactor proposed in #438.

Since, the API client operation methods no longer return a built Request, (e.g.  `GetObjectRequest`), the paginator needs be refactored.

## Proposed API Client Paginator Refactor

The proposed refactor of the SDK's clients is split into four parts, API Client, Paginators, Waiters, and Request Presigner helpers. This proposal covers the API client Paginator changes.

The following is an example of the existing S3 ListObjects Paginator.

```go
req := client.GetObjectRequest(params)

p := s3.NewListObjectsPaginator(req)
```

The paginator's constructor would be updated to take an interface for the paginated operation client, in addition to the operation's input parameters.  The constructor would continue to return the operation paginator type, (e.g.  `ListObjectsPaginator`). The paginator's `NextPage`) method would be updated to take functional options for modifying the request per pages, in addition to a `Context` value.

The following is an example of the proposed construction of a S3 ListObjects Paginator.

```go
p := s3.NewListObjectsPaginator(client, params)
```

The following is an example of the updated ListObjectsPaginator's constructor and method signatures. The client parameter is a unnamed interface to prevent additional types defined for each operation paginator and circular package imports with the `api` package where the `ListObjectClient` interface is defined.

```go
// NewListObjectsPaginator returns a paginator for the ListObjects operation
// with the input parameters specified.
func NewListObjectsPaginator(input *ListObjectsInput,
	client interface{
		ListObjects(context.Context, *ListObjectsInput, ...func(*aws.Request)) (*ListObjectsResponse, error)
	},
) *ListObjectsPaginator { 
	// implementation
}

type ListObjectsPaginator struct {
}

func (p *ListObjectsPaginator) NextPage(context.Context, ...func(*aws.Request)) bool {
	// implementation
}

func (p *ListObjectsPaginator) CurrentPage() *ListObjectsResponse {
	// implementation
}

func (p *ListObjectsPaginator) Err() error {
	// implementation
}
```

## Using the refactored paginator

The application creates a paginator for an operation by calling the paginator's constructor, passing in the client and input parameters. Once the application has created a paginator, (e.g.  `ListObjectsPaginator`), the application can iterate over the response pages using the operation paginator's `NextPage`, `CurrentPage`, and `Err` methods.

[...]

Issue #438, opened 2019-11-22
Refactoring API client operation methods for better usability


jasdel:
This issue details how the v2 AWS SDK for Go's API client operations can be updated to reduce the difficulty of mocking the v2 SDK's API clients in your applications. The proposed changes in this design seek to improve the v2 SDK by reducing the complexity of the API client's operation methods, paginators and waiters.

We'd like to hear your feedback on this proposed refactor to the v2 SDK!

## Background

Issues such as #70, highlight the difficulty and pain writing unit test for applications which use the v2 AWS SDK for Go API clients. Specifically due to the v2 SDK API client operation methods returning a concrete structure with a method that must be invoked before a response is returned.

```go
req := client.GetObjectRequest(params)

result, err := req.Send(context.Background())
```

Since the API operation request's build, and send behavior are split into two statements, the application's tests must mock both the API client's operation method, and the request `struct`, (e.g. `GetObjectRequest`). Returning an interfaces instead of a concrete type would not significantly improved the issue, since the application tests would still need to mock out the `interface` type in addition to the client operation method as well.

## Proposed API Client Refactor

The proposed refactor of the SDK's clients is split into four parts, API Client, Paginators, Waiters, and Request Presigner helpers. This proposal covers the API client changes.

The refactor would replace all of the API client operation `Request` builder methods, (e.g. `GetObjectRequest`) with methods that invoke the operation directly. This pattern is similar to the v1 SDK's `WithContext` operation methods, (e.g.  [GetObjectWithContext]).  The new method would initialize the operation's underlying behavior, apply the application's optional customizations, and invoke the operation. The response of the operation would be returned, or error.

The following example invokes the Amazon S3 `GetObject` operation with the operation's `GetObjectResponse` or `error` returned via the `resp`, or `err` variables. The `GetObjectOutput` parameters can be 

```go
client := s3.New(cfg)

resp, err := client.GetObject(context.Background(), params)
```

This pattern enables request customization via [functional options]. Functional options are an idiomatic Go pattern for creating APIs with optional configuration. Common options can be defined by the SDK for applications to customize operation call, (e.g. add HTTP headers).

The following example is similar to the above `GetObject` operation call, with the addition of a custom HTTP header `x-custom-header` with `some value` on the request.

```go
client := s3.New(cfg)

resp, err := client.GetObject(context.Background(), params,
	aws.WithHTTPHeader("x-custom-header","some value"))
```

[...]

Issue #70, opened 2017-12-22
Mocking out new API?


jellevandenhooff:
A major change in the v2 API is the requirement to use Request methods (https://github.com/aws/aws-sdk-go-v2/pull/33). We currently use the `iface` packages to test our code using the v1 API with the strategy described in https://github.com/aws/aws-sdk-go-v2/blob/ac26486d921d336e20fec49e2afb348c801f0003/service/acm/acmiface/interface.go#L42, made slightly nicer using Go's `mockgen`. 

Could you please ensure that this mocking remains possible? Perhaps the Request API can return a Request object that doesn't go out to the web and instead returns a mock Response? Other alternative mock strategies would be equally welcome.

To tack on, the suggestion in https://github.com/aws/aws-sdk-go-v2/issues/50 to split up the `iface` packages would likely make mocking for us more painful as we would have to generate more  interfaces.

jasdel:
Thanks for the feedback @jellevandenhooff. With the change to making the v2 SDK repo public I had to recreate PR #50 as PR #72.

We plan on keeping the ability to mock out API requests. This is an important way of using the SDK that many users, and the SDK's on unit tests, rely on. How the API's are mocked is a little bit different now, and I think there is room for improvement. One way mocking experience could be improved is for the SDK to provide a utility that does just what you mentioned, configures the API Request to return a specific value without actually making a network request.  This utility would also help ensure that any custom handlers installed by the application are exercised. The V1 SDK's mocking ignores custom handlers. 

I'd like to hear more about the impact on your application if the `iface`s are split into different APIs. Could you go into more details the impact this would have?

jasdel:
Thanks again for you feedback @jellevandenhooff lets use #50 and #72 to track this. I've copied your feedback into #50. I'll close this as a duplicate of #50, and we can track the discussion there.

aaron0browne:
@jasdel I think you missed the main portion of this issue, which is that since the new service APIs return request objects, which are in turn used to fetch results, testing behavior requires more than just mocking the service. It also requires a way to mock the request object. Does that make sense?

jasdel:
[...]

PR #530, opened 2020-04-15
Add hand written prototype for updated client API design.

jasdel:
# Generated API client design

Addresses issue #70, #438, #439

## Operation generated types

The following are a set of types that must be generated by smithy for an operation, that are not defined in the model. These types are needed by the API client to wrap and encapsulate the modeled operation types. The Smithy generator must ensure that all novel generated type do not conflict with modeled shapes.

**Operation input**: Smithy must generate a unique input type for each operation, regardless if an input shape is modeled. The naming of the input shape will be <Operation>Input, where <Operation> is the exported name of the operation. The input shape is a renamed clone of the modeled operation’s input shape, if any. If the operation does not have a modeled input shape, the a new empty shape will be created and used as the input shape.

**Operation output**: Smithy must generate a unique output type for each operation, regardless if an output shape is modeled. The name of the output type will be <Operation>Output, where <Operation> is the exported name of the operation. The output shape is a renamed clone of the modeled operation’s output shape, if any. If the operation does not have a modeled output shape, the a new empty output shape will be created and used as the input shape.

**Operation Paginator**: Smithy must generate operation specific paginator helpers for operations with modeled paginators. The paginator is a novel type that provides pagination iterator behavior. The name of the type will be <Operation>Paginator, where <Operation> is the exported name of the operation.

**API Client**: Smithy must generate a type named Client for each API client package. There may only be one API client per package. The client type will have additional configuration option utility types and functions generated as well.

## Client

Smithy will generate a `Client` type in the API client package. This type will have methods for each modeled operation. The client's constructor takes functional options for configuring how the client operates. If customers would like to modify the client's behavior for all operation they can use the Client's functional options to do this.

```go
// Client provides the client for interacting with the Amazon Lex Runtime Service.
type Client struct {
    options Options
}
﻿﻿
// New returns an initialized client with the client options used to specify the behavior
// of the client. The user has the ability to provide the options parameters
// directly without being required to load configuraiton.
func New(Options) *Client { ... }

// NewFromConfig returns an initialized Client based on the passed in config, and
// functional options. Provide additional function options to further configure the
// behavior of the client. Such as changing the client's region, endpoint, or
// adding custom middleware behavior.
func NewFromConfig(cfg aws.Config, opts ...func(*Options)) *Client {...}

// ServiceID returns the name of the identifier for the service API.
func (c *Client) ServiceID() string { return "ServiceID" }

// ServiceName returns the full service name.
func (c *Client) ServiceName() string { return "Service Name" }
```

### Client options

A client options is a type unique to the Client, that defines the set of options that a customer can modify when they create a client via the `New` function.

If the SDK were to generate functional option helpers the `New` function's `opts` parameter could be defined as an exported type that all functional options return. This would ensure functional option helpers are grouped with the option type in API reference docs.

```go
// Options provides the set of configurations that can be applied to the
// Client. Use functional options to modify this set when creating the client,
// or when invoking API operations.
type Options struct {
	// The AWS regional endpoint the client will make API operation calls to.
	// Required by the client.
	RegionID string

	// Behavior to resolve endpoints. Defaults to client's default endpoint
	// resolution if nil.
	EndpointResolver aws.EndpointResolver

	// The HTTP client to invoke API calls with. Defaults to client's default
	// HTTP implementation if nil.
	HTTPClient HTTPClient

	// Credentials will be used to create the client's default signer if one is
	// not provided. If signer is provided Credentials are ignored.
	Credentials aws.CredentialsProvider

	// If nil, will default to client's signer using provided credentials.
	// Credentials must also be set.
	Signer      HTTPSigner
	SigningName string

	// Set of options to modify how an operation is invoked. These apply to all
	// operations invoked for this client. Use functional option on operation
	// call to modify this list for per operation behavior.
	APIOptions []func(*middleware.Stack)

	Retryer  aws.Retryer
	LogLevel aws.LogLevel
	Logger   aws.Logger
}
```

## Operations

The SDK will generate a method on the API Client for each supported operation. The operation will need to take three parameters, context.Context the modeled input shape, and a variadic argument of functional options to modify operation being invoked. 

The functional options allow customers to modify how the individual operation is invoked. If the customer wants to modify all operations of a client, they can use the `Options.APIOptions` member.

```go
// GetSession invokes the GetSession S3 API operation with the input parameters,
// and context provided. The passed in functional options can be used to modify
// how the operation is invoked. A result will be returned on success, or error
// if the operation invoke fails.
func (c *Client) GetSession(
    ctx context.Context, params *GetSessionInput, opts ...func(*Options),
) (
    *GetSessionOutput, error,
) {
    stack := middleware.NewStack("GetSession stack", smithyhttp.NewStackRequst)
    // op specific middleware
    
    // client options applied to stack
    // user func options applied to stack
    // decorate client handler with stack
    
    result, err := handler.Handle(ctx, params)
    if err != nil {
        return nil, err
    }
    return res.(*GetSessionOutput), nil
}
```

### Operation input and output shape

Since Go does not support method overloading, and an operation can be updated to have a modeled input or output shape, all generated operation methods must have a stable input and output type generated. In order to ensure a stable name, this types must be `<Operation>`Input and `<Operation>`Output, where `<Operation>` is the name of the generated operation method.

If a modeled input or output shape in an orphan and is not used by any other operation, or nested within a shape, the original shape should not be generated.

If a modeled input or output shape is reused by another operation, or nested, the original shape should be generated. The generator should also generate copy utilities to copy the cloned input/output shape to and from the original modeled shape. This allows customers to easily convert the input or output shape into its original shape for round tripping.
 
For example, the Lex Runtime `GetSession` operation is modeled with `GetSessionRequest` input shape. Smithy will generated the `GetSessionInput` type with all the same parameters as the modeled `GetSessionRequest` shape. If the `GetSessionRequest` shape is not used by any other shape or operation, it would not be generated. The reason to not generate these unused types is, because it would cause noise, and be confusion.

### Operation output type result metadata

The operation output generated type will have additional members and methods generated. These member's provide access to result metadata such as retry attempts and access to the raw protocol response message.

The operation output type will also have metadata helper methods generated on it such as `ServiceRequestID()`, and `HTTPResponse()`.

For example the Lex Runtime GetSession operation has a modeled output shape of `GetSessionResponse`, which becomes the operation output type of `GetSessionOutput`. The `GetSessionOutput` type will have the additional member `ResultMetadata` generated. This member will be used to store result metadata of invoking the operation.

```go
type GetSessionOutput struct {
	// Members from modeled output shape, GetSessionResponse.
	DialogAction            *types.DialogAction
	RecentIntentSummaryView []types.IntentSummary
	SessionAttributes       map[string]string
	SessionId               *string

	// Additional member for result metadata
	ResultMetadata middleware.Metadata
}
```

## Mocking API Client operations

The Smithy generator will not generate interfaces for the API client. Customers will need to write these interfaces them selves. Depending on customer demand this decision might be changed, but this is a two-way door decision to not generate the API client interfaces. The exception to this are operation clients that are needed for Waiters and Paginators. 

Customers are able to define their own interfaces for operations they use. These interfaces can be composed, adding multiple operation methods to the same interface as they need. This style is the idiomatic style of not defining interfaces for behavior not used by the package.

```go
type GetSessionAPIClient interface {
    GetSession(context.Context, *lexruntimeservice.GetSessionInput,
          ...func(*Options))
}
```

## Paginated operations

Smithy must generate operation specific paginator helpers for operations with modeled paginators. The paginator is a novel type that provides pagination iterator behavior. The name of the type will be `<OperationName>Paginator`.

When a paginator or other utility that needs a client operation, an interface in the client package should be generated for that operation. The interface should be named `<OperationName>APIClient.`

```go
// ListSessionsClient provides the interface for a client that implements the
// ListSessions API operation. Implemented by the package's Client type.
type ListSessionsAPIClient interface {
    ListSessions(context.Context, *ListSessionsInput, ...func(*lexruntimeservice.Options)) (
        *ListSessionsOutput, error,
    )
}
```
The generated paginator utility will take the client operation interface, and input parameters as input.

```go
// ListSessionsPaginator provides the paginator for the ListSessions API operation.
type ListSessionsPaginator struct {
    client ListSessionsAPIClient
    input *ListSessionsInput
    
    // members for pagination state
}

// NewListSessionsPaginator returns a ListSessionsPaginator configured for the
// API operation client, and input parameters.
func NewListSessionsPaginator(
    ListSessionsAPIClient, *ListSessionsInput, ...func(*Options),
) *ListSessionsPaginator

// HasMorePages returns if there may be more pages to retrieve.
func (p *ListSessionsPaginator) HasMorePages() bool

// NextPage returns the next page from the API, or error.
func (p *ListSessionsPaginator) NextPage(context.Context, ...func(*Options)) (
    *ListSessionsOutput, error,
)
```

The following is an example of the paginator being used to iterate over pages.

```go
func ExampleClient_ListSessions_pagination() {
    if err != nil {
        log.Fatalf("unable to load configuration, %v", err)
    }

    client := lexruntimeservice.NewClient(cfg)

    // Create a paginator with the client and input parameters.
    p := lexruntime.NewListSessionsPaginator(client, &lexruntime.ListSessions{
        BotAlias: aws.String("botAlias"),
        BotName:  aws.String("botName"),
        UserId:   aws.String("userID"),
    })

    // Use the paginator to iterate over pages returned by the API. HasMorePages
    // will return false when there are no more pages, or NextPage failed.
    for p.HasMorePages() {
        o, err := p.NextPage(context.TODO())
        if err != nil {
            log.Fatalf("failed to get next page, %v", err)
        }
        fmt.Println("Page:", o)
    }
}
```

**Outstanding Question**: Should HasMorePages continue to return true even if NextPage fails"? The reasoning behind this would be so that in the case of temporary server side error. This would give the user the ability to recover mid pagination, without needing to restart from the beginning, or rebuilding the input parameters with the current token.

### DIFF
--- a/service/smithyprototype/lexruntimeservice/api_client.go
+++ b/service/smithyprototype/lexruntimeservice/api_client.go
@@ -82,6 +82,7 @@ type HTTPClient interface {
 // the behavior of the client.
 func New(opts Options) *Client {
 	// Fill options with default values.
+	// TODO move this to helper function.
 	if opts.EndpointResolver == nil {
 		opts.EndpointResolver = newEndpointResolver()
 	}
@@ -103,6 +104,11 @@ func New(opts Options) *Client {
 // in the AWS Config, and functional options provided.
 func NewFromConfig(cfg aws.Config, optFns ...func(*Options)) *Client {
 	var o Options
+
+	// TODO Copy cfg into options into options value
+	o.RegionID = cfg.Region
+	o.EndpointResolver = cfg.EndpointResolver
+
 	for _, fn := range optFns {
 		fn(&o)
 	}


skmcgrail:
> Operation input: Smithy must generate a unique input type for each operation, regardless if an input shape is modeled.

Might be good to enumerate that we effectivly clone shapes targeted by input operations. This would not affect that shapes usage as part of another member, but would still be treated as effectivly different named shapes but with same members.

skmcgrail:
> 	// This service uses AWS Sigv4a signing, and needs Credentials values.
>    Credentials aws.Credentials
>
>	// Signer is the signer for the client.
>    Signer HTTPSigner

May need to rethink the signer some, as today a signer is constructed with the aws.Credentials, maybe these should be passed to the SignHTTP/PresignHTTP functions instead?

skmcgrail:
> If the GetSessionRequest shape is not used by any other shape, it must not be generated, because it would be unusable, cause noise, and be confusion.

May want to enumerate the keywords we will strip off when adding the `Input` suffix.

For example `Request` => `Input`

skmcgrail:
> Note: An alternative would be to copy the output shape just like the SDK does for the input shape. The metadata would a novel member generated into the output shape. Metadata helpers would need some prefix, (e.g. SDK) to reduce collision potential.

My thought was that Metadata access helpers would be provided by the packages where middleware that produces said artifacts resides. Thoughts? Or were you thinking that we would have more specific metadata for services specifically?

jasdel:
Closing this since prototype is no longer needed. Using smithy to generate based on this design

PR #544, opened 2020-04-24
`smithyprototype`: adds prototype for deserializer middleware

skotambkar:
Rev 1
---
* Adds prototyped lex-runtime service operations with deserializer middleware
* This PR uses the new SDK layout, and middleware/(de)serializer naming scheme

### DIFF
--- a/service/smithyprototype/lexruntimeservice/deserialize.go
+++ b/service/smithyprototype/lexruntimeservice/deserialize.go
@@ -1,0 +1,1459 @@
+package lexruntimeservice
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/sdkio"
+	"github.com/aws/aws-sdk-go-v2/service/smithyprototype/lexruntimeservice/types"
+	"github.com/awslabs/smithy-go/middleware"
+	smithyHTTP "github.com/awslabs/smithy-go/transport/http"
+)
+
+type awsRestJson1_deserializeOpPutSession struct{}
+
+// ID is the middleware identifier
+func (p awsRestJson1_deserializeOpPutSession) ID() string {
+	return "awsRestJson1_deserializeOpPutSession"
+}
+
+func (p awsRestJson1_deserializeOpPutSession) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (
+	out middleware.DeserializeOutput, metadata middleware.Metadata, err error,
+) {
+	out, metadata, err = next.HandleDeserialize(ctx, in)
+	if err != nil {
+		return out, metadata, err
+	}
+	response, ok := out.RawResponse.(*smithyHTTP.Response)
+	if !ok {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("unknown transport type %T", out.RawResponse),
+		}
+	}
+
+	output, ok := out.Result.(*PutSessionOutput)
+	if !ok {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("unknown output result type %T", out.Result),
+		}
+	}
+
+	if err := awsRestJson1_deserializeRestPutSessionOutput(output, response); err != nil {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("failed to decode response with invalid REST bindings, %w", err),
+		}
+	}
+
+	// TODO: add error response deserializer function
+
+	// body of type blob
+	if err := awsRestJson1_deserializeJsonPutSessionOutput(output, response.Body); err != nil {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("failed to decode response body with invalid JSON, %w", err),
+		}
+	}
+
+	return out, metadata, err
+}
+
+func awsRestJson1_deserializeRestPutSessionOutput(v *PutSessionOutput, response *smithyHTTP.Response) error {
+	if v == nil {
+		return fmt.Errorf("unsupported deserialization of nil %T", v)
+	}
+
+	if v.ContentType != nil {
+		val := response.Header.Get("Content-Type")
+		v.ContentType = &val
+	}
+
+	if len(v.DialogState) > 0 {
+		val := response.Header.Get("x-amz-lex-dialog-state")
+		v.DialogState = types.DialogState(val)
+	}
+
+	if v.IntentName != nil {
+		val := response.Header.Get("x-amz-lex-intent-name")
+		v.IntentName = &val
+	}
+
+	if v.Message != nil {
+		val := response.Header.Get("x-amz-lex-message")
+		v.Message = &val
+	}
+
+	if len(v.MessageFormat) > 0 {
+		val := response.Header.Get("x-amz-lex-message-format")
+		v.MessageFormat = types.MessageFormatType(val)
+	}
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ LINTIGNOREDEPS='vendor/.+\.go'
 LINTIGNOREPKGCOMMENT='service/[^/]+/doc_custom.go:.+package comment should be of the form'
 LINTIGNOREENDPOINTS='aws/endpoints/defaults.go:.+(method|const) .+ should be '
 LINTIGNORESINGLEFIGHT='internal/sync/singleflight/singleflight.go:.+error should be the last type'
+LINTIGNOREPROTO="service/smithyprototype/.+\.go:.+don't use underscores in Go names"
```
jasdel, line 11: Rebasing with feature/SmithyAPIClient should take care of these changes.

service/smithyprototype/lexruntimeservice/deserialize.go
```diff
@@ -0,0 +1,1444 @@
...
+			Err: fmt.Errorf("unknown transport type %T", out.RawResponse),
+		}
+	}
+
+	output, ok := out.Result.(*PutSessionOutput)
+	if !ok {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("unknown output result type %T", out.Result),
+		}
+	}
```
jasdel, line 43: We'll want to consider which middleware will populate `out.Result` with the operation's output type. The operation serialize probably is a good opportunity for that. Alternatively a separate deserialize middleware that does nothing more than set the `out.Result` member to an empty op output type value.

service/smithyprototype/lexruntimeservice/deserialize.go
```diff
@@ -0,0 +1,1444 @@
...
+		val := response.Header.Get("x-amz-lex-slots")
+		b, err := base64.StdEncoding.DecodeString(val)
+		if err != nil {
+			return err
+		}
+		m := aws.JSONValue{}
+		err = json.Unmarshal(b, &m)
+		if err != nil {
+			return err
+		}
```
jasdel, line 125: not a part of this PR but this is a good example how the SDK's current `aws.JSONValue` could be better. There is no way for a user to serialize the value into their own type, or provide their own serialize. When we end up replacing `JSONValue` we probably want a type that is lazy deserialize. This would give applications opportunity to deserialize the value into their own type.

service/smithyprototype/lexruntimeservice/deserialize.go
```diff
@@ -0,0 +1,1444 @@
...
+			Err: fmt.Errorf("failed to decode response with invalid REST bindings, %w", err),
+		}
+	}
+
+	// body of type blob
+	if err := awsRestJson1_deserializeJsonPutSessionOutput(output, response.Body); err != nil {
+		return out, metadata, &aws.DeserializationError{
+			Err: fmt.Errorf("failed to decode response body with invalid JSON, %w", err),
+		}
+	}
```
jasdel, line 56: Nit outside of the scope for this PR, but should would be handy to add a TODO before the `awsRestJson1_deserializeRestPutSessionOutput` for error deserializing.

service/smithyprototype/lexruntimeservice/deserialize.go
```diff
@@ -0,0 +1,1444 @@
...
+
+	buff := make([]byte, 1024)
+	ringBuffer := sdkio.NewRingBuffer(buff)
+
+	// wrap a TeeReader to read from response body & write on snapshot
+	body := io.TeeReader(response.Body, ringBuffer)
+	defer response.Body.Close()
+	decoder := json.NewDecoder(body)
+
+	if err := awsRestJson1_deserializeJsonPostTextOutput(output, decoder); err != nil {
```
jasdel, line 177: probably also want to configure the `json.Decoder` to decode all numbers as `json.Number` type. so we don't have to worry about it decoding them as float vs int. The generated SDK's deserialize logic can take care of converting the `json.Number` to the modeled type.

https://pkg.go.dev/encoding/json?tab=doc#Decoder.UseNumber

service/smithyprototype/lexruntimeservice/serializers.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+
+	jsonEncoder := json.NewEncoder()
+	if err := awsRestJson1_serializeJsonPutSessionInput(input, jsonEncoder.Value); err != nil {
+		return out, metadata, &aws.SerializationError{Err: err}
+	}
+
+	if request, err = request.SetStream(bytes.NewReader(jsonEncoder.Bytes())); err != nil {
+		return out, metadata, &aws.SerializationError{Err: err}
+	}
```
jasdel, line 201: Should `json.NewEncoder` take a buffer to write to instead of having its own internal buffer?  This way that buffer can be sent to set stream directly without creating the new `bytes.Reader`?  This would also allow us to write to any underlying io.Writer, making the Encoder more flexible, and code gen could pre-allocate estimated size preventing multiple future allocations during serialization.

```go
// could optimistically pre-allocate a buffer if we have rough idea of size, prevent multiple
// allocations later.
streamBuf := bytes.NewBuffer(nil)

jsonEncoder := json.NewEncoder(streamBuf)
if err := awsRestJson1_serializeJsonPutSessionInput(input, jsonEncoder.Value); err != nil {
	return out, metadata, &aws.SerializationError{Err: err}
}

if request, err = request.SetStream(streamBuf); err != nil {
	return out, metadata, &aws.SerializationError{Err: err}
}
```

skotambkar: This feedback concerns serializer implementation. @skmcgrail can you take a look?

service/smithyprototype/lexruntimeservice/deserialize.go
skotambkar: out.Result is an `interface{}`and here we are just type casting it to `*PutSessionOutput` before passing it to protocol deserializer funcs.

If you were talking about populating `out.RawResponse`, then I think a seperate deserialize middleware or a client middleware that actually takes in the serialized request, send it to server and returns a response can handle that.

service/smithyprototype/lexruntimeservice/deserialize.go
skotambkar: That can be added as a TODO but probably not here.

service/smithyprototype/lexruntimeservice/deserialize.go
skotambkar: Will update to use Json.Number instead; however that is an interface{} instead of float64 and will need type casting everytime, isn't that more expensive than type casting if int64?

skotambkar: Will update to use json.Number

service/smithyprototype/lexruntimeservice/deserialize.go
jasdel: Cool in that case, yes a separate middleware added by the operation building the stack would need to take care of populating the Result member with a empty output type value.

skotambkar: I don't see a reason to keep the middleware that set's the `out.Result` separate. If in case some customization needs it in the future, we can add the middleware, we can change this. 

I will go ahead and set the type in this middleware, instead of trying to type assert it.

skotambkar: Updated the deserializer middleware to initialize the `out.Result` with output type value. This did not need to be a separate middleware. Initializing with an empty output type value ensures that we overwrite any existing value in `out.Result`

jasdel:
Dropping hand written prototype PRs in favor of code generation ones

PR #559, opened 2020-05-05
 Prototype Smithy generated protocol tests

jasdel:
Adds a set of example protocol tests that would be generated based on
the Smithy protocol tests model.

Note: this PR is based on the generated API client prototype, and includes several changes that will be resolved when the API client prototype PR is merged.

adds `api_errors_test.go` and `api_op_GetSession_test.go` as example generated protocol tests


jasdel:
Closing prototype as code is being generated for smithy models.

PR #562, opened 2020-05-07
service/dynamodb/expression: fix empty expression returned when unset

### LINKED ISSUE #554: [bug] service/dynamodb/expression#Expression.returnExpression does not return nil if unset
  
**Describe the bug**
The code here is incorrect: https://github.com/aws/aws-sdk-go-v2/blob/241222ec3c705bcc2b729dfe8d0cc4e01bd70b80/service/dynamodb/expression/expression.go#L487-L492

It should be:
```
if e.expressionMap == nil {
  return nil
}
if _, ok := e.expressionMap[expressionType]; !ok {
  return nil
}
return aws.String(e.expressionMap[expressionType])
```
This will make the behavior consistent with the documentation. For example, `Filter()` docs specify that it returns null if the filter doesn't exist:
```
// Filter returns the *string corresponding to the Filter Expression of the
// argument Expression. This method is used to satisfy the members of DynamoDB
// input structs. If the Expression does not have a filter expression this
// method returns nil.
//
// Example:
//
//     // let expression be an instance of Expression{}
//
//     queryInput := dynamodb.QueryInput{
//       KeyConditionExpression:    expression.KeyCondition(),
//       FilterExpression:          expression.Filter(),
//       ExpressionAttributeNames:  expression.Names(),
//       ExpressionAttributeValues: expression.Values(),
//       TableName: aws.String("SomeTable"),
//     }
func (e Expression) Filter() *string {
	return e.returnExpression(filter)
}
```

With the way the code is currently written, calls to dynamo with an unset filter (but another set expression, e.g. projection) will fail with: `ValidationException: Invalid FilterExpression: The expression can not be empty`

Example: v1.29.22
* get SDK version by printing the output of `aws.SDKVersion` in your code after importing `"github.com/aws/aws-sdk-go-v2/aws"`

go version go1.13.7 darwin/amd64

Described above

**Expected behavior**
Described above

**Additional context**

alextbok:
Fixes a big in the expression builder that returns an empty string expression value when the expression has not been set.

Fixes #554


alextbok:
@jasdel @skotambkar another small PR for you. This time a bug fix.

ref https://github.com/aws/aws-sdk-go-v2/issues/554

alextbok:
Bump @jasdel @skotambkar

jasdel:
Hi @alextbok  thanks for taking the time to create this PR. We'll review this and update the issue with feedback. 

From initial glance this looks good. DynamoDB recently made updates to their API to allow empty maps/lists/strings in their APIs. Will need to check and see if that impacts the Filter behavior as well.

PR #568, opened 2020-05-27
Adds RestJson deserializer and middleware generation support

skotambkar:
*Description of changes*

* Adds Rest Json deserializer generation support 
* Adds Rest Json deserializer middleware generation support

Related to awslabs/smithy-go#48

### DIFF
--- a/aws/protocol/json/decoder_util.go
+++ b/aws/protocol/json/decoder_util.go
@@ -34,6 +34,7 @@ func GetErrorInfo(decoder *json.Decoder) (errorType string, message string, err 
 
 		switch st := t.(string); {
 		case strings.EqualFold(st, "code"):
+			fallthrough
 		case strings.EqualFold(st, "__type"):
 			target = &errorType
 		case strings.EqualFold(st, "message"):
@@ -58,6 +59,9 @@ func GetErrorInfo(decoder *json.Decoder) (errorType string, message string, err 
 	if t, ok := endToken.(json.Delim); !ok || t.String() != "}" {
 		return "", "", fmt.Errorf("expected end token to be }")
 	}
+
+	// sanitize error
+	errorType = SanitizeErrorCode(errorType)
 	return errorType, message, nil
 }
 
@@ -111,17 +115,17 @@ func DiscardUnknownField(decoder *json.Decoder) error {
 // Takes in json decoder, and error Code string as args. The function retrieves error message
 // and error code from the decoder body. If errorCode of length greater than 0 is passed in as
 // an argument, it is used instead.
-func GetSmithyGenericAPIError(decoder *json.Decoder, errorCode string) (smithy.GenericAPIError, error) {
+func GetSmithyGenericAPIError(decoder *json.Decoder, errorCode string) (*smithy.GenericAPIError, error) {
 	errorType, message, err := GetErrorInfo(decoder)
 	if err != nil {
-		return smithy.GenericAPIError{}, err
+		return nil, err
 	}
 
 	if len(errorCode) == 0 {
-		errorCode = SanitizeErrorCode(errorType)
+		errorCode = errorType
 	}
 
-	return smithy.GenericAPIError{
+	return &smithy.GenericAPIError{
 		Code:    errorCode,
 		Message: message,
 	}, nil
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
@@ -469,6 +469,8 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
 
         // checks if response has an error and retrieve the error code from the response
         writer.openBlock("if response.StatusCode < 200 || response.StatusCode >= 300 {", "}", () -> {
+            // Retrieve error shape name from response. For REST JSON protocol, the error shape name can be found either
+            // at Header `X-Amzn-Errortype` or a body field with the name `code`, or a body field named `__type`.
             writer.write("errorType := response.Headers.Get($S)","X-Amzn-Errortype" );
             writer.write("errorType = jsonprotocol.SanitizeErrorCode(errorType)");
             writer.write("");
@@ -506,8 +508,6 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
             writer.write("defer response.Body.Close()");
             writer.write("");
 
-            // Retrieve error shape name from response. For REST JSON protocol, the error shape name can be found either
-            // at Header `X-Amzn-Errortype` or a body field with the name `code`, or a body field named `__type`.
             writer.addUseImports(GoDependency.JSON);
             writer.write("decoder := json.NewDecoder(body)");
             writer.write("decoder.UseNumber()");
@@ -515,6 +515,9 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
 
             writer.write("var errorMessage string");
             writer.addUseImports(GoDependency.AWS_JSON_PROTOCOL_ALIAS);
+
+            // If errorType is empty, look for error type in a body field with the name `code`,
+            // or a body field named `__type`.
             writer.openBlock("if len(errorType) == 0 {", "}", () -> {
                 writer.write("errorType, errorMessage, err = jsonprotocol.GetErrorInfo(decoder)");
                 writer.openBlock("if err != nil {", "}", () -> {
@@ -525,7 +528,6 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
                         writer.write("Snapshot: snapshot.Bytes(),");
                     });
                 });
-                writer.write("errorType = jsonprotocol.SanitizeErrorCode(errorType)");
             });
 
             writer.write("");


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -373,4 +372,764 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+        String funcName = isOutputShape? ProtocolGenerator.getDocumentDeserializerFunctionName(shape, getProtocolName())
+                : ProtocolGenerator.getDocumentDeserializerFunctionName(shape, getProtocolName());
+
+        // since payload trait can only be applied to a single member in a shape
+        MemberShape memberShape = shape.members().iterator().next();
+        String memberName = symbolProvider.toMemberName(memberShape);
+        Shape targetShape = model.expectShape(memberShape.getTarget());
+        if (
+                targetShape.getType().equals(ShapeType.STRING) ||
+                        targetShape.getType().equals(ShapeType.BLOB)
```
JordonPhillips, line 692: You can do things like `shape.isStringShape()`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -373,4 +372,764 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            Boolean isOutputShape
+    ) {
+        Symbol jsonDecoder = SymbolUtils.createPointableSymbolBuilder("Decoder", GoDependency.JSON).build();
+        Symbol shapeSymbol = symbolProvider.toSymbol(shape);
+        String functionName = isOutputShape?
+                ProtocolGenerator.getDocumentOutputDeserializerFunctionName(shape, getProtocolName()) :
+                ProtocolGenerator.getDocumentDeserializerFunctionName(shape, getProtocolName());
+
+        writer.addUseImports(GoDependency.FMT);
+        switch (shape.getType()) {
```
JordonPhillips, line 739: switch statements should always have a default, even if it's just break

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -373,4 +372,764 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            SymbolProvider symbolProvider,
+            MemberShape memberShape
+    ) {
+        String memberName = symbolProvider.toMemberName(memberShape);
+        Shape targetShape = model.expectShape(memberShape.getTarget());
+        Symbol targetSymbol = symbolProvider.toSymbol(targetShape);
+        writer.write("val, err := decoder.Token()");
+        writer.write("if err != nil { return err }");
+        writer.write("st, ok := val.(string)");
+        writer.openBlock("if !ok {", "}", () -> {
```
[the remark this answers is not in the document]
skotambkar: True, if the service incorrectly returns `123` for a member modeled as string. We would incorrectly parse the number as `"123"`. I don't see a clear way to handle this at the moment. There is a relevant [thread](https://github.com/golang/go/issues/34472) at golang/go, which is in decision pending state. 

Adding additional checks seems over optimizing, this should be handled by the service team contract to send correctly formatted data?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -373,4 +372,764 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+        Symbol targetSymbol = symbolProvider.toSymbol(targetShape);
+
+        writer.write("bs := $P{}", targetSymbol);
+        writer.write("err := decoder.Decode(&bs)");
+        writer.write("if err != nil { return err }");
+        return "bs";
+    }
+
+    // Generates deserializer for delegator for structure member shape.
+    private String generateDocumentBindingStructureMemberDeserializer(
```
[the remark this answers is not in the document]
skotambkar: There's a slight difference between the return type for Structure member deserializer vs collection/map member deserializer.

Can combine the Collection member and Map member deserializer methods

aws/protocol/json/decoder_util.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		switch t {
+		case "code":
+			v, err := decoder.Token()
+			if err != nil {
+				return "", "", err
+			}
+			code = v.(string)
+			break
```
jasdel, line 43: no extra breaks needed in Go because it has a sane switch behavior ;)

aws/protocol/json/decoder_util.go
```diff
@@ -0,0 +1,123 @@
...
+		case "message":
+			v, err := decoder.Token()
+			if err != nil {
+				return "", "", err
+			}
+			msg = v.(string)
+			break
+		case "__type":
```
jasdel, line 51: decent amount of duplicated code here between these values. can this be collapsed a bit to share some of the logic?  Using a `continue` in the default case will help allow the bottom of the `for` statement to be used for generic logic.  Especially since the `v` decoded is always used as a string.


```go
for decoder.More() {
     var target *string
     switch t {
     case strings.EqualFold("code"):
              target = &code

     case strings.EqualFold("__type"):
              target = &typeCode

     case strings.EqualFold("message"):
              target = &msg

    default:
           DiscardUnknownField(decoder)
           continue
     }

     v, err := decoder.Token()
     if err != nil { return "", "", err }
     
     *target = v.(string)
}

File: aws/protocol/json/decoder_util.go [hunk 7]
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return code, msg, nil
+}
+
+// SanitizeErrorCode sanitizes the errorCode string .
+// The rule for sanitizing is if a `:` character is present, then take only the
+// contents before the first : character in the value.
+// If a # character is present, then take only the contents after the
+// first # character in the value.
+func SanitizeErrorCode(errorCode string) string {
```
jasdel, line 84: Should these sanitize utilities be used by the above `GetErrorInfo` function?

aws/protocol/json/decoder_util.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil{
+			return err
+		}
+		if _, ok := endToken.(json.Delim); !ok {
+			return fmt.Errorf("invalid JSON : expected json delimiter, found %T %v",
+				endToken, endToken)
+		}
+	}
+
+	return err
```
jasdel, line 122: should this be nil?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -299,16 +304,15 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
 
             Optional<MediaTypeTrait> mediaTypeTrait = payloadShape.getTrait(MediaTypeTrait.class);
             mediaTypeTrait.ifPresent(typeTrait -> writer.write("restEncoder.SetHeader(\"Content-Type\").String($S)",
                     typeTrait.getValue()));
 
-            if (shapeType == ShapeType.BLOB) {
+            if (payloadShape.isBlobShape()) {
                 writer.write("documentPayload = input.$L", memberName);
-            } else if (shapeType == ShapeType.STRING) {
+            } else if (payloadShape.isStringShape()) {
```
jasdel, line 315: Curious why `isBlobShape` method vs ShapeType.BLOB?  is one better to use than the other?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            }
+
+            writer.write("buff := make([]byte, 1024)");
+            writer.write("ringBuffer := sdkio.NewRingBuffer(buff)");
+            writer.write("");
+
+            writer.write("var errorBody bytes.Buffer");
+
+            writer.addUseImports(GoDependency.IO);
+            writer.write("_, err := io.Copy(errorBody, response.Body)");
```
jasdel, line 505: nit "io.Copy(`&`errorBody, response.Body)"

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+                writer.write("io.Copy(&snapshot, ringBuffer)");
+                writer.openBlock("return out, metadata, &aws.DeserializationError {", "}", () -> {
+                    writer.write("Err: fmt.Errorf(\"failed to decode response error with invalid JSON, %w\", err),");
+                    writer.write("Snapshot: snapshot.Bytes(),");
+                });
+            });
+
+            writer.openBlock("if len(errorType) == 0 {", "}", () -> {
+                writer.write("errorType = response.Headers.Get($S)", "X-Amzn-Errortype");
+            });
```
jasdel, line 534: Nit should check the `X-Amzn-Errortype` header first. If the code/type exist there, you don't need to double decode the error body.  The current impl will double decode the body even if the error code can be found in the header.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            GoWriter writer,
+            MemberShape memberShape
+    ) {
+        writer.write("val, err := decoder.Token()");
+        writer.write("if err != nil { return err }");
+        writer.write("nt, err := val.(json.Number).Int64()");
+        writer.write("if err != nil { return err }");
+        switch (memberShape.getType()) {
+            case BYTE:
+                writer.write("st := byte(nt)");
```
jasdel, line 990: nit this should be `int8` smithy's Bytes are signed 8 bit integers, SDK will render shape as int8 type.  Go's `byte` is a `uint8`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            case INTEGER:
+                writer.write("st := int32(nt)");
+                break;
+            case LONG:
+                writer.write("st := nt");
+                break;
+            case BIG_INTEGER:
+                writer.addUseImports(GoDependency.BIG);
+                writer.write("st := big.NewInt(nt)");
+                break;
```
jasdel, line 1003: It might be safer for big int to use `i, ok := new(big.Int).SetString(val.(string), 10)` instead of truncating to int64.

This will prevent unexpected behavior and aligns with the SDK's serialization strategy with regard to JSON and bigs.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            case FLOAT:
+                writer.write("st := float32(nt)");
+                break;
+            case DOUBLE:
+                writer.write("st := nt");
+                break;
+            case BIG_DECIMAL:
+                writer.addUseImports(GoDependency.BIG);
+                writer.write("st := big.NewFloat(nt)");
+                break;
```
jasdel, line 1029: It might be safer for big int to use `f, ok := big.ParseFloat(val.(string), 10, 200, big.ToNearestAway)` instead of truncating to float64.

This will prevent unexpected behavior and aligns with the SDK's serialization strategy with regard to JSON and bigs.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -374,4 +374,762 @@ private String getSerializedMemberName(MemberShape memberShape) {
...
+            GoWriter writer,
+            MemberShape memberShape
+    ) {
+        writer.write("val, err := decoder.Token()");
+        writer.write("if err != nil { return err }");
+        writer.write("nt, err := val.(json.Number).Int64()");
+        writer.write("if err != nil { return err }");
+
+        writer.addUseImports(GoDependency.TIME);
+        writer.write("ts := time.Unix(nt, 0).UTC()");
```
jasdel, line 1046: Nit this should use the new `smithytime.ParseEpochSeconds(val.(string))` utility. Will also ensure that it handles fractional seconds too.

jasdel: Also if smithytime.ParseEpochSeconds is updated to return UTC the extra UTC call here won't be needed.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
skotambkar: Even if we get the error type from header, we still need to look for message in the json body. So decided to look for message, and error code in json body first, as that seems more optimal.

aws/protocol/json/decoder_util.go
skotambkar: Want to keep it separate as we might find error type in X-Amzn-Errortype, which would also need to be sanitized.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
skotambkar: Just syntactical sugar.

aws/protocol/json/decoder_util.go
```diff
@@ -28,37 +26,28 @@ func GetErrorInfo(decoder *json.Decoder) (string, string, error) {
...
-				return "", "", err
-			}
-			typeCode = v.(string)
-			break
+		switch st := t.(string); {
+		case strings.EqualFold(st, "code"):
+		case strings.EqualFold(st, "__type"):
+			target = &errorType
+		case strings.EqualFold(st, "message"):
+			target = &message
```
jasdel, line 40: :) nit, need to use fallthrough, or comma separated case list.  

```go
case strings.EqualFold(std, "code"),
     strings.EqualFold(st, "__type"):
     target = &errorType
```

aws/protocol/json/decoder_util.go
```diff
@@ -119,5 +104,25 @@ func DiscardUnknownField(decoder *json.Decoder) error {
...
 
-	return err
+	return nil
+}
+
+// GetSmithyGenericAPIError returns smithy generic api error and an error interface.
+// Takes in json decoder, and error Code string as args. The function retrieves error message
+// and error code from the decoder body. If errorCode of length greater than 0 is passed in as
+// an argument, it is used instead.
+func GetSmithyGenericAPIError(decoder *json.Decoder, errorCode string) (smithy.GenericAPIError, error) {
```
jasdel, line 114: I think smithy.GenericAPIError needs to be a pointer to since its Error method is pointer receiver

aws/protocol/json/decoder_util.go
```diff
@@ -119,5 +104,25 @@ func DiscardUnknownField(decoder *json.Decoder) error {
...
+	errorType, message, err := GetErrorInfo(decoder)
+	if err != nil {
+		return smithy.GenericAPIError{}, err
+	}
+
+	if len(errorCode) == 0 {
+		errorCode = SanitizeErrorCode(errorType)
+	}
```
jasdel, line 122: It looks like the error code value is sanitized in a few spots. Not a big deal, but this could probably be cleaned up a bit, Maybe the utility that does the GetErrorInfo might be a better spot.

PR #578, opened 2020-06-03
 fixups for generated import usage 

jasdel:
Updates smithy API client generation for AWS SDK to use updated import path and pkg names. Depends on Smithy PR https://github.com/awslabs/smithy-go/pull/59


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -620,9 +629,11 @@ protected void writeMiddlewareErrorDeserializer(
             writer.openBlock("if len(errorType) == 0 {", "}", () -> {
                 writer.write("errorType, errorMessage, err = restjson.GetErrorInfo(decoder)");
                 writer.openBlock("if err != nil {", "}", () -> {
+                    writer.addUseImports(GoDependency.BYTES);
```
skotambkar, line 632: Thanks, had missed it! :)

PR #581, opened 2020-06-04
fixup API client code generation compile errors for type assignments.

jasdel:
* fixup bytes and smithyjson dep references
* fixup API client code generation compile errors for type assignments. 
* add stubbing out output var in case unused


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -683,6 +685,7 @@ private void writeErrorShapeDeserializerDelegator(GoWriter writer, Model model,
             writer.openBlock("if errorType == $S {", "}", errorShapeId.getName(), () -> {
                 writer.write("errResult := &$T{}", errorSymbol);
                 writer.write("output := errResult");
+                writer.write("_ = output");
```
skotambkar, line 688: what is this used for?

jasdel: yeah this is needed because it is possible for the `output` member never to be referenced. 

```go
		if errorType == "FooError" {
			errResult := &types.FooError{}
			output := errResult
			_ = output

			response.StatusCode = 500
			response.Status = http.StatusText(500)

			return out, metadata, errResult
		}
```

PR #586, opened 2020-06-10
`codegen`: adds operation middleware to middleware stack

skotambkar:
* Adds Operation middleware for retry, signer into middleware stack
* Related to awslabs/smithy-go#69

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
@@ -38,7 +38,7 @@ public class AssembleMiddlewareStack implements GoIntegration{
     ) {
         String st = String.format("%s.%s.Add(%s, middleware.After)", stackOperand, stackstep, content);
         writer.write(st, middlewareSymbol);
-        writer.addUseImports(AwsGoDependency.AWS_MIDDLEWARE);
+        writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
     }
 
     /**
@@ -58,7 +58,7 @@ public class AssembleMiddlewareStack implements GoIntegration{
     ) {
         String st = String.format("%s.%s.Add(%s, middleware.After)", stackOperand, stackstep, content);
         writer.write(st, middlewareName);
-        writer.addUseImports(AwsGoDependency.AWS_MIDDLEWARE);
+        writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
     }
 
     /**
@@ -78,7 +78,7 @@ public class AssembleMiddlewareStack implements GoIntegration{
     ) {
         String st = String.format("%s.%s.Add(%s, middleware.Before)", stackOperand, stackstep, content);
         writer.write(st, middlewareSymbol);
-        writer.addUseImports(AwsGoDependency.AWS_MIDDLEWARE);
+        writer.addUseImports(SmithyGoDependency.SMITHY_MIDDLEWARE);
     }
 
     @Override
@@ -92,7 +92,7 @@ public class AssembleMiddlewareStack implements GoIntegration{
                                 "RequestInvocationIDMiddleware", AwsGoDependency.AWS_MIDDLEWARE
                         ).build();
 
-                        writeAddMiddlewareAfter(writer, INITIALIZE_MIDDLEWARE,
+                        writeAddMiddlewareAfter(writer, BUILD_MIDDLEWARE,
                                 RequestInvocationIDMiddleware, "$T{}", stackOperand);
                     }).build(),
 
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -0,0 +1,145 @@
...
+                "NewSignHTTPRequestMiddleware", AwsGoDependency.AWS_V4SIGNER_MIDDLEWARE
+        ).build();
+
+        return ListUtils.of(
+                // Add RequestInvocationIDMiddleware to operation stack
+                RuntimeClientPlugin.builder()
+                    .buildMiddlewareStack((writer, service, operation, protocolGenerator, stackOperand) -> {
+                        writer.write("$L.Initialize.Add($T{}, $T)", stackOperand,
+                                RequestInvocationIDMiddleware, middlewareStepAfter);
+                    }).build(),
```
jasdel, line 60: Not saying this needs to change, but from a readability point of view, this style with everything predefined as symbols personally is much more difficult to read vs more inline code. 

It might help to go a step further with the abstraction had have helpers for the `writer.write` behavior using enums for the `Initialize`, etc statements, because from the looks of these this is the only things that are really changing between several of these statements.

The `After` could also be an enum too. or the helper already does this, e.g. 

```
writeAddMiddlewareAfter(writer, INITIALIZE_MIDDLEWARE, "&$L{}, middlewareSymbol)
````

File: codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java [hunk 2]
```diff
@@ -0,0 +1,145 @@
...
+                "UnsignedPayloadMiddleware", AwsGoDependency.AWS_V4SIGNER_MIDDLEWARE
+        ).build();
+
+        Symbol computePayloadSHA256Middleware = SymbolUtils.createValueSymbolBuilder(
+                "ComputePayloadSHA256Middleware", AwsGoDependency.AWS_V4SIGNER_MIDDLEWARE
+        ).build();
+
+        Symbol newSignHTTPRequestMiddleware = SymbolUtils.createValueSymbolBuilder(
+                "NewSignHTTPRequestMiddleware", AwsGoDependency.AWS_V4SIGNER_MIDDLEWARE
+        ).build();
```
jasdel, line 52: Minor styling preference, but what do you think about moving these into the context of where they are used, instead of all listed above?  Or instead of doing the symbol, just using the `addUsedImport` statement, might be easier to read.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -0,0 +1,145 @@
...
+        Symbol middlewareStepBefore = SymbolUtils.createValueSymbolBuilder(
+                "Before", SmithyGoDependency.SMITHY_MIDDLEWARE
+        ).build();
+
+        Symbol RequestInvocationIDMiddleware = SymbolUtils.createValueSymbolBuilder(
+                "RequestInvocationIDMiddleware", AwsGoDependency.AWS_MIDDLEWARE
+        ).build();
+
+        Symbol attemptClockSkewMiddleware = SymbolUtils.createValueSymbolBuilder(
+                "AttemptClockSkewMiddleware", AwsGoDependency.AWS_MIDDLEWARE
```
jasdel, line 31: Looks like the Smithy or AWS middleware pkgs, need to define an alias because these are not being imported correct. Both are being imported as the "middleware" package which is causing the generated code to only include one of the import paths. an import alias on one of the two should fix the issue.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
jasdel, line 60: The RequestInvocationIDMiddleware is a Build middleware

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -0,0 +1,145 @@
...
+                                    attemptClockSkewMiddleware, middlewareStepAfter);
+                        })
+                        .build(),
+
+                // Add newAttempt middleware to operation stack
+                RuntimeClientPlugin.builder()
+                        .buildMiddlewareStack((writer, service, operation, protocolGenerator, stackOperand) -> {
+                            writer.write("$L.Finalize.Add($T(options.Retryer), $T)", stackOperand,
+                                    newAttemptMiddleware, middlewareStepAfter);
+                        })
```
jasdel, line 101: the `NewAttemptMiddleware` shoudl take the request cloner as the second parameter, Doesn't look like the smithy-go/transport/http pkg has this implemented, but it probably should and be used here.

PR #591, opened 2020-06-13
aws.Config Constructor for Smithy Clients

skmcgrail:
Generate aws.Config universal configuration options on service clients options. Solution can be extended to easily add specific parameters to the service client options that should be mapped from the aws.Config structure.

Example: https://gist.github.com/skmcgrail/ebec7045825f7073f234559102846473

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
@@ -15,26 +15,34 @@
 
 package software.amazon.smithy.aws.go.codegen;
 
-import java.util.Optional;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.function.Consumer;
-import java.util.stream.Collectors;
+import java.util.logging.Logger;
+import software.amazon.smithy.aws.traits.auth.SigV4Trait;
 import software.amazon.smithy.codegen.core.Symbol;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.GoSettings;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SymbolUtils;
 import software.amazon.smithy.go.codegen.TriConsumer;
+import software.amazon.smithy.go.codegen.integration.ConfigField;
 import software.amazon.smithy.go.codegen.integration.GoIntegration;
+import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.ServiceIndex;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.utils.ListUtils;
 import software.amazon.smithy.utils.SetUtils;
-import software.amazon.smithy.utils.SmithyBuilder;
 
 /**
  * Registers additional AWS specific client configuration fields
  */
 public class AddAwsConfigFields implements GoIntegration {
+    private static final Logger LOGGER = Logger.getLogger(AddAwsConfigFields.class.getName());
+
     private static final String REGION_CONFIG_NAME = "Region";
     private static final String CREDENTIALS_CONFIG_NAME = "Credentials";
     private static final String ENDPOINT_RESOLVER_CONFIG_NAME = "EndpointResolver";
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -0,0 +1,282 @@
...
+    private static final String ENDPOINT_RESOLVER_CONFIG_NAME = "EndpointResolver";
+    private static final String HTTP_CLIENT_CONFIG_NAME = "HTTPClient";
+    private static final String LOGGER_CONFIG_NAME = "Logger";
+    private static final String LOG_LEVEL_CONFIG_NAME = "LogLevel";
+    private static final String RETRYER_CONFIG_NAME = "Retryer";
+
+    private static final List<ConfigField> UNIVERSAL_FIELDS = new ArrayList<>();
+
+    static {
+        UNIVERSAL_FIELDS.add(ConfigField.builder(REGION_CONFIG_NAME, getUniversalSymbol("string"))
```
JordonPhillips, line 48: `SetUtils.of` into a `TreeSet` would work a bit better here

skmcgrail: Good call, changed!

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -0,0 +1,285 @@
...
+ */
+public class AddAwsConfigFields implements GoIntegration {
+    private static final String REGION_CONFIG_NAME = "Region";
+    private static final String CREDENTIALS_CONFIG_NAME = "Credentials";
+    private static final String ENDPOINT_RESOLVER_CONFIG_NAME = "EndpointResolver";
+    private static final String HTTP_CLIENT_CONFIG_NAME = "HTTPClient";
+    private static final String LOGGER_CONFIG_NAME = "Logger";
+    private static final String LOG_LEVEL_CONFIG_NAME = "LogLevel";
+    private static final String RETRYER_CONFIG_NAME = "Retryer";
+    private static final Set<ConfigField> UNIVERSAL_FIELDS = new TreeSet<>(SetUtils.of(
```
JordonPhillips, line 44: I don't know how this comment got dropped from my last set, but wouldn't it generally be simpler just to put the `aws.Config` struct directly into the client configuration rather than trying to break it out?

skmcgrail: Had a discussion yesterday to with Jason around the design some more, and we came to the conclusion that we would like to continue down this path of generating the corresponding options on the service client. The reason being is aws.Config is really a property grab bag of parameters loaded from shared config and is not always applicable to a given service client. By generating the options on the service client we can be smarter about which options are exposed/required and which ones are not. For example if a client were to support some new authentication scheme we could generate the specific type of credential provider required.

JordonPhillips: 👍

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -0,0 +1,148 @@
...
+            ConfigField.builder()
+                    .name(LOG_LEVEL_CONFIG_NAME)
+                    .type(getAwsCoreSymbol("LogLevel"))
+                    .documentation("An integer value representing the logging level.")
+                    .build(),
+            ConfigField.builder()
+                    .name(LOGGER_CONFIG_NAME)
+                    .type(getAwsCoreSymbol("Logger"))
+                    .documentation("The logger writer interface to write logging messages to.")
+                    .build()
```
jasdel, line 80: We should create github issues to track the design ideas we discussed for Logger improvements with the SDK and manager.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -0,0 +1,148 @@
...
+                            plugin.getConfigFields().forEach(configField -> {
+                                writer.write("$L: cfg.$L,", configField.getName(), configField.getName());
+                            });
+                        }
+                    });
+                    writer.write("");
+                    writer.openBlock("for _, fn := range optFns {", "}", () -> {
+                        writer.write("fn(&opts)");
+                    });
+                    writer.write("return New(opts)");
```
jasdel, line 144: I think the generated API client generic constructor needs an update as well to not return `error`

skmcgrail:
> Looks good only minor change is the smithy-go generated API client constructor needs to be updated not to return error.

Created https://github.com/awslabs/smithy-go/pull/75

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
skmcgrail: I've created https://github.com/aws/aws-sdk-go-v2/issues/594 to track this design requirement.

PR #596, opened 2020-06-18
Updates middleware addition on operation stack to take use middleware helpers

skotambkar:
* Adds static middleware helpers for aws sdk middlewares. 
* Updates to new run-time client plugin to use these static middleware helpers. 
* Related to awslabs/smithy-go#79

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/aws/aws-sdk-go-v2
 
 require (
-	github.com/awslabs/smithy-go v0.0.0-20200617230012-dcb49a2a302d
+	github.com/awslabs/smithy-go v0.0.0-20200625204536-9749f335749f
 	github.com/go-sql-driver/mysql v1.5.0
 	github.com/google/go-cmp v0.4.1
 	github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,7 @@
 github.com/awslabs/smithy-go v0.0.0-20200617230012-dcb49a2a302d h1:kLefSezMjHYpy9ayv5M7trxeERY8HV94m4n+b4tXqi0=
 github.com/awslabs/smithy-go v0.0.0-20200617230012-dcb49a2a302d/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
+github.com/awslabs/smithy-go v0.0.0-20200625204536-9749f335749f h1:exFiNX12QoMjLwwtH70VxP1rrz+TxWD9jhDg3I6RoMY=
+github.com/awslabs/smithy-go v0.0.0-20200625204536-9749f335749f/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/go-sql-driver/mysql v1.5.0 h1:ozyZYNQW3x3HtqT1jira07DN2PArx2v7/mN66gGcHOs=


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -83,78 +84,63 @@ private void writeAddMiddlewareBefore(
...
-                        .buildMiddlewareStack((writer, service, operation, protocolGenerator, stackOperand) -> {
-                            if (protocolGenerator == null) {
-                                return;
-                            }
-                            String deserializerMiddlewareName = ProtocolGenerator.getDeserializeMiddlewareName(operation.getId(),
-                                    protocolGenerator.getProtocolName());
-                            writeAddMiddlewareAfter(writer, DESERIALIZE_MIDDLEWARE,
-                                    deserializerMiddlewareName, "&$L{}", stackOperand);
-                        }).build(),
+                        .writeMiddlewareHelpers((writer, service, operation, protocolGenerator) -> {
```
JordonPhillips, line 101: I didn't catch this on the first round, but we should be adding the serde functions to all operations across protocols, so this should be generated in `OperationGenerator`.

aws/signer/v4/middleware.go
```diff
@@ -58,6 +58,12 @@ func (m *UnsignedPayloadMiddleware) HandleFinalize(ctx context.Context, in middl
 	return next.HandleFinalize(ctx, in)
 }
 
+// AddUnsignedPayloadMiddlewares adds UnsignedPayloadMiddleware to the operation middleware stack
+func AddUnsignedPayloadMiddlewares(stack *middleware.Stack) {
+ stack.Finalize.Add(&UnsignedPayloadMiddleware{}, middleware.After)
+}
+
```
jasdel, line 65: nit, gofmt

aws/signer/v4/middleware.go
```diff
@@ -128,6 +134,17 @@ func (s *SignHTTPRequestMiddleware) HandleFinalize(ctx context.Context, in middl
...
 
+// SigV4MiddlewareConfig interface for sigv4 middleware config
+type SigV4MiddlewareConfig interface {
+	GetSigner() Signer
+}
+
+// AddV4SignerMiddlewares adds SigV4 signer middlewares to operation stack
+func AddV4SignerMiddlewares(stack *middleware.Stack, cfg SigV4MiddlewareConfig) {
+	stack.Finalize.Add(&ComputePayloadSHA256Middleware{}, middleware.Before)
+	stack.Finalize.Add(NewSignHTTPRequestMiddleware(cfg.GetSigner()), middleware.After)
```
jasdel, line 145: nit these should be named `HTTPSigner` for the middleware and config. To match the constructor for HTTP request signer.

In addition, like `retryer` pkg and its middleware, this middleware helper probably needs to be in a separate package form the specific signer implementation. In addition it should take the `v4.HTTPSigner` interface as input not the concrete signer type.

Basically the two middleware for  `retryrer` and `v4` need to be somewhere else not in their concrete implementations, because the middleware are supposed to be generic, because the input parameter from the API client's options are abstract interfaces.

aws/retry/middleware.go
```diff
@@ -173,3 +173,14 @@ func getRetryMetadata(ctx context.Context) (metadata retryMetadata, ok bool) {
...
+// AwsRetryMiddlewareConfig interface for retry middleware config
+type AwsRetryMiddlewareConfig interface {
+	GetRetryer() Retryer
+}
+
+// AddRetryMiddlewares adds retry middleware to operation middleware stack
+func AddRetryMiddlewares(stack *smithymiddle.Stack, cfg AwsRetryMiddlewareConfig) {
+	stack.Finalize.Add(NewAttemptMiddleware(cfg.GetRetryer(), http.RequestCloner), smithymiddle.After)
+	stack.Finalize.Add(MetricsHeaderMiddleware{}, smithymiddle.After)
+}
```
jasdel, line 186: These probably need to refer to the interface `Retryer` not the concrete implementation. These interfaces would actually not be compatible with API client's `aws.Retryer`.  Most likely this means that the middleware cannot be in the `retryer` pkg.  this probably is the same feedback for signer.

skotambkar: In the above case, AwsRetryMiddlewareConfig has GetRetryer() which retry a `retry.Retryer` interface. Should this be a smithyhttp.Retryer interface instead?

aws/middleware/middleware.go
```diff
@@ -101,3 +101,23 @@ func GetResponseMetadata(metadata middleware.Metadata) (v ResponseMetadata) {
...
+}
+
+// AddAttemptClockSkewMiddleware adds AttemptClockSkewMiddleware to the middleware stack
+func AddAttemptClockSkewMiddleware(stack *middleware.Stack, optFns ...func(*AttemptClockSkewMiddleware)) {
+	attemptClockSkeyMiddleware := AttemptClockSkewMiddleware{}
+	for _, fn := range optFns {
+		fn(&attemptClockSkeyMiddleware)
+	}
+
+	stack.Deserialize.Add(attemptClockSkeyMiddleware, middleware.After)
```
jasdel, line 122: Do these helpers need functional options? It probably will be confusing to have these when there are no options that could be set.

aws/stack/register_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+
+// HTTPSignerMiddlewareConfig interface for HTTP signer middleware config
+type HTTPSignerMiddlewareConfig interface {
+	GetSigner() v4.HTTPSigner
+}
+
+// AddHTTPSignerMiddlewares adds HTTP signer middleware's to operation stack
+func AddHTTPSignerMiddlewares(stack *middleware.Stack, cfg HTTPSignerMiddlewareConfig,
+	optFns ...func( *v4.ComputePayloadSHA256Middleware, *v4.SignHTTPRequestMiddleware)) {
+	computePayloadSHA256Middleware := v4.ComputePayloadSHA256Middleware{}
```
jasdel, line 46: Probably want this functional option to be a structure that contains that contains pointers members to these values. That way additional middleware could be added to this helper.

aws/stack/register_middleware.go
```diff
@@ -0,0 +1,54 @@
...
+
+// AddUnsignedPayloadMiddlewares adds UnsignedPayloadMiddleware to the operation middleware stack
+func AddUnsignedPayloadMiddlewares(stack *middleware.Stack, optFns ...func(*v4.UnsignedPayloadMiddleware)) {
+	unsignedPayloadMiddleware := v4.UnsignedPayloadMiddleware{}
+	for _, fn := range optFns {
+		fn(&unsignedPayloadMiddleware)
+	}
+
+	stack.Finalize.Add(&unsignedPayloadMiddleware, middleware.After)
+}
```
jasdel, line 36: Probably won't want functional args to be the default when they don't  have anything to configure.

aws/stack/register_middleware.go
```diff
@@ -0,0 +1,54 @@
+package stack
```
jasdel, line 1: Does this need to be a separate package vs `aws/middleware`?  What is the difference between the two packages?

skotambkar: Can't keep the middleware helpers in `aws/middleware` as `aws/retry` and `aws/signer/v4` depend on `aws/middleware`, thus resulting in a *cyclic dependency*.

This package is supposed to be for the middleware helpers.

aws/stack/register_middleware.go
```diff
@@ -12,26 +12,30 @@ type AwsRetryMiddlewareConfig interface {
 	GetRetryer() retry.Retryer
 }
 
+// AwsRetryMiddlewares represents the Aws middleware's for Retry
+type AwsRetryMiddlewares struct {
+	*retry.AttemptMiddleware
+	*retry.MetricsHeaderMiddleware
+}
```
jasdel, line 19: probably want these as named fields instead of anonymous. You'll get unexpected results if either structure has clashing members.

aws/stack/register_middleware.go
```diff
@@ -40,13 +44,21 @@ type HTTPSignerMiddlewareConfig interface {
 	GetSigner() v4.HTTPSigner
 }
 
+// HTTPSignerMiddlewares represent the middleware's for HTTPSigner
+type HTTPSignerMiddlewares struct {
+	*v4.ComputePayloadSHA256Middleware
+	*v4.SignHTTPRequestMiddleware
+}
```
jasdel, line 51: named members instead of anonymous

aws/retry/middleware.go
```diff
@@ -173,3 +173,29 @@ func getRetryMetadata(ctx context.Context) (metadata retryMetadata, ok bool) {
...
+
+// AwsRetryMiddlewares represents the Aws middleware's for Retry
+type AwsRetryMiddlewares struct {
+	attemptMiddleware       *AttemptMiddleware
+	metricsHeaderMiddleware *MetricsHeaderMiddleware
+}
+
+// AddRetryMiddlewares adds retry middleware to operation middleware stack
+func AddRetryMiddlewares(stack *smithymiddle.Stack, cfg AwsRetryMiddlewareConfig, optsFn ...func(*AwsRetryMiddlewares)) {
```
jasdel, line 189: nit shouldn't these members be exported?  Its not possible to set the values from an API client if they are not exported.

aws/signer/v4/middleware.go
```diff
@@ -144,3 +144,35 @@ func SetPayloadHash(ctx context.Context, hash string) context.Context {
 	ctx = context.WithValue(ctx, payloadHashKey{}, hash)
 	return ctx
 }
+
+// AddUnsignedPayloadMiddlewares adds UnsignedPayloadMiddleware to the operation middleware stack
+func AddUnsignedPayloadMiddlewares(stack *middleware.Stack) {
+	unsignedPayloadMiddleware := UnsignedPayloadMiddleware{}
+	stack.Finalize.Add(&unsignedPayloadMiddleware, middleware.After)
+}
```
jasdel, line 152: `Middlewares` should probably be singular, not plural?

aws/signer/v4/middleware.go
```diff
@@ -144,3 +144,35 @@ func SetPayloadHash(ctx context.Context, hash string) context.Context {
...
+
+// HTTPSignerMiddlewares represent the middleware's for HTTPSigner
+type HTTPSignerMiddlewares struct {
+	computePayloadSHA256Middleware *ComputePayloadSHA256Middleware
+	signHTTPRequestMiddleware      *SignHTTPRequestMiddleware
+}
+
+// AddHTTPSignerMiddlewares adds HTTP signer middleware's to operation stack
+func AddHTTPSignerMiddlewares(stack *middleware.Stack, cfg HTTPSignerMiddlewareConfig, optFns ...func(*HTTPSignerMiddlewares)) {
```
jasdel, line 166: Should these members be exported?

aws/signer/v4/middleware.go
```diff
@@ -144,3 +144,35 @@ func SetPayloadHash(ctx context.Context, hash string) context.Context {
...
+	computePayloadSHA256Middleware := ComputePayloadSHA256Middleware{}
+	signHTTPRequestMiddleware := NewSignHTTPRequestMiddleware(cfg.GetSigner())
+	for _, fn := range optFns {
+		fn(&HTTPSignerMiddlewares{
+			computePayloadSHA256Middleware: &computePayloadSHA256Middleware,
+			signHTTPRequestMiddleware:      signHTTPRequestMiddleware,
+		})
+	}
```
jasdel, line 174: probably want to create a variable for the `HTTPSignerMiddlewares` value first, then iterate through the `optFns` with a pointer to it. Will prevent creating a new `HTTPSignerMiddlewares` each iteration. i.e initialize `HTTPSignerMiddlewares` once, then iterate with the variable pointer.

Also should only really need to take the address of `computePayloadSHA256Middleware` once. Looks like that is being done multiple times here.  Creating the variable original as an pointer will simplify this.

```go
 	computePayloadSHA256Middleware := &ComputePayloadSHA256Middleware{} 
````

File: codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RegisterServiceMetadataMiddleware.java [hunk 15]
```diff
@@ -0,0 +1,110 @@
...
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.utils.ListUtils;
+
+public final class RegisterServiceMetadataMiddleware implements GoIntegration {
+    List<RuntimeClientPlugin> runtimeClientPlugins = new ArrayList<>();
+
+    @Override
+    public byte getOrder(){
+        return 30;
+    }
```
jasdel, line 28: `30` is same order as AssembleMiddlewareStack, will that cause any ordering issues?

skotambkar: No, previous PR's assigned AssembleMiddlewareStack -40 as order ID.

aws/retry/middleware.go
skotambkar: Yes should have been exported

aws/signer/v4/middleware.go
skotambkar: updated to
```
// AddHTTPSignerMiddlewares adds HTTP signer middleware's to operation stack
func AddHTTPSignerMiddlewares(stack *middleware.Stack, cfg HTTPSignerMiddlewareConfig, optFns ...func(*HTTPSignerMiddlewares)) {
	m := HTTPSignerMiddlewares {
		ComputePayloadSHA256Middleware: &ComputePayloadSHA256Middleware{},
		SignHTTPRequestMiddleware:      NewSignHTTPRequestMiddleware(cfg.GetSigner()),
	}
	for _, fn := range optFns {
		fn(&m)
	}

	stack.Finalize.Add(m.ComputePayloadSHA256Middleware, middleware.Before)
	stack.Finalize.Add(m.SignHTTPRequestMiddleware, middleware.After)
}



```

PR #598, opened 2020-06-18
Metadata and Endpoint Middleware, Signing Middleware Changes

skmcgrail:
* Introduces a metadata middleware to inject values into the middleware context, such that middleware downstream can perform runtime based introspection if required.
* Adds an endpoint serializer middleware that will resolve the the endpoint and signing region for the service using the provided signing. For HTTP transports the endpoint and path will be parse and configure on the http.Request.URL member.
* Changes to the signer middleware to reflect with the changes introduced with the metadata middleware.

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
@@ -45,6 +45,6 @@ public final class AwsGoDependency {
     }
 
     private static final class Versions {
-        private static final String AWS_SDK = "v0.0.0-20200622225525-a0b748e6c297";
+        private static final String AWS_SDK = "v0.0.0-20200622232612-3b8c27eea891";
     }
 }


aws/middleware/endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+func (r ResolveServiceEndpoint) HandleSerialize(
+	ctx context.Context, in middleware.SerializeInput, next middleware.SerializeHandler,
+) (out middleware.SerializeOutput, metadata middleware.Metadata, err error) {
+	req, ok := in.Request.(*smithyhttp.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown transport type %T", in.Request)
+	}
+
+	var endpoint aws.Endpoint
+	endpoint, err = r.Resolver.ResolveEndpoint(GetEndpointID(ctx), GetRegion(ctx))
```
jasdel, line 33: nit probably want to name `endpointId` endpointPrefix for now. Hopefully we can replace this with service id in the future.

aws/middleware/endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return out, metadata, fmt.Errorf("failed to parse endpoint URL: %w", err)
+	}
+
+	if len(endpoint.SigningName) > 0 && (len(GetSigningName(ctx)) == 0 || !endpoint.SigningNameDerived) {
+		ctx = SetSigningName(ctx, endpoint.SigningName)
+	}
+	ctx = SetSigningRegion(ctx, endpoint.SigningRegion)
+
+	return next.HandleSerialize(ctx, in)
```
jasdel, line 48: We'll probably want special behavior that handles the case where a endpoint contains a standard port, e.g. 80/443. These ports need to be stripped off the "host" header before signing, or ths signature might become invalid.

aws/middleware/endpoints_test.go
```diff
@@ -0,0 +1,195 @@
...
+func TestResolveServiceEndpoint(t *testing.T) {
+	initContext := func(signingName, endpointID, region string) context.Context {
+		ctx := context.Background()
+		ctx = SetSigningName(ctx, signingName)
+		ctx = setEndpointID(ctx, endpointID)
+		ctx = setRegion(ctx, region)
+		return ctx
+	}
+
+	cases := []struct {
```
jasdel, line 33: Nit it would be great to use map[string]struct here so these test cases can be named. helps as description and debugging

aws/signer/v4/middleware_test.go
```diff
@@ -124,10 +125,11 @@ func TestSignHTTPRequestMiddleware(t *testing.T) {
...
 
-			ctx := SetSigningMetadata(context.Background(), SigningMetadata{
-				SigningName:   signingName,
-				SigningRegion: signingRegion,
-			})
+			ctx := func() context.Context {
+				ctx := awsmiddleware.SetSigningName(context.Background(), signingName)
+				ctx = awsmiddleware.SetSigningRegion(ctx, signingRegion)
+				return ctx
+			}()
```
jasdel, line 132: Why the closure?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -78,28 +86,67 @@ private void writeAddMiddlewareBefore(
...
+                            }
+                            if (sigV4Trait.isPresent()) {
+                                SigV4Trait trait = sigV4Trait.get();
+                                builder.append(String.format("SigningName: \"%s\",\n", trait.getName()));
+                            }
+                            builder.append(String.format("OperationName: \"%s\",\n", operation.getId().getName()));
+                            builder.append("}");
+
+                            writeAddMiddlewareBefore(writer, INITIALIZE_MIDDLEWARE, ServiceMetadataProvider,
+                                    builder.toString(), stackOperand);
```
jasdel, line 122: Is the middleware utilities the reason that need to use string format here to build a string instead of using the writer directly with $T, $L etc?

Would it make sense for a version of writeAddMiddlewareBefore to be defined that takes  Consummable<<GoWriter> ?

aws/middleware/endpoints.go
skmcgrail: https://github.com/aws/aws-sdk-go-v2/issues/606 to track this work.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
skmcgrail: Yes this is due to the nature of writeAddMiddlewareBefore has been defined. Comments regarding this design should probably be addressed by Shantanu's relevant PRs that are refactoring this area,

aws/signer/v4/middleware_test.go
skmcgrail: I felt that it made the code cleaner to read, but if there is confusion around the closure then I will just inline everything.

jasdel: Not a big deal since a part of a test and not runtime. It surprised me and had to read a few times to figure out if i was missing something

aws/middleware/endpoints.go
```diff
@@ -0,0 +1,69 @@
...
+	req, ok := in.Request.(*smithyhttp.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown transport type %T", in.Request)
+	}
+
+	var endpoint aws.Endpoint
+	endpoint, err = r.Resolver.ResolveEndpoint(GetEndpointPrefix(ctx), GetRegion(ctx))
+	if err != nil {
+		return out, metadata, fmt.Errorf("failed to resolve service endpoint ")
+	}
```
jasdel, line 36: probably want to return an error if `r.Resolver` is nil

PR #601, opened 2020-06-19
Add preliminary developer changelog tool

aggagen:
I added a preliminary library and CLI for managing change metadata. I still need to add documentation and testing, but I wanted to get some feedback before that to make sure I'm going in the right direction.

I added `golang.org/x/mod`, which is hopefully fine. We'll also need to figure out where exactly to put this within the repo and come up with a better name for the tool.

### DIFF
--- a/internal/tools/changes/change.go
+++ b/internal/tools/changes/change.go
@@ -13,8 +13,10 @@ import (
 type ChangeType string
 
 const (
-	FeatureChangeType ChangeType = "feature" // FeatureChangeType is a constant change type for a new feature.
-	BugFixChangeType  ChangeType = "bugfix"  // BugFixChangeType is a constant change type for a bug fix.
+	// FeatureChangeType is a constant change type for a new feature.
+	FeatureChangeType ChangeType = "feature"
+	// BugFixChangeType is a constant change type for a bug fix.
+	BugFixChangeType  ChangeType = "bugfix"
 )
 
 // ParseChangeType attempts to parse the given string v into a ChangeType, returning an error if the string is invalid.
@@ -114,7 +116,7 @@ func NewChanges(modules []string, changeType ChangeType, description string) ([]
 
 	for _, module := range modules {
 		changes = append(changes, Change{
-			ID:            generateId(module, changeType),
+			ID:            generateID(module, changeType),
 			SchemaVersion: SchemaVersion,
 			Module:        module,
 			Type:          changeType,
@@ -152,7 +154,7 @@ func ChangeToTemplate(change Change) ([]byte, error) {
 	return append(templateBytes, []byte(changeTemplateSuffix)...), nil
 }
 
-func generateId(module string, changeType ChangeType) string {
+func generateID(module string, changeType ChangeType) string {
 	module = strings.ReplaceAll(module, "/", ".")
 
 	return fmt.Sprintf("%s-%s-%v", module, changeType, time.Now().UnixNano())
--- a/internal/tools/changes/cmd/changetool/change.go
+++ b/internal/tools/changes/cmd/changetool/change.go
@@ -222,7 +222,7 @@ func rmCmd(metadata *changes.Metadata, id string) error {
 		return fmt.Errorf("failed to remove change: %v", err)
 	}
 
-	err = metadata.RemoveChangeById(change.ID)
+	err = metadata.RemoveChangeByID(change.ID)
 	if err != nil {
 		return fmt.Errorf("failed to remove change: %v", err)
 	}
@@ -238,10 +238,10 @@ func selectChange(metadata *changes.Metadata, id string) (changes.Change, error)
 	index, err := strconv.Atoi(id)
 	if err == nil {
 		if index < 0 || index >= len(metadata.Changes) {
-			return changes.Change{}, fmt.Errorf("failed to get change with index %d: index out of range\n", index)
+			return changes.Change{}, fmt.Errorf("failed to get change with index %d: index out of range", index)
 		}
 		return metadata.Changes[index], nil
 	}
 
-	return metadata.GetChangeById(id)
+	return metadata.GetChangeByID(id)
 }
--- a/internal/tools/changes/metadata.go
+++ b/internal/tools/changes/metadata.go
@@ -71,8 +71,8 @@ func (m *Metadata) AddChanges(changes []Change) error {
 	return nil
 }
 
-// GetChangeById returns the pending Change with the given id.
-func (m *Metadata) GetChangeById(id string) (Change, error) {
+// GetChangeByID returns the pending Change with the given id.
+func (m *Metadata) GetChangeByID(id string) (Change, error) {
 	_, c, err := m.getChange(id)
 	return c, err
 }
@@ -87,7 +87,7 @@ func (m *Metadata) getChange(id string) (int, Change, error) {
 	return 0, Change{}, fmt.Errorf("couldn't find change with id %s", id)
 }
 
-// ListChanges returns all pending Changes with a module matching the given module. If module is empty, returns all Changes.
+// GetChanges returns all pending Changes with a module matching the given module. If module is empty, returns all Changes.
 func (m *Metadata) GetChanges(module string) []Change {
 	if module == "" {
 		return m.Changes
@@ -113,25 +113,25 @@ func (m *Metadata) SaveChange(c Change) error {
 func (m *Metadata) UpdateChangeFromTemplate(oldChange Change, template []byte) ([]Change, error) {
 	newChanges, err := TemplateToChanges(template)
 	if err != nil {
-		return nil, fmt.Errorf("failed to modify change: %v\n", err)
+		return nil, fmt.Errorf("failed to modify change: %v", err)
 	}
 
 	err = m.AddChanges(newChanges)
 	if err != nil {
-		return nil, fmt.Errorf("failed to modify change: %v\n", err)
+		return nil, fmt.Errorf("failed to modify change: %v", err)
 	}
 
-	err = m.RemoveChangeById(oldChange.ID)
+	err = m.RemoveChangeByID(oldChange.ID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to remove old change with id %s: %v\n", oldChange.ID, err)
+		return nil, fmt.Errorf("failed to remove old change with id %s: %v", oldChange.ID, err)
 	}
 
 	return newChanges, nil
 }
 
-// RemoveChangeById removes the Change with the specified id from the Metadata's Changes and also removes the Change
+// RemoveChangeByID removes the Change with the specified id from the Metadata's Changes and also removes the Change
 // from the .changes/next-release directory.
-func (m *Metadata) RemoveChangeById(id string) error {
+func (m *Metadata) RemoveChangeByID(id string) error {
 	i, _, err := m.getChange(id)
 	if err != nil {
 		return fmt.Errorf("failed to remove change: %v", err)
@@ -155,7 +155,7 @@ func (m *Metadata) ClearChanges() error {
 	}
 
 	for _, id := range ids {
-		err := m.RemoveChangeById(id)
+		err := m.RemoveChangeByID(id)
 		if err != nil {
 			return err
 		}


changes/change.go
```diff
@@ -0,0 +1,107 @@
...
+type Change struct {
+	Id            string
+	SchemaVersion string
+	Modules       []string
+	Type          string
+	Description   string
+	SetVersion    string
+}
+
+func NewChange(modules []string, changeType, description string) *Change {
```
skmcgrail, line 26: For each module targeted for a change, would it be simpler to write out a change description for each module. This simplifies later discovery of pending changes for an individual module. That's not to say you couldn't have the ability to write a changelog as user that targets multiple modules, but when stored in the repository would it make more sense to write out a file per module?

changes/change.go
```diff
@@ -0,0 +1,107 @@
...
+			if len(parts) != 2 {
+				return fmt.Errorf("template is incorrectly formatted at line: %s", l)
+			}
+
+			switch parts[0] {
+			case "modules":
+				trimmedMods := strings.Trim(parts[1], "[] ")
+				change.Modules = strings.Split(trimmedMods, ",")
+			case "change type":
+				change.Type = parts[1]
```
skmcgrail, line 75: We should probably have some simple validation that normalizes the parsed value and compares it to a set of constants.

changes/change.go
```diff
@@ -0,0 +1,107 @@
...
+			case "description":
+				change.Description = parts[1]
```
skmcgrail, line 77: Should description allowed to be multi-line? Since the description will be embedded in a markdown file. It would be ideal to allow the developer to write notes that span lines,  and form bulleted lists etc.

changes/cmd/changetool/main.go
```diff
@@ -0,0 +1,139 @@
...
+		modifyCmd(metadata, os.Args[2])
+	case "rm", "delete":
+		if len(os.Args) < 3 {
+			usage()
+			os.Exit(1)
+		}
+
+		rmCmd(metadata, os.Args[2])
+	case "release":
+		fmt.Println(metadata.CreateRelease("2020-06-19", []changes.VersionBump{}))
```
skmcgrail, line 54: Should release date be a flag argument that defaults to the value for todays date?

changes/change.go
```diff
@@ -0,0 +1,107 @@
...
+	c.SchemaVersion = version
+}
+
+func (c *Change) isValid() bool {
+	return len(c.Modules) > 0 && c.Type != "" && c.Description != ""
+}
+
+// TemplateToChange parses the provided filledTemplate into the provided Change. If Change has no Id, TemplateToChange
+// will set the Id.
+func TemplateToChange(filledTemplate string, change *Change) error {
```
skmcgrail, line 56: Would be good to have some unit-tests that cover the parsing and validation.

changes/cmd/changetool/main.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+func addCmd(metadata *changes.Metadata, module string) {
+	if module == "" {
+		if currentModule, err := changes.GetCurrentModule(); err == nil {
+			module = currentModule
+		}
```
skmcgrail, line 62: Should this return an error if we aren't able to resolve a module?

changes/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/aggagen/changes
```
skmcgrail, line 1: This should be under `github.com/aws/aws-sdk-go-v2`

changes/metadata.go
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+// GetChangeById returns the pending Change with the given id.
+func (m *Metadata) GetChangeById(id string) (*Change, error) {
+	_, c, err := m.getChange(id)
+	return c, err
+}
+
+func (m *Metadata) getChange(id string) (int, *Change, error) {
```
skmcgrail, line 57: What was your thought for returning the index value here?

changes/modules.go
```diff
@@ -0,0 +1,34 @@
...
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	return modfile.ParseLax(path, data, nil)
+}
+
+func shortenModPath(modulePath string) string {
+	return strings.TrimPrefix(modulePath, "github.com/aggagen/")
```
skmcgrail, line 33: Needs correct repo path.

changes/modules.go
```diff
@@ -0,0 +1,34 @@
+package changes
+
+import (
+	"golang.org/x/mod/modfile"
```
skmcgrail, line 4: Something worth considering, but the `go mod` command can return JSON structures back that describe the go mod file contents. We could avoid pulling in a dependency library if we structure ourselves around that.

changes/release.go
```diff
@@ -0,0 +1,18 @@
+package changes
+
+type VersionBump struct {
+	Module string
+	From   string
+	To     string
+}
+
+type Release struct {
+	Id            string
```
skmcgrail, line 10: Minor, but for the SDK code base we would typically format this as `ID` rather then `Id`

changes/metadata.go
aggagen: RemoveChangeById uses the index to remove the Change from m.Changes.

changes/cmd/changetool/main.go
aggagen: Yeah I think so. The release command is currently more of a placeholder that I was using to test some of the Metadata operations.

changes/change.go
aggagen: Switched to YAML, so we now support multi-line descriptions.

changes/modules.go
aggagen: If the tool ends up being its own submodule, its dependencies will be isolated from the rest of the SDK. In that case, it seems beneficial to avoid running and parsing the output of `go` commands by pulling in some of the Go packages for working with modules.

changes/modules.go
aggagen: Yeah, I'll update this and the import statement to use the right import path once we figure out where the tool will be/what it will be named.

changes/cmd/changetool/editor.go
```diff
@@ -0,0 +1,45 @@
...
+		editor = "vim"
+	}
+
+	f, err := ioutil.TempFile(".", "tmp-template-entry")
+	if err != nil {
+		return "", err
+	}
+	defer os.Remove(f.Name())
+
+	_, err = f.Write([]byte(template))
```
skmcgrail, line 22: In general, should we pass around the template as just `[]byte`, given that most of the usage for marshaling / unmarshaling is using a byte slice rather then a string? This also reduces excess allocations.

changes/metadata.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+// SchemaVersion defines the current JSON schema version for persistent data types (Change, Release, ...)
+const SchemaVersion = "1.0"
+
+// Metadata is a representation of the change metadata stored in a .changes directory.
+type Metadata struct {
+	ChangePath string     // ChangePath is the relative path from the current directory to .changes
+	Changes    []*Change  // Changes holds all pending change metadata in the .changes/next-release directory
+	Releases   []*Release // Releases contains all releases in the .changes/releases directory
```
skmcgrail, line 20: One thing to consider is the potential growth rate of the number of releases we do perform per year. Likely we don't want/need to load all the history about releases into memory. We should probably consider moving this to a  separate method that we can use to discover the release files, and lazily load their contents if we want to inspect them.

changes/cmd/changetool/change.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	switch args[0] {
+	case "add", "new":
+		addFlags.Parse(args[1:])
+		addCmd(metadata, changeParams.module)
+	case "ls", "list":
+		lsFlags.Parse(args[1:])
+		lsCmd(metadata, changeParams.module)
+	case "modify", "edit":
```
skmcgrail, line 49: This likely didn't get discussed originally, but we should probably have a non visual mode that would allow our release automation system to feed in the change descriptions for service modules that are being updated. Maybe just pure CLI flags to feed in the module, type, and description.

changes/change.go
```diff
@@ -0,0 +1,93 @@
...
+	ID            string // ID is a unique identifier for this Change
+	SchemaVersion string // SchemaVersion is the version of the library's types used to create this Change.
+	Module        string // Module is the Go module affected by this Change.
+	Type          string // Type indicates what category of Change was made. Type may be either "feature" or "bugfix".
+	Description   string // Description is a human readable description of this Change meant to be included in a CHANGELOG.
+}
+
+// NewChanges returns a Change slice containing a Change with the given type and description for each of the specified
+// modules.
+func NewChanges(modules []string, changeType, description string) ([]*Change, error) {
```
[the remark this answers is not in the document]
aggagen: Yeah, that's correct. I did it this way since it's easier to enter the shortened relative module path (like `aws/module`) for the developer. It might make sense to take the developer entered short module path and store it in a Change as the full module import path.

changes/util.go
```diff
@@ -0,0 +1,82 @@
...
+
+		return true, nil
+	} else if os.IsNotExist(err) {
+		return false, nil
+	} else {
+		return false, err
+	}
+}
+
+func findFile(fileName string, dir bool) (string, error) {
```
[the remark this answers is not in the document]
aggagen: We are trying to find some root directory/file (in this case, that's currently a `.changes` directory or `go.mod`) that is in a parent of the current directory. It seems that `filepath.Walk` does the opposite, searching from a given root directory for some child file.

changes/metadata.go
```diff
@@ -0,0 +1,225 @@
...
+				return nil, err
+			}
+
+			change := &Change{}
+			err = json.Unmarshal(changeData, change)
+			if err != nil {
+				return nil, err
+			}
+
+			changes = append(changes, change)
```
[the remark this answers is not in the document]
aggagen: I added a check looking for whether the unmarshaled `SchemaVersion` matches the current `SchemaVersion`. In the future we could add some behavior here that attempts to correct for older versions, though the nature of the changes JSON files are pretty short lived, so I'm not sure that'll be necessary.

Old changes are stored in release JSON files, though, so if we end up consuming those in the future we'll have to be careful about the Change `SchemaVersion`s

changes/metadata_test.go
```diff
@@ -0,0 +1,391 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	// testdata already has .changes-static, but make another .changes directory so that we can easily cleanup after
+	// tests have run.
+	err := os.MkdirAll("testdata/.changes/next-release", 0777)
```
skmcgrail, line 14: Since this will be dynamic data, should we just create a temporary directory using `ioutil.TempDir`? `testdata` is good for populating static data to be used for tests, but if we are going to be manipulating the directory contents then we should use a temporary directory in a different location that we can cleanup. This will prevent accidental changes to the static testdata that might get committed int the repository if cleanup fails.

internal/tools/changes/release.go
```diff
@@ -0,0 +1,69 @@
...
+
+	sort.Strings(sectionsList)
+
+	for _, section := range sectionsList {
+		entry += headerPrefix + "## " + changeHeaders[section] + "\n"
+		entry += sections[section] + "\n"
+	}
+
+	return entry
+}
```
jasdel, line 52: Might be easier to follow and write to investigating using [text/template#Template](https://pkg.go.dev/text/template?tab=doc) for the changelog. It also allows easier formatting that can be viewed at a glance.

internal/tools/changes/repository.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Repository is a representation of a git repository containing multiple Go modules.
+type Repository struct {
+	RootPath string    // RootPath is the path to the root of the repository.
+	Metadata *Metadata // Metadata is the repository's .changes metadata.
+}
+
+// NewRepository loads the repository at the given path.
+func NewRepository(path string) (*Repository, error) {
+	metadata, err := LoadMetadata(filepath.Join(path, ".changes"))
```
jasdel, line 16: You probably want to put the file names into constants instead of referring to them inline.

internal/tools/changes/repository_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	for _, changelogType := range []string{"pending", "regular"} {
+		cases := getTestChangelogCases(t, changelogType)
+		for _, tt := range cases {
+			pending := false
+			fileName := "testdata/CHANGELOG.md"
+			if changelogType == "pending" {
+				pending = true
+				fileName = "testdata/CHANGELOG_PENDING.md"
+			}
```
jasdel, line 33: nit, these should use `filepath.Join` instead of inline `/`  These tests probably won't work on windows.

internal/tools/changes/testdata/changes/test-feature-2b.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ID": "test-feature-2",
+  "SchemaVersion": "1.0",
+  "Module": "other",
+  "Type": "bugfix",
+  "Description": "test description"
```
jasdel, line 6: Should have some tests with multi line and rich descriptions for changes.

internal/tools/changes/metadata_test.go
```diff
@@ -0,0 +1,402 @@
...
+	m := getMetadata(t)
+
+	for id, tt := range testCases {
+		t.Run(id, func(t *testing.T) {
+			changes, err := m.AddChangesFromTemplate(tt.template)
+			if len(changes) == 0 {
+				if err == nil {
+					t.Errorf("expected non-nil err, got nil")
+				}
+			} else {
```
jasdel, line 205: If changes are expected or not probably should be annotated in the Test cases (`tt`)

internal/tools/changes/util.go
```diff
@@ -0,0 +1,82 @@
...
+	pathParts := strings.Split(path, string(os.PathSeparator))
+	prefix := ""
+	if strings.HasPrefix(path, string(os.PathSeparator)) {
+		prefix = string(os.PathSeparator)
+	}
+
+	for i := len(pathParts); i > 0; i-- {
+		path = prefix + filepath.Join(append(pathParts[:i], fileName)...)
+
+		found, err := fileExists(path, dir)
```
jasdel, line 71: I think you could simplify this logic with [filepath#Base](https://pkg.go.dev/path/filepath?tab=doc#Base)  The Base function could be called in a loop without needing to manually split the paths.

```go
currpath, err := os.Getwd()
if err != nil { return "", err }

for {
     currPath = filepath.Base(currPath)

     targetFilepath := filepath.Join(currPath, fileName)
     found, err := fileExists(targetFilepath, dir)
     if err != nil {
          return "", err // <- probably want to wrap errors like this with meaningful context.
     }

     if found {
          return targetFilepath, nil
     }

     if currPath == "." || currPath == "/" {
           return "", fmt.Errorf("<error message>")
     }
}
```

internal/tools/changes/.changes/releases/2020-06-26.json
```diff
@@ -0,0 +1,30 @@
...
+   "Module": "changes",
+   "Type": "feature",
+   "Description": "this is a test"
+  },
+  {
+   "ID": "test-feature-1593193458080838000",
+   "SchemaVersion": "1.0",
+   "Module": "test",
+   "Type": "feature",
+   "Description": "this is a test"
```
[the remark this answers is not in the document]
aggagen: Yeah, you can add markdown when creating a change. I added a test case that tests/demonstrates this. The one potential issue is that each change file is mapped to a single bullet point in the generated CHANGELOG, although in theory nothing is actually stopping a developer from entering two bullet points in the description:

```
modules:
  - module
type: feature
description: |-
  first bullet point <-- first bullet point * is already added by the tool
  * second bullet point
```

internal/tools/changes/change.go
```diff
@@ -0,0 +1,96 @@
...
+		Modules:     []string{change.Module},
+		Type:        change.Type,
+		Description: change.Description,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return append(templateBytes, []byte(changeTemplateSuffix)...), nil
+}
```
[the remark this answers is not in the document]
aggagen: The tool lets you edit a single change at a time, which is why this function takes a single change as an argument. I looked into having `Description` pre-rendered as a multi line string: when modifying a description with multiple lines, it will correctly render as a multi-line yaml string in your editor. I couldn't find a good way to get it to render the initial change creation template with the multi line yaml syntax there. (so currently the developer has to replace the `""` the yaml parser generates for description with `>-`)

internal/tools/changes/cmd/changetool/change.go
```diff
@@ -0,0 +1,186 @@
...
+		return fmt.Errorf("failed to load .changes directory: %v", err)
+	}
+
+	switch args[0] {
+	case "add", "new":
+		addFlags.Parse(args[1:])
+		return addCmd(metadata, changeParams.module, changeParams.changeType, changeParams.description)
+	case "ls", "list":
+		lsFlags.Parse(args[1:])
+		return lsCmd(metadata, changeParams.module)
```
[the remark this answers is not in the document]
aggagen: It might make sense to add this later. Right now the tool doesn't know what packages a module provides, but I think once the release tool part is developer that information should be available.

internal/tools/changes/cmd/changetool/release.go
```diff
@@ -0,0 +1,52 @@
...
+	repo, err := changes.NewRepository(args[0])
+	if err != nil {
+		return fmt.Errorf("couldn't load repository: %v", err)
+	}
+
+	switch args[1] {
+	case "update-pending":
+		return updatePendingCmd(repo)
+	case "demo-release":
+		release, err := repo.Metadata.CreateRelease("2020-06-26", map[string]changes.VersionBump{
```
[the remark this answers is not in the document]
aggagen: Yeah, I'll add this as I develop the release tool portion.

internal/tools/changes/cmd/changetool/release.go
```diff
@@ -0,0 +1,52 @@
...
+
+func releaseSubcmd(args []string) error {
+	if len(args) < 2 {
+		usage()
+	}
+
+	repo, err := changes.NewRepository(args[0])
+	if err != nil {
+		return fmt.Errorf("couldn't load repository: %v", err)
+	}
```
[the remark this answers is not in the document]
aggagen: I broke the positional CLI arguments into separate variables instead of using `args[x]` in multiple locations. I suppose these could also just be part of the `flags.FlagSet`s, but that seems like it would make the tool more difficult to use.

internal/tools/changes/metadata.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (m *Metadata) getChange(id string) (int, *Change, error) {
+	for i, c := range m.Changes {
+		if c.ID == id {
+			return i, c, nil
+		}
+	}
+
+	return 0, nil, fmt.Errorf("couldn't find change with id %s", id)
+}
```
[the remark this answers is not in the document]
aggagen: Having `m.Changes` be loaded in sorted order allows the tool to have consistent change indices between runs. This is useful since `tool change ls` prints each change's index in `m.Changes`. That index can then be used as an argument to `tool change edit <index>` and `tool change rm <index>` instead of providing the full Change ID.

internal/tools/changes/release.go
aggagen: I switched this function over to using a `text/template`. It's still not the most clear thing to look at, but I think it's definitely better at a glance than the original function.

changes/change_test.go
```diff
@@ -0,0 +1,139 @@
...
+func assertChangeEqual(t *testing.T, want, got *Change) bool {
+	if !changeEquals(t, want, got) {
+		t.Errorf("expected change like %v, got %v", want, got)
+		return false
+	}
+
+	return true
+}
+
+func changeEquals(t *testing.T, want, got *Change) bool {
```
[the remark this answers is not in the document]
jasdel: Also another good comparison for structured data is go-cmp's Diff for testing it does a great job of giving structure deltas

internal/tools/changes/.changes/releases/2020-06-26.json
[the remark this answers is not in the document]
jasdel: It would be really great if we can avoid the conditional first bullet point added by tool complexity. such as if first item has bullet point accept, if not add.

PR #625, opened 2020-06-26
Update signer middleware for content sha and order with retry

jasdel:
Updates the signer middleware helper to only add the signer. The compute SHA256 middleware is broken out into its own header.

Adds a new middleware for setting X-Amz-Content-Sha256 header that will be paired with the unsigned payload middleware code generation by default. Other services that need to be customizes for this header can add the middleware as well, e.g. S3 and Glacier.

Reorders the clock skew middleware to be after retry.

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
@@ -56,9 +56,8 @@ public class AssembleMiddlewareStack implements GoIntegration {
                 // Add unsigned payload middleware to operation stack
                 RuntimeClientPlugin.builder()
                         .operationPredicate((model, service, operation) -> {
-                            boolean hasSigV4Auth = hasSigV4AuthScheme(model, service, operation);
-
-                            return hasSigV4Auth && operation.hasTrait(UnsignedPayloadTrait.class);
+                            return hasSigV4AuthScheme(model, service, operation)
+                                    && operation.hasTrait(UnsignedPayloadTrait.class);
                         })
                         .registerMiddleware(MiddlewareRegistrar.builder()
                                 .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
@@ -70,9 +69,8 @@ public class AssembleMiddlewareStack implements GoIntegration {
                 // Add signed payload middleware to operation stack
                 RuntimeClientPlugin.builder()
                         .operationPredicate((model, service, operation) -> {
-                            boolean hasSigV4Auth = hasSigV4AuthScheme(model, service, operation);
-
-                            return hasSigV4Auth && !operation.hasTrait(UnsignedPayloadTrait.class);
+                            return hasSigV4AuthScheme(model, service, operation)
+                                    && !operation.hasTrait(UnsignedPayloadTrait.class);
                         })
                         .registerMiddleware(MiddlewareRegistrar.builder()
                                 .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
@@ -84,9 +82,8 @@ public class AssembleMiddlewareStack implements GoIntegration {
                 // Add content-sha256 payload header middleware to operation stack
                 RuntimeClientPlugin.builder()
                         .operationPredicate((model, service, operation) -> {
-                            boolean hasSigV4Auth = hasSigV4AuthScheme(model, service, operation);
-
-                            return hasSigV4Auth && operation.hasTrait(UnsignedPayloadTrait.class);
+                            return hasSigV4AuthScheme(model, service, operation)
+                                    && operation.hasTrait(UnsignedPayloadTrait.class);
                         })
                         .registerMiddleware(MiddlewareRegistrar.builder()
                                 .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
@@ -130,8 +127,8 @@ public class AssembleMiddlewareStack implements GoIntegration {
     /**
      * Returns if the SigV4Trait is a auth scheme for the service and operation.
      *
-     * @param model model definition
-     * @param service service shape for the API
+     * @param model     model definition
+     * @param service   service shape for the API
      * @param operation operation shape
      * @return if SigV4Trait is an auth scheme for the operation and service.
      */


aws/signer/v4/middleware.go
```diff
@@ -145,33 +176,24 @@ func SetPayloadHash(ctx context.Context, hash string) context.Context {
...
 
 // HTTPSignerMiddlewares represent the middleware's for HTTPSigner
 type HTTPSignerMiddlewares struct {
-	ComputePayloadSHA256Middleware *ComputePayloadSHA256Middleware
-	SignHTTPRequestMiddleware      *SignHTTPRequestMiddleware
+	SignHTTPRequestMiddleware *SignHTTPRequestMiddleware
 }
 
 // AddHTTPSignerMiddlewares adds HTTP signer middleware's to operation stack
 func AddHTTPSignerMiddlewares(stack *middleware.Stack, cfg HTTPSignerMiddlewareConfig, optFns ...func(*HTTPSignerMiddlewares)) {
```
skotambkar, line 190: Nit: singular vs plural Middlewares

aws/signer/v4/middleware.go
```diff
@@ -55,13 +61,38 @@ func (m *UnsignedPayloadMiddleware) ID() string {
...
 }
 
 // ComputePayloadSHA256Middleware computes sha256 payload hash to sign
 type ComputePayloadSHA256Middleware struct{}
 
+// AddComputePayloadSHA256Middleware adds ComputePayloadSHA256Middleware to the
+// operation middleware stack
+func AddComputePayloadSHA256Middleware(stack *middleware.Stack) {
+	unsignedPayloadMiddleware := ComputePayloadSHA256Middleware{}
+	stack.Finalize.Add(&unsignedPayloadMiddleware, middleware.After)
```
skotambkar, line 93: So it's Build or finalize? Computing Payload SHA can be build

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -41,51 +41,57 @@ public byte getOrder() {
...
                         .registerMiddleware(MiddlewareRegistrar.builder()
                                 .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
-                                        "AddAttemptClockSkewMiddleware", AwsGoDependency.AWS_MIDDLEWARE)
+                                        "AddUnsignedPayloadMiddleware", AwsGoDependency.AWS_SIGNER_V4)
                                         .build())
                                 .build())
                         .build(),
 
-                // Add newAttempt middleware to operation stack
+                // Add signed payload middleware to operation stack
```
skotambkar, line 64: Description doesn't match with AddComputePayloadSHA256Middleware addition

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -41,51 +41,57 @@ public byte getOrder() {
...
+                            if (service.hasTrait(SigV4Trait.class) && !operation.hasTrait(UnsignedPayloadTrait.class)) {
+                                return true;
+                            }
+                            return false;
+                        })
                         .registerMiddleware(MiddlewareRegistrar.builder()
-                                .resolvedFunction( SymbolUtils.createValueSymbolBuilder(
-                                        "AddRetryMiddlewares", AwsGoDependency.AWS_RETRY)
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(
+                                        "AddComputePayloadSHA256Middleware", AwsGoDependency.AWS_SIGNER_V4)
```
skotambkar, line 74: Should AddComputePayloadSHA256Middleware be before UnsignedPayload Middleware addition?

jasdel:
Test failing because the use of [Header.Values](https://golang.org/pkg/net/http/#Header.Values) that wasn't added until go 1.14

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
jasdel: Ideally they are mutually exclusive, so cannot be present at the same time.

aws/signer/v4/middleware.go
jasdel: So this is one that I'd like to suggest both this and unsigned payload SHA256 middleware be moved form `Finalize` to `Build` to better signal that they are middleware that are only performed once. Where as middleware in `Finalize` could expect to be called multiple times per request attempt.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
skotambkar: Oh I thought the change was required as they can exist together. Why separate it from HTTPSignerMiddlewares then?

aws/signer/v4/middleware.go
skotambkar: Yeah I think both of them should be moved to Build.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
jasdel: Mainly because only one could be used at a time. Either Checksum or unsigned middleware, if both were present the wrong behavior might occur.  This was the reasoning for breaking out checksum middleware from the signer middleware. So that the signer could always be used separately from how the payload checksum is or is not set.

aws/signer/v4/middleware.go
jasdel: Also we should move setting the content-sha256 header to its own middleware probably

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -41,61 +41,91 @@ public byte getOrder() {
...
                         .build(),
 
                 // Add HTTPSigner middleware to operation stack
                 RuntimeClientPlugin.builder()
                         .operationPredicate((model, service, operation) -> {
-                            if (service.hasTrait(SigV4Trait.class) && (!operation.hasTrait(UnsignedPayloadTrait.class))
+                            if (service.hasTrait(SigV4Trait.class)
                                     && (operation.hasTrait(SigV4Trait.class) || !operation.hasTrait(AuthTrait.class))
-                            ){
+                            ) {
```
skotambkar, line 109: Does AuthTrait.class also include UnsignedPayloadTrait.class? Would this predicate work if service is decorated with SigV4 trait and an operation is decorated with `UnsignedPayloadTrait`?

jasdel: No so UnsignedPayloadTrait is a standalone trait that is not coupled to the authentication scheme. `[auth trait](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#auth-trait) is a list of authentication schemes that are intended to apply direclty to the service or subset of auth schemes for an operation.

I think the way the `SigV4Trait` is used on the operation is wrong here, and should be updated to check the values of the `AuthTrait` instead.  It looks like [authDefintionTrait](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#authdefinition-trait) aka Auth schemes are only supposed to decorate services not operations.

Looks like the logic for these middleware needs to be tweeked with the expectation of decorating traits, and auth schemes.

jasdel: switched to use the `ServiceIndex` and `getEffectiveAuthSchemes` utilities to check for auth scheme.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -41,62 +48,100 @@ public byte getOrder() {
...
+                                        .build())
                                 .useClientOptions()
                                 .build())
                         .build(),
 
-                // Add attemptClockSkew middleware to operation stack
+                // Add unsigned payload middleware to operation stack
                 RuntimeClientPlugin.builder()
+                        .operationPredicate((model, service, operation) -> {
+                            boolean hasSigV4Auth = hasSigV4AuthScheme(model, service, operation);
```
JordonPhillips, line 59: nit: This assignment isn't necessary

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -41,62 +48,100 @@ public byte getOrder() {
...
+                                        "AddUnsignedPayloadMiddleware", AwsGoDependency.AWS_SIGNER_V4)
                                         .build())
                                 .build())
                         .build(),
 
-                // Add newAttempt middleware to operation stack
+                // Add signed payload middleware to operation stack
                 RuntimeClientPlugin.builder()
+                        .operationPredicate((model, service, operation) -> {
+                            boolean hasSigV4Auth = hasSigV4AuthScheme(model, service, operation);
```
JordonPhillips, line 73: ditto

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
jasdel: mainly just to make the lines shorter, but can drop extra variable. :)

PR #627, opened 2020-06-30
AwsRestJson1 Deserializer Improvements

skmcgrail:
* Fixes nested payload structures by delegating to the correct document deserializer where appropriate.
* Removes the need return a boolean indicator on whether a non-null value was deserialized from the underlying JSON document
  * Structures, Maps, and List/Set shape types are now passed by pointers to their respective smithy representation.
  * This allows for the function that is delegated to decide whether to allocate a value or not based on the returned tokens from the JSON decoder.
  * The delegated function will not allocate a new value if the value pointed to is not null. This allows for the top level output operation shapes to be created ahead of time to allow for http binding deserialization.

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
@@ -607,9 +607,7 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
             MemberShape memberShape = model.getKnowledge(HttpBindingIndex.class)
                     .getResponseBindings(operation, HttpBinding.Location.PAYLOAD).stream()
                     .findFirst()
-                    .orElseThrow(() -> {
-                        throw new CodegenException("Expected payload binding member");
-                    })
+                    .orElseThrow(() -> new CodegenException("Expected payload binding member"))
                     .getMember();
 
             Shape payloadShape = model.expectShape(memberShape.getTarget());


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -1064,10 +1082,13 @@ private void generateDocumentBindingStructureShapeDeserializer(
             Predicate<MemberShape> filterMemberShapes
     ) {
         writeJsonTokenizerStartStub(writer, shape);
+        writer.openBlock("if *v == nil {", "}", () -> {
+            writer.write("*v = &$T{}", symbolProvider.toSymbol(shape));
+        });
```
jasdel, line 1087: Should this always happen regardless if the underlying `*T` is nil?  This will ensure that deserializing an empty object will overwrite any existing values in T that may be present. Kind of contrived given the context of deserialization. But I'd expect it to be more of an error if T already had a value.

internal/protocoltest/awsrestjson/api_op_EmptyInputAndEmptyOutput_test.go
```diff
@@ -120,7 +120,7 @@ func TestClient_EmptyInputAndEmptyOutput_awsRestjson1Deserialize(t *testing.T) {
 		"RestJsonEmptyInputAndEmptyOutput": {
 			StatusCode:    200,
 			BodyMediaType: "application/json",
-			Body:          []byte(``),
+			Body:          []byte(`{}`),
 			ExpectResult:  &EmptyInputAndEmptyOutputOutput{},
```
jasdel, line 124: might want to check if smithy core was updated to include this updated test case. @JordonPhillips mentioned that it should be.

internal/protocoltest/awsrestjson/deserializers.go
```diff
@@ -686,69 +418,69 @@ func (m *awsRestjson1_deserializeOpJsonMaps) HandleDeserialize(ctx context.Conte
...
+		return err
 	}
 	if startToken == nil {
-		return false, nil
+		return nil
 	}
 	if t, ok := startToken.(json.Delim); !ok || t.String() != "{" {
-		return false, fmt.Errorf("expect `{` as start token")
+		return fmt.Errorf("expect `{` as start token")
 	}
```
jasdel, line 438: Outside the scope of this PR, but just noticed this. All of these start/end token comparisons can do value comparison instead of converting the delim rune to a string.

```go
if t, ok := startToken, (json.Delim); !ok || t != json.Delim('{') {
```

Minor, but may save the additional allocations during deserialization turning the rune into a string potentially.

internal/protocoltest/awsrestjson/deserializers.go
```diff
@@ -886,70 +591,221 @@ func awsRestjson1_deserializeHttpBindingsHttpPayloadTraitsWithMediaTypeOutput(v
...
+			(*v).IntegerValue = st
+
+		case "longValue":
+			val, err := decoder.Token()
+			if err != nil {
+				return err
+			}
+			var st *int64
+			if val != nil {
+				jtv, ok := val.(json.Number)
```
jasdel, line 731: Also outside the scope of this PR, but  these `if val != nil` checks could be simplified a little if wanted so that they return early instead of intended.

```go
var st *int64
if val == nil {
    continue
}
// get value
(*v).LongValue = st
```

skmcgrail: Ah good call, as this is generated within the context of a loop

internal/protocoltest/awsrestjson/api_op_EmptyInputAndEmptyOutput_test.go
skmcgrail: Correct the test case model was consumed so this regeneration picked up that change.

internal/protocoltest/awsrestjson/deserializers.go
skmcgrail: Looking at this more, we would require a bit more changes to the deserializer if we wanted to simplify this further. As calling continue is not always valid unless there is further inspection on the context of the parent caller.

PR #632, opened 2020-07-07
Auto-generate build.json for aws models

JordonPhillips:
This auto-generates the `smithy-build.json` so any model can be put into the `aws-models` directory to have it start generating.


codegen/sdk-codegen/build.gradle.kts
```diff
@@ -35,8 +38,53 @@ tasks.create<SmithyBuild>("buildSdk") {
...
+                    // as part of the actual build task done by the smithy gradle plugin.
+                    .assemble().result.get();
+            val services = model.shapes(ServiceShape::class.javaObjectType).sorted().toList();
+            if (services.size != 1) {
+                throw Exception("There must be exactly one service in each aws model file, but found " +
+                        "${services.size} in ${file.name}: ${services.map { it.id }}");
+            }
+            val service = services[0]
+            var (sdkId, version, remaining) = file.name.split(".")
+            sdkId = sdkId.replace("-", "").toLowerCase();
```
jasdel, line 64: Should spaces be stripped out as well?  ServiceIds aren't supposed to contain white space, but wasn't sure if Smithy validated this. Several of the old c2j modules contain spaces within their serviceId values.

JordonPhillips: This is based on the filename, which already has any set of consecutive whitespace chars replaced with `-`

PR #635, opened 2020-07-10
Substitute string values "GET", "POST" to net/http constants

heartwilltell:
It is better to use constants rather than string values.

### DIFF
--- a/aws/signer/v2/v2.go
+++ b/aws/signer/v2/v2.go
@@ -57,7 +57,7 @@ func SignSDKRequest(req *aws.Request) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


jasdel:
Thanks for taking the time to create this PR @heartwilltell. The v2 SDK is currently going through a major refactor iteration in the [feature/SmithyAPIClient](https://github.com/aws/aws-sdk-go-v2/tree/feature/SmithyAPIClient) feature branch.  The `aws/signer/v2` package will probably be dropped, but it looks like the SDK commonly uses `"GET"`/`"POST"`/etc instead of the http.Methodxxx constants.

We should take this idea and create an issue to update the rest of the SDK to use these constants as well.

PR #642, opened 2020-07-15
Fixed typo in interface.go example

Doug-AWS:
Typo in mockClientClient struct definition.

### DIFF
--- a/service/ssm/ssmiface/interface.go
+++ b/service/ssm/ssmiface/interface.go
@@ -41,7 +41,7 @@ import (
 //
 //    // Define a mock struct to be used in your unit tests of myFunc.
 //    type mockClientClient struct {
-//        ssmiface.ClientPI
+//        ssmiface.ClientAPI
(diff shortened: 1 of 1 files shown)


service/ssm/ssmiface/interface.go
```diff
@@ -41,7 +41,7 @@ import (
 //
 //    // Define a mock struct to be used in your unit tests of myFunc.
 //    type mockClientClient struct {
-//        ssmiface.ClientPI
```
jasdel, line 44: This for pointing this out. This issue is actually apart of the SDK's code generation. With the `feature/SmithyAPIClient` branch we're working these iface packages will be removed.

jasdel:
Thanks for creating this PR, but the future changes to the v2 SDK are going to be produced from `feature/SmithyAPIClient` branch once it is merged into master. In the smithy generated API clients the API interfaces will not be generated so will no be needed.

PR #643, opened 2020-07-15
Per Service Endpoint Metadata Generator

skmcgrail:
This PR adds support for generating a unique endpoint resolver and corresponding middleware per service client.

**Note:** As to avoid a significant amount of refactoring the changes to the internalized view of the endpoints model has been moved into `aws/endpoints/v2`, but the goal would be to replace the existing contents in `aws/endpoints` with this new version.

Lex Public Interface: https://github.com/skmcgrail/aws-sdk-go-v2/blob/endpoints/service/lexruntimeservice/endpoints.go
Lex Internal Data and Implementation:  https://github.com/skmcgrail/aws-sdk-go-v2/blob/endpoints/service/lexruntimeservice/internal/endpoints/endpoints.go

Additionally I've rebased the Lex test client PR onto this branch, and have verified that the default endpoint resolver is used when the client is not configured, and that the service endpoints resolve correctly.


aws/endpoints/v2/ednpoints.go
```diff
@@ -0,0 +1,177 @@
+package endpoints
```
jasdel, line 1: We can actually replace the existing endpoints implementation since all API clients will be regenerated with this pattern. No need to preserve the old behavior.

jasdel: Also this file name has minor typo in it.

service/lexruntimeservice/endpoints_test.go
```diff
@@ -0,0 +1,444 @@
...
+				"us-west-2": {
+					Hostname:  "runtime.lex.us-west-2.amazonaws.com",
+					Protocols: []string{"https"},
+				},
+				"us-west-2-fips": {
+					Hostname:  "runtime-fips.lex.us-west-2.amazonaws.com",
+					Protocols: []string{"https"},
+					InferredCapabilities: &Capabilities{
+						FIPS: true,
+					},
```
jasdel: nit should this be a pointer since all of the other types are values?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/EndpointGenerator.java
```diff
@@ -0,0 +1,456 @@
...
+        final String internalAlias = "serviceEndpoints";
+
+        Symbol awsEndpointSymbol = SymbolUtils.createValueSymbolBuilder("Endpoint", AwsGoDependency.AWS_CORE).build();
+        Symbol internalEndpointsSymbol = getInternalEndpointsSymbol(getResolverImplementationName(), true,
+                internalAlias).build();
+
+        // Generate Resolver Interface
+        writer.openBlock("type $L interface {", "}", getResolverInterfaceName(), () -> {
+            writer.write("ResolveEndpoint(region string) ($T, error)", awsEndpointSymbol);
+        });
```
jasdel, line 247: For future capabilities does it make sense for the ResolveEndpoint to take the same options as the EndpointResolver interface? Also allows the same resolver value to be shared across all operations

internal/protocoltest/awsrestjson/internal/endpoints/endpoints.go
```diff
@@ -0,0 +1,92 @@
...
+var DefaultPartitions = endpoints.Partitions{
+	{
+		ID: "aws",
+		Defaults: endpoints.Endpoint{
+			Hostname:          "restjson.{region}.amazonaws.com",
+			Protocols:         []string{"https"},
+			SignatureVersions: []string{"v4"},
+		},
+		RegionRegex: func() *regexp.Regexp {
+			r, _ := regexp.Compile("^(us|eu|ap|sa|ca|me|af)\\-\\w+\\-\\d+$")
```
jasdel, line 35: Maybe this should be a must compile to capture regex modeling errors during CI/CD? The must compile would also get rid of the extra inline function call.

service/lexruntimeservice/endpoints_test.go
skmcgrail: this looks like old artifact history

internal/protocoltest/awsrestjson/internal/endpoints/endpoints.go
skmcgrail: Yeah I was debating on this, it would require that we attempted to execute the code to ensure that the `MustCompile` checks pass.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -178,4 +199,81 @@ private void writeAwsConfigConstructor(Model model, ServiceShape service, GoWrit
...
+        public Optional<Symbol> getResolverFunction() {
+            return Optional.ofNullable(resolveFunction);
+        }
+
+        public static Builder builder() {
+            return new Builder();
+        }
+
+        private static class Builder extends ConfigField.Builder {
+            private boolean generatedOnClient;
```
JordonPhillips, line 232: Probably best to just default this to `true` since that's the actual value almost always

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -134,42 +151,46 @@ public void writeAdditionalFiles(
...
-                                .build()))
-                        .servicePredicate((model, serviceShape) -> model.getKnowledge(ServiceIndex.class)
-                                .getAuthSchemes(serviceShape).values().stream().anyMatch(trait -> trait.getClass()
-                                        .equals(SigV4Trait.class)))
-                        .build()
-        );
+        List<RuntimeClientPlugin> plugins = new ArrayList<>();
+
+        List<ConfigField> allClients = AWS_CONFIG_FIELDS.stream().filter(AwsConfigField::isGeneratedOnClient)
+                .filter(field -> !field.getServicePredicate().isPresent())
```
JordonPhillips, line 157: Why is this filter being applied?

skmcgrail: This is strictly to group all config fields that don't have a specified service predicate into a single RuntimeClientPlugin. Versus generating a RuntimeClientPlugin per config field.

JordonPhillips: That's fine, just add a comment and it's good

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -13,6 +13,21 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
 package software.amazon.smithy.aws.go.codegen;
```
jasdel, line 31: duplicated copywrite

skmcgrail: I feel like this happens to me a lot in Intellij for some reason. Not sure if I'm hitting some hot key that does it.

PR #647, opened 2020-07-16
Add generated dynamodb client

JordonPhillips:
This adds a generated dynamodb client with the service id changed to suffix "gov2" so that it can exist alongside the older generated client. The intent is to make the client available for testing while the work to untangle the older version from tests is ongoing.


JordonPhillips:
Go V2

PR #649, opened 2020-07-18
Regenerate protocol tests with streaming support

jasdel:
Regenerates the SDK's protocol tests based on the fixes in smithy-go https://github.com/awslabs/smithy-go/pull/119. To fix generation of protocol tests with streaming shapes.


jasdel:
Once awslabs/smithy-go#119 is merged in and a smithy-go Go module hash update is in, can update this PR with the correct module hashes.

PR #652, opened 2020-07-22
Implement query protocol serialization

JordonPhillips:
This adds a query encoder and implements query protocol serialization.

The biggest difference between the query encoder and the json encoder interfaces is that in query there's a distinction between a map and an "object" (structure/union).

### DIFF
--- a/internal/protocoltest/awsrestjson/go.mod
+++ b/internal/protocoltest/awsrestjson/go.mod
@@ -4,6 +4,6 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 	github.com/google/go-cmp v0.4.1
 )
--- a/internal/protocoltest/ec2query/go.mod
+++ b/internal/protocoltest/ec2query/go.mod
@@ -4,5 +4,5 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 )
--- a/internal/protocoltest/jsonrpc/go.mod
+++ b/internal/protocoltest/jsonrpc/go.mod
@@ -4,6 +4,6 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 	github.com/google/go-cmp v0.4.1
 )
--- a/internal/protocoltest/query/go.mod
+++ b/internal/protocoltest/query/go.mod
@@ -4,5 +4,6 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
+	github.com/google/go-cmp v0.4.1
 )
--- a/internal/protocoltest/restxml/go.mod
+++ b/internal/protocoltest/restxml/go.mod
@@ -4,6 +4,6 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 	github.com/google/go-cmp v0.4.1
 )
--- a/service/dynamodbgov2/go.mod
+++ b/service/dynamodbgov2/go.mod
@@ -4,5 +4,5 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 )
--- a/service/lexruntimeservice/go.mod
+++ b/service/lexruntimeservice/go.mod
@@ -4,5 +4,5 @@ go 1.14
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200720171838-25cc9a8769cf
-	github.com/awslabs/smithy-go v0.0.0-20200723171755-012806b992a2
+	github.com/awslabs/smithy-go v0.0.0-20200805220509-75306686f16c
 )


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsQuery.java
```diff
@@ -0,0 +1,101 @@
...
+        String functionName = ProtocolGenerator.getDocumentSerializerFunctionName(input, getProtocolName());
+        writer.addUseImports(AwsGoDependency.AWS_QUERY_PROTOCOL);
+
+        writer.write("bodyEncoder := query.NewEncoder()");
+        writer.write("body := bodyEncoder.Object()");
+        writer.write("body.Key(\"Action\").String($S)", operation.getId().getName());
+        writer.write("body.Key(\"Version\").String($S)", context.getService().getVersion());
+        writer.write("");
+
+        if (input.members().size() != 0) {
```
kstich, line 58: Nit: Should be `!input.members().isEmpty()`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
```diff
@@ -0,0 +1,228 @@
...
+
+    public QueryShapeSerVisitor(GenerationContext context, Predicate<MemberShape> memberFilter) {
+        super(context);
+        this.memberFilter = memberFilter;
+    }
+
+    private DocumentMemberSerVisitor getMemberSerVisitor(MemberShape member, String source, String dest) {
+        // Get the timestamp format to be used, defaulting to epoch seconds.
+        Format format = member.getMemberTrait(getContext().getModel(), TimestampFormatTrait.class)
+                .map(TimestampFormatTrait::getFormat).orElse(DEFAULT_TIMESTAMP_FORMAT);
```
kstich, line 59: Nit: Can the `.orElse` be on its own line?

aws/protocol/query/encoder.go
```diff
@@ -0,0 +1,27 @@
+package query
+
+import "net/url"
+
+// Encoder is a Query encoder that supports construction of Query body
+// values using methods.
+type Encoder struct {
```
skotambkar, line 7: Should the query encoder be a part of smithy-go?

aws/protocol/query/array.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Array represents the encoding of Query lists and sets
+type Array struct {
+	values url.Values
+	prefix string
+	flat bool
+	memberName string
+	size int32
+}
```
skotambkar, line 15: I don't know what prefix member refers to for query. Is this like the xml headers prefix?

aws/protocol/query/encoder_test.go
```diff
@@ -0,0 +1,50 @@
...
+	smithytesting.AssertURLFormEqual(t, []byte(`map.entry.1.key=bar&map.entry.1.value=baz&map.entry.2.key=foo&map.entry.2.value=bin`), encoder.Bytes())
+}
+
+func TestEncodeFlatMap(t *testing.T) {
+	encoder := NewEncoder()
+	mapValue := encoder.Object().FlatKey("map").Map("key", "value")
+	mapValue.Key("bar").String("baz")
+	mapValue.Key("foo").String("bin")
+	smithytesting.AssertURLFormEqual(t, []byte(`map.1.key=bar&map.1.value=baz&map.2.key=foo&map.2.value=bin`), encoder.Bytes())
+}
```
skotambkar, line 50: Nit: These tests can be a single test function that has different cases in form of a  `map[string]struct{...}`. This is generally the pattern followed by the SDK. 

An example would be : https://github.com/awslabs/smithy-go/blob/012806b992a232f8c0805c5a6600bc4fb8142985/testing/document_test.go#L8

aws/protocol/query/map.go
```diff
@@ -0,0 +1,54 @@
...
+	prefix string
+	// Whether the map is flat or not.
+	flat bool
+	// The location name of the key. In most cases this should be "key".
+	keyLocationName string
+	// The location name of the value. In most cases this should be "value".
+	valueLocationName string
+	// Elements are stored in values, so we keep track of the list size here.
+	size int32
+}
```
skotambkar, line 22: Would be helpful for `Array` type to have comments similar to this.

aws/protocol/query/array.go
```diff
@@ -0,0 +1,37 @@
...
+	return &Array{
+		values: values,
+		prefix: prefix,
+		flat: flat,
+		memberName: memberName,
+	}
+}
+
+// Value adds a new element to the Query Array. Returns a Value type used to
+// encode the array element.
```
skotambkar, line 27: Would be helpful to document an example for each `flattened` vs `unflattened` array  and it's encoded value. 

Can optionally provide a link to smithy docs query serialization for further reference.

aws/protocol/query/map.go
```diff
@@ -0,0 +1,54 @@
...
+		values:            values,
+		prefix:            prefix,
+		flat:              flat,
+		keyLocationName:   keyLocationName,
+		valueLocationName: valueLocationName,
+	}
+}
+
+// Key adds the given named key to the Query map.
+// Returns a Value encoder that should be used to encode a Query value type.
```
skotambkar, line 35: Same comment as Array value. Would be helpful to document examples for encoded maps (both flattened, unflattened).

aws/protocol/query/map.go
```diff
@@ -0,0 +1,54 @@
...
+	var value string
+	if m.flat {
+		key = fmt.Sprintf("%s.%d.%s", m.prefix, m.size, m.keyLocationName)
+		value = fmt.Sprintf("%s.%d.%s", m.prefix, m.size, m.valueLocationName)
+	} else {
+		key = fmt.Sprintf("%s.entry.%d.%s", m.prefix, m.size, m.keyLocationName)
+		value = fmt.Sprintf("%s.entry.%d.%s", m.prefix, m.size, m.valueLocationName)
+	}
+
+	// The key can only be a string, so we just go ahead and set it here
```
skotambkar, line 49: Unrelated to this CR: Is this generally true for XML maps too?

aws/protocol/query/object.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (o *Object) key(name string, flatValue bool) Value {
+	var v Value
+	if o.prefix != "" {
+		v = newValue(o.values, fmt.Sprintf("%s.%s", o.prefix, name), flatValue)
+	} else {
+		v = newValue(o.values, name, flatValue)
+	}
+	return v
```
skotambkar, line 43: Nit: can just return newValue(...) for each case. 

Also first check should be `len(o.prefix) != 0`. 

```
if len(o.prefix) != 0 {
   return newValue(...)
} 
return newValue(...)

```

aws/protocol/query/encoder.go
JordonPhillips: No reason for it to - it's an AWS specific protocol.

aws/protocol/query/encoder_test.go
JordonPhillips: I wasn't sure how to square the fact that I'm building everything up and then having to call `.Bytes()` at the end, so I can't just make it one line.

aws/protocol/query/map.go
JordonPhillips: Modeled maps in smithy can only have string keys

aws/protocol/query/encoder.go
```diff
@@ -0,0 +1,27 @@
...
+
+// String returns the string output of the Query encoder
+func (e Encoder) String() string {
+	return e.values.Encode()
+}
+
+// Bytes returns the []byte slice of the Query encoder
+func (e Encoder) Bytes() []byte {
+	return []byte(e.values.Encode())
+}
```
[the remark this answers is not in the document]
JordonPhillips: Sure, I mostly did this to match the json encoder.

aws/protocol/query/encoder_test.go
skotambkar: The building part can have a member `func(encoder Encoder) []byte`. Not an important change though, just consistent.

aws/protocol/query/value.go
```diff
@@ -0,0 +1,46 @@
...
+	values url.Values
+	// The value's key, which will form the prefix for complex types.
+	key string
+	// Whether the value should be flattened or not if it's a flattenable type.
+	flat bool
+	httpbinding.QueryValue
+}
+
+func newValue(values url.Values, key string, flat bool) Value {
+	return Value{values, key, flat, httpbinding.NewQueryValue(values, key, false)}
```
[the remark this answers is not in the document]
JordonPhillips: Yeah, my concern was that I didn't want to duplicate the code that's already in `QueryValue`. Would there be a better way to re-use it?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
```diff
@@ -0,0 +1,229 @@
...
+            writer.write("om := object.Key(key)");
+
+            // Null values should be omitted for query.
+            if (!target.hasTrait(EnumTrait.class)) {
+                writer.openBlock("if vv := v[key]; vv == nil {", "}", () -> {
+                    writer.write("continue");
+                });
+            }
+
+            target.accept(getMemberSerVisitor(shape.getValue(), "v[key]", "om"));
```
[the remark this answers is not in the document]
JordonPhillips: That's a good question, I'll have to check.

internal/protocoltest/query/api_op_EmptyInputAndEmptyOutput_test.go
```diff
@@ -0,0 +1,184 @@
...
+			ExpectMethod:  "POST",
+			ExpectURIPath: "/",
+			ExpectQuery:   []smithytesting.QueryItem{},
+			ExpectHeader: http.Header{
+				"Content-Type": []string{"application/x-www-form-urlencoded"},
+			},
+			BodyMediaType: "application/x-www-form-urlencoded",
+			BodyAssert: func(actual io.Reader) error {
+				return smithytesting.CompareURLFormReaderBytes(actual, []byte(`Action=EmptyInputAndEmptyOutput
+			&Version=2020-01-08`))
```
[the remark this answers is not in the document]
JordonPhillips: The compare function does just that sorting

aws/protocol/query/map.go
```diff
@@ -0,0 +1,54 @@
...
+	// Query lists start a 1, so adjust the size first
+	m.size++
+	var key string
+	var value string
+	if m.flat {
+		key = fmt.Sprintf("%s.%d.%s", m.prefix, m.size, m.keyLocationName)
+		value = fmt.Sprintf("%s.%d.%s", m.prefix, m.size, m.valueLocationName)
+	} else {
+		key = fmt.Sprintf("%s.entry.%d.%s", m.prefix, m.size, m.keyLocationName)
+		value = fmt.Sprintf("%s.entry.%d.%s", m.prefix, m.size, m.valueLocationName)
```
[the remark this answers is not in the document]
JordonPhillips: I'll add comments to that effect into the docs for the prefix.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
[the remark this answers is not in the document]
JordonPhillips: It looks like there isn't a representation for empty lists/maps. Wouldn't the for loop handle ignoring empty lists though?

internal/protocoltest/query/serializers.go
```diff
@@ -0,0 +1,1503 @@
...
+		return out, metadata, &smithy.SerializationError{Err: fmt.Errorf("unknown input parameters type %T", in.Parameters)}
+	}
+
+	request.Request.URL.Path = "/"
+	request.Request.Method = "POST"
+	httpBindingEncoder, err := httpbinding.NewEncoder(request.URL.Path, request.URL.RawQuery, request.Header)
+	if err != nil {
+		return out, metadata, &smithy.SerializationError{Err: err}
+	}
+	httpBindingEncoder.SetHeader("Content-Type").String("application/x-www-form-urlencoded")
```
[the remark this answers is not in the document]
JordonPhillips: Is this in the context of pre-signed requests? I think you'd either need another entry point to enable that, or you'd need to move the body into the url as a second step.

aws/protocol/query/encoder.go
```diff
@@ -0,0 +1,27 @@
...
+
+import "net/url"
+
+// Encoder is a Query encoder that supports construction of Query body
+// values using methods.
+type Encoder struct {
+	// The query values that will be built up to manage encoding.
+	values url.Values
+	Value
+}
```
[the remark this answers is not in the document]
JordonPhillips: The difficulty there is that we pass around `Value` to all of the serializer functions, and `Object` doesn't embed `Value`. We could possibly solve this issue by unrolling the top level serialization into the operation's handler method. We'd potentially have some duplicated code, but I'm not sure it matters.

JordonPhillips:
> If we want to optimize this serialization without a new version of this protocol utility we'll have a couple things limiting us. `NewEncoder` doesn't take an `io.Writer` ...

In this case should I just accept one even though we're not using it?

skotambkar:
Currently this is using stdlib [net.url#Encode](https://golang.org/src/net/url/url.go?s=26139:26170#L903) directly which creates a buffer internally. An alternative is to type alias the stdlib net.url#Value, and modify the Encode method on `type aliased Value` to use the passed in io.Writer.

JordonPhillips:
> An alternative is to type alias the stdlib net.url#Value, and modify the Encode method on type aliased Value to use the passed in io.Writer.

So I can do that,  but then the exposed interface is super different from the json encoder. Rather than returning a value it's just adding into the writer. Is that the direction we want to go?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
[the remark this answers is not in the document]
jasdel: It should, I think my concern was mostly in the context if a prefix key was defined would an entry in the `url.Values` be created? If no, then probably don't need to add any additional checks.

internal/protocoltest/query/serializers.go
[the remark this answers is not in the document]
jasdel: When I asked, I was thinking there was a Query service that used GET for its requests. But, on looking back it looks like that was [Amazon CloudSearch Domain](https://github.com/aws/aws-sdk-go/pull/749/files) API with REST-JSON that used GETs with very long query strings.

Yeah, for presigned URLs we'll need to define some other kind of custom behavior. The current design is to generate a "presign" API client on a inclusive set of operations that we know support or need presign functionality.

aws/protocol/query/encoder.go
```diff
@@ -1,26 +1,62 @@
...
+			}
+			if _, err := e.writer.Write([]byte(`=`)); err != nil {
+				return err
+			}
+			if _, err := e.writer.Write([]byte(url.QueryEscape(value))); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
```
jasdel: To simplify this you could continue to use the `url.Value.Encode` method, and write the output of that to the `writer` the Query `Encoder` was created with.  Can also optimize for the `WriteString` method that many writer implementations implement, e.g. [byte.Buffer.WriteString](https://pkg.go.dev/bytes?tab=doc#Buffer.WriteString).

```go
func (e *Encoder) Encode() error {
     encoded := e.values.Encode()
     if ws, ok := e.writer.(interface{ WriteString(string) (int, error) }); ok {
         // Optimize for WriteString to prevent additional byte array being allocated.
         return ws.WriteString(encoded)
     }

     return e.writer.Write([]byte(encoded))
}
```

The idea behind taking an `io.Writer` as input is that later, if we want to improve performance of the query `Encoder` the utility would be writing to a stream of bytes instead of using the `url.Values` utility. `url.Values` as an intermediate encoding datatype will cause unneeded extra allocations per member written.

JordonPhillips: Not deferring to `url.Values.Encode` was a suggestion from @skotambkar to eliminate the extra allocation from the buffer it creates. Is that just not much of a concern here, or is there something else I'm not seeing?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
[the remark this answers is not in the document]
JordonPhillips: I don't *think* that'll be an issue since the collection types don't put anything into the list until you start adding entries. I'll double check.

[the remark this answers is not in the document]
JordonPhillips: I updated it to put the entry object creation after the null check, which does actually fix this issue for maps. For lists it's a readability improvement. I also added a short-circuit check for maps to avoid the extra work/allocations done to sort them.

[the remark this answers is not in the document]
JordonPhillips: I also added one for lists just to avoid creating the extra allocation

aws/protocol/query/encoder.go
jasdel: An i see. In that case it would be good to remove the need for the extra `[]byte` casts ,because there is a good chance each one of those will do an allocation, because a byte slice is mutable.

One of way to address this is to test if the `io.Writer` passed into `Encoder`'s constructor has the `WriteString` method up front, and if not wrap it with a type that does for convenience.

```go
type wrapWriterString struct {
      w io.Writer
}
func (w wrapWriterString) WriteString(v string) (int, error) {
      return w.Write([]byte(v))
}
```

Then can update the `Encode` method to check for a string writer, and if not wrap the writer.  The SDK should always be using an io.Writer implementation that has the `WriteString` method, e.g. `bytes.Buffer`

```go
func (e Encoder) Encode() error {
     ws, ok := e.writer.(interface{ WriteString(string) (int, error) })
     if !ok {
          // Falls back to less optimal byte slice casting if not available.
          ws = &wrapWriteString{w: e.writer}
     }

     // ...
     //  use as string writer, ws.WriteString(escapedKey)
     // ...
}

Comment (JordonPhillips) [reply aws/protocol/query/encoder.go]: 👍

File: aws/protocol/query/encoder_test.go (hunk 5 above)
Comment (jasdel) [reply aws/protocol/query/encoder_test.go:120]: If you were to use [table testing]() with these cases to squash a couple different tests together into a group it could look something like:

```go
func TestEncode(t *testing.T) {
     cases := map[string]struct{
          Encode func(*Encoder) error
          Expect []byte
     } {
         "list": {
               Encode: func(e Encoder) error {
                    list := e.Object().Key("list").Array("spam")
                    list.Value().String("spam")
                    list.Value().String("eggs")
                    return e.Encode()
               },
               Expect: []byte(`list.spam.1=spam&list.spam.2=eggs`),
         },
         "flat list": {
               Encode: func(e Encoder) error {
                   list := encoder.Object().FlatKey("list").Array("spam")
                   list.Value().String("spam")
                   list.Value().String("eggs")
                    return e.Encode()
               },
               Expect: []byte(`list.1=spam&list.2=eggs`),
         },
        // ...
     }

     for name, c := range cases {
         t.Run(name, func(t *testing.T) {
               var buff bytes.Buffer
	       encoder := NewEncoder(&buff)

               if err := c.Encode(encoder); err != nil {
                        t.Fatalf("failed to encode, %v", err)
               }

                smithytesting.AssertURLFormEqual(t, c.Expect, buff.Bytes())
          })
     }
}

File: codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java [hunk 19]
```diff
@@ -0,0 +1,233 @@
...
+        writer.write("");
+
+        writer.openBlock("for i := range v {", "}", () -> {
+            // Null values should be omitted for query.
+            if (!target.hasTrait(EnumTrait.class)) {
+                writer.openBlock("if vv := v[i]; vv == nil {", "}", () -> {
+                    writer.write("continue");
+                });
+            }
+
```
jasdel, line 88: this implies that query does not support sparse collections, correct?

internal/protocoltest/query/serializers.go
```diff
@@ -0,0 +1,1672 @@
...
```
jasdel, line 45: Minor callout, with REST-JSON v1 SDK serializer it was discovered that several of the SDKs will write a Content-Type header even though the operation has no input parameters, https://github.com/awslabs/smithy-go/issues/110 Does Query protocol have any considerations we need to handle, or is it "safe" to always send the content type

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/QueryShapeSerVisitor.java
JordonPhillips: Right. There's more debate going on as to whether they should be supported at all, even where it is technically accepted. At any rate this can be easily changed if need be, though I'm not sure how we would serialize a null for query.

internal/protocoltest/query/serializers.go
JordonPhillips: python / cli always set it and have done for at least six years so I don't think there's any issues. The only fiddly bit is there are some sigv2 services that can have issues if you don't add the charset, but we aren't supporting sigv2.

PR #661, opened 2020-07-24
xml document serializer implementation

skotambkar:
To only run serializer protocol test associated with this change, use: 
`go test -run 'Test.*Serialize' ./internal/protocoltest/restxml`

or configure your IntelliJ run configuration's Go tool argument as `-run 'Test.*Serialize'` 

* Implementation for code generating XML document serializers. 
* Depends on https://github.com/awslabs/smithy-go/pull/124, which adds xml assertion and xml encoder libraries.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/aws/aws-sdk-go-v2
 
 require (
-	github.com/awslabs/smithy-go v0.0.0-20200819234228-98aeffe1013a
+	github.com/awslabs/smithy-go v0.0.0-20200820060017-084b57e08122
 	github.com/go-sql-driver/mysql v1.5.0
 	github.com/google/go-cmp v0.4.1
 	github.com/jmespath/go-jmespath v0.0.0-20180206201540-c2b33e8439af
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,5 @@
-github.com/awslabs/smithy-go v0.0.0-20200702221640-b1c2d4088d13 h1:YfjQNH1IJhR1SLM1Mme/7G5i1hHMNsJFBjY9nPsbAcg=
-github.com/awslabs/smithy-go v0.0.0-20200702221640-b1c2d4088d13/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
+github.com/awslabs/smithy-go v0.0.0-20200820060017-084b57e08122 h1:bTe1aDzVUqWuTf2YOMFLam9uPrWIonyNCU6qtpJ4Tlg=
+github.com/awslabs/smithy-go v0.0.0-20200820060017-084b57e08122/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/go-sql-driver/mysql v1.5.0 h1:ozyZYNQW3x3HtqT1jira07DN2PArx2v7/mN66gGcHOs=


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberSerVisitor.java
```diff
@@ -0,0 +1,228 @@
...
+import software.amazon.smithy.model.shapes.ShapeVisitor;
+import software.amazon.smithy.model.shapes.ShortShape;
+import software.amazon.smithy.model.shapes.StringShape;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.model.shapes.TimestampShape;
+import software.amazon.smithy.model.shapes.UnionShape;
+import software.amazon.smithy.model.traits.EnumTrait;
+import software.amazon.smithy.model.traits.TimestampFormatTrait;
+
+public class XmlMemberSerVisitor implements ShapeVisitor<Void> {
```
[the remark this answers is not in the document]
skotambkar: Yeah this is a copy. I will look at your PRs, and see how we can de-dupe this.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
```diff
@@ -162,4 +172,91 @@ public static void handleDecodeError(GoWriter writer, String returnExtras) {
...
+            XmlNamespaceTrait ns = xmlNamespaceTrait.get();
+            writer.write("$L = append($L, smithyxml.NewNamespaceAttribute($S, $S))",
+                    attrName, attrName,
+                    ns.getPrefix().isPresent() ? ns.getPrefix().get() : "", ns.getUri()
+            );
+        }
+
+        // Traverse member shapes to get attributes
+        shape.members().stream().forEach(memberShape -> {
+            Optional<XmlAttributeTrait> xmlAttributeTrait = memberShape.getTrait(XmlAttributeTrait.class);
```
JordonPhillips, line 200: You're not using the value, so you should just use `hasTrait`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
```diff
@@ -162,4 +172,91 @@ public static void handleDecodeError(GoWriter writer, String returnExtras) {
...
+        if (clone.isPresent()) {
+            SyntheticClone cl = clone.get();
+            shapeName = cl.getArchetype().getName();
+        }
+
+        // check if shape is member shape
+        Optional<MemberShape> member = shape.asMemberShape();
+        if (member.isPresent()) {
+            MemberShape memberShape = member.get();
+            return getSerializedXMLMemberName(memberShape);
```
JordonPhillips, line 256: nit: you can just put `member.get()` here

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
```diff
@@ -162,4 +172,91 @@ public static void handleDecodeError(GoWriter writer, String returnExtras) {
...
+        }
+
+        Optional<XmlNameTrait> xmlNameTrait = shape.getTrait(XmlNameTrait.class);
+        return xmlNameTrait.isPresent() ? xmlNameTrait.get().getValue() : shapeName;
```
JordonPhillips, line 260: Something like the following would be better:

`shape.getTrait(XmlNameTrait.class).map(XmlNameTrait::getValue).orElse(shapeName)`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -21,24 +31,41 @@
         super(true);
     }
 
+
+    @Override
+    public void generateSharedDeserializerComponents(GenerationContext context) {
+        super.generateSharedDeserializerComponents(context);
```
JordonPhillips, line 37: You don't need to pull in this method since you're not changing it

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -21,24 +31,41 @@
...
+                .map(HttpBinding::getMember)
+                .collect(Collectors.toSet());
+
+        if (documentBindings.size() == 0) {
+            return;
+        }
+        Shape inputShape = ProtocolUtils.expectInput(context.getModel(), operation);
+        inputShape.accept(new XmlShapeSerVisitor(context,
+            memberShape -> documentBindings.contains(memberShape) &&
+                    !memberShape.hasTrait(XmlAttributeTrait.class))
```
JordonPhillips, line 67: This can be on one line

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -47,43 +74,123 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
+
+        AwsProtocolUtils.generateXMLStartElement(context, inputShape, "root", "input");
+        writer.openBlock("if err := $L(input, xmlEncoder.RootElement(root)); err != nil {", "}",
+                functionName, () -> {
+            writer.write("return out, metadata, &smithy.SerializationError{Err: err}");
+        });
+        writer.insertTrailingNewline();
+
+        writer.openBlock("if request, err = request.SetStream(bytes.NewReader(xmlEncoder.Bytes())); " +
+                        "err != nil {", "}", () -> {
```
JordonPhillips, line 99: Nit: The `+` should be on the new line

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -47,43 +74,123 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
     }
 
     @Override
-    protected void writeMiddlewarePayloadSerializerDelegator(
-            GenerationContext context,
-            OperationShape operation,
-            MemberShape memberShape,
-            GoStackStepMiddlewareGenerator generator
+    protected void generateDocumentBodyShapeSerializers(
+            GenerationContext context, Set<Shape> shapes
```
JordonPhillips, line 155: Nit: This function declaration can fit on one line

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -102,8 +209,19 @@ protected void generateDocumentBodyShapeDeserializers(
...
     }
 
+
+
     @Override
-    public ShapeId getProtocol() {
-        return null;
+    protected void writeErrorMessageCodeDeserializer(GenerationContext context) {
+        context.getWriter().writeDocs("TODO: implement error message / code deser");
+        context.getWriter().write("_= errorBody");
```
JordonPhillips, line 217: Nit: there should probably be a space after the underscore

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
```diff
@@ -0,0 +1,221 @@
...
+import software.amazon.smithy.utils.FunctionalUtils;
+
+final class XmlShapeSerVisitor extends DocumentShapeSerVisitor {
+    private static final TimestampFormatTrait.Format DEFAULT_TIMESTAMP_FORMAT = TimestampFormatTrait.Format.DATE_TIME;
+    private static final Logger LOGGER = Logger.getLogger(JsonShapeSerVisitor.class.getName());
+
+    private final Predicate<MemberShape> memberFilter;
+
+    public XmlShapeSerVisitor(
+            GenerationContext context
```
JordonPhillips, line 38: Nit: this constructor can fit on one line

internal/protocoltest/restxml/serializers.go
```diff
@@ -3,156 +3,204 @@
...
+
 	if request.Request, err = restEncoder.Encode(request.Request); err != nil {
 		return out, metadata, &smithy.SerializationError{Err: err}
 	}
 	in.Request = request
 
 	return next.HandleSerialize(ctx, in)
 }
-func awsRestxml_serializeHttpBindingsRecursiveShapesInput(v *RecursiveShapesInput, encoder *httpbinding.Encoder) error {
+func awsRestxml_serializeHttpBindingsAllQueryStringTypesInput(v *AllQueryStringTypesInput, encoder *httpbinding.Encoder) error {
```
jasdel, line 65: Are these being generated in stable order?  This change suggests that the order is different.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
```diff
@@ -162,4 +172,88 @@ public static void handleDecodeError(GoWriter writer, String returnExtras) {
...
+
+    /**
+     * generateXMLStartElement generates the XML start element for a shape. It is used to generate smithy xml's startElement.
+     *
+     * @param context  is the generation context.
+     * @param shape    is the Shape for which xml start element is to be generated.
+     * @param dst      is the operand name which holds the generated start element.
+     * @param inputSrc is the input variable for the shape with values to be serialized.
+     */
+    public static void generateXMLStartElement(GenerationContext context, Shape shape, String dst, String inputSrc) {
```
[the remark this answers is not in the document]
skotambkar: I think that pattern of spliting utils is good to be followed with go files. Not sure if we want/need to do this with Java classes.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
```diff
@@ -0,0 +1,217 @@
...
+            writer.insertTrailingNewline();
+
+            // Null values in maps should be serialized as such. Enums can't be null
+            if (!targetValue.hasTrait(EnumTrait.class)) {
+                writer.openBlock("if vv := v[key]; vv == nil {", "}", () -> {
+                    writer.write("entry.Close()");
+                    writer.write("continue");
+                });
+            }
+
```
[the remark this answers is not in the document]
skotambkar: There isn't a distinction between null and empty string value. 

Can you clarify the second part about unset enums. What would an unset enum look like?

internal/protocoltest/restxml/serializers.go
skotambkar: I couldn't reproduce any order change now. This change might be because earlier when this file was checked in with base code, the order was not stable.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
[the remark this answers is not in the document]
jasdel: Mainly so that they are easier to find, Having a XMLProtocolUtils.java class would make it easier to find XML relevant utilities, instead of seraching a generic AwsProtocolsUtils.  The benefit of this is that these XMlProtocollUtils could be moved to the smithy-go repo as well since they are not AWS specific.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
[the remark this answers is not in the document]
jasdel: unset enum is the same as empty string.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -47,47 +65,125 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
             GoStackStepMiddlewareGenerator generator
     ) {
-
+        GoWriter writer = context.getWriter();
+        writer.addUseImports(SmithyGoDependency.SMITHY);
+        writer.addUseImports(SmithyGoDependency.SMITHY_XML);
+
+        writer.write("restEncoder.SetHeader(\"Content-Type\").String($S)", getDocumentContentType());
+        writer.write("");
+
```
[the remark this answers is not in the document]
jasdel: There is the potential issue with API services if they inspect content-type prior to content-length, and incorrectly attempt to deserialize the request from the client when the service shouldn't. This is an edge case and more of a friendly neighbor kind of issue.  It shouldn't block this PR, but a general topic. Related to https://github.com/awslabs/smithy-go/issues/110

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
```diff
@@ -162,4 +175,154 @@ public static void handleDecodeError(GoWriter writer, String returnExtras) {
...
+                    throw new CodegenException("Unknown timestamp format");
+            }
+            return;
+        }
+
+        if (target.isStringShape()) {
+            writer.addUseImports(SmithyGoDependency.SMITHY_XML);
+            writer.write("$L = smithyxml.EscapeString(*$L)", dest, src);
+            return;
+        }
```
jasdel, line 253: Shouldn't escaping be handled by the encoder utility not the serializer before the encoder value is called?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -88,50 +88,24 @@ protected void generateOperationDocumentSerializer(
...
-                    () -> {
-                        writer.write("return out, metadata, &smithy.SerializationError{Err: err}");
-                    });
+        String functionName = ProtocolGenerator.getDocumentSerializerFunctionName(payloadShape,
+                getProtocolName());
+        writer.addUseImports(SmithyGoDependency.SMITHY_JSON);
+        writer.write("jsonEncoder := smithyjson.NewEncoder()");
+        writer.openBlock("if err := $L($L, jsonEncoder.Value); err != nil {", "}", functionName,
+                operand, () -> {
+            writer.write("return out, metadata, &smithy.SerializationError{Err: err}");
```
jasdel, line 106: Should the usage of the encoder check for encoding error? For example in the case, where the generated code incorrectly uses the encoder, and does not close and object or list?

internal/protocoltest/restxml/serializers.go
```diff
@@ -3,156 +3,204 @@
...
+type awsRestxml_serializeOpAllQueryStringTypes struct {
 }
 
-func (*awsRestxml_serializeOpRecursiveShapes) ID() string {
+func (*awsRestxml_serializeOpAllQueryStringTypes) ID() string {
 	return "OperationSerializer"
 }
 
-func (m *awsRestxml_serializeOpRecursiveShapes) HandleSerialize(ctx context.Context, in middleware.SerializeInput, next middleware.SerializeHandler) (
+func (m *awsRestxml_serializeOpAllQueryStringTypes) HandleSerialize(ctx context.Context, in middleware.SerializeInput, next middleware.SerializeHandler) (
```
jasdel, line 26: Is this file being generated in stable order? I wouldn't expect the order of these parameters to change.

skotambkar: I think they change because the last time this file was committed (long time ago), the order was not stable.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsProtocolUtils.java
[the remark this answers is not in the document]
JordonPhillips: I wonder how much we want to expose our XML stuff. The only XML protocols we have are considered deprecated. In the industry as a whole XML hasn't been popular for quite some time. We necessarily have to expose our XML go module for it to work, but I wonder how useful external customers would find it.

All in all I'm +0 on breaking out XML utilities and -0 on having them in smithy-go.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
skotambkar: This comment was made on restjson protocol gen, but still relevant.

In general if the generated code incorrectly uses the encoder then it should be caught by the protocol tests. We may consider adding some util to verify the correctness of json/xml later.

PR #666, opened 2020-08-12
Fix bug of not serialized for self closing bool

fuku2014:
XML containing self-closing boolean cannot be serialized.

Case:
```
<Element/>
```
Error:
```
expect no error, got strconv.ParseBool: parsing "": invalid syntax
```

### DIFF
--- a/private/protocol/xml/xmlutil/unmarshal.go
+++ b/private/protocol/xml/xmlutil/unmarshal.go
@@ -249,11 +249,13 @@ func parseScalar(r reflect.Value, node *XMLNode, tag reflect.StructTag) error {
 		}
 		r.Set(reflect.ValueOf(b))
 	case bool:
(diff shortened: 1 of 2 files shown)


jasdel:
Thanks for reporting this issue, and creating a PR for it @fuku2014. Were you using an AWS service API when you discovered this issue? If so, it would be very helpful to investigate this issue further. In the SDK's current working branch `feature/SmithyAPIClient` we were working under the impression that self closing tags were only valid for API parameters with `string` types.

If you have a reproduction of this, it would be every helpful in tracking down the expected behavior with regard to empty/self closing tags.

jasdel:
Thanks for creating this PR @fuku2014 the SDK's v0.25.0 refresh included a fix for this behavior internally and now correctly supports self closing tags.

PR #667, opened 2020-08-12
RestXml Deserialization support

skotambkar:
Associates with awslabs/smithy-go#131 that fixes a protocol test.

To run rest-xml deserialization specific test use: 

`go test -run 'Test.*Deserialize' ./internal/protocoltest/restxml/...`

Note: we might want to move few things around, as we deal with EC2Query and Query deserializers


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -0,0 +1,370 @@
...
+        });
+        return null;
+    }
+
+    /**
+     * Consumes a single token into the variable "val", returning on any error.
+     */
+    private void consumeToken(Shape shape) {
+        if (!filter.test(shape)) {
+         return;
```
JordonPhillips, line 95: Indentation. Also, won't this result in `val` not being defined at all and thus breaking the various nil checks that rely on it?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -0,0 +1,370 @@
...
+    @Override
+    public Void setShape(SetShape shape) {
+        return collectionShape(shape);
+    }
+
+    private Void collectionShape(CollectionShape shape) {
+        if (isFlattened) {
+            writeUnwrappedDelegateFunction(shape);
+        } else {
+        writeDelegateFunction(shape);
```
JordonPhillips, line 334: indentation

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        return new XmlMemberDeserVisitor(getContext(), dataDest, format, filter, isFlattened);
+    }
+
+    @Override
+    protected void deserializeCollection(GenerationContext context, CollectionShape shape) {
+        GoWriter writer = context.getWriter();
+        SymbolProvider symbolProvider = context.getSymbolProvider();
+        Symbol symbol = symbolProvider.toSymbol(shape);
+
+        // Initialize the value now that the start stub has verified that there's something there.
```
JordonPhillips, line 87: This seems to be copied over from the json version, but there's no start stub here.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        writer.write("var sv $P", symbol);
+        writer.openBlock("if *v == nil {", "", () -> {
+            writer.write("sv = make($P, 0)", symbol);
+            writer.openBlock("} else {", "}", () -> {
+                writer.write("sv = *v");
+            });
+        });
+
+        // Iterate through the decoder. The member visitor will handle popping tokens.
```
JordonPhillips, line 96: This does not seem to be entirely true as you are popping a token on each iteration, unless `.Token` doesn't consume in XML

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        writer.openBlock("for {", "}", () -> {
+            writer.write("t, done, err := decoder.Token()");
+            writer.write("if err != nil { return err }");
+            writer.write("if done { break }");
+
+            MemberShape member = shape.getMember();
+            Shape target = context.getModel().expectShape(member.getTarget());
+            String serializedMemberName = getSerializedMemberName(member);
+            writer.openBlock("switch t.Name.Local {", "}", () -> {
```
JordonPhillips, line 105: This switch appears to only have one case and a default, so it could probably just be an if

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+                            });
+                        });
+                    } else {
+                        target.accept(getMemberDeserVisitor(member, "col", FunctionalUtils.alwaysTrue()));
+                    }
+                    writer.write("sv = append(sv, col)");
+                });
+
+                writer.openBlock("default:", "", () -> {
+                    writer.write("// Do nothing and ignore");
```
JordonPhillips, line 143: You should use `writeDocs`. Also if you're going to write a comment here, I would say more about what's happened. This is a case of an unexpected tag right?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+                                    writer.write("decoder = smithydecoding.NewXMLNodeDecoder(memberDecoder.Decoder, t)");
+                                    writer.openBlock("for {", "}", () -> {
+                                        writer.write("var mv $P", targetMemberSymbol);
+                                        nestedTarget.accept(getMemberDeserVisitor(targetMember, "mv", FunctionalUtils.alwaysTrue()));
+                                        writer.write("col = append(col, mv)");
+                                    });
+                                });
+                                writer.write("decoder = originalDecoder");
+                            });
+                        });
```
JordonPhillips, line 135: Can you be more descriptive in your comments? I'm not sure why this is needed, and it seems like it'll break if there's more than one level of nesting.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+
+                    // in case of nested collections we will have nested `member` element tags
+                    if (target.isListShape() || target.isSetShape()) {
+                        writer.write("memberDecoder := smithydecoding.NewXMLNodeDecoder(decoder.Decoder, t)");
+                        writer.openBlock("for {", "}", () -> {
+                            writer.write("t, done, err = memberDecoder.Token()");
+                            writer.write("if err != nil { return err }");
+                            writer.write("if done { break }");
+
+                            CollectionShape targetAsCollection = getCollectionShape(target);
```
JordonPhillips, line 117: You should just cast

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+                writer.openBlock("default:", "", () -> {
+                    writer.write("// Do nothing and ignore");
+                });
+            });
+        });
+        writer.write("*v = sv");
+        writer.write("return nil");
+    }
+
+    private CollectionShape getCollectionShape(Shape shape) {
```
JordonPhillips, line 151: You should cast instead of using this function

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        writer.openBlock("func $L(v *$P, decoder *smithydecoding.XMLNodeDecoder) error {", "}",
+                getUnwrappedMapDelegateFunctionName(context, shape), symbol, () -> {
+                    // Initialize the value now that the start stub has verified that there's something there.
+                    writer.write("var sv $P", symbol);
+                    writer.openBlock("if *v == nil {", "", () -> {
+                        writer.write("sv = make($P, 0)", symbol);
+                        writer.openBlock("} else {", "}", () -> {
+                            writer.write("sv = *v");
+                        });
+                    });
```
JordonPhillips, line 183: This exact code block seems to be used in a lot of places with the same variable names, so it probably deserves to be its own function. Json could probably use a similar treatment alongside making variable names consistent across the collection types. (but don't worry about doing that second part here)

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+
+            // non-flattened maps
+            writer.openBlock("switch t.Name.Local {", "}", () -> {
+                writer.write("case \"entry\":");
+                writer.write("entryDecoder := smithydecoding.NewXMLNodeDecoder(decoder.Decoder, t)");
+                // delegate to unwrapped map deserializer function
+                writer.openBlock("if err := $L(&sv, entryDecoder); err != nil {", "}",
+                        getUnwrappedMapDelegateFunctionName(context, shape), () -> {
+                    writer.write("return err");
+                    });
```
JordonPhillips, line 230: indentation

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        });
+
+        // Iterate through the decoder. The member visitor will handle popping tokens.
+        writer.openBlock("for {", "}", () -> {
+            writer.write("t, done, err := decoder.Token()");
+            writer.write("if err != nil { return err }");
+            writer.write("if done { break }");
+
+            // non-flattened maps
+            writer.openBlock("switch t.Name.Local {", "}", () -> {
```
JordonPhillips, line 223: Similar as before, this should probably just be an if

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+                            // set the key value pair in map
+                            writer.write("sv[*ek] = ev");
+                            writer.write("break");
+                        });
+
+                        writer.openBlock("switch t.Name.Local {", "}", () -> {
+                            MemberShape keyShape = shape.getKey();
+                            writer.openBlock("case $S:", "", getSerializedMemberName(keyShape), () -> {
+                                String dest = "ek";
+                                context.getModel().expectShape(keyShape.getTarget()).accept(getMemberDeserVisitor(keyShape, dest,((shape1)-> !shape1.hasTrait(XmlAttributeTrait.ID))));
```
JordonPhillips, line 276: This looks like more than 120 characters. Also lambdas with one argument can be be set up without parens, e.g. `s -> s.foo` vs `(s) -> s.foo`. Also, only structure members can have the `xmlAttribute` trait.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+        // TODO: Refactor this to be more inline and DRY when handling xml attributes
+        Predicate<Shape> hasXmlAttributeTraitMember = shape1 -> {
+            Collection<MemberShape> members = shape1.members();
+            for (MemberShape member: members) {
+                if (member.hasTrait(XmlAttributeTrait.ID)) {
+                    return true;
+                }
+            }
+            return false;
+        };
```
JordonPhillips, line 323: This could just be a function yes

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+
+        if (hasXmlAttributeTraitMember.test(shape)) {
+            writer.openBlock("for _, attr := range decoder.StartEl.Attr {", "}", () -> {
+                writer.write("val := []byte(attr.Value)");
+                writer.openBlock("switch attr.Name.Local {", "}", () -> {
+                    Set<MemberShape> members = new TreeSet<>(shape.members());
+                    for (MemberShape member : members) {
+                        // check if member is not a document binding or does not have a xmlAttribute trait
```
JordonPhillips, line 331: Technically the member filter could be used for other things

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -0,0 +1,403 @@
...
+                        if (!memberFilter.test(member) || !member.hasTrait(XmlAttributeTrait.ID)) {
+                            continue;
+                        }
+
+                        String memberName = symbolProvider.toMemberName(member);
+                        String serializedMemberName = getSerializedMemberName(member);
+                        writer.openBlock("case $S:", "", serializedMemberName, () -> {
+                            String dest = "sv." + memberName;
+                            context.getModel().expectShape(member.getTarget()).accept(
+                                    getMemberDeserVisitor(member, dest, ((shape1) -> shape1.hasTrait(XmlAttributeTrait.ID))));
```
JordonPhillips, line 341: This is testing the target shape, which cannot have the xml attribute trait since it only applies to members.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -51,59 +65,309 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
+                visitor.generateFlattenedMapDeserializer(context, shape.asMapShape().get());
+            }
+
+            if (shape.isListShape()) {
+                visitor.generateFlattenedCollectionDeserializer(context, shape.asListShape().get());
+            }
 
+            if (shape.isSetShape()) {
+                visitor.generateFlattenedCollectionDeserializer(context, shape.asSetShape().get());
+            }
```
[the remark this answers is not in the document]
skotambkar: Can you give an example on how to do it? The documentation for DocumentShapeSerVisitor says that  
* Visitor methods for aggregate types are final and will generate functions that dispatch
 * their loading from the body to the matching abstract method.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
skotambkar: Will try to clarify the usage or extract the val variable out (if possible).

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -0,0 +1,370 @@
...
+    private final Format timestampFormat;
+    private final Predicate<Shape> filter;
+    private final Boolean isFlattened;
+
+
+    public XmlMemberDeserVisitor(GenerationContext context, String dataDest, Format timestampFormat, Predicate<Shape> filter, Boolean isFlattened) {
+        this.context = context;
+        this.dataDest = dataDest;
+        this.timestampFormat = timestampFormat;
+        this.filter = filter;
```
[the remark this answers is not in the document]
skotambkar: The filter is to identify xml attributes. In case of xml attributes we do not need to decode a value enclosed between the start/end tags, but rather the attribute value enclosed within the start element. 

I have renamed filter to `xmlAttributeMemberFilter` and will add more documentation

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
skotambkar: Just kept it switch to be consistent and not have to change the code flow specifically for this case. Switches will have similar performance to if-else in Go if not more.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
skotambkar: Yes, the default can also be an additional element received in response which is not modeled.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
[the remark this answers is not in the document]
JordonPhillips: Ah I had forgotten about that. You could un-finalize them, I'm not sure it matters all that much.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
JordonPhillips: This does seem to be a pattern though - where else besides the error dispatcher do you need the switch?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
skotambkar: Updated and modified the handling of nested lists.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
skotambkar: updated

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -51,59 +65,283 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
+    /**
+     * Generates code to retrieve error code or error message from the error response body
+     * This method is used by generateErrorDispatcher to generate operation specific error handling functions
+     *
+     * @param context the generation context
+     * @see <a href="https://awslabs.github.io/smithy/1.0/spec/aws/aws-restxml-protocol.html#operation-error-serialization">Rest-XML operation error serialization.</a>
+     */
+    private void writeErrorMessageCodeDeserializer(GenerationContext context) {
+        GoWriter writer = context.getWriter();
+        boolean isNoErrorWrapping = false;
```
JordonPhillips, line 145: You don't need this anymore

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -51,59 +65,283 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
+        if (payloadBindings.size() == 0) {
+            return;
+        }
+
+        writePayloadBindingDeserializer(context, outputShape, payloadBindings::contains);
+        writer.write("");
     }
 
+
+
```
JordonPhillips, line 234: excess whitespace

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    /**
+     * Consumes a single token into the variable "val", returning on any error.
+     */
+    private void consumeToken() {
+        GoWriter writer = context.getWriter();
+        if (isXmlAttributeMember) {
+            writer.write("val := []byte(attr.Value)");
+         return;
```
JordonPhillips, line 101: indentation is still messed up for this line

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -90,12 +94,13 @@ public Void booleanShape(BooleanShape shape) {
     /**
      * Consumes a single token into the variable "val", returning on any error.
      */
-    private void consumeToken(Shape shape) {
-        if (!filter.test(shape)) {
+    private void consumeToken() {
+        GoWriter writer = context.getWriter();
+        if (isXmlAttributeMember) {
```
JordonPhillips, line 99: Can you add a comment explaining why this is treated differently

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -67,32 +69,78 @@ public XmlShapeDeserVisitor(GenerationContext context, Predicate<MemberShape> me
         return Collections.singletonMap("decoder", "*smithydecoding.XMLNodeDecoder");
     }
 
-    private XmlMemberDeserVisitor getMemberDeserVisitor(MemberShape member, String dataDest, Predicate<Shape> filter) {
-        // Get the timestamp format to be used, defaulting to epoch seconds.
+    private XmlMemberDeserVisitor getMemberDeserVisitor(MemberShape member, String dataDest, Boolean isXmlAttibuteMember) {
```
JordonPhillips, line 72: You should use little-b `boolean` here, since you don't want null values

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -67,32 +69,78 @@ public XmlShapeDeserVisitor(GenerationContext context, Predicate<MemberShape> me
...
+            }
             writer.openBlock("} else {", "}", () -> {
                 writer.write("sv = *v");
             });
         });
+    }
+
+    @Override
+    public Void mapShape(MapShape shape) {
+        generateDeserFunction(shape, (c, s) -> deserializeMap(c, s.asMapShape().get()));
```
JordonPhillips, line 101: You should call the super here instead so you pick up any changes that happen there. Same for the others

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
skotambkar: Sure. 

The difference is that in case of Xml Attribute members, we already have the value in the root element's start element. ie. <root attr="value">

Thus, we do not need to pull out the next XML element from the body which we would do normally. ie. <root><attr>value</attr></root>

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -51,59 +66,238 @@ protected void writeMiddlewareDocumentSerializerDelegator(
...
+        writer.write("buff := make([]byte, 1024)");
+        writer.write("ringBuffer := smithyio.NewRingBuffer(buff)");
+        writer.insertTrailingNewline();
+
+        writer.addUseImports(SmithyGoDependency.IO);
+        writer.write("body := io.TeeReader(response.Body, ringBuffer)");
+        writer.write("defer response.Body.Close()");
+        writer.insertTrailingNewline();
+
+        writer.openBlock("if decoder == nil {", "}", () -> {
```
jasdel, line 283: I think the `decoder` here might be from an older implementation, since the decoder here in this flow can

PR #675, opened 2020-08-17
Smithy User-Agent Middleware

skmcgrail:
Adds a Smithy Build step middleware for constructing the request `User-Agent`.

### DIFF
--- a/aws/middleware/user_agent_test.go
+++ b/aws/middleware/user_agent_test.go
@@ -95,16 +95,8 @@ func TestRequestUserAgent_HandleBuild(t *testing.T) {
 
 	for name, tt := range cases {
 		t.Run(name, func(t *testing.T) {
-			environ := os.Environ()
-			os.Clearenv()
-			defer func() {
-				os.Clearenv()
-				for _, v := range environ {
-					split := strings.SplitN(v, "=", 2)
-					key, value := split[0], split[1]
-					os.Setenv(key, value)
-				}
-			}()
+			restoreEnv := clearEnv()
+			defer restoreEnv()
 			for k, v := range tt.Env {
 				os.Setenv(k, v)
 			}
@@ -119,7 +111,23 @@ func TestRequestUserAgent_HandleBuild(t *testing.T) {
 	}
 }
 
+func clearEnv() func() {
+	environ := os.Environ()
+	os.Clearenv()
+	return func() {
+		os.Clearenv()
+		for _, v := range environ {
+			split := strings.SplitN(v, "=", 2)
+			key, value := split[0], split[1]
+			os.Setenv(key, value)
+		}
+	}
+}
+
 func TestAddUserAgentKey(t *testing.T) {
+	restoreEnv := clearEnv()
+	defer restoreEnv()
+
 	b := newRequestUserAgent()
 	stack := middleware.NewStack("testStack", smithyhttp.NewStackRequest)
 	err := stack.Build.Add(b, middleware.After)
@@ -149,6 +157,9 @@ func TestAddUserAgentKey(t *testing.T) {
 }
 
 func TestAddUserAgentKeyValue(t *testing.T) {
+	restoreEnv := clearEnv()
+	defer restoreEnv()
+
 	b := newRequestUserAgent()
 	stack := middleware.NewStack("testStack", smithyhttp.NewStackRequest)
 	err := stack.Build.Add(b, middleware.After)
@@ -178,6 +189,9 @@ func TestAddUserAgentKeyValue(t *testing.T) {
 }
 
 func TestAddUserAgentKey_AddToStack(t *testing.T) {
+	restoreEnv := clearEnv()
+	defer restoreEnv()
+
 	stack := middleware.NewStack("testStack", smithyhttp.NewStackRequest)
 	bi := middleware.BuildInput{Request: &smithyhttp.Request{Request: &http.Request{Header: map[string][]string{}}}}
 	stack.Build.Add(middleware.BuildMiddlewareFunc("testInit", func(ctx context.Context, input middleware.BuildInput, handler middleware.BuildHandler) (o middleware.BuildOutput, m middleware.Metadata, err error) {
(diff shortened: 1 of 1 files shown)


skmcgrail:
Current test failures are unrelated:
```
./deserializers.go:1800:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:1843:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:1886:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:1929:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:1945:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:1961:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHTTPDateTimestampHeaderListValues
./deserializers.go:3292:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
./deserializers.go:3413:24: undefined: "github.com/awslabs/smithy-go/transport/http".SplitHeaderListValues
```

aws/middleware/user_agent.go
```diff
@@ -0,0 +1,116 @@
...
+//
+// Default Example:
+//   aws-sdk-go/2.3.4 GOOS/linux GOARCH/amd64 GO/go1.14
+func NewRequestUserAgent() *RequestUserAgent {
+	uab := httpbinding.NewUserAgentBuilder()
+	uab.AddKeyValue(aws.SDKName, aws.SDKVersion)
+	uab.AddKeyValue("GOOS", runtime.GOOS)
+	uab.AddKeyValue("GOARCH", runtime.GOARCH)
+	uab.AddKeyValue("GO", runtime.Version())
+	return &RequestUserAgent{uab: uab}
```
skmcgrail, line 35: Add exec-env support here

aws/middleware/user_agent.go
```diff
@@ -0,0 +1,116 @@
...
+	AddKey(string)
+}) error) func(*middleware.Stack) error {
+	return func(stack *middleware.Stack) error {
+		requestUserAgent, err := getOrAddRequestUserAgent(stack)
+		if err != nil {
+			return err
+		}
+		return callback(requestUserAgent)
+	}
+}
```
jasdel, line 49: I think the callback pattern is equivalent to invoking `requestUserAgent.AddKey(v)` if the `AddUserAgentKey` function was updated to take `v string`.

but is potentially a little bit similiar because I wouldn't need to define another function closure. (This also changes `AddKey` to return an error.

```go
func AddUserAgentKey(key string) func(*middleware.Stack) error {
     return func(stack *middleware.Stack) error {
          requestUserAgent, err := getOrAddRequestUserAgent(stack)
          if err != nil { return err }

          return requestUserAgent.AddKey(key)
}
```

with the usage of the two patterns being:

```go
client.DoThing(ctx, params, middleware.AddUserAgentKey(func(ua interface{AddKey(string)}) error {
       ua.AddKey("foo abc")
       return nil
}))
```

Vs without the callback

```go
client.DoThing(ctx, params, middleware.AddUserAgentKey("foo abc"))
```

aws/middleware/user_agent.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return err
+		}
+		requestUserAgent.AddKeyValue(key, value)
+		return nil
+	}
+}
+
+func getOrAddRequestUserAgent(stack *middleware.Stack) (*RequestUserAgent, error) {
+	id := (&RequestUserAgent{}).ID()
```
jasdel, line 70: If you want you could change this to work with a nil `RequestUserAgent` value since `ID` method doesn't use receiver value.  Or alternatively, hoist the ID value to a `const` that is referenced.  Could even keep the ID and `RequestUserAgent` middleware  all as unexported to prevent direct usage outside of the exposed helpers.

```go
id := (*RequestUserAgent)(nil).ID()
```

aws/middleware/user_agent.go
```diff
@@ -0,0 +1,121 @@
...
+		return out, metadata, fmt.Errorf("unknown transport type %T", in)
+	}
+
+	current := req.Header.Get("User-Agent")
+	if v := u.uab.Build(); len(current) > 0 {
+		current += " " + v
+	} else {
+		current = v
+	}
+	req.Header.Set("User-Agent", current)
```
jasdel, line 118: Do we want to SDK's user-agent components to always be first, or suffixed?  I was thinking we wanted it first for consistency across SDKs, but i don't know how strict that is.

aws/middleware/user_agent_test.go
```diff
@@ -0,0 +1,186 @@
...
+					"User-Agent": {expectedAgent},
+				}},
+			}},
+			Next: func(t *testing.T, expect middleware.BuildInput) middleware.BuildHandler {
+				return mockBuilderHandler(func(ctx context.Context, input middleware.BuildInput) (o middleware.BuildOutput, m middleware.Metadata, err error) {
+					if diff := cmp.Diff(input, expect, cmpopts.IgnoreUnexported(http.Request{}, smithyhttp.Request{})); len(diff) > 0 {
+						t.Error(diff)
+					}
+					return o, m, err
+				})
```
jasdel, line 49: All of the `Next` member values look to be very similar if not the same. Would it make sense to move these all into a single implementation that can be referred for the tests cases? I think the `error for unknown type` case could still use the same handler.

aws/middleware/user_agent.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+const execEnvVar = `AWS_EXECUTION_ENV`
+const execEnvUAKey = `exec-env`
+
+// UserAgent is an interface for modifying the User-Agent of a request.
+type UserAgent interface {
+	AddKey(name string)
+	AddKeyValue(name, version string)
+}
```
jasdel, line 22: I don't think this is used any more.

PR #680, opened 2020-08-19
Add CloseResponseMiddleware to the SDK's HTTP protocols

jasdel:
Adds the CloseResponseMiddleware to all of the SDK's HTTP protocols to ensure that the HTTP client's response body is closed after the request has been deserialized.

Requires: https://github.com/awslabs/smithy-go/pull/139


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddProtocols.java
```diff
@@ -18,6 +20,19 @@ public byte getOrder() {
         return -10;
     }
 
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        List<RuntimeClientPlugin> plugins = ListUtils.copyOf(
```
JordonPhillips, line 25: I don't think you need to do this copy

PR #692, opened 2020-08-24
adds deserializer impl for Query and Ec2Query protocol

skotambkar:
Adds logic to deserialize Query and Ec2Query protocol based services. 

Remaining Work:

- [ ] Handle deserialization of members corresponding to self closing tags in xml response body. 

Pairs with https://github.com/awslabs/smithy-go/pull/148


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/Ec2QueryShapeDeserVisitor.java
```diff
@@ -0,0 +1,22 @@
+package software.amazon.smithy.aws.go.codegen;
+
+import java.util.Optional;
+import java.util.function.Predicate;
+import software.amazon.smithy.aws.traits.protocols.Ec2QueryNameTrait;
+import software.amazon.smithy.go.codegen.integration.ProtocolGenerator.GenerationContext;
+import software.amazon.smithy.model.shapes.MemberShape;
+
+final class Ec2QueryShapeDeserVisitor extends XmlShapeDeserVisitor {
```
JordonPhillips, line 9: You don't seem to need this class

aws/protocol/ec2query/error_utils_test.go
```diff
@@ -0,0 +1,41 @@
...
+	cases := map[string]struct {
+		errorResponse          io.Reader
+		expectedErrorCode      string
+	}{
+		"Invalid Greeting": {
+			errorResponse: bytes.NewReader([]byte(`<Response>
+			    <Errors>
+			        <Error>
+			            <Code>InvalidGreeting</Code>
+			            <Message>Hi</Message>
```
JordonPhillips, line 20: Any particular reason to not try to extract the message?

skotambkar: Here the intention was to only extract the error code to be able to direct the error response to correct error handling function specific to that error code.

But yeah I see the use-case where if the message is set and the error code is unknown ie. an error is unmodeled, we should be retrieving this message.

Is it safe to assume this Message is the errorMessage that we want to send for Unmodeled Errors?

skotambkar: I think this might need change in XML too.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/Ec2QueryShapeDeserVisitor.java
skotambkar: yeah it seems so. Will remove.

aws/protocol/ec2query/error_utils_test.go
JordonPhillips: Yeah, Message for sure. If you can make it case insensitive that's even better

aws/protocol/ec2query/error_utils.go
```diff
@@ -0,0 +1,28 @@
...
+
+// GetEc2QueryResponseErrorCode returns the error code, error message from an ec2query error response body
+func GetResponseErrorCode(r io.Reader) (code string, message string, err error) {
+	rb, err := ioutil.ReadAll(r)
+	if err != nil {
+		return "", "", err
+	}
+
+	var er errorResponse
+	if err := xml.Unmarshal(rb, &er); err != nil {
```
jasdel, line 24: What do you think about using [xml.NewDecoder](https://pkg.go.dev/encoding/xml?tab=doc#NewDecoder) instead of Unmarshal?  This will prevent allocating an additional buffer for the error message.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -327,6 +326,8 @@ protected void deserializeStructure(GenerationContext context, StructureShape sh
             writer.write("t, done, err := decoder.Token()");
             writer.write("if err != nil { return err }");
             writer.write("if done { break }");
+            // TODO: fix self closing tags deser
+            writer.write("_ = t");
```
jasdel, line 330: Is this still outstanding?

skotambkar: yeah for now we don't know the behavior in case of self closing tags for numbers, boolean, time etc. 

That will be a separate follow up PR that touches XML as well and shouldn't block this PR.

aws/protocol/ec2query/error_utils.go
skotambkar: I was hoping to make a decision on this when we do benchmarking. How would using xml.NewDecoder save an allocation here? I chose Unmarshal as we have a defined structure for modeled errors.

jasdel: The allocation will happen when creating the `rb` buffer via the `ioutil.ReadAll` call. This allocation will be in addition to any other allocations created by the xml Unmarshaler's internal behavior. `xml.Decoder` will stream the bytes in without creating the additional byte slice buffer for the overall contents.

PR #701, opened 2020-09-02
fix for deserializing xml empty or self closed tags

skotambkar:
Fixes bug to correctly deserialize empty or self closed xml tags in ec2query, query, and rest-xml protocol services.

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
@@ -112,14 +112,20 @@ public class XmlMemberDeserVisitor implements ShapeVisitor<Void> {
         });
     }
 
-    // handles zero value assignment in case a empty xml tag is sent for the response member
+    // handles zero value assignment in case a empty or self closed xml tag is sent for the response member
     private void handleDone(Shape shape) {
         GoWriter writer = context.getWriter();
 
-        // Only string shapes can have an empty xml tag in response. If an empty xml tag is
-        // found for a string shape, we deserialize and the member value to a empty string.
+        // Member shapes string, blob, collection and map can have an empty xml or self closed tag in response.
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -106,9 +106,22 @@ private void consumeToken() {
...
 
-        // Since we got the matching end element we skip setting the value
-        writer.write("if done { break }");
+    // handles zero value assignment in case a empty xml tag is sent for the response member
+    private void handleDone(Shape shape) {
+        GoWriter writer = context.getWriter();
+
+        // Only string shapes can have an empty xml tag in response. If an empty xml tag is
+        // found for a string shape, we deserialize and the member value to a empty string.
+        if (shape.isStringShape() && !shape.hasTrait(EnumTrait.ID)) {
```
JordonPhillips, line 121: Blob shapes can also be empty. Are collection shapes transparently handled? I'd imagine so.

JordonPhillips (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
skotambkar: Blobs will be handled correctly by this change. Collection shapes are handled already. 

I have tested them with protocol tests added in https://github.com/awslabs/smithy/pull/547

JordonPhillips: The first line of your comment is still technically wrong. It might be good to call out that that those other cases are handled and mention where.

JordonPhillips:
Looks good. Lets get the updated protocol tests generated before we merge this in though

jasdel (approved):
Looks good, agree would be good to have protocol tests before merging this in.

PR #702, opened 2020-09-03
Smithy Performance Fixes

skmcgrail:
This PR captures several performance fixes for the smithy feature branch.

* The SigV4 signer has been refactored to remove a number of string concatenations that were utilizing a non-trivial amount of CPU and memory.
* The SigV4 signer will now cache derived credentials used for signing, this reduces the need to recompute the derived signing key when the credentials have been cached. Subsequent calls reduce the HMAC-SHA256 computation to one call, which saves on CPU and memory allocations.
* A number of `fmt.Sprintf` cases have been removed in favor of string concatenation where appropriate.
* Removes the SDKs internal UUID implementation in favor of the smithy-go implementaiton.


jasdel (approved):
Looks good, just need to rebase

PR #708, opened 2020-09-03
Fix API client generation issues discovered in new model builds

jasdel:
Fixes issues with the API client codegen that were discovered in the latest set of models added to the SDK.

* Fixes `bytes` package not being imported for document payload serializers.
* Fixup IdempotencyToken resolver added in correct scope. If IdempotencyToken was used as a member, not on the top level shape it would cause the resolver to be generated, but the client option was never set.

Regenerates API models for operation input/output (de)serialization helper function names.


skotambkar (approved):
For sts client, we see a missing dependency of strconv. 

Can you add that in this PR, or can be separate; doesn't have to block this pr. 

# github.com/aws/aws-sdk-go-v2/service/sts [github.com/aws/aws-sdk-go-v2/service/sts.test]
./deserializers.go:1598:17: undefined: strconv

skotambkar (approved):

jasdel:
Looks like also missed transcribe streaming, 

```
Testing github.com/aws/aws-sdk-go-v2/service/transcribestreaming
# github.com/aws/aws-sdk-go-v2/service/transcribestreaming [github.com/aws/aws-sdk-go-v2/service/transcribestreaming.test]
./serializers.go:60:38: cannot use payload (type types.AudioStream) as type "io".Reader in argument to request.SetStream:
	types.AudioStream does not implement "io".Reader (missing Read method)
./serializers.go:121:74: undefined: smithyjson
./serializers.go:133:75: undefined: smithyjson
FAIL	github.com/aws/aws-sdk-go-v2/service/transcribestreaming [build failed]
```

PR #709, opened 2020-09-04
Adds s3 client, model

skotambkar:
Uses smithy model to add s3 client. 

Note: This model does not have `SelectObjectContent` operation, and is removed till we support eventstream.

Needs rebase after https://github.com/aws/aws-sdk-go-v2/pull/718 is merged

### DIFF
--- a/internal/protocoltest/awsrestjson/go.mod
+++ b/internal/protocoltest/awsrestjson/go.mod
@@ -4,6 +4,6 @@ go 1.15
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200828232643-9cee6f194dae
-	github.com/awslabs/smithy-go v0.0.0-20200831191241-f0896471bee5
+	github.com/awslabs/smithy-go v0.0.0-20200908004534-2685ba1f252f
 	github.com/google/go-cmp v0.4.1
 )
--- a/internal/protocoltest/ec2query/deserializers.go
+++ b/internal/protocoltest/ec2query/deserializers.go
@@ -1738,7 +1738,7 @@ func awsEc2query_deserializeDocumentRenamedListMembersUnwrapped(v *[]*string, de
 	}
 
 	switch {
-	case true:
+	default:
 		var mv *string
 		t := decoder.StartEl
 		_ = t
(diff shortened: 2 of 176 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -190,7 +190,7 @@ public void generateFlattenedCollectionDeserializer(GenerationContext context, C
                 getUnwrappedMapDelegateFunctionName(context, shape), symbol, () -> {
                     // initialize the output member variable
                     generatesIntializerForOutputVariable(context, shape);
-                    writer.openBlock("for {", "}", () -> {
+                    writer.openBlock(" switch { case true: ", "}", () -> {
```
skmcgrail, line 193: What's the reason for this being a `switch` rather then a `for`. Why have this in a block if its always going to be executed?

skotambkar: We don't need a loop here, so replaced `for` with `switch`. 

The reason I opted for switch was to reuse the deser visitor methods. For simple shapes, we generate a break statement and break expects to be scoped within `for`, `switch` or `select` block.  

We can refactor this, if this affects performance (don't know how much it will affect, did this as a quick fix to unblock on customizations).

skmcgrail: Makes sense, I would personally think that generating it as:
```
switch {
default:
    // generated code
}
```
Would be clearer since there is value being compared against. So it looks less like an accidental generation error. At least in my opinion. Otherwise changes look good.

skotambkar: Yeah I can change to using default. Looks much cleaner.

skotambkar: updated

skmcgrail (approved):

PR #720, opened 2020-09-14
S3/customizations pathstyle

skotambkar:
Adds change for s3 customization related to Virtual host style. 

This PR doesn't handle any other S3 customizations

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
@@ -18,11 +18,9 @@
 package software.amazon.smithy.aws.go.codegen.customization;
 
 import java.util.List;
-import java.util.Optional;
-import software.amazon.smithy.aws.go.codegen.AwsGoDependency;
-import software.amazon.smithy.aws.go.codegen.AwsSignatureVersion4;
+import java.util.stream.Collectors;
 import software.amazon.smithy.aws.traits.ServiceTrait;
-import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.GoDelegator;
 import software.amazon.smithy.go.codegen.GoSettings;
@@ -36,10 +34,13 @@ import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
-import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.utils.ListUtils;
 
+/**
+ * S3UpdateEndpoint integration serves to apply customizations for S3 service,
+ * and modifies the resolved endpoint based on S3 client config or input shape values.
+ */
 public class S3UpdateEndpoint implements GoIntegration {
     private static final String USE_PATH_STYLE_OPTION = "UsePathStyle";
     private static final String UPDATE_ENDPOINT_ADDER = "addUpdateEndpointMiddleware";
@@ -85,12 +86,18 @@ public class S3UpdateEndpoint implements GoIntegration {
                     OperationShape operation = model.expectShape(operationId, OperationShape.class);
                     StructureShape input = model.expectShape(operation.getInput().get(), StructureShape.class);
 
-                    // get the member that targets "com.amazonaws.s3#BucketName" shape
-                    input.getAllMembers().values().stream().forEach((shape)->{
-                        if (shape.getTarget().getName().equals("BucketName")) {
-                            writer.write("case $P: return i.$L, true", symbolProvider.toSymbol(input), shape.getMemberName());
-                        }
-                    });
+                    List<MemberShape> targetBucketShape = input.getAllMembers().values().stream()
+                            .filter(m -> m.getTarget().getName().equals("BucketName"))
+                            .collect(Collectors.toList());
+                    // if model has multiple top level shapes targeting `BucketName`, we throw a codegen exception
+                    if (targetBucketShape.size()>1) {
+                        throw new CodegenException("BucketName shape should be targeted by only one input member, found " +
+                                targetBucketShape.size() +" for Input shape: "+ input.getId());
+                    }
+
+                    if (!targetBucketShape.isEmpty()) {
+                        writer.write("case $P: return i.$L, true", symbolProvider.toSymbol(input), targetBucketShape.get(0).getMemberName());
+                    }
                 });
                 writer.write("default: return nil, false");
             });


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -0,0 +1,154 @@
...
+            // TODO: Look for alternatives so that these Getters are NOT exported?
+            Symbol inputSymbol = symbolProvider.toSymbol(input);
+            writer.writeDocs("GetBucket retrieves the Bucket member value if provided");
+            writer.openBlock("func (s $P) GetBucket() (v string) {", "}", inputSymbol, () -> {
+                writer.write("if s.Bucket == nil { return v }");
+                writer.write("return *s.Bucket");
+            });
+            writer.insertTrailingNewline();
+        }
+    }
```
jasdel, line 94: The v1 SDK used the getter pattern since the types, client, and customizations were all in the same package so they could be private. V2 Doesn't have this benefit, and would need to export any helper methods.

Because of this, i think we want to look at a different design strategy for the SDK's middleware customizations to get access to the API parameter's members. This strategy would be similar to how operation middleware is generated per operation. Similar idea could be applied here. 

The generator could generate a helper function to extract the `BucketName` member from a relevant input type.  This helper could either be generated per operation, or shared across multiple operations using a type switch to selectively grab the inputs.  Generating the member helpers per operation probably is easier from a type point of view, but probably more complicated when looking at this from the middleware point of view.

From the middleware point of view it would make sense to have a "generic" generated accessor function for multiple input types that knows the member of the specific input type to extract the value from.

With this design change, no additional methods need to be generated on the types, and all accessor helpers stay private and only used internally by the middleware helper that is generated.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -0,0 +1,154 @@
...
+                                        .documentation("Set this to `true` to enable S3 Accelerate feature. For all operations "
+                                                + "compatible with S3 Accelerate will use the accelerate endpoint for "
+                                                + "requests. Requests not compatible will fall back to normal S3 requests. "
+                                                + ""
+                                                + "The bucket must be enable for accelerate to be used with S3 client with "
+                                                + "accelerate enabled. If the bucket is not enabled for accelerate an error "
+                                                + "will be returned. The bucket name must be DNS compatible to also work "
+                                                + "with accelerate."
+                                        )
+                                        .build()
```
jasdel, line 141: Probably want to leave accelerate out of this for now since it adds complexity. Will help focus on just virtual host addressing.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -0,0 +1,154 @@
...
+                        .configFields(ListUtils.of(
+                                ConfigField.builder()
+                                        .name(USE_PATH_STYLE_OPTION)
+                                        .type(SymbolUtils.createValueSymbolBuilder("bool")
+                                                .putProperty(SymbolUtils.GO_UNIVERSE_TYPE, true)
+                                                .build())
+                                        .documentation("Allows you to enable the client to use path-style addressing, "
+                                                + "i.e., `http://s3.amazonaws.com/BUCKET/KEY`. By default, the S3 client"
+                                                + "will use virtual hosted bucket addressing when possible"
+                                                + "(`http://BUCKET.s3.amazonaws.com/KEY`).")
```
jasdel, line 125: nit `https` in the example.

service/s3/api_client.go
```diff
@@ -151,3 +173,7 @@ func addHTTPSignerV4Middleware(stack *middleware.Stack, o Options) {
 	signer := v4.Signer{}
 	stack.Finalize.Add(v4.NewSignHTTPRequestMiddleware(o.Credentials, signer), middleware.After)
 }
+
+func addUpdateEndpointMiddleware(stack *middleware.Stack, options Options) {
+	s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{UsePathStyle: options.UsePathStyle, UseAccelerate: options.UseAccelerate, Region: options.Region})
+}
```
jasdel, line 179: Region could be changed by some other middleware, probably want to not put region here and let the middleware access it from the context.

service/s3/internal/customizations/updateEndpoint.go
```diff
@@ -0,0 +1,156 @@
+package customizations
```
jasdel, line 1: nit file naming pattern between implementation and test is inconsistent. would be good if they followed the same pattern, e.g. `update_endpoint.go`, `update_endpoint_test.go`

service/s3/internal/customizations/updateEndpoint.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/awslabs/smithy-go/middleware"
+	"github.com/awslabs/smithy-go/transport/http"
+)
+
+// UpdateEndpointOptions provides the options for the UpdateEndpoint middleware setup.
+type UpdateEndpointOptions struct {
+	UsePathStyle  bool
+	UseAccelerate bool
+
+	Region string
```
jasdel, line 20: Probably want to get `region` from the Context since this could change by some other middleware in the stack.

service/s3/internal/customizations/updateEndpoint.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// UpdateEndpoint adds the middleware to the middleware stack based on the UpdateEndpointOptions.
+func UpdateEndpoint(stack *middleware.Stack, options UpdateEndpointOptions) {
```
jasdel, line 24: Instead of this middleware filing the endpoint directly, maybe this middleware should be generated as the  endpoint resolver, instead of the default one for other API clients.

service/s3/internal/customizations/updateEndpoint.go
```diff
@@ -0,0 +1,156 @@
...
+var reDomain = regexp.MustCompile(`^[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]$`)
+var reIPAddress = regexp.MustCompile(`^(\d+\.){3}\d+$`)
+
+// dnsCompatibleBucketName returns true if the bucket name is DNS compatible.
+// Buckets created outside of the classic region MUST be DNS compatible.
+func dnsCompatibleBucketName(bucket string) bool {
+	return reDomain.MatchString(bucket) &&
+		!reIPAddress.MatchString(bucket) &&
+		!strings.Contains(bucket, "..")
+}
```
jasdel, line 156: regex is really slow in Go, this is an area we could consider replacing with a for loop.

jasdel (changes requested):
Good start to getting this customization added. Its a fairly complicated one.

Two primary feedback:
* instead of generating exported helper methods on input types, It would be better to generate private helper functions that know how to access the specific members from the input types. These could be passed as function pointer to the middleware, and return the value to be extracted. This allows the customizations to remain model driven, while still ensuring other middleware that might mutate the input member value elsewhere will still work.
* We should consider if this middleware should replace the generic API client endpoint resolver. There is a lot going on here that could easily collide with the endpoint resolver. When we start looking at this in the context of `access points` i think that will be even more important.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
skmcgrail, line 94: So a couple of thoughts here:
* Customizations that require manipulation of the request based on the value of bucket specified in the top level of the shape should be targeted based on members that are of the `com.amazonaws.s3#BucketName` shape type. So we shouldn't be assuming the bucket is always named `Bucket`.
* Instead of generating a `Getter` on the type shape, we instead should generate an exported method in the client package called `getBucketName`. This method definition would be code generated based on the model to correctly retrieve the top-level bucket parameter from the given operation input shape. As an example
```go

// getBucketNameFromInput retrieves the bucket input parameter from the shape. Returns false if the provided input does not contain a bucket parameter.
func getBucketNameFromInput(input interface{}) (*string, bool) {
	switch i := input.(type) {
	case *GetObject:
		return i.Bucket, true
	case *PutObject
		return i.Bucket, true
	default:
		return nil, false
	}
}
```
Then your customization middleware would continue to remain generic, and would use the helper to retrieve the bucketName shape from the input appropriately. This can then be further used for accesspoint ARNs.  Additionally this  same pattern would be used for s3-control to retrieve fields marked as the accountId.

service/s3/internal/customizations/updateEndpoint.go
```diff
@@ -0,0 +1,156 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/awslabs/smithy-go"
+	"github.com/awslabs/smithy-go/middleware"
+	"github.com/awslabs/smithy-go/transport/http"
+)
+
+// UpdateEndpointOptions provides the options for the UpdateEndpoint middleware setup.
+type UpdateEndpointOptions struct {
```
skmcgrail, line 16: Do we need this type? Can we just pass the two boolean value directly to the constructor function?

skotambkar: We want to keep the struct here. The struct helps with adding function pointers and possible options without having to change the constructor functions. Also we want to stay consistent with other customizations. This is all internal, so can be modified later if needed

service/s3/internal/customizations/updateEndpoint.go
skotambkar: region is not needed for now, so removed

service/s3/internal/customizations/updateEndpoint.go
skotambkar: We may switch to this when we do other customization, but for now this stays a little ambiguous on how we would modify the endpoint with the multiple customization's in place.

service/s3/internal/customizations/update_endpoint.go
```diff
@@ -7,33 +7,33 @@ import (
...
 // UpdateEndpoint adds the middleware to the middleware stack based on the UpdateEndpointOptions.
 func UpdateEndpoint(stack *middleware.Stack, options UpdateEndpointOptions) {
 	stack.Serialize.Insert(&updateEndpointMiddleware{
-		usePathStyle:  options.UsePathStyle,
-		useAccelerate: options.UseAccelerate,
-		region:        options.Region,
+		getBucketFromInput: options.GetBucketFromInput,
+		usePathStyle:       options.UsePathStyle,
 	}, "OperationSerializer", middleware.After)
 }
```
jasdel: Should this be inserted after the endpoint resolver?  or does endpoint resolving happen before operation serializing?

jasdel (changes requested):
Looks good change, only feedback is to use table testing to collapse the`TestPathStyleBucketBuild` and `TestHostStyleBucketBuild`.  I think this will be helpful when additional parameters are added like accesspoint, accelerate, etc.  

You could use an expect endpoint URL as the expect condition for these test cases.

service/s3/internal/customizations/update_endpoint.go
skotambkar: Endpoint resolving happens before operation serializing

jasdel (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -0,0 +1,143 @@
...
+                service.getAllOperations().forEach((operationId)-> {
+                    OperationShape operation = model.expectShape(operationId, OperationShape.class);
+                    StructureShape input = model.expectShape(operation.getInput().get(), StructureShape.class);
+
+                    // get the member that targets "com.amazonaws.s3#BucketName" shape
+                    input.getAllMembers().values().stream().forEach((shape)->{
+                        if (shape.getTarget().getName().equals("BucketName")) {
+                            writer.write("case $P: return i.$L, true", symbolProvider.toSymbol(input), shape.getMemberName());
+                        }
+                    });
```
skmcgrail, line 93: We should probably ensure that only one member in the top level of the shape is of type BucketName and throw a codgen exception if there is two or more.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -0,0 +1,143 @@
...
+import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
+import software.amazon.smithy.utils.ListUtils;
+
+public class S3UpdateEndpoint implements GoIntegration {
```
skmcgrail, line 43: Should have a javadoc string explaining the intent of this integration.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
skotambkar: I thought of this, but was not inclined to assert service models in the SDK gen. This should be handled by Smithy IMO. 

I will add the assertion here.

skmcgrail (approved):

PR #721, opened 2020-09-14
JSON Deserialization Improvements

skmcgrail:
Improves the Smithy deserialization code by removing the need to tokenize the JSON document being parsed.

### With Tokenization + Other Performance Improvements
BenchmarkGetSession/old-36         	  150000	     30939 ns/op	   34336 B/op	     671 allocs/op
BenchmarkGetSession/smithy-36      	  150000	     42823 ns/op	   41883 B/op	    1409 allocs/op

### Without Tokenization + Other Performance Improvements
BenchmarkGetSession/old-16         	  150000	     37711 ns/op	   34336 B/op	     671 allocs/op
BenchmarkGetSession/smithy-16      	  150000	     35620 ns/op	   35912 B/op	     562 allocs/op


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -215,6 +215,12 @@ protected void writeMiddlewareDocumentDeserializerDelegator(
         writer.addUseImports(SmithyGoDependency.JSON);
         writer.write("decoder := json.NewDecoder(body)");
         writer.write("decoder.UseNumber()");
+        writer.write("var shape interface{}");
+        writer.addUseImports(SmithyGoDependency.IO);
+        writer.openBlock("if err := decoder.Decode(&shape); err != nil && err != io.EOF {", "}", () -> {
+            writer.addUseImports(SmithyGoDependency.SMITHY);
+            writer.write("return out, metadata, &smithy.DeserializationError{Err: err}");
+        });
```
jasdel, line 223: not significant, but can that be shared across the implementations, it looks like this duplicated several times.

Sometimes it's deserialization error has a fmt.Errorf and sometimes it doesn't, probably want consistent there where it makes sense.

jasdel (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
skmcgrail: There are some subtitles here due to how de-serialization expectations are different about returns for errors versus returns from the document body path. I will refactor where possible.

jasdel (approved):

PR #725, opened 2020-09-15
Refactor credentials and external config to be own modules

jasdel:
###  Move credential providers to their own module

Moves the credential providers to their own module, `credentials`. This
sets the ground work for the `aws/external` package to be broken out
into its own module, `config` at the root of the repository too.
Breaking out the `credentials` module, allows credential providers to
depend on API clients.

Renamed the `SafeCredentialProvider` to `CredentialsCache`. Also the
credential providers were updated to not embed the credential cache
directly in their implementations. This change allows other credential
caching schemes to be used in the future, such as disk cache, or other
caching strategies. The SDK's external config loading is responsible for
ensuring that the credential providers it resolves are always wrapped
with the credential cache.

Users using the credential provider directly will need to wrap it with
the cache for the credentials to be cached.

###  Move aws/external to config as its own Go submodule

Moves the aws/external package out of the aws core module and into the
config submodule at the root of the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,8 @@ import (
 	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
+	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 )
 
@@ -70,7 +70,7 @@ func main() {
 	// Using the SDK's default configuration, loading additional config
 	// and credentials values from the environment variables, shared
 	// credentials, and shared configuration files
-	cfg, err := external.LoadDefaultConfig()
+	cfg, err := config.LoadDefaultConfig()
 	if err != nil {
 		panic("unable to load SDK config, " + err.Error())
 	}


aws/credentials.go
```diff
@@ -2,38 +2,25 @@ package aws
...
+//// when making service API calls. For example, when accessing public
+//// s3 buckets.
+////
+////     s3Cfg := cfg.Copy()
+////     s3cfg.Credentials = AnonymousCredentials
+////
+////     svc := s3.New(s3Cfg)
+//var AnonymousCredentials = StaticCredentialsProvider{
+//	Value: Credentials{Source: "AnonymousCredentials"},
+//}
```
skmcgrail, line 23: I know we've discussed this briefly, do we want to continue to have an Anonymous  sentinel value, or do we want make the statement that nil credentials == anonymous credentials?

service/s3/internal/configtesting/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/aws/aws-sdk-go-v2/service/s3/internal/configtesting
```
skmcgrail, line 1: What was your reasoning for wanting to keep this separate?

jasdel: This depends on the `config` module to assert that the s3 external config resolvers are valid interface.  If It was not separate then the S3 package would depend on the `config` module, which we probably don't want. This is similar reason `integration` module under `service/dynamodb` module are a broken out as well.

aws/credentials.go
jasdel: Thanks for pointing that out. I think API client's signer should treat nil credentials as anonymous.

This is slightly more complicated of a story for the `config.LoadDefaultAWSConfig` behavior, since `nil` `aws.Config.Credentials` is treated as unset, and can be populated.  This usability story probably shouldn't drive the behavior downstream at the API client level though, and we can solve separately.

jasdel: Discussed this further, and decided to keep the anonymous credentials. the signer middleware should not perform signing if credentials are `nil` or `Anonymous`. Having the `Anonymous` sentinel, is very helpful for configuring the SDK's external config loading.

config/resolve_credentials.go
```diff
@@ -74,13 +83,18 @@ func ResolveCredentialChain(cfg *aws.Config, configs Configs) (err error) {
...
 		err = assumeWebIdentity(cfg, envConfig.WebIdentityTokenFilePath, envConfig.RoleARN, envConfig.RoleSessionName, configs)
 	default:
 		err = resolveCredsFromProfile(cfg, envConfig, sharedConfig, other)
 	}
 
+	// Wrap the resolved provider in a cache so the SDK will cache credentials.
+	if err != nil {
+		cfg.Credentials = &aws.CredentialsCache{Provider: cfg.Credentials}
+	}
+
```
skmcgrail, line 97: Should this be if `err == nil`?

jasdel: Thanks! yes it should of been.

skmcgrail (approved):

config/example_test.go
```diff
@@ -34,10 +35,10 @@ func ExampleWithSharedConfigProfile() {
 }
 
 func ExampleWithCredentialsProvider() {
-	cfg, err := external.LoadDefaultAWSConfig(
+	cfg, err := config.LoadDefaultAWSConfig(
```
skotambkar, line 38: Nit: Does LoadDefaultAWSConfig need to have AWS in it? What other type of Config can it load?

skotambkar (approved):
Looks good. Few questions/nit picks

credentials/processcreds/provider_test.go
```diff
@@ -10,17 +10,18 @@ import (
...
 	"testing"
 	"time"
-
-	"github.com/aws/aws-sdk-go-v2/aws/external"
-	"github.com/aws/aws-sdk-go-v2/aws/processcreds"
-	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
 )
 
-func TestProcessProviderFromSessionCfg(t *testing.T) {
+// TODO these tests belong in the external configuration loading module, not
```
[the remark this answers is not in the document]
jasdel: yeah this is still outstanding.

credentials/processcreds/provider_test.go
```diff
@@ -540,16 +505,12 @@ func TestProcessProviderAltConstruct(t *testing.T) {
...
-	provider := processcreds.NewProvider(
+	provider := NewProvider(
 		fmt.Sprintf(
 			"%s %s",
 			getOSCat(),
-			strings.Join(
-				[]string{"testdata", "static.json"},
-				string(os.PathSeparator))))
+			filepath.Join("testdata", "static.json"),
+		))
```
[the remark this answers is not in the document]
jasdel: In this case its OK, because it won't be included in the metrics because`b.ResetTimer` is called before the benchmark iteration starts.

aws/credentials.go
```diff
@@ -2,39 +2,72 @@ package aws
...
+//
+// For example you can use this CredentialsProvider with an API client's
+// Options to instruct the client not to sign a request for accessing public
+// S3 bucket objects.
+//
+// The following example demonstrates using the AnonymousCredentials to prevent
+// SDK's external config loading attempt to resolve credentials.
+//
+//     cfg, err := config.LoadDefaultAWsConfig(
+//          config.WithCredentialsProvider(aws.AnonymousCredentials{}))
```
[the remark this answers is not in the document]
jasdel: Thanks!

aws/credentials.go
```diff
@@ -2,39 +2,72 @@ package aws
...
+type AnonymousCredentials struct{}
+
+// Retrieve implements the CredentialsProvider interface, but will always
+// return error, and cannot be used to sign a request. The AnonymousCredentials
+// type is used as a sentinel type instructing the AWS request signing
+// middleware to not sign a request.
+func (AnonymousCredentials) Retrieve(context.Context) (Credentials, error) {
+	return Credentials{Source: "AnonymousCredentials"},
+		fmt.Errorf("the AnonymousCredentials is not a valid credential provider, and cannot be used to sign AWS requests with")
 }
```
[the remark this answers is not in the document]
jasdel: AnonymousCredentials allows the config's resolver's behavior correctly with the user's intent. If the type didn't exist there would be no way for customers to use `config.LoadDefaultAWSConfig` with out resolving the default credential chain, when they really want anonymous credentials.

config/example_test.go
jasdel: Mainly had `AWS` in there to clarify the `AWSConfig` as the AWS shared config, shared credentials. Its kind of implied, but helped emphasis this.  But in theory this could support other.

credentials/processcreds/provider_test.go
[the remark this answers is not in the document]
jasdel: Moved these to the config module.

README.md
```diff
@@ -70,7 +70,7 @@ func main() {
 	// Using the SDK's default configuration, loading additional config
 	// and credentials values from the environment variables, shared
 	// credentials, and shared configuration files
-	cfg, err := external.LoadDefaultAWSConfig()
+	cfg, err := external.LoadDefaultConfig()
```
skotambkar, line 73: external here needs to be renamed to config too

skotambkar: Oh this example is stale. Needs update/removal

jasdel: thanks missed that one

skotambkar (approved):

PR #727, opened 2020-09-16
`service/apigateway`: Accept header customization

skotambkar:
* Adds customization for apigateway service to always add accept header to it's request qith value as `application/json`. 
* Adds unit test for testing the behavior

### DIFF
--- a/service/apigateway/internal/customizations/accept_header.go
+++ b/service/apigateway/internal/customizations/accept_header.go
@@ -29,7 +29,8 @@ func (m *acceptHeaderMiddleware) HandleBuild(
 	}
 
 	// all APIGateway operations must have Accept header set to application/json
-	req.Header.Add("Accept", "application/json")
+	const conHeader = "Accept"
+	req.Header[conHeader] = append(req.Header[conHeader][:0], "application/json")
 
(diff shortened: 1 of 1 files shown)


service/apigateway/internal/customizations/accept_header.go
```diff
@@ -0,0 +1,35 @@
...
+	ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (
+	out middleware.BuildOutput, metadata middleware.Metadata, err error,
+) {
+	req, ok := in.Request.(*http.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown transport type %T", in.Request)
+	}
+
+	// all APIGateway operations must have Accept header set to application/json
+	req.Header.Add("Accept", "application/json")
```
skmcgrail, line 32: nit: Manipulate the `req.Header` map directly so we don't need to perform a string allocation to canonicalize the `Accept` header.

```go
const conHeader = "Accept"
req.Header[conHeader] = append(req.Header[conHeader], "application/json")
```

jasdel (approved):

skmcgrail (approved):

PR #731, opened 2020-09-17
Draft: AWS Response Error  implementation and usage

skotambkar:
- [x] Adds aws response error type
- [x] Add aws response error wrapper middleware
- [x] Update codegen to use aws response error wrapper middleware
- [ ] Update deser codegen to add request id with response metadata when extracted from raw response
- [ ] Migrate to standard smithy version and regen protocol tests for this change

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
@@ -52,6 +52,6 @@ public class AwsGoDependency {
     }
 
     private static final class Versions {
-        private static final String AWS_SDK = "v0.0.0-20200917111458-b3170caebebe";
+        private static final String AWS_SDK = "v0.0.0-20200918012104-3f8a30e45bdb";
     }
 }
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsQuery.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsQuery.java
@@ -43,6 +43,19 @@ class AwsQuery extends HttpRpcProtocolGenerator {
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


aws/transport/http/response_error_middleware.go
```diff
@@ -0,0 +1,65 @@
...
+
+// AddResponseErrorWrapper adds response error wrapper middleware
+func AddResponseErrorWrapper(stack *middleware.Stack) {
+	// add error wrapper middleware before operation deserializers so that it can wrap the error response
+	// returned by operation deserializers
+	stack.Deserialize.Insert(&errorWrapperMiddleware{}, "OperationDeserializer", middleware.Before)
+}
+
+// RequestID is the identifier for Request Id set in response metadata
+const RequestID = "RequestID"
```
jasdel, line 19: probably want to use getter/setter helpers for the metadata that use struct keys instead of strings. This way we can ensure that there are no collisions with names.  similar to how the SDK is using context to get and store values.

aws/transport/http/response_error_middleware.go
```diff
@@ -0,0 +1,65 @@
...
+	// Wrap the returned smithy error with the request id retrieved from the metadata
+	if err != nil {
+		var respErr *smithyhttp.ResponseError
+		if errors.As(err, &respErr) {
+			err = &ResponseError{
+				ResponseError: respErr,
+				RequestID:     reqID,
+			}
+		}
+	}
```
jasdel, line 62: The SDK's version of this middleware probably can do all of the work of wrapping the underlying error without extra layer doing the `smithyhttp.ResponseError` work.  I think ideally this was the intention so that the SDK's deserializer middleware would be codegen to use the "correct" response error for the api, e.g. smithy client, vs aws client, vs s3 client.

Mainly because using the `errors.As` approach here in theory could cause loss of data if the top most error is not a `smtihyhttp.ResponseError.`  Not 100% aposed to this approach, but it does create the potential for loss of data.

jasdel (changes requested):

skotambkar:
Replaced by https://github.com/aws/aws-sdk-go-v2/pull/736

PR #734, opened 2020-09-17
Endpoint Credentials Provider and Client

skmcgrail:
Adds the endpoint credential provider to the credentials Go module. Currently contains an internal hand-written Smithy client for interacting with the configured customer endpoint.

### DIFF
--- a/credentials/endpointcreds/internal/client/client.go
+++ b/credentials/endpointcreds/internal/client/client.go
@@ -45,6 +45,11 @@ func (o Options) Copy() Options {
 	return to
 }
 
+// GetRetryer returns the configured retryer
+func (o Options) GetRetryer() retry.Retryer {
+	return o.Retryer
+}
+
 // Client is an client for retrieving AWS credentials from an endpoint
 type Client struct {
 	options Options
@@ -89,6 +94,7 @@ func (c *Client) GetCredentials(ctx context.Context, params *GetCredentialsInput
 	stack.Serialize.Add(&serializeOpGetCredential{}, smithymiddleware.After)
 	stack.Build.Add(&buildEndpoint{Endpoint: options.Endpoint}, smithymiddleware.After)
 	stack.Deserialize.Add(&deserializeOpGetCredential{}, smithymiddleware.After)
+	retry.AddRetryMiddlewares(stack, options)
 	middleware.AddUserAgentKey(ServiceID)
 	smithyhttp.AddErrorCloseResponseBodyMiddleware(stack)
 	smithyhttp.AddCloseResponseBodyMiddleware(stack)
--- a/credentials/go.mod
+++ b/credentials/go.mod
@@ -3,9 +3,9 @@ module github.com/aws/aws-sdk-go-v2/credentials
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200921195903-c1c5f2a1c25c
-	github.com/aws/aws-sdk-go-v2/service/sts v0.0.0-20200921195903-c1c5f2a1c25c
-	github.com/awslabs/smithy-go v0.0.0-20200917082847-627658712072
+	github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200921214008-bd52d2b3309e
+	github.com/aws/aws-sdk-go-v2/service/sts v0.0.0-20200921214008-bd52d2b3309e
+	github.com/awslabs/smithy-go v0.0.0-20200921163940-68b34a89572d
 )
 
 replace (
--- a/credentials/go.sum
+++ b/credentials/go.sum
@@ -2,6 +2,8 @@ github.com/awslabs/smithy-go v0.0.0-20200914213924-b41e7bef5d4f h1:+ORlmI9IRG6Z/
 github.com/awslabs/smithy-go v0.0.0-20200914213924-b41e7bef5d4f/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
 github.com/awslabs/smithy-go v0.0.0-20200917082847-627658712072 h1:5pOwJyf2x7UjzLTKkD/xnBOKRc0EWuUcHAD21yLCgrM=
 github.com/awslabs/smithy-go v0.0.0-20200917082847-627658712072/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
+github.com/awslabs/smithy-go v0.0.0-20200921163940-68b34a89572d h1:fRnPNtr70h7FNjTpxr7E9KCiWGYbtTPORb7yXgUaKhM=
+github.com/awslabs/smithy-go v0.0.0-20200921163940-68b34a89572d/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/google/go-cmp v0.4.1 h1:/exdXoGamhu5ONeUJH0deniYLWYvQwW66yvlfiiKTu0=


credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+func (o Options) Copy() Options {
+	to := o
+	to.APIOptions = append([]func(*smithymiddleware.Stack) error{}, o.APIOptions...)
+	return to
+}
+
+// Client is an client for retrieving AWS credentials from an endpoint
+type Client struct {
+	options Options
+}
```
skotambkar, line 53: The usage will be `client.Client`, seems redundant. Though this is internal so may not matter much

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+
+// ErrorMessage is the error message returned by the endpoint
+func (e *EndpointError) ErrorMessage() string {
+	return e.Message
+}
+
+// ErrorFault indicates error fault classification
+func (e *EndpointError) ErrorFault() smithy.ErrorFault {
+	return smithy.FaultUnknown
```
skotambkar, line 146: Seems odd that this will always be Unknown fault.

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+
+// GetCredentialsOutput is the response from the credential endpoint
+type GetCredentialsOutput struct {
+	Expiration      *time.Time
+	AccessKeyID     string
+	SecretAccessKey string
+	Token           string
+}
+
+// EndpointError is an error returned from the endpoint service
```
skotambkar, line 123: Question: which service is `the endpoint service`? Are you referring to the endpoints.json?

credentials/endpointcreds/internal/client/middleware.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return next.HandleBuild(ctx, in)
+}
+
+type serializeOpGetCredential struct {
+	AuthorizationToken string
+}
+
+func (s *serializeOpGetCredential) ID() string {
+	return "OperationSerializer"
```
skotambkar, line 49: Is the ID value intentional? Will this clash as the `OperationSerializer` as ID is used by Operation serializers?

credentials/endpointcreds/provider.go
```diff
@@ -0,0 +1,125 @@
...
+	creds := aws.Credentials{
+		AccessKeyID:     resp.AccessKeyID,
+		SecretAccessKey: resp.SecretAccessKey,
+		SessionToken:    resp.Token,
+		Source:          ProviderName,
+	}
+
+	if resp.Expiration != nil {
+		creds.CanExpire = true
+		creds.Expires = resp.Expiration.Add(-p.options.ExpiryWindow)
```
skotambkar, line 114: You might want to check if provider is nil. Either here or at line 100

skotambkar (commented):
Minor comments/questions

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+	stack := smithymiddleware.NewStack("GetCredentials", smithyhttp.NewStackRequest)
+	options := c.options.Copy()
+	for _, fn := range optFns {
+		fn(&options)
+	}
+
+	stack.Serialize.Add(&serializeOpGetCredential{AuthorizationToken: options.AuthorizationToken}, smithymiddleware.After)
+	stack.Build.Add(&buildEndpoint{Endpoint: options.Endpoint}, smithymiddleware.After)
+	stack.Deserialize.Add(&deserializeOpGetCredential{}, smithymiddleware.After)
+
```
jasdel, line 101: probably want to also add the SDK's user agent to the request.

credentials/endpointcreds/internal/client/client.go
jasdel: in this context its any http endpoint that can return credentials in a JSON document.  unrelated to the `endpoints.json` model.

credentials/endpointcreds/internal/client/client.go
jasdel: guessing based on HTTP status code i think is only alternative.

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,147 @@
...
+	Do(*http.Request) (*http.Response, error)
+}
+
+// Options is the endpoint client configurable options
+type Options struct {
+	// The endpoint to retrieve credentials from
+	Endpoint string
+
+	// The authorization token to pass in the Authorization header with the request
+	AuthorizationToken string
```
jasdel, line 27: What do you think about the authorization value being a parameter of the operation input parameter instead of on the client?  Mainly suggesting this to separate state from the client, which is generally stateless.   Though I guess this is similar to the sdk's credentials.

credentials/endpointcreds/internal/client/client_test.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+}
+
+func TestClient_GetCredentials_NoEndpoint(t *testing.T) {
+	client := New(Options{})
+
+	_, err := client.GetCredentials(context.Background())
+	if err == nil {
+		t.Fatalf("expect ~ got none")
+	}
```
jasdel, line 193: is the `~` here intentional?  Guessing might be hold over from template

credentials/endpointcreds/internal/client/middleware.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	if response.StatusCode < 200 || response.StatusCode >= 300 {
+		return out, metadata, deserializeError(response)
+	}
+
+	var shape *GetCredentialsOutput
+	if err = json.NewDecoder(response.Body).Decode(&shape); err != nil {
+		return out, metadata, &smithy.DeserializationError{Err: fmt.Errorf("failed to deserialize json response, %w", err)}
+	}
```
jasdel, line 97: nit need to close the response body.

credentials/endpointcreds/provider.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return creds, nil
+}
+
+func (p *Provider) getCredentials(ctx context.Context) (*client.GetCredentialsOutput, error) {
+	return p.client.GetCredentials(ctx, func(options *client.Options) {
+		options.Endpoint = p.options.Endpoint
+		options.AuthorizationToken = p.options.AuthorizationToken
+	})
+}
```
jasdel, line 125: This is probably a good case were passing the `AuthorizationToken` in as an input parameter to the operation call may make sense.

credentials/endpointcreds/provider.go
```diff
@@ -0,0 +1,125 @@
...
+	Endpoint string
+
+	// Optional authorization token value if set will be used as the value of
+	// the Authorization header of the endpoint credential request.
+	AuthorizationToken string
+}
+
+// New returns a credentials Provider for retrieving AWS credentials
+// from arbitrary endpoint.
+func New(cfg aws.Config, endpoint string, options ...func(*ProviderOptions)) *Provider {
```
jasdel, line 82: Nit `ProviderOptions` can be just `Options` since this is the only provider in this package.

credentials/endpointcreds/provider.go
```diff
@@ -0,0 +1,125 @@
...
+	// The AWS Client to make HTTP requests to the endpoint with. The endpoint
+	// the request will be made to is provided by the aws.Config's
+	// EndpointResolver.
+	client getCredentials
+
+	options ProviderOptions
+}
+
+// ProviderOptions is structure of configurable options for Provider
+type ProviderOptions struct {
```
jasdel, line 60: The SDK's `Retryer` and `HTTPClient` probably should be passed down from the Provider to the internal client

jasdel (changes requested):

credentials/endpointcreds/internal/client/client.go
skmcgrail: Yeah I was treating this similar to the authentication scheme's credentials which is usually stored on the client.

credentials/endpointcreds/internal/client/client_test.go
skmcgrail: I think this was accidental 👍

credentials/endpointcreds/provider.go
skmcgrail: I don't think that is applicable here. Retrieve is never meant to be called on a nil provider value. So a panic is actually the expected behavior.

credentials/endpointcreds/provider.go
skmcgrail: Discussed offline, these are pulled in from the provided `aws.Config`

credentials/endpointcreds/internal/client/client.go
skmcgrail: I agree it's redundant, I was rolling with it since its internal. If we were to expose this client ever we would change the package name and path appropriately.

credentials/endpointcreds/internal/client/middleware.go
skmcgrail: OperationSerializer is the standard ID we have use for serialization middleware. This client is entirely handwritten as is the registered stack, so no risk of collision.

credentials/endpointcreds/internal/client/client.go
skmcgrail: Can update it to base the fault on client / server based on HTTP Status code

credentials/endpointcreds/internal/client/middleware.go
skmcgrail: Will use the smithy http middleware to handle this.

credentials/endpointcreds/internal/client/middleware.go
```diff
@@ -86,6 +89,7 @@ func (d *deserializeOpGetCredential) HandleDeserialize(ctx context.Context, in s
 	if !ok {
 		return out, metadata, &smithy.DeserializationError{Err: fmt.Errorf("unknown transport type %T", out.RawResponse)}
 	}
+	defer response.Body.Close()
```
skotambkar, line 92: Can you verify this, probably don't need to close the body here since you have AddCloseResponseBodyMiddleware to close the response body.

skmcgrail: Ah thanks for catching this I thought I had removed these.

skotambkar (approved):

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,135 @@
...
+
+// GetCredentialsInput is the input to send with the endpoint service to receive credentials.
+type GetCredentialsInput struct {
+	AuthorizationToken string
+}
+
+// GetCredentials retrieves credentials from credential endpoint
+func (c *Client) GetCredentials(ctx context.Context, params *GetCredentialsInput, optFns ...func(*Options)) (*GetCredentialsOutput, error) {
+	stack := smithymiddleware.NewStack("GetCredentials", smithyhttp.NewStackRequest)
+	options := c.options.Copy()
```
jasdel, line 81: Should the `Client` provide an option for `APIOptions` to allow user to mutate the client operation's stack?  Might be helpful for tracing.

credentials/endpointcreds/provider.go
```diff
@@ -0,0 +1,137 @@
...
+	// 10 seconds before the credentials are actually expired.
+	//
+	// If ExpiryWindow is 0 or less it will be ignored.
+	ExpiryWindow time.Duration
+
+	// Endpoint to retrieve credentials from. Required
+	Endpoint string
+
+	// HTTPClient to handle sending HTTP requests to the target endpoint.
+	HTTPClient client.HTTPClient
```
jasdel, line 76: probably want to define this interface locally in the `endpointscreds` package, technically the reference to the `internal/client` package could be satisfied, there would be no public docs for the interface type.

credentials/endpointcreds/internal/client/client.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, fn := range optFns {
+		fn(&options)
+	}
+
+	stack.Serialize.Add(&serializeOpGetCredential{}, smithymiddleware.After)
+	stack.Build.Add(&buildEndpoint{Endpoint: options.Endpoint}, smithymiddleware.After)
+	stack.Deserialize.Add(&deserializeOpGetCredential{}, smithymiddleware.After)
+	middleware.AddUserAgentKey(ServiceID)
+	smithyhttp.AddErrorCloseResponseBodyMiddleware(stack)
+	smithyhttp.AddCloseResponseBodyMiddleware(stack)
```
jasdel, line 94: does this client need any retry?

jasdel (approved):

PR #738, opened 2020-09-18
STS Credential Providers

skmcgrail:
This PR re-introduces the STS AsumeRole and WebIdentity providers the the credentials module.

### DIFF
--- a/credentials/go.mod
+++ b/credentials/go.mod
@@ -3,9 +3,9 @@ module github.com/aws/aws-sdk-go-v2/credentials
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200918204210-51fd26a6f991
-	github.com/aws/aws-sdk-go-v2/service/sts v0.0.0-20200918170804-ed5c823eb142
-	github.com/awslabs/smithy-go v0.0.0-20200914213924-b41e7bef5d4f
+	github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200921195903-c1c5f2a1c25c
+	github.com/aws/aws-sdk-go-v2/service/sts v0.0.0-20200921195903-c1c5f2a1c25c
+	github.com/awslabs/smithy-go v0.0.0-20200917082847-627658712072
 )
 
 replace (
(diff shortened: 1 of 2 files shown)


credentials/stscreds/assume_role_provider.go
```diff
@@ -0,0 +1,314 @@
...
+		options: o,
+	}
+}
+
+// Retrieve generates a new set of temporary credentials using STS.
+func (p *AssumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
+	// Apply defaults where parameters are not set.
+	if len(p.options.RoleSessionName) == 0 {
+		// Try to work out a role name that will hopefully end up unique.
+		p.options.RoleSessionName = fmt.Sprintf("%d", time.Now().UTC().UnixNano())
```
jasdel, line 271: I think the CLI (and maybe other SDKs) prefix this unique id with a value identifying it came from the SDK. That is something we could consider for this.

credentials/stscreds/web_identity_provider.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+var invalidIdentityTokenExceptionCode = (&types.InvalidIdentityTokenException{}).ErrorCode()
+
+const (
+	// WebIdentityProviderName is the web identity provider name
+	WebIdentityProviderName = "WebIdentityCredentials"
+)
+
+type AssumeRoleWithWebIdentity interface {
```
jasdel, line 24: Should document

credentials/stscreds/assume_role_provider.go
```diff
@@ -0,0 +1,314 @@
...
+	_, err := fmt.Scanln(&v)
+
+	return v, err
+}
+
+// ProviderName provides a name of AssumeRole provider
+const ProviderName = "AssumeRoleProvider"
+
+// AssumeRole represents the minimal subset of the STS client API used by this provider.
+type AssumeRole interface {
```
jasdel, line 140: What do you think about `AssumeRoleAPIClient` for this? I think a name similar to that was proposed for higher level features like paginators would use.

credentials/stscreds/web_identity_provider.go
jasdel: similar to AssumeRole, should this be, `AssumeRoleWithWebIdentityAPIClient`?

credentials/stscreds/web_identity_provider.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to read file at %s: %v", string(j), err)
+	}
+
+	return b, nil
+}
+
+// NewWebIdentityRoleProvider will return a new WebIdentityRoleProvider with the
+// provided stsiface.ClientAPI
+func NewWebIdentityRoleProvider(options WebIdentityRoleOptions, optFns ...func(*WebIdentityRoleOptions)) *WebIdentityRoleProvider {
```
jasdel, line 96: Should required options be input parameters for the constructor?

jasdel (approved):

PR #742, opened 2020-09-21
Moves the supporting endpoint types to correct import path

skmcgrail:
Moves the endpoint supporting types from `github.com/aws/aws-sdk-go-v2/aws/endpoints/v2` to `github.com/aws/aws-sdk-go-v2/aws/endpoints`.

### DIFF
--- a/config/go.mod
+++ b/config/go.mod
@@ -2,10 +2,9 @@ module github.com/aws/aws-sdk-go-v2/config
 
 go 1.15
 
-require github.com/aws/aws-sdk-go-v2 v0.0.0-20200917190052-bb89e83d660c
-
 require (
-	github.com/aws/aws-sdk-go-v2/credentials v0.0.0-20200917190052-bb89e83d660c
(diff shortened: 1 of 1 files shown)


credentials/go.mod
```diff
@@ -2,6 +2,6 @@ module github.com/aws/aws-sdk-go-v2/credentials
 
 go 1.15
 
-require github.com/aws/aws-sdk-go-v2 v0.0.0-20200917190009-a19312abfd63
+require github.com/aws/aws-sdk-go-v2 v0.24.1-0.20200921191018-1a09374b6cac
```
jasdel, line 5: Curious why sometimes its adds the v0.24.1 prefix and sometimes doesn't

jasdel (approved):

credentials/go.mod
skmcgrail: `v0.24.1-0` is a made up version based on the last tagged replace in the repository that this commit is after. This is happening cause often I will use `go get` to update the hash versus using the `git log` command.

PR #743, opened 2020-09-21
Adds AWS specific RequestError wrapper's  

skotambkar:
This PR contains changes for 
- [x] Adds aws request error wrapper that used request id. 
- [x] Adds support for fetching request id from error response document body. 
- [x] Adds support for fetching request id from response header's

Customization:
- [x] Adds customization for S3 response error wrapper.  

Fixes:
- [x] Fix xml error response message not being deserialized. 

Depends on https://github.com/awslabs/smithy-go/pull/178

### DIFF
--- a/service/s3control/go.mod
+++ b/service/s3control/go.mod
@@ -4,5 +4,11 @@ go 1.15
 
 require (
 	github.com/aws/aws-sdk-go-v2 v0.0.0-20200924095642-be147c6e7568
+	github.com/aws/aws-sdk-go-v2/service/internal/s3shared v0.0.0-20200924095642-be147c6e7568
 	github.com/awslabs/smithy-go v0.0.0-20200924081159-7ac2e6483c86
+)
+
+replace (
+	github.com/aws/aws-sdk-go-v2/service/internal/s3shared => ./../internal/s3shared
+	github.com/aws/aws-sdk-go-v2/service/s3control => ./../s3control
 )


aws/transport/http/request_id.go
```diff
@@ -0,0 +1,40 @@
+package http
```
jasdel, line 1: Not sure if this will or won't be an HTTP specific concept, so far its been generic across AWS request.

aws/transport/http/request_id.go
```diff
@@ -0,0 +1,40 @@
...
+
+// SetRequestIDMetadataFromReader takes in dst metadata on which metadata needs to be set,
+// and a src metadata reader from which the metadata is read.
+func SetRequestIDMetadataFromReader(dst *middleware.Metadata, src middleware.MetadataReader) error {
+	id, ok := src.Get(requestID{}).(string)
+	if !ok {
+		return fmt.Errorf("expected requestID to present as string in middleware metadata")
+	}
+	SetRequestIDMetadata(dst, id)
+	return nil
```
jasdel, line 39: A reader wouldn't be needed if the deserialization function took `*middleware.Metadata` as input. This would also remove the need for the additional `middleware.Metadata` `deserMetadata` return value.

aws/transport/http/response_error_middleware.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+
+	"github.com/awslabs/smithy-go/middleware"
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+)
+
+// AddResponseErrorWrapper adds response error wrapper middleware
+func AddResponseErrorWrapper(stack *middleware.Stack) {
```
jasdel, line 11: probably better named as ` AddResponseErrorMiddleware`

aws/transport/http/response_error_middleware.go
```diff
@@ -0,0 +1,63 @@
...
+	reqID := GetRequestIDMetadata(metadata)
+	if len(reqID) == 0 {
+		// check for request id in header `X-Amzn-Requestid`
+		if v := resp.Header.Get("X-Amzn-Requestid"); len(v) != 0 {
+			reqID = v
+			// set reqID on metadata for successful responses.
+			SetRequestIDMetadata(&metadata, reqID)
+		}
+	}
+
```
jasdel, line 50: There is also the `X-Amz-RequestId` header as well i think.

Also this middleware probably shouldn't do double duty and check for the header it self. It should rely on the metadata to have a value for the RequestID. Duplicating this behavior makes it difficult to reason about where things are coming from.

aws/transport/http/response_error_middleware.go
```diff
@@ -0,0 +1,63 @@
...
+	// Wrap the returned smithy error with the request id retrieved from the metadata
+	if err != nil {
```
jasdel, line 52: This additional error check is unneeded, the original check on the `next.Handle` call above takes care if this case.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/Ec2Query.java
```diff
@@ -34,9 +34,21 @@ protected void generateDocumentBodyShapeDeserializers(GenerationContext context,
...
-        writer.write("if err != nil { return err }");
+        writer.write("errorComponents, err := ec2query.GetErrorResponseComponents(errorBody)");
+        writer.write("if err != nil { return m, err }");
+        writer.insertTrailingNewline();
+
+        writer.addUseImports(AwsGoDependency.AWS_HTTP_TRANSPORT);
+        writer.write("awshttp.SetRequestIDMetadata(&m, errorComponents.RequestID)");
+        writer.insertTrailingNewline();
+
+        // assign middleware metadata to metadata var
```
jasdel, line 47: Why duplicate the `metadata` parameter?  Won't this risk deleting any other metadata that may of been stored on the `metadata` from further down the stack?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
```diff
@@ -199,6 +200,21 @@ protected void writeErrorMessageCodeDeserializer(GenerationContext context) {
...
+    protected void processDeserializedErrorMetadata(GenerationContext context) {
+        GoWriter writer = context.getWriter();
+        writer.openBlock("if errorMetadata, ok := deserMetadata.(middleware.MetadataReader); ok {", "}", () -> {
+            writer.write("awshttp.SetRequestIDMetadataFromReader(&metadata, errorMetadata)");
+
+            // TODO: does this need to be a RuntimeClient plugin?
+            if (context.getService().getId().getName().equals("S3")) {
+                writer.addUseImports(AwsCustomGoDependency.S3_CUSTOMIZATION);
+                writer.write("s3cust.SetHostIDMetadataFromReader(&metadata, errorMetadata)");
+            }
```
jasdel, line 213: this header maybe be present for both s3 and s3control

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XMLProtocolUtils.java
```diff
@@ -267,9 +269,30 @@ public static void writeXmlErrorMessageCodeDeserializer(ProtocolGenerator.Genera
...
+            writer.write("errorComponents, err := awsxml.GetErrorResponseComponents(errorBody, $L)", isNoErrorWrapping);
+            writer.write("if err != nil { return m, err }");
+        }
+
+        writer.addUseImports(AwsGoDependency.AWS_HTTP_TRANSPORT);
+        writer.write("awshttp.SetRequestIDMetadata(&m, errorComponents.RequestID)");
+        writer.insertTrailingNewline();
+
+        // assign middleware metadata to metadata var
+        writer.write("metadata = m");
```
jasdel, line 292: Same comment as non-s3 case, why duplicate/replace metadata?

How does this work with the operation middleware for `deserMetadata` variable created by smithy-go?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ResponseErrorWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public void writeAdditionalFiles(
+            GoSettings settings,
+            Model model,
+            SymbolProvider symbolProvider,
+            GoDelegator goDelegator
+    ) {
+        ServiceShape service = settings.getService(model);
+        if (!isS3Service(model, service)) {
```
jasdel, line 45: Will want this for both s3 and s3control.

service/s3/internal/customizations/response_error_middleware.go
```diff
@@ -0,0 +1,80 @@
...
+
+	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
+	"github.com/awslabs/smithy-go/middleware"
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+
+	s3http "github.com/aws/aws-sdk-go-v2/service/s3/transport/http"
+)
+
+// AddResponseErrorWrapper adds response error wrapper middleware
+func AddResponseErrorWrapper(stack *middleware.Stack) {
```
jasdel: nit `AddResposneErrorMiddleware`

service/s3/internal/customizations/response_error_middleware.go
```diff
@@ -0,0 +1,80 @@
...
+	stack.Deserialize.Insert(&errorWrapperMiddleware{}, "OperationDeserializer", middleware.Before)
+}
+
+type errorWrapperMiddleware struct {
+}
+
+// ID returns the middleware identifier
+func (m *errorWrapperMiddleware) ID() string {
+	return "AWSResponseErrorWrapperMiddleware"
+}
```
jasdel: Should this middleware have a different ID than the generic AWSResponseErrorWrapperMiddleware?

service/s3/internal/customizations/response_error_middleware.go
```diff
@@ -0,0 +1,80 @@
...
+	// look for request id in metadata
+	reqID := awshttp.GetRequestIDMetadata(metadata)
+	if len(reqID) == 0 {
+		// check for request id in header `X-Amzn-Requestid`
+		if v := resp.Header.Get("X-Amzn-Requestid"); len(v) != 0 {
+			reqID = v
+			// set reqID on metadata for successful responses.
+			awshttp.SetRequestIDMetadata(&metadata, reqID)
+		}
+	}
```
jasdel: Same as generic response error wrapper, this shouldn't duplicate the request ID.

service/s3/internal/customizations/response_error_middleware.go
```diff
@@ -0,0 +1,80 @@
...
+		// check for host id in header `X-Amz-Id-2`
+		if v := resp.Header.Get("X-Amz-Id-2"); len(v) != 0 {
+			hostID = v
+			// set hostID on metadata for successful responses.
+			s3http.SetHostIDMetadata(&metadata, hostID)
+		}
+	}
+
+	// Wrap the returned smithy error with the request id retrieved from the metadata
+	if err != nil {
```
jasdel: error check not needed.

service/s3/transport/http/host_id.go
```diff
@@ -0,0 +1,40 @@
+package http
```
jasdel: Probably want to reconsider this package location, `transport/http` is a little awkward nested under the s3 module. Similar to RequestID, this isn't a HTTP specific metadata, but across all AWS services.

Maybe this would make sense under a `s3metadata` package?

Or alternative would be to generate it into the S3 client package

service/s3/transport/http/response_error.go
```diff
@@ -0,0 +1,33 @@
+package http
```
jasdel: Same comment as host_id.go file, not really http specific, but higher level concept. maybe better under `s3metadata`, or just `metadata` too.

jasdel (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XMLProtocolUtils.java
skotambkar: The metadata here isn't the middleware.Metadata that is being passed around in operation stack. 

The naming should be changed to prevent the confusion. The metadata here is the return argument (an empty interface) for top level error deserializer function.

skotambkar: metadata is actually the `deserMetadata`. It is used as a return argument here

aws/transport/http/response_error_middleware.go
skotambkar: Not sure why this is unneeded. This middleware should always be inserted even in case of successful responses. Or am I missing something here?

skotambkar:
For review, please refer to the top commit. https://github.com/aws/aws-sdk-go-v2/pull/743/commits/99e2bf2a3e33071ab48f781d7b3696476a8b6810

The consequent commits are hash update, minor lint fix/ comment update.

The last commit is generated code change.

aws/request_id.go
```diff
@@ -0,0 +1,28 @@
...
+type requestID struct {
+}
+
+// SetRequestIDMetadata sets the provided request id over middleware metadata
+func SetRequestIDMetadata(metadata *middleware.Metadata, id string) {
+	metadata.Set(requestID{}, id)
+}
+
+// GetRequestIDMetadata retrieves the request id from middleware metadata
+func GetRequestIDMetadata(metadata middleware.Metadata) string {
```
jasdel, line 17: Did we already discuss if these getters should return `(T, bool)` to indicated empty value vs unset?

aws/request_id_retriever.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/awslabs/smithy-go/middleware"
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+)
+
+// AddRequestIDRetrieverMiddleware adds request id retrievr middleware
+func AddRequestIDRetrieverMiddleware(stack *middleware.Stack) {
+	// add error wrapper middleware before operation deserializers so that it can wrap the error response
+	// returned by operation deserializers
+	stack.Deserialize.Insert(&requestIDRetrieverMiddleware{}, "OperationDeserializer", middleware.Before)
+}
```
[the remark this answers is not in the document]
skotambkar: Do all aws specific middleware go in aws/middleware pkg? User agent is in aws/middleware, so sure will move this too

aws/request_id_retriever.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// Different header which can map to request id
+	requestIDHeaderList := []string{"X-Amzn-Requestid", "X-Amz-RequestId"}
+
+	for _, h := range requestIDHeaderList {
+		// check for headers known to contain Request id
+		if v := resp.Header.Get(h); len(v) != 0 {
+			// set reqID on metadata for successful responses.
+			SetRequestIDMetadata(&metadata, v)
```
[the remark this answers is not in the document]
skotambkar: The document deserializers need access to "SetRequestIDMetadata" to set parsed request id from document body.

jasdel (approved):

jasdel:
Should decide if `service/internal/s3shared` is its own module or not

PR #744, opened 2020-09-21
Use service index everywhere sigv4 is checked

JordonPhillips:
This updates the service metadata generation to use the service index to check for the sigv4 trait. While this is technically correct, it may not be necessary, but I figured the PR can be a discussion point

### DIFF
--- a/service/cognitoidentityprovider/api_op_ChangePassword.go
+++ b/service/cognitoidentityprovider/api_op_ChangePassword.go
@@ -74,7 +74,6 @@ func newServiceMetadataMiddleware_opChangePassword(region string) awsmiddleware.
 	return awsmiddleware.RegisterServiceMetadata{
 		Region:        region,
 		ServiceID:     ServiceID,
-		SigningName:   "cognito-idp",
 		OperationName: "ChangePassword",
 	}
 }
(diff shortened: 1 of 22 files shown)


service/cognitoidentityprovider/api_op_ChangePassword.go
```diff
@@ -74,7 +74,6 @@ func newServiceMetadataMiddleware_opChangePassword(region string) awsmiddleware.
 	return awsmiddleware.RegisterServiceMetadata{
 		Region:        region,
 		ServiceID:     ServiceID,
-		SigningName:   "cognito-idp",
```
skotambkar, line 77: Just need clarification, was looking for why Signing Name is getting removed with getEffectiveAuthSchemes.

The documentation states that: 
```
If the operation does not define an {@code smithy.api#auth} trait, then the effective auth schemes of the service is returned (that is, the return value of {@link #getEffectiveAuthSchemes(ToShapeId)}).

``` 

The service here has sigv4 auth trait, the operation has the auth trait but is assigned as `[]`. Does this indicate no auth for the operation (even if service has auth trait on it)? I think that would justify the removal of signing name.

JordonPhillips: Yeah, an empty auth set on an operation indicates that it must not be authenticated. That said I don't know if it really matters in this case since it's just metadata whose presence or lack thereof won't have a material impact other than there being one less string to allocate.

skotambkar (approved):

service/cognitoidentityprovider/api_op_ChangePassword.go
skotambkar: Ok if it isn't needed, we shouldn't generate it. Looks good

PR #747, opened 2020-09-23
Add Glacier Tree Hash customization

JordonPhillips:
This ports over the Glacier tree hash computation middleware. Glacier still needs a few other customizations, but I figured one at a time is best.

### DIFF
--- a/service/glacier/api_op_AbortMultipartUpload.go
+++ b/service/glacier/api_op_AbortMultipartUpload.go
@@ -7,6 +7,7 @@ import (
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	glaciercust "github.com/aws/aws-sdk-go-v2/service/glacier/internal/customizations"
 	smithy "github.com/awslabs/smithy-go"
 	"github.com/awslabs/smithy-go/middleware"
 	smithyhttp "github.com/awslabs/smithy-go/transport/http"
@@ -49,6 +50,7 @@ func (c *Client) AbortMultipartUpload(ctx context.Context, params *AbortMultipar
 	smithyhttp.AddCloseResponseBodyMiddleware(stack)
 	addOpAbortMultipartUploadValidationMiddleware(stack)
 	stack.Initialize.Add(newServiceMetadataMiddleware_opAbortMultipartUpload(options.Region), middleware.Before)
+	glaciercust.AddTreeHashMiddleware(stack)
 
 	for _, fn := range options.APIOptions {
 		if err := fn(stack); err != nil {
--- a/service/glacier/api_op_AbortVaultLock.go
+++ b/service/glacier/api_op_AbortVaultLock.go
@@ -7,6 +7,7 @@ import (
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	glaciercust "github.com/aws/aws-sdk-go-v2/service/glacier/internal/customizations"
 	smithy "github.com/awslabs/smithy-go"
 	"github.com/awslabs/smithy-go/middleware"
 	smithyhttp "github.com/awslabs/smithy-go/transport/http"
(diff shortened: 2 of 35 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/GlacierCustomizations.java
```diff
@@ -0,0 +1,36 @@
...
+        return 127;
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                RuntimeClientPlugin.builder()
+                        .servicePredicate(GlacierCustomizations::isGlacier)
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(TREE_HASH_ADDER).build())
```
jasdel, line 27: This symbol is probably supposed to refer to the `glacier/internal/customizations` pkg. e.g. https://github.com/aws/aws-sdk-go-v2/blob/feature/SmithyAPIClient/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsCustomGoDependency.java

jasdel, line 27: This symbol is probably supposed to refer to the `glacier/internal/customizations` pkg. e.g. https://github.com/aws/aws-sdk-go-v2/blob/feature/SmithyAPIClient/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsCustomGoDependency.java

service/glacier/internal/customizations/treehash.go
```diff
@@ -0,0 +1,144 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"github.com/awslabs/smithy-go"
+	"github.com/awslabs/smithy-go/middleware"
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+	"io"
+)
+
+const bufsize = 1024 * 1024
```
jasdel, line 14: not significant but this could be defined in the `computeHashes` function body as a `const` instead of at the package level. Prevents extra pkg `const`s that aren't clear on their usage.

service/glacier/internal/customizations/treehash.go
```diff
@@ -0,0 +1,144 @@
...
+
+func addChecksum(req *smithyhttp.Request) error {
+	if req.GetStream() == nil || req.Header.Get("X-Amz-Sha256-Tree-Hash") != "" {
+		return nil
+	}
+
+	h := computeHashes(req.GetStream())
+	if err := req.RewindStream(); err != nil {
+		return err
+	}
```
jasdel, line 58: Probably need to check before computing the hash if the stream is seekable or not, and fail when the reader is not seakable with a descriptive error.

To do this smithy-go should expose an accessor to check if the stream is seekable,

https://github.com/awslabs/smithy-go/blob/dab44aa9975953da6d6c9b6ab922800fb6b3be79/transport/http/request.go#L17

service/glacier/internal/customizations/treehash.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// ComputeHashes computes the tree-hash and linear hash of a reader r.
+//
+// See http://docs.aws.amazon.com/amazonglacier/latest/dev/checksum-calculations.html for more information.
+func ComputeHashes(r io.ReadSeeker) Hash {
+	start, _ := r.Seek(0, io.SeekCurrent) // Read the whole stream
+	defer r.Seek(start, io.SeekStart)     		// Rewind stream at end
+	return computeHashes(r)
+}
```
jasdel, line 83: `ComputeHashes` and `computeHashes` is a little confusing since it looks like only the lower case `computeHashes` is used.

service/glacier/internal/customizations/treehash_test.go
```diff
@@ -0,0 +1,150 @@
...
+		"all zeroes": {
+			Body:   createTestReader(),
+			Linear: "68aff0c5a91aa0491752bfb96e3fef33eb74953804f6a2f7b708d5bcefa8ff6b",
+			Tree:   "154e26c78fd74d0c2c9b3cc4644191619dc4f2cd539ae2a74d5fd07957a3ee6a",
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			actual := ComputeHashes(c.Body)
```
jasdel, line 95: Since `ComputeHashes` is only used for tests, it should be defined in this `treehash_test.go` file instead of the `treehash.go` file so its not accidentally used when it shouldn't be.

jasdel (changes requested):

service/glacier/internal/customizations/treehash.go
JordonPhillips: Upper case is what's exposed for customers to use. These functions are exposed in the existing version of V2 so I wanted to keep that. Not exposing them would definitely be simpler if that's where we want to go.

jasdel: In this case this func is nested under an `internal` package so won't be visible to them.

service/glacier/internal/customizations/treehash.go
JordonPhillips: https://github.com/awslabs/smithy-go/pull/182

jasdel (approved):

PR #748, opened 2020-09-23
`s3`: Adds customization to disable s3 auto gzip decode

skotambkar:
Adds customization to disable s3 auto gzip decode
- [x] Adds codegen customization for AWS Config field to set transport's `DisableCompression` to true.
- [x] Generate code change for s3/api_client.go

### DIFF
--- a/service/s3/api_client.go
+++ b/service/s3/api_client.go
@@ -8,6 +8,7 @@ import (
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+	acceptencodingcust "github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding"
 	"github.com/aws/aws-sdk-go-v2/service/internal/s3shared"
 	s3cust "github.com/aws/aws-sdk-go-v2/service/s3/internal/customizations"
 	"github.com/awslabs/smithy-go/middleware"
@@ -359,4 +360,8 @@ func addResponseErrorMiddleware(stack *middleware.Stack) {
 
 func addMetadataRetrieverMiddleware(stack *middleware.Stack) {
 	s3shared.AddMetadataRetrieverMiddleware(stack)
+}
+
+func disableAcceptEncodingGzip(stack *middleware.Stack) {
+	acceptencodingcust.AddAcceptEncodingGzip(stack, acceptencodingcust.AddAcceptEncodingGzipOptions{})
 }
--- a/service/dynamodb/api_client.go
+++ b/service/dynamodb/api_client.go
@@ -11,6 +11,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	ddbcust "github.com/aws/aws-sdk-go-v2/service/dynamodb/internal/customizations"
+	acceptencodingcust "github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding"
 	"github.com/awslabs/smithy-go/middleware"
 	smithyrand "github.com/awslabs/smithy-go/rand"
 	"net/http"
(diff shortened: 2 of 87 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -88,6 +89,26 @@
...
+        String serviceId= service.expectTrait(ServiceTrait.class).getSdkId();
+        // services that require modified HTTP Client
+        String []serviceList = {"S3"};
+        for (String s : serviceList) {
+            if (serviceId.equalsIgnoreCase(s)) {
+                return false;
+            }
+        }
+        return true;
+    }
```
jasdel, line 110: Is there anyway this can be triggered by GoIntegrations?  sprinkling the customizations throughout the generator code will make it difficult to track down and understand what all is being done.

service/s3/api_client.go
```diff
@@ -150,18 +150,22 @@ func NewFromConfig(cfg aws.Config, optFns ...func(*Options)) *Client {
...
+	if o.HTTPClient != nil {
 		return
 	}
-	o.Retryer = retry.NewStandard()
+	client := aws.NewBuildableHTTPClient()
+	// set client transport's DisableCompression to true. This will instruct the HTTP
+	// Client to not auto decode the GZIP response.
+	client.GetTransport().DisableCompression = true
+	o.HTTPClient = client
 }
```
jasdel, line 169: We'll want to document this behavior. What do you think about using a pattern similar to DynamoDB's disable compression customization? If the customer were to create a custom HTTP client or a custom transport this method would not work, and the behavior of the S3 API client would change significantly, in a non-obvious way.  This was the reason the DDB's customization set the `Accept-Encoding` header.

Its not 100% clear if s3 should have the option to enable compression or not though. future behavior of the s3 client like trailing checksum validation might get overly complicated if the API client support enabling decompressing responses.

jasdel (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
skotambkar: This was not needed any more, so removed.

service/s3/api_client.go
skotambkar: Moved the dynamodb middleware to service/internal/customizations so as to reuse it ; For s3 we won't generate Enable option and default to disabling

service/internal/customizations/accept_encoding_gzip_test.go
```diff
@@ -196,3 +196,20 @@ func (*stubOpDeserializer) HandleDeserialize(
 ) {
 	return next.HandleDeserialize(ctx, input)
 }
+
```
jasdel, line 199: What do you think about  putting this under a more specific customization package so it doesn't get lost. or becomes a dumping area.  maybe something like `service/internal/customizations/accept-encoding-gzip`.

Also the `dynamodb/internal/customization/doc.go` might need a tweak to to say it refers to this customization maybe?

skotambkar: moved to `service/internal/accept-encoding`. Also made the accept-encoding a separate go mod

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsCustomGoDependency.java
```diff
@@ -29,8 +29,11 @@
             "service/apigateway/internal/customizations", "agcust");
     public static final GoDependency GLACIER_CUSTOMIZATION = aws(
             "service/glacier/internal/customizations", "glaciercust");
+
     public static final GoDependency S3_SHARED_CUSTOMIZATION = awsModuleDep(
             "service/internal/s3shared", null, "s3shared");
+    public static final GoDependency ACCEPT_ENCODING_CUSTOMIZATION = awsModuleDep(
+            "service/internal/accept-encoding", null, "acceptencodingcust");
```
jasdel, line 36: Won't this initially generate an invalid hash e.g. `v0.25.1-` ? since these are new modules. Maybe the `awsModuleDep` should use a different version, e,g just regular `gohash` instead of sharing the `AWS_SDK` value?

skotambkar: Updated awsModuleDep to use a separate version from AWS_SDK.

jasdel (approved):

PR #751, opened 2020-09-24
Add glacier service version and account id customizations

JordonPhillips:
This adds customizations for setting the default account id and setting the glacier service version

### DIFF
--- a/service/glacier/api_client.go
+++ b/service/glacier/api_client.go
@@ -197,3 +197,210 @@ func addRequestIDRetrieverMiddleware(stack *middleware.Stack) {
 func addResponseErrorMiddleware(stack *middleware.Stack) {
 	awshttp.AddResponseErrorMiddleware(stack)
 }
+
+// setDefaultAccountID sets the AccountID to the given value if the current value
+// is nil
+func setDefaultAccountID(input interface{}, accountID string) interface{} {
+	switch i := input.(type) {
+	case *GetVaultNotificationsInput:
+		if i.AccountId == nil {
+			i.AccountId = &accountID
+		}
+		return i
+
+	case *AddTagsToVaultInput:
+		if i.AccountId == nil {
+			i.AccountId = &accountID
+		}
+		return i
+
+	case *InitiateMultipartUploadInput:
+		if i.AccountId == nil {
+			i.AccountId = &accountID
+		}
+		return i
+
+	case *DeleteVaultInput:
+		if i.AccountId == nil {
+			i.AccountId = &accountID
+		}
+		return i
+
+	case *AbortVaultLockInput:
+		if i.AccountId == nil {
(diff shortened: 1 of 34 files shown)


service/glacier/internal/customizations/apiversion.go
```diff
@@ -0,0 +1,43 @@
...
+const glacierAPIVersionHeaderKey = "X-Amz-Glacier-Version"
+
+// AddGlacierAPIVersionMiddleware explicitly add handling for the Glacier api version
+// middleware to the operation stack.
+func AddGlacierAPIVersionMiddleware(stack *middleware.Stack, apiVersion string) {
+	stack.Serialize.Add(&GlacierAPIVersionMiddleware{apiVersion: apiVersion}, middleware.Before)
+}
+
+// GlacierAPIVersionMiddleware handles automatically setting Glacier's API version header.
+type GlacierAPIVersionMiddleware struct{
```
skotambkar, line 20: `GlacierAPIVersionMiddleware` doesn't need to be exported

service/glacier/internal/customizations/accountid.go
```diff
@@ -0,0 +1,34 @@
...
+// AddDefaultAccountIDMiddleware adds the DefaultAccountIDMiddleware to the stack using
+// the given options.
+func AddDefaultAccountIDMiddleware(stack *middleware.Stack, setDefaultAccountID setDefaultAccountID) {
+	stack.Initialize.Add(&DefaultAccountIDMiddleware{
+		setDefaultAccountID: setDefaultAccountID,
+	}, middleware.Before)
+}
+
+// DefaultAccountIDMiddleware sets the account ID to "-" if it isn't already set
+type DefaultAccountIDMiddleware struct {
```
skotambkar, line 19: `DefaultAccountIDMiddleware` should be unexported

service/glacier/api_client.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 const ServiceID = "Glacier"
+const ServiceAPIVersion = "2012-06-01"
```
skotambkar, line 17: We probably won't need `ServiceAPIVersion` exported

skotambkar (changes requested):
Minor comments on not exporting middlewares var/constant

service/glacier/api_client.go
JordonPhillips: This is generated by smithy-go

jasdel: It was added in another PR to smithy-go. Not sure benefit to user so much but useful for debugging potentially.  Is this an attribute that should be included in the user agent for the client? or does the request already capture that value?

jasdel (approved):

skotambkar (approved):

PR #753, opened 2020-09-24
Add utility to run Go commands for all modules in the repo

jasdel:
Adds a new utility `internal/repotools/cmd/eachmodule` that is used by the Make file to invoke commands on each module within the SDK repo. This utility walks the file tree searching for go modules, and then invokes the command on it. The utility can also run command in parallel.

Also moved the `internal/cmd/makerelative` to `internal/repotools/cmd/makerelative` to group under the same `repotools` module.

Updated the makefile with additional recipes to make it easier to run tests,build,bench on various modules in the repo.  Each task as a prefix `<job>-modules-` followed by the folder path relative to the repo's root that modules should be searched for. Since `/` are not valid make recipie values, you must use `_` instead. e.g. `test-modules-service_dynamodb` would search for modules in the `service/dynamodb` folder and invoke go tests on each module.

* `vet-modules-%`: runs go vet on the modules
* `build-modules-%`:  runs go build command on each module ensuring that it compiles.
* `test-modules-%`:  runs Go tests on the modules
* `test-modules-race-%`: run Go test with race enabled on the modules
* `integ-modules-%`: runs go test with integration on modules with test run prefix of `^TestInteg_`


skotambkar (approved):
LGTM

Makefile
```diff
@@ -38,25 +30,60 @@ gen-config-asserts:
...
+
+test-modules-%:
+	@# Test command that uses the pattern to define the root path that the
+	@# module testing will start from. Strips off the "test-modules-" and
+	@# replaces all "_" with "/".
+	@#
+	@# e.g. test-modules-internal_protocoltest
+	cd ./internal/repotools/cmd/eachmodule \
+		&& go run . -p $(subst _,/,$(subst test-modules-,,$@)) \
+		'go test ${BUILD_TAGS} -run NONE ./...' \
```
skmcgrail, line 79: Just a styling perspective, but since the `-run` argument takes a regex should we just pass `'^$'` as the match to actually truely specify none?

jasdel: Sure can switch to that, `NONE` was just a convention I'd been using the other can be used instead

internal/repotools/repo_root.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// FindRepoRoot returns the absolute path to the root directory of the
+// repository, or error. If the dir passed in is a relative path it will be
+// used relative to the current working directory of the executable.
+func FindRepoRoot(dir string) (string, error) {
```
skmcgrail, line 12: Should `makerelative` command be updated  to share the same findRepoRoot logic?

jasdel:
Updated per feedbac, and pushed changes but running into failure with `service/s3|s3control` can manually update that module until thats fixed.

internal/repotools/repo_root.go
jasdel: good point, i missed that

internal/repotools/walk.go
```diff
@@ -0,0 +1,51 @@
+package repotools
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// Boots was made for walking the file tree searching for modules.
```
skmcgrail, line 8: 🤦

skmcgrail (approved):

PR #755, opened 2020-09-24
Adds support to add go mod dependency for customization modules

skotambkar:
add support for adding go mod dependency for customization modules

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
@@ -45,11 +45,37 @@ public class AwsGoDependency {
     }
 
     protected static GoDependency aws(String relativePath, String alias) {
-        String importPath = AWS_SOURCE_PATH;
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
```diff
@@ -45,11 +45,37 @@ protected static GoDependency aws(String relativePath) {
...
+            String relativePath,
+            String version,
+            String alias
+    ) {
+        String importPath = moduleImportPath;
         if (relativePath != null) {
             importPath = importPath + "/" + relativePath;
         }
-        return GoDependency.moduleDependency(AWS_SOURCE_PATH, importPath, Versions.AWS_SDK, alias);
+        return GoDependency.moduleDependency(moduleImportPath, importPath, version, alias);
```
jasdel, line 78: can this be inherited from the `smithy-go`'s GoDependency?  Maybe that `module` should be made public if it cannot be inherited?

skotambkar: smithy-go class is final, so cannot extend it.

jasdel (approved):

PR #760, opened 2020-09-25
`S3`: Adds support for `X-Amz-Content-Sha25` header 

skotambkar:
Adds customization to s3 to add `X-Amz-Content-Sha25` header for all operations


jasdel:
Should this customization be applied to the s3control client as well?

jasdel (approved):

PR #767, opened 2020-09-28
codegen: Add changing s3 size member to long 

jasdel:
Updates the Amazon S3 model to use an long (int64) for `Size` type instead of an integer (int32).


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/BackfillS3ObjectSizeMemberShapeType.java
```diff
@@ -0,0 +1,79 @@
...
+        ShapeId serviceId = settings.getService();
+        if (!SERVICE_TO_SHAPE_MAP.containsKey(serviceId)) {
+            return model;
+        }
+
+        Set<ShapeId> shapeIds = SERVICE_TO_SHAPE_MAP.get(serviceId);
+
+        Model.Builder builder = model.toBuilder();
+        for (ShapeId shapeId : shapeIds) {
+            Shape shape = model.expectShape(shapeId);
```
mtdowling, line 62: I'd check if the shape exists and not do anything if it doesn't exist. You're already accounting for them possibly fixing these shapes in the future, so I think it's reasonable to account for the shape not existing (e.g., they rename a primitive shape which has basically no impact to anything else).

```
OptionalUtils.ifPresentOrElse(
    model.getShape(shapeId),
    shape -> {
        // do shape changes
    },
    () -> LOGGER.warning("shape " + shapeId + " not found in " + serviceId + " model");
);
```

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/BackfillS3ObjectSizeMemberShapeType.java
```diff
@@ -0,0 +1,79 @@
...
+
+            if (shape.isLongShape()) {
+                LOGGER.warning("shape is already modeled as an LONG does not require backfill");
+                continue;
+            }
+
+            LongShape.Builder longBuilder = LongShape.builder();
+            longBuilder.id(shape.getId());
+            // TODO all documentation, and traits are lost, those should be preserved.
```
mtdowling, line 71: You can do `longBuilder.addTraits(shape.getAllTraits().values());`

mtdowling (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/BackfillS3ObjectSizeMemberShapeType.java
jasdel: Thanks! this was what i was missing

skmcgrail (approved):

PR #772, opened 2020-09-29
Add read timeout to Kinesis GetRecords

JordonPhillips:
This ports over the Kinesis read timeout customization. As far as I can tell this only was ever added to GetRecords, so that's where it is.

### DIFF
--- a/aws/transport/http/timeout_read_closer.go
+++ b/aws/transport/http/timeout_read_closer.go
@@ -1,4 +1,4 @@
-package aws
+package http
 
 import (
 	"context"
--- a/aws/transport/http/timeout_read_closer_test.go
+++ b/aws/transport/http/timeout_read_closer_test.go
@@ -1,4 +1,4 @@
-package aws
+package http
 
 import (
 	"context"
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/KinesisCustomizations.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/KinesisCustomizations.java
@@ -31,7 +31,7 @@ public class KinesisCustomizations implements GoIntegration {
                         .operationPredicate(KinesisCustomizations::isGetRecords)
                         .registerMiddleware(MiddlewareRegistrar.builder()
                                 .resolvedFunction(SymbolUtils.createValueSymbolBuilder(READ_TIMEOUT_ADDER,
-                                        AwsGoDependency.AWS_CORE).build())
+                                        AwsGoDependency.AWS_HTTP_TRANSPORT).build())
                                 .functionArguments(ListUtils.of(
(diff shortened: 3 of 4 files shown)


skotambkar (changes requested):
Minor assertion change

aws/timeout_read_closer_test.go
```diff
@@ -53,3 +57,42 @@ func TestTimeoutReadCloserSameDuration(t *testing.T) {
...
+
+	response, ok := out.RawResponse.(*smithyhttp.Response)
+	if !ok {
+		t.Errorf("expect smithy response, got %T", out.RawResponse)
+	}
+
+	b := make([]byte, 100)
+	_, err = response.Body.Read(b)
+	if err != nil {
+		t.Log(err)
```
skotambkar, line 94: Thanks for the update. Instead of logging the error, we should assert the type and expected error. 

```
readTimeoutError := &ReadTimeoutError{}
if !errors.Is(err, &readTimeoutError) {
	t.Fatalf("expect %T error, got %T", readTimeoutError, err)
}

```

skotambkar (approved):

aws/timeout_read_closer.go
```diff
@@ -56,3 +60,44 @@ func (r *timeoutReadCloser) Read(b []byte) (int, error) {
 func (r *timeoutReadCloser) Close() error {
 	return r.reader.Close()
 }
+
+// AddResponseReadTimeoutMiddleware adds a middleware to the stack that wraps the
+// response body so that a read that takes too long will return an error.
+func AddResponseReadTimeoutMiddleware(stack *middleware.Stack, duration time.Duration) {
+	stack.Deserialize.Add(&readTimeoutMiddleware{duration: duration}, middleware.After)
+}
```
jasdel, line 68: This middleware probably should go in the SDK's `aws/transport/http` package since it is specific to `smithyhttp.Request`.  Ideally the `aws` package is transport agnostic.

jasdel (changes requested):

aws/timeout_read_closer.go
```diff
@@ -56,3 +60,44 @@ func (r *timeoutReadCloser) Read(b []byte) (int, error) {
...
+
```
jasdel, line 69: This middleware probably should go in the SDK's `aws/transport/http` package since it is specific to `smithyhttp.Request`.  Ideally the `aws` package is transport agnostic.

Our strategy for middleware pkg location isn't really clear, we need to decide where these things go more concretely.

jasdel (changes requested):

jasdel (approved):

service/kinesis/internal/customizations/readtimeout.go
```diff
@@ -0,0 +1,8 @@
+package customizations
```
jasdel, line 1: Nit it would be helpful to have a `doc.go` with package godoc explaining this customization and why it is here, github linked issues woudl be awesome too.

PR #775, opened 2020-09-30
codegen: Add skipping protocol test cases that are not supported

jasdel:
Adds skipping of protocol test cases that are not supported, e.g. document, and opinionated protocol tests like deserialize prefix headers as empty map vs nil value.

Depends on https://github.com/awslabs/smithy-go/pull/190

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ matrix:
 
     - language: java
       jdk: openjdk8
+      env:
+        - JAVA_TOOL_OPTIONS="-Xmx1g"
       before_script:
           - rm -rf /tmp/smithy-go
           - git clone https://github.com/awslabs/smithy-go /tmp/smithy-go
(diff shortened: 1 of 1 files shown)


jasdel:
Looks like windows CI failure are unrelated release tooling test failures: https://travis-ci.org/github/aws/aws-sdk-go-v2/jobs/731455786#L1586

smithy build fails because smithy-go change isn't committed yet.

Makefile
```diff
@@ -32,7 +35,7 @@ all: generate unit
 generate: smithy-generate gen-config-asserts gen-repo-mod-replace tidy-modules-. add-module-license-files
 
 smithy-generate:
-	cd codegen && ./gradlew clean build -Plog-tests && ./gradlew clean
+	cd codegen && ./gradlew clean build -Plog-tests
```
skmcgrail, line 38: I thought this was hear to avoid having build directories being present in the repo and being iterated over?

jasdel: The eachmodule now automatically ignores the `/codegen` since there is not really any case we want any module in there to be evaluated.

jasdel: added new `smithy-build` that don't clean `smith-generate` will now do the clean after the build.

skmcgrail (approved):

skotambkar (approved):

PR #776, opened 2020-09-30
`s3`: Adds customization for s3 error responses with 2xx status code

skotambkar:
* Adds customization for `CopyObject`, `CompleteMultipartUpload`, and `UploadPartCopy` operations to support deserializing error responses with 2xx status code.

### DIFF
--- a/service/s3/internal/customizations/doc.go
+++ b/service/s3/internal/customizations/doc.go
@@ -1,0 +1,44 @@
+/*
+Package customizations provides customizations for the Amazon S3 API client.
+
+This package provides support for following S3 customizations
+
+    UpdateEndpoint Middleware: Virtual Host style url addressing
+
+    processResponseWith200Error Middleware: Deserializing response error with 200 status code
+
+
+Virtual Host style url addressing
+
+Since serializers serialize by default as path style url, we use customization
+to modify the endpoint url when `UsePathStyle` option on S3Client is unset or
+false.
+
+UpdateEndpoint middleware handler for virtual host url addressing needs to be
+executed after request serialization.
+
+    Middleware layering:
+
+    HTTP Request -> operation serializer -> Update-Endpoint customization -> next middleware
+
+Customization option:
+    UsePathStyle (Disabled by Default)
+
+Handle Error response with 200 status code
+
+S3 operations: CopyObject, CompleteMultipartUpload, UploadPartCopy can have an
+error Response with status code 2xx. The processResponseWith200Error middleware
+customizations enables SDK to check for an error within response body prior to
+deserialization.
+
+As the check for 2xx response containing an error needs to be performed earlier
+than response deserialization. Since the behavior of Deserialization is in
+reverse order to the other stack steps its easier to consider that "after" means
+"before".
+
+    Middleware layering:
+
+	HTTP Response -> handle 200 error customization -> deserialize
+
+*/
+package customizations


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Gets the sort order of the customization from -128 to 127, with lowest
+     * executed first.
+     *
+     * @return Returns the sort order, defaults to -40.
+     */
+    @Override
+    public byte getOrder() {
+        return 127;
```
JordonPhillips, line 36: Can you add a comment explaining why this has to go towards the end?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
```diff
@@ -0,0 +1,97 @@
...
+    private static boolean supports200Error(Model model, ServiceShape service, OperationShape operation){
+        if (!isS3Service(model, service)) {
+            return false;
+        }
+
+        for (String op:customizedOperations) {
+            if (operation.getId().getName().equalsIgnoreCase(op)) {
+                return true;
+            }
+        }
```
JordonPhillips, line 88: A Set + contains would work better here

service/s3/api_client.go
```diff
@@ -360,3 +360,7 @@ func addResponseErrorMiddleware(stack *middleware.Stack) {
 func addMetadataRetrieverMiddleware(stack *middleware.Stack) {
 	s3shared.AddMetadataRetrieverMiddleware(stack)
 }
+
+func handleResponseErrorWith200Status(stack *middleware.Stack) {
+	s3cust.HandleResponseErrorWith200Status(stack)
+}
```
JordonPhillips, line 366: Why is this needed if it's effectively just a function alias?

service/s3/internal/customizations/handle_200_error.go
```diff
@@ -0,0 +1,126 @@
...
+// HandleResponseErrorWith200Status check for S3 200 error response.
+// If an s3 200 error is found, status code for the response is modified temporarily to
+// 5xx response status code.
+func HandleResponseErrorWith200Status(stack *middleware.Stack) {
+	stack.Deserialize.Insert(&processResponseFor200ErrorMiddleware{}, "OperationDeserializer", middleware.After)
+	stack.Deserialize.Insert(&reset200StatusCodeMiddleware{}, "OperationDeserializer", middleware.Before)
+}
+
+// middleware to process raw response and look for error response with 200 status code
+type processResponseFor200ErrorMiddleware struct{}
```
JordonPhillips, line 27: You don't have to do this, but I wonder if this could be a single middleware registered to multiple stack steps? Theoretically such a middleware could maintain its own internal state so metadata passing isn't necessary.

JordonPhillips (changes requested):

service/s3/api_client.go
jasdel: Could call the `s3cust` helper directly without the additional function helper since there is no additional setup needed.

service/s3/internal/customizations/handle_200_error.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	var readBuff bytes.Buffer
+	body := io.TeeReader(response.Body, &readBuff)
+
+	rootDecoder := xml.NewDecoder(body)
+	t, err := smithyxml.FetchRootElement(rootDecoder)
+	if err == io.EOF {
+		return out, metadata, nil
+	}
```
jasdel, line 58: If a 200 status code is received, but not response body is received the SDK should consider that an error. Since the service is required to provide some kind of response document for these operation.

service/s3/internal/customizations/handle_200_error.go
```diff
@@ -0,0 +1,126 @@
...
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+	smithyxml "github.com/awslabs/smithy-go/xml"
+)
+
+// HandleResponseErrorWith200Status check for S3 200 error response.
+// If an s3 200 error is found, status code for the response is modified temporarily to
+// 5xx response status code.
+func HandleResponseErrorWith200Status(stack *middleware.Stack) {
+	stack.Deserialize.Insert(&processResponseFor200ErrorMiddleware{}, "OperationDeserializer", middleware.After)
+	stack.Deserialize.Insert(&reset200StatusCodeMiddleware{}, "OperationDeserializer", middleware.Before)
```
jasdel, line 23: I don't think this is correct. The SDK should consider any 200 error as a 500 status code failure. swapping back to 200 can easily cause unexpected behavior.  Especially for the response error wrapper which maybe report a 200 error when the SDK should be reporting a 500 error back to the caller.

service/s3/internal/customizations/handle_200_error_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+)
+
+type EndpointResolverFunc func(region string, options s3.ResolverOptions) (aws.Endpoint, error)
+
+func (fn EndpointResolverFunc) ResolveEndpoint(region string, options s3.ResolverOptions) (endpoint aws.Endpoint, err error) {
+	return fn(region, options)
+}
+
+func TestErrorResponseWith200StatusCode(t *testing.T) {
```
jasdel, line 25: Would be helpful to have two additional test cases

* No response document received at all, the SDK can use the unexpected eof error for this case. Or some other similar error that might be more descriptive.
* Partial response document received, that should fail with unexpected eof.

jasdel (changes requested):

service/s3/internal/customizations/handle_200_error.go
jasdel: The same middleware helper would need to be registered multiple times which is not supported with the middleware stack steps. Using the metadata to communicate to the `before` middleware probably is a good way to upstream that information.

service/s3/internal/customizations/handle_200_error.go
jasdel: To help clarify this a little. We should consider if it matters at all to inform the user that this was a 200 error vs just regular error.

Most likely the user doesn't need to know that a 200 was received but contained an error response. Most likely the user only needs to know the operation failed, the rest is internal details.

So with that said I suggest removing the 200 status reset middleware completely, and not replace it with anything.  Leave the rest of the middleware the way it is.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
```diff
@@ -0,0 +1,72 @@
...
+import software.amazon.smithy.utils.ListUtils;
+
+/**
+ * Adds middleware to handle S3 response errors with 200 ok status code.
+ */
+public class S3ErrorWith200Status implements GoIntegration {
+    private static String ADD_ERROR_HANDLER_INTERNAL = "HandleResponseErrorWith200Status";
+
+    // list of operations for which this customization is valid.
+    private static Set<String> customizedOperations = Stream.of(
```
JordonPhillips, line 30: `SetUtils.of`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
```diff
@@ -0,0 +1,72 @@
...
+    // returns true if the operation supports error response with 200 ok status code
+    private static boolean supports200Error(Model model, ServiceShape service, OperationShape operation){
+        if (!isS3Service(model, service)) {
+            return false;
+        }
+
+        if (customizedOperations.contains(operation.getId().getName())) {
+            return true;
+        }
+        return false;
```
JordonPhillips, line 64: You can just return the result of contains

JordonPhillips (changes requested):
Did you remove the generated code?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
skotambkar: Isn't that a Java9 thing? Thought we were limited to Java8

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
skotambkar: Right, I missed refactoring this. Thanks

JordonPhillips (approved):
Looks good on the code side but still needs the generated code added back in

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3ErrorWith200Status.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                RuntimeClientPlugin.builder()
+                        .operationPredicate(S3ErrorWith200Status::supports200Error)
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(ADD_ERROR_HANDLER_INTERNAL,
+                                        AwsCustomGoDependency.S3_CUSTOMIZATION).build())
+                                .build())
```
jasdel, line 42: nit line length.

service/s3/internal/customizations/handle_200_error.go
```diff
@@ -0,0 +1,71 @@
+package customizations
```
jasdel, line 1: It would be good to update the `s3/internal/customizations/doc.go` to include information about this customization

jasdel (approved):

Issue #777, opened 2020-09-30
aws/to_ptr  same address?


hacker65536:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
This code gives only one result.
```go
package main

import (
	"context"
	"encoding/json"
	"fmt"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/ec2"
)

func main() {

	cfg, err := config.LoadDefaultConfig()
	if err != nil {
		panic("unable to load SDK config, " + err.Error())
	}

	svc := ec2.NewFromConfig(cfg)

	ids := []string{
		"i-0a70c779ab9ea03b3",
		"i-06dde7896483e8f86",
	}

	params := &ec2.DescribeInstancesInput{
		InstanceIds: aws.StringSlice(ids),
	}
	resp, err := svc.DescribeInstances(context.Background(), params)
	if err != nil {
		panic("failed to describe instances, " + err.Error())
	}

	r, _ := json.Marshal(resp.Reservations)
	fmt.Println(string(r))
}
```

Is this correct?
https://github.com/aws/aws-sdk-go-v2/blob/4a1d0a5a322eda3296ed32a9fa7b811db5f6c684/aws/to_ptr.go#L66-L75

**Version of AWS SDK for Go?**
v0.25.0

**Version of Go (`go version`)?**
 go1.15.2 linux/amd64

**To Reproduce (observed behavior)**

https://play.golang.org/p/3P1K30uzW-W

**Expected behavior**

https://play.golang.org/p/FdZT_wMM0iR

**Additional context**

skmcgrail:
Thanks for the report @hacker65536  this has been addressed and included in yesterdays [v0.26.0](https://github.com/aws/aws-sdk-go-v2/releases/tag/v0.26.0) release.

PR #778, opened 2020-09-30
fix to_ptr address

shogo82148:
fixes https://github.com/aws/aws-sdk-go-v2/issues/777

### DIFF
--- a/aws/to_ptr.go
+++ b/aws/to_ptr.go
@@ -14,6 +14,7 @@ func Bool(v bool) *bool {
 func BoolSlice(vs []bool) []*bool {
 	ps := make([]*bool, len(vs))
 	for i, v := range vs {
+		v := v
 		ps[i] = &v
 	}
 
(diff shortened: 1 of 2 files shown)


jasdel (commented):
Thanks for taking the time to create this PR @shogo82148 in both the v2 SDK and `smithy-go` repository. I'd like to see if it is possible to use Go aliases to refer to the `smithy-go` `ptr` pkg's implementation without needing to copy/redefine it here in the SDK repository. There may be godoc issues that prevent this though, which in that case we can just keep the copy pattern.

jasdel:
Thanks for creating this PR @shogo82148 I went a head and change the v2 SDK's behavior to rely on code gen wrappers around the `smithy-go` implementation so we only have one implementation to be concerned about.

Lets use https://github.com/aws/aws-sdk-go-v2/pull/781 for fixing this issue in the v2 SDK.

PR #781, opened 2020-09-30
 aws: Update From/To pointer utilities to delegate to smithy-go

jasdel:
Updates the From and To pointer helper utilities to delegating to the smithy-go implementation instead of duplicating it. The wrappers are code generated based on the smithy-go/ptr definitions.

Related to https://github.com/aws/aws-sdk-go-v2/pull/778 and https://github.com/awslabs/smithy-go/pull/191

### DIFF
--- a/aws/doc.go
+++ b/aws/doc.go
@@ -55,5 +55,7 @@
 // value of the SDK's Config type when creating a Session or service client.
 package aws
 
+// generate.go uses a build tag of "ignore", go run doesn't need to specify
+// this because go run ignores all build flags when running a go file directly.
 //go:generate go run -tags codegen generate.go
 //go:generate gofmt -w -s .


aws/generate.go
```diff
@@ -0,0 +1,117 @@
+// +build ignore
```
skmcgrail, line 1: Should this be codegen? See the `go:generate` command in `doc.go`.

jasdel: not i this case the `generate.go` file is supposed to use a magical `+build ignore` tag that will only be enabled when `go generate` is called, which will implicitly add the tag.

jasdel: will update docs for `go:generate` command to clarify this behavior with `ignore`

skmcgrail (approved):

PR #782, opened 2020-09-30
Override ml endpoint with PredictEndpoint if set

JordonPhillips:

### DIFF
--- a/service/machinelearning/internal/customizations/doc.go
+++ b/service/machinelearning/internal/customizations/doc.go
@@ -1,0 +1,13 @@
+/*
+Package customizations provides customizations for the Machine Learning API client.
+
+The Machine Learning API client uses one customization to support the PredictEndpoint
+input parameter.
+
+Predict Endpoint
+
+The predict endpoint customization runs after normal endpoint resolution happens. If
+the user has provided a value for PredictEndpoint then this customization will
+overwrite the request's endpoint with that value.
+ */
+package customizations


service/machinelearning/internal/customizations/predictendpoint.go
```diff
@@ -0,0 +1,61 @@
...
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+	"net/url"
+)
+
+type fetchPredictEndpointFunc func(interface{}) (*string, error)
+
+// AddPredictEndpointMiddleware adds the middleware required to set the endpoint
+// based on Predict's PredictEndpoint input member.
+func AddPredictEndpointMiddleware(stack *middleware.Stack, endpoint fetchPredictEndpointFunc) {
+	stack.Serialize.Insert(&predictEndpointMiddleware{}, "ResolveEndpoint", middleware.After)
```
jasdel, line 17: At some point we need to figure out a strategy for these middleware IDs and be constants somehow. Potentially with middleware slots, this may be easier. https://github.com/awslabs/smithy-go/issues/156

service/machinelearning/internal/customizations/predictendpoint.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if endpoint != nil && len(*endpoint) != 0 {
+		uri, err := url.Parse(*endpoint)
+		if err != nil {
+			return out, metadata, &smithy.SerializationError{
+				Err: fmt.Errorf("unable to parse predict endpoint, %v", err),
+			}
+		}
+		req.URL = uri
+		in.Request = req
```
jasdel, line 57: this line isn't really needed because the `req` value is a pointer, which is already stored in the `in.Request` interface value.  Kind of like a pointer to a pointer. So any updates to `req`'s members will be present in `in.Request`

service/machinelearning/internal/customizations/predictendpoint.go
```diff
@@ -0,0 +1,58 @@
+package customizations
```
jasdel, line 1: Nit it would be helpful to have a doc.go with package godoc explaining this customization and why it is here.

jasdel (approved):

PR #785, opened 2020-09-30
Add route53 uri sanitizer middleware

JordonPhillips:

### DIFF
--- a/service/route53/internal/customizations/sanitizeurl.go
+++ b/service/route53/internal/customizations/sanitizeurl.go
@@ -1,0 +1,57 @@
+package customizations
+
+import (
+	"context"
+	"fmt"
+	"github.com/awslabs/smithy-go"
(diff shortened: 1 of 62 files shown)


service/route53/internal/customizations/sanitizeurl.go
```diff
@@ -0,0 +1,57 @@
+package customizations
```
jasdel, line 1: It would be helpful to have a `doc.go` with package godoc describing this customization and why its here. Linked issues references would be awesome too if possible.

jasdel (changes requested):
After discussing this feature further the V2 SDK's implementation probably would be better to follow `boto`'s more focused approach to only applying the customizations to `ID` members serialized to the URI path directly, instead of a blanket customizations across all operations of the API.

PR #789, opened 2020-10-02
Strip Monotonic Clock Readings when Comparing Credential Expiry Time

skmcgrail:
Strip monotonic clock readings from credential cache implementation when comparing expiration time.

### DIFF
--- a/aws/credentials.go
+++ b/aws/credentials.go
@@ -90,7 +90,9 @@ type Credentials struct {
 // Expired returns if the credentials have expired.
 func (v Credentials) Expired() bool {
 	if v.CanExpire {
-		return !v.Expires.After(sdk.NowTime().Truncate(0))
+		// Calling Round(0) on the current time will truncate the monotonic reading only. Ensures credential expiry
+		// time is always based on reported wall-clock time.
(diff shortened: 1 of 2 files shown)


shogo82148:
I don't understand why stripping is needed.
Could you explain to me, please?

skmcgrail:
> I don't understand why stripping is needed.
> Could you explain to me, please?

The reason for stripping off monotonic readings when comparing the expiry time is to prevent issues where Go's time functions will utilize the monotonic clock readings if present on both time values. The monotonic clock reading can be incorrect in cases where the host system is hibernated (for example using [EC2 Hibernate](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Hibernate.html)). On Linux the monotonic clock readings post hibernation will not correctly represent the time that elapsed while the host system was hibernated. To ensure that credential expiration time comparisons are occurring based on the wall-clock time, we chose to call `Truncate(0)` on one of the time values being compared, which will strip off the monotonic reading but keep the wall-clock the same.

jasdel (approved):

shogo82148:
> The monotonic clock reading can be incorrect in cases where the host system is hibernated (for example using EC2 Hibernate).

I got it, thanks!

PR #791, opened 2020-10-02
codegen: Update SDK modeled endpoints

jasdel:
Updates's the SDK's modeled endpoints with the latest version. Regenerates the SDK's API client modules to also use the latest enpdoint version.


skmcgrail (approved):

Makefile
```diff
@@ -27,6 +27,10 @@ SDK_ALL_PKGS=${SDK_COMPA_PKGS} ${SDK_EXAMPLES_PKGS}
 RUN_NONE=-run '^$$'
 RUN_INTEG=-run '^TestInteg_'
 
+CODEGEN_RESOURCES_PATH=$(shell pwd)/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen
+ENDPOINTS_JSON=${CODEGEN_RESOURCES_PATH}/endpoints.json
+ENDPOINT_PREFIX_JSON=${CODEGEN_RESOURCES_PATH}/endpoint-prefix.json
+
 LICENSE_FILE=$(shell pwd)/LICENSE.txt
 
 all: generate unit
```
skotambkar, line 36: Should `all` call sync-endpoint-models?

skotambkar (approved):

Makefile
jasdel: I wasn't sure, how we wanted to do that since this really is a prep for releasing a update. I err'ed on the side of caution to not make it get the v1 SDK when uneeded. If the SDK had a similar location it could sync the Smithy API models from I'd vote it be included with that set of tasks.

PR #792, opened 2020-10-02
Route53  deserializer customization

skotambkar:
Adds customization to address route53's custom error deserializer behavior.

### DIFF
--- a/service/route53/go.mod
+++ b/service/route53/go.mod
@@ -3,7 +3,7 @@ module github.com/aws/aws-sdk-go-v2/service/route53
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2 v0.0.0-20201001231852-1fc1ab173989
+	github.com/aws/aws-sdk-go-v2 v0.0.0-20201005175632-36f8dc6bcc52
 	github.com/awslabs/smithy-go v0.1.1
 )
 
--- a/service/route53/go.sum
+++ b/service/route53/go.sum
@@ -1,5 +1,3 @@
-github.com/aws/aws-sdk-go-v2 v0.0.0-20201002220303-ffee47617303 h1:WvnHYDpFtCsvvaih/t1GB+kXFhcAtjQvxaMTfH7MyOE=
-github.com/aws/aws-sdk-go-v2 v0.0.0-20201002220303-ffee47617303/go.mod h1:KAhGbTQM7VGrReUuwuRBIp+SSiB01L/jdfvwjtCJtEo=
 github.com/awslabs/smithy-go v0.0.0-20200930175536-2cd7f70a8c2f/go.mod h1:hPOQwnmBLHsUphH13tVSjQhTAFma0/0XoZGbBcOuABI=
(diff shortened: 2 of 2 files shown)


service/route53/internal/customizations/custom_error_deser_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func TestCustomErrorDeserialization(t *testing.T) {
+	cases := map[string]struct {
+		responseStatus     int
+		responseBody       []byte
+		expectedError      string
+		expectedRequestID  string
+		expectedResponseID string
+	}{
+		"invalidChangeBatchError": {
```
jasdel, line 24: Not significant but adding a newline between the test case struct initialization would help readability. It took a little bit to see there were multiple test cases here.

service/route53/internal/customizations/doc.go
```diff
@@ -0,0 +1,41 @@
+/*
+Package customizations provides customizations for the Amazon Route53 API client.
+
+This package provides support for following customizations
+    Process Response Middleware: used for custom error deserializing
```
jasdel, line 5: nit white spaces seems a bit off.

jasdel, line 5: lint seems to complain (on some platforms) about this style. We switched to doing `// ` prefix instead of `/*` to not have lint complain.

service/route53/internal/customizations/custom_error_deser_test.go
```diff
@@ -0,0 +1,122 @@
...
+				}
+			}
+
+		})
+	}
+}
+
+type responseError interface {
+	ServiceRequestID() string
+}
```
jasdel, line 122: Since this is only used once in the test case it could be defined inline instead of a named type.

service/route53/internal/customizations/custom_error_deser.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// set request id in metadata
+	if len(err.RequestID) != 0 {
+		awsmiddle.SetRequestIDMetadata(metadata, err.RequestID)
+	}
+
+	return &types.InvalidChangeBatch{
+		Message:  ptr.String("ChangeBatch errors occurred"),
+		Messages: ptr.StringSlice(err.Messages),
+	}
```
jasdel, line 92: That's interesting, is `InvalidChangeBatch` a modeled shape? its error format just isn't valid? Convenient that it exists but odd.

skotambkar: Yes it is a modeled exception shape. The error format used to return error response over the wire isn't consistent with RestXML standards.

service/route53/internal/customizations/doc.go
```diff
@@ -1,41 +1,39 @@
...
-
-In case the returned error response has `InvalidChangeBatch` format, the error is
-deserialized and returned. The operation deserializer does not attempt to deserialize
-as an error is returned by the process response error middleware.
-
-*/
+// Package customizations provides customizations for the Amazon Route53 API client.
+//
+// This package provides support for following customizations
+//  Process Response Middleware: used for custom error deserializing
```
jasdel, line 4: nit, one extra space :)

jasdel (approved):

PR #793, opened 2020-10-03
`s3`: fix bug for s3 200 error customization

skotambkar:
1. Corrects behavior for success response for s3 operations that support 200 error. 
2. Fixes typos, and adds HostID in S3 response error string.


jasdel (approved):
Good find! we should look into porting the v1 integration tests over for v2. Especially once the `s3manager` is brought back.

PR #799, opened 2020-10-06
codegen: Add support for presigned URL API customizations

jasdel:
Adds support for presigned URL customizations for API clients.

* [x] Generates `SourceRegion` and `DestinationRegion` input parameters for operations with presigned URL customization.
* [x] Skips serializing the struct members with custom `NoSerializeTrait`.
  * Depends on https://github.com/awslabs/smithy-go/pull/201
* [x] Static presign URL customization middleware
* [x] Generate Presign URL api client
  * Depends on https://github.com/awslabs/smithy-go/pull/205
* [x] Generate accessor funcs for `PresignedURL`, `SourceRegion` and `DestinationRegion` members.
* [x] Generate customize middleware for PresignedURL operations.
* [x] Testing of presign middleware and generated presign clients.


aws/protocol/query/middleware.go
```diff
@@ -0,0 +1,63 @@
+package query
```
skmcgrail, line 1: What was the thought around putting this middleware in the query protocol package?

jasdel: Primarily didn't have another place to put it, and the behavior was specific to the `query` protocol.

service/internal/presigned-url/middleware_test.go
```diff
@@ -0,0 +1,14 @@
+package presignedurl
+
+import "testing"
+
+func TestPresignMiddleware(t *testing.T) {
```
skmcgrail, line 5: Probably want a GitHub issue to track this.

skmcgrail (approved):

aws/signer/v4/presign_middleware.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	credentials, err := s.credentialsProvider.Retrieve(ctx)
+	if err != nil {
+		return out, metadata, &SigningError{
+			Err: fmt.Errorf("failed to retrieve credentials: %w", err),
+		}
+	}
+
+	u, h, err := s.presigner.PresignHTTP(ctx, credentials,
```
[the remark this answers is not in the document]
jasdel: Good point didn't include that. yeah this should have a similar check as the regular signer to do nothing if not credentials are available.

service/internal/presigned-url/middleware_test.go
jasdel: Thanks! that was completely missed. I'll fix this up...

PR #801, opened 2020-10-07
Fix empty response body deser in case of error response

skotambkar:
Fixes xml deser util to correctly handle empty response body in case of an error response.

Fixes ec2query and s3shared xml deser utils too.


jasdel (commented):
What is the SDK's behavior if there is a modeled response with response document/payload bindings, but the HTTP response body is empty?  This SDK should not have an error if this occurs.

skotambkar:
The SDK should handle that correctly. I will add test cases for success response case.

jasdel (approved):

PR #806, opened 2020-10-09
generated s3, s3control client which support checksumrequired trait

skotambkar:
Regenerated s3, s3control client supporting md5 header


jasdel (approved):

skotambkar:
Replaced by #812

PR #818, opened 2020-10-14
service/s3: Derive error code, message from HTTP status code for no body/header responses.

skotambkar:
* Updates s3 xml error util to derive error code, message from HTTP status code if none is found in error response.
* Adds customization integration for S3 HeadObject, that removes all error shapes associated with that operation.

### DIFF
--- a/service/internal/s3shared/xml_utils_test.go
+++ b/service/internal/s3shared/xml_utils_test.go
@@ -1,54 +1,56 @@
 package s3shared
 
 import (
-	"bytes"
-	"io"
 	"strings"
 	"testing"
 )
 
 func TestGetResponseErrorCode(t *testing.T) {
+	const xmlErrorResponse = `<Error>
+    <Type>Sender</Type>
+    <Code>InvalidGreeting</Code>
+    <Message>Hi</Message>
+    <HostId>bar-id</HostId>
+    <RequestId>foo-id</RequestId>
+</Error>`
+
 	cases := map[string]struct {
-		isS3Service            bool
-		status                 int
-		errorResponse          io.Reader
+		getErr                 func() (ErrorComponents, error)
 		expectedErrorCode      string
 		expectedErrorMessage   string
 		expectedErrorRequestID string
 		expectedErrorHostID    string
 	}{
 		"standard xml error": {
-			isS3Service: true,
-			status:      400,
-			errorResponse: bytes.NewReader([]byte(`<Error>
-    <Type>Sender</Type>
-    <Code>InvalidGreeting</Code>
-    <Message>Hi</Message>
-    <HostId>bar-id</HostId>
-    <RequestId>foo-id</RequestId>
-</Error>`)),
+			getErr: func() (ErrorComponents, error) {
+				errResp := strings.NewReader(xmlErrorResponse)
+				return GetErrorResponseComponents(errResp)
+			},
 			expectedErrorCode:      "InvalidGreeting",
 			expectedErrorMessage:   "Hi",
 			expectedErrorRequestID: "foo-id",
 			expectedErrorHostID:    "bar-id",
 		},
+
 		"s3 no response body": {
-			isS3Service:          true,
-			status:               400,
-			errorResponse:        bytes.NewReader([]byte(``)),
+			getErr: func() (ErrorComponents, error) {
+				errResp := strings.NewReader("")
+				return GetS3ErrorResponseComponents(errResp, 400)
+			},
 			expectedErrorCode:    "BadRequest",
 			expectedErrorMessage: "Bad Request",
 		},
 		"s3control no response body": {
-			isS3Service:   false,
-			status:        400,
-			errorResponse: bytes.NewReader([]byte(``)),
+			getErr: func() (ErrorComponents, error) {
+				errResp := strings.NewReader("")
+				return GetErrorResponseComponents(errResp)
+			},
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
```diff
@@ -81,6 +81,6 @@ protected static GoDependency module(
     }
 
     private static final class Versions {
-        private static final String AWS_SDK = "v0.0.0-20201006075021-8b185f9d6dff";
+        private static final String AWS_SDK = "v0.0.0-20201014161145-592e24615bd4";
```
jasdel, line 84: Should this be using the pseudo versions, `v0.26.1-0.20201014161145-592e24615bd4` ?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsCustomGoDependency.java
```diff
@@ -45,7 +45,7 @@ private AwsCustomGoDependency() {
     }
 
     private static final class Versions {
-        private static final String INTERNAL_S3SHARED = "v0.1.0";
+        private static final String INTERNAL_S3SHARED = "v0.0.0-20201014161145-592e24615bd4";
```
jasdel, line 48: nit probably should be using the pseudo version `v0.1.1-0.`... Those seem to be the most compatible with existing modules.

codegen/smithy-aws-go-codegen/src/main/resources/META-INF/services/software.amazon.smithy.go.codegen.integration.GoIntegration
```diff
@@ -24,3 +24,4 @@ software.amazon.smithy.aws.go.codegen.customization.S3AcceptEncodingGzip
 software.amazon.smithy.aws.go.codegen.customization.KinesisCustomizations
 software.amazon.smithy.aws.go.codegen.customization.S3ErrorWith200Status
 software.amazon.smithy.aws.go.codegen.customization.Route53Customizations
+software.amazon.smithy.aws.go.codegen.customization.S3HeadObjectCustomizations
```
jasdel, line 27: would be nice to  group same service customizations together.

service/s3control/deserializers.go
```diff
@@ -89,7 +89,7 @@ func awsRestxml_deserializeOpErrorCreateAccessPoint(response *smithyhttp.Respons
 	errorCode := "UnknownError"
 	errorMessage := errorCode
 
-	errorComponents, err := s3shared.GetErrorResponseComponents(errorBody)
+	errorComponents, err := s3shared.GetErrorResponseComponents(errorBody, response.StatusCode)
```
jasdel, line 92: Should s3 control have this customization behavior to, or just S3?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
skotambkar: Seems like the pseudo version was only required for `Smithy-go` dependency.

service/s3control/deserializers.go
skotambkar: just s3

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XMLProtocolUtils.java
```diff
@@ -275,8 +275,13 @@ public static void writeXmlErrorMessageCodeDeserializer(ProtocolGenerator.Genera
 
         if (requiresS3Customization(service)) {
             writer.addUseImports(AwsCustomGoDependency.S3_SHARED_CUSTOMIZATION);
-            writer.write("errorComponents, err := s3shared.GetErrorResponseComponents(errorBody)");
+            String getErrorComponents = String.format("errorComponents, err := "
+                            + "s3shared.GetErrorResponseComponents(errorBody, response.StatusCode, %s)",
+                    isS3Service(service));
+            writer.write(getErrorComponents);
```
jasdel, line 281: This might be a good case where having two different functions would be better than an additional `bool` parameter. `s3control` can use the "generic" version of this function, where `s3` could use the one that wraps the "generic" and adds the additional behavior.

This would result in a if/else here not to bad, but have the benefit of the error functions be specific to their use cases. Individual functions also lets us more finely modify the use cases separately too.

service/internal/s3shared/xml_utils.go
```diff
@@ -16,10 +18,19 @@ type ErrorComponents struct {
 }
 
 // GetErrorResponseComponents returns the error fields from an xml error response body
-func GetErrorResponseComponents(r io.Reader) (ErrorComponents, error) {
+func GetErrorResponseComponents(r io.Reader, statusCode int, isS3service bool) (ErrorComponents, error) {
```
jasdel, line 21: Does it make sense to split these into two different functions? With the S3 version wrapping the "generic" one? That would remove the need for the boolean, and both client's being updated. S3 could have its own special version of this function

jasdel (approved):

PR #821, opened 2020-10-14
codegen: Regenerate APIs

jasdel:
Regenerates SDK API clients based on codegen change https://github.com/awslabs/smithy-go/pull/216


JordonPhillips (commented):
Looks good, though you should probably rebase and re-run after the dependant PR is merged. There have been other doc fixes introduced

skmcgrail (approved):

Issue #832, opened 2020-10-19
unexpected panics instead of errors in >=v0.25.0


jcarter3:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
Getting a panic when listing backup vault access policies.

```
panic: interface conversion: json.Token is nil, not string

goroutine 15 [running]:
github.com/aws/aws-sdk-go-v2/aws/protocol/restjson.GetErrorInfo(0xc002a54160, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jcarter/go/pkg/mod/github.com/aws/aws-sdk-go-v2@v0.27.0/aws/protocol/restjson/decoder_util.go:53 +0x557
github.com/aws/aws-sdk-go-v2/service/backup.awsRestjson1_deserializeOpErrorGetBackupVaultAccessPolicy(0xc000697b70, 0xc0006cf0d0, 0xc005206330, 0x6291200)
	/home/jcarter/go/pkg/mod/github.com/aws/aws-sdk-go-v2/service/backup@v0.27.0/deserializers.go:3855 +0x3e8
github.com/aws/aws-sdk-go-v2/service/backup.(*awsRestjson1_deserializeOpGetBackupVaultAccessPolicy).HandleDeserialize(0x8693380, 0x6d917a0, 0xc005206330, 0x6291200, 0xc0051c3e40, 0x6bc12e0, 0xc002d80040, 0x0, 0xc00321dc00, 0x40, ...)
	/home/jcarter/go/pkg/mod/github.com/aws/aws-sdk-go-v2/service/backup@v0.27.0/deserializers.go:3798 +0x935
github.com/awslabs/smithy-go/middleware.decoratedDeserializeHandler.HandleDeserialize(...)
	/home/jcarter/go/pkg/mod/github.com/awslabs/smithy-go@v0.2.0/middleware/step_deserialize.go:201
[...]

PR #837, opened 2020-10-20
`rest-json`: updates rest-json error code retriever util

skotambkar:
* Updates error code getter util to use json.Decoder's decode method directly. 
* Adds tests

Fixes https://github.com/aws/aws-sdk-go-v2/issues/832

### DIFF
--- a/aws/protocol/restjson/decoder_util.go
+++ b/aws/protocol/restjson/decoder_util.go
@@ -14,9 +14,9 @@ import (
 // identify the error code, msg in a REST JSON error response.
 func GetErrorInfo(decoder *json.Decoder) (errorType string, message string, err error) {
 	var errInfo struct {
-		Code    string `json:"code"`
-		Type  string `json:"__type"`
-		Message string `json:"message"`
+		Code    string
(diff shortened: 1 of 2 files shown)


aws/protocol/restjson/decoder_util.go
```diff
@@ -2,68 +2,48 @@ package restjson
...
 // identify the error code, msg in a REST JSON error response.
 func GetErrorInfo(decoder *json.Decoder) (errorType string, message string, err error) {
-	startToken, err := decoder.Token()
-	if err == io.EOF {
-		return "", "", nil
+	var errInfo struct {
+		Code    string `json:"code"`
+		Type  string `json:"__type"`
+		Message string `json:"message"`
 	}
```
jasdel, line 20: Are we positive that there are no `Code` (capital `C`) cases?  Might be better not to have struct tags for these members so that the decoder's caseless field matching will be used.

skotambkar: Thanks for pointing it out. Fixed.

jasdel (approved):

PR #838, opened 2020-10-20
example: add example for S3 ListObjectsV2 with paginator

jasdel:
Adds a simple example for Amazon S3 ListObjectsV2 API call with a hand written paginator. Paginator should be updated to use the SDK's generated paginators once available.

### DIFF
--- a/example/service/s3/listObjects/listObjects.go
+++ b/example/service/s3/listObjects/listObjects.go
@@ -53,13 +53,14 @@ func main() {
 	if len(objectDelimiter) != 0 {
 		params.Delimiter = &objectDelimiter
 	}
-	if v := int32(maxKeys); v != 0 {
-		params.MaxKeys = &v
-	}
 
 	// TODO replace this with the code generate paginator when available
 	// s3.NewListObjectsV2Paginator()
-	p := NewS3ListObjectsV2Paginator(client, params)
+	p := NewS3ListObjectsV2Paginator(client, params, func(o *S3ListObjectsV2PaginatorOptions) {
+		if v := int32(maxKeys); v != 0 {
+			o.Limit = v
+		}
+	})
 
 	// Iterate through the S3 object pages, printing each object returned.
 	var i int
@@ -87,9 +88,18 @@ type S3ListObjectsV2APIClient interface {
 	ListObjectsV2(context.Context, *s3.ListObjectsV2Input, ...func(*s3.Options)) (*s3.ListObjectsV2Output, error)
 }
 
+// S3ListObjectsV2PaginatorOptions provides the options for configuring the
+// ListObjectsV2Paginator.
+type S3ListObjectsV2PaginatorOptions struct {
+	// The maximum number of keys to return per page.
+	Limit int32
+}
+
 // S3ListObjectsV2Paginator provides the paginator to paginate S3 ListObjectsV2
 // response pages.
 type S3ListObjectsV2Paginator struct {
+	options S3ListObjectsV2PaginatorOptions
+
 	client S3ListObjectsV2APIClient
 	params s3.ListObjectsV2Input
 
@@ -99,8 +109,13 @@ type S3ListObjectsV2Paginator struct {
 
 // NewS3ListObjectsV2Paginator initializes a new S3 ListObjectsV2 Paginator for
 // paginating the ListObjectsV2 respones.
-func NewS3ListObjectsV2Paginator(client S3ListObjectsV2APIClient, params *s3.ListObjectsV2Input) *S3ListObjectsV2Paginator {
+func NewS3ListObjectsV2Paginator(client S3ListObjectsV2APIClient, params *s3.ListObjectsV2Input, optFns ...func(*S3ListObjectsV2PaginatorOptions)) *S3ListObjectsV2Paginator {
+	var options S3ListObjectsV2PaginatorOptions
+	for _, fn := range optFns {
+		fn(&options)
+	}
 	p := &S3ListObjectsV2Paginator{
+		options:   options,
 		client:    client,
 		firstPage: true,
 	}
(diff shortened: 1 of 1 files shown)


example/service/s3/listObjects/README.md
```diff
@@ -0,0 +1,35 @@
+# Amazon S3 ListObjectsV2 Example
+
+This is an example using the AWS SDK for Go to list objects' key in a S3 bucket.
```
skotambkar, line 3: Nit: is `'` intended after list objects

example/service/s3/listObjects/listObjects.go
```diff
@@ -0,0 +1,129 @@
...
+) {
+	if !p.HasMorePages() {
+		return nil, fmt.Errorf("no more pages available")
+	}
+
+	params := *p.params
+	result, err := p.client.ListObjectsV2(ctx, &params)
+	if err != nil {
+		return nil, err
+	}
```
skotambkar, line 123: You don't have to take the value of `p.params` here, as you are sending the `&params` in the next step

If the intention is to use a copy of input; would it make sense for Paginator struct to not take a input value (rather than a pointer).

example/service/s3/listObjects/listObjects.go
```diff
@@ -0,0 +1,129 @@
...
+	// s3.NewListObjectsV2Paginator()
+	p := NewS3ListObjectsV2Paginator(client, params)
+
+	// Iterate through the S3 object pages, printing each object returned.
+	var i int
+	log.Println("Objects:")
+	for p.HasMorePages() {
+		i++
+
+		page, err := p.NextPage(context.TODO())
```
skotambkar, line 67: Nit: Can you add a comment on taking a new context each time, this avoids timeouts between page retrieval and seems good to call out.

example/service/s3/listObjects/listObjects.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			log.Fatalf("failed to get page %v, %v", i, err)
+		}
+
+		// Log the objects found
+		for _, obj := range page.Contents {
+			fmt.Println("Object:", *obj.Key)
+		}
```
skotambkar, line 75: Nit: using `println` would probably result in a different formatted log than mentioned in the Readme example. 

Not important, but if this was a runnable example it probably fails?

skotambkar (commented):
Minor feedback/questions/comments

example/service/s3/listObjects/listObjects.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	params := *p.params
+	result, err := p.client.ListObjectsV2(ctx, &params)
+	if err != nil {
+		return nil, err
+	}
+
+	p.firstPage = false
+	p.nextToken = result.NextContinuationToken
```
skmcgrail, line 126: minor: Technically we should check `IsTruncated` first, and if that is true then set the `nextToken` with the value from `NextContinuationToken`

example/service/s3/listObjects/README.md
jasdel: nope that would be a typo

example/service/s3/listObjects/listObjects.go
jasdel: Thanks missed that, will update the example forgot about the `IsTruncate` alternative flow.

example/service/s3/listObjects/listObjects.go
jasdel: The intention is to create a copy of `p.params` that will not mutate the original.

Taking a value for params at input instead of a pointer could be helpful, though it is asymmetric to the other API calls. I don't think we've discussed this yet with the design of the Paginators.

One benefit of taking a pointer as input, is that for operations that have no required members you can pass in `nil` instead. If values were taken here that would be another asymmetric usage of an input vs API.

skotambkar (approved):

example/service/s3/listObjects/listObjects.go
```diff
@@ -0,0 +1,129 @@
...
+	client S3ListObjectsV2APIClient
+	params *s3.ListObjectsV2Input
+
+	nextToken *string
+	firstPage bool
+}
+
+// NewS3ListObjectsV2Paginator initializes a new S3 ListObjectsV2 Paginator for
+// paginating the ListObjectsV2 respones.
+func NewS3ListObjectsV2Paginator(client S3ListObjectsV2APIClient, params *s3.ListObjectsV2Input) *S3ListObjectsV2Paginator {
```
JordonPhillips, line 102: This should probably have a page size parameter

jasdel: Page size is currently a parameter of the `s3.ListObjectsV2Input` type. I'm not sure if this should be duplicated here or not since it is an optional parameter. I'd think if the user wants to specify this value with a custom size then it should be done on the input value.

What are your thoughts on this?

JordonPhillips: The reason it's an attribute on the paginator trait is so that clients can expose a consistent interface for providing that value. Between operations and services the member could be any number of things, like `Limit`, `MaxItems`, `PageSize`, etc.

jasdel: yup though its still an optional parameter in all APIs. By putting it as a parameter on the constructor the SDK is paginator is effectively making it required. The handling for required, but not required parameter, would be to treat the limit parameter as unset if its value is 0.  Minor, but this approach still requires the user to pass in a `0` for the default behavior.

JordonPhillips: Well you can always make a pointer so people pass it in as nil, or have a function to set it as a second step that takes a non-pointer

jasdel: Adding functional options to add configuration to the paginator might be a good way to do this.

```go
type ListObjectsV2PaginatorOptions struct {
     Limit int32 // generic name for all paginators
}

func NewListObjectsV2Paginator(client ListObjectsV2APIClient, params *s3.ListObjectsV2Input,
     optFns ...func(*ListObjectsV2PaginatorOptions)
```

```go
p := s3.NewListObjectsV2Paginator(client, params, func(o *ListObjectsV2PaginatorOptions) {
      o.Limit = 10
})
```

jasdel: updated to include this pattern 👍

JordonPhillips (approved):

PR #846, opened 2020-10-21
route53 : fix sanitizeURL customization to handle leading `/`

### LINKED ISSUE #843: route53 ListResourceRecordSetsInput doesn't work when HostedZoneId contains "/hostedzone/"
- [ x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [ x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x ] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
Using `/hostedzone/xxxxx` in HostedZoneId field in ListResourceRecordSetsInput results in unknown error from AWS API.
I know there is this [sanitizeHostedZoneIDInput](https://github.com/aws/aws-sdk-go-v2/blob/1049b73d5c17334b9dbcc675b565d91cf201eff6/service/route53/api_client.go#L206) function in the code base but somehow I had to trim `/hostedzone/` prefix to avoid the error.

v0.27.0

go version go1.15.3 darwin/amd64

```
                awscfg, err := config.LoadDefaultConfig()
                svc := route53.NewFromConfig(awscfg)
                zoneID := "/hostedzone/xxxxxxxxxxx"
		result, err := svc.ListResourceRecordSets(
			context.Background(),
			&route53.ListResourceRecordSetsInput{
				HostedZoneId:    aws.String(zoneID),
				StartRecordName: token,
			})
```
This resulted in error:
```
2020/10/20 15:32:04 failed because of an API error, Code: UnknownError, Message: UnknownError
panic: operation error Route 53: ListResourceRecordSets, https response error StatusCode: 400, RequestID: , api error UnknownError: UnknownError
```
It works if I trim the prefix like `aws.String(strings.ReplaceAll(zoneID, "/hostedzone/", ""))` 

**Expected behavior**
Given ListHostedZones() now returns hostedZoneID in the format of `/hostedzone/xxxxx`, the module should parse it correctly. 

**Additional context**

skotambkar:
This PR updates route53 customization that sanitizes URL to take handle a leading `/` in `hostedzone`

Fixes #843

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsGoDependency.java
@@ -82,6 +82,6 @@ public class AwsGoDependency {
     }
 
     private static final class Versions {
-        private static final String AWS_SDK = "v0.27.1-0.20201021184031-883c13b4b7a8";
+        private static final String AWS_SDK = "v0.27.1-0.20201021211102-ffda7ace423e";
     }
 }
--- a/service/route53/go.mod
+++ b/service/route53/go.mod
@@ -3,7 +3,7 @@ module github.com/aws/aws-sdk-go-v2/service/route53
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2 v0.27.1-0.20201021184031-883c13b4b7a8
+	github.com/aws/aws-sdk-go-v2 v0.27.1-0.20201021211102-ffda7ace423e
 	github.com/awslabs/smithy-go v0.2.0
 )
 


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/Route53Customizations.java
```diff
@@ -110,7 +110,7 @@ private void writeHostedZoneIDInputSanitizer(
                             writer.addUseImports(SmithyGoDependency.STRINGS);
                             for (MemberShape member : hostedZoneIDMembers) {
                                writer.openBlock("if i.$L != nil {", "}", member.getMemberName(), () -> {
-                                writer.write("values := strings.SplitN(*i.$L, \"/\", 2)", member.getMemberName());
+                                writer.write("values := strings.Split(*i.$L, \"/\")", member.getMemberName());
```
jasdel, line 113: We can simplify this and use [strings.LastIndex](https://pkg.go.dev/strings#LastIndex), and either use the value return or if `-1` fallback to `0`. The the write becomes`v := *i.$L[idx:]`. Removes the need for the slice of strings created.

```go
idx := strings.LastIndex(*i.$L, `/`) // using ` for string literals removes need for escaping in codegen too
if idx == -1 { idx = 0 }
v := *i.$L[idx:]
i.$L = &v
```

service/route53/api_client.go
```diff
@@ -205,188 +205,188 @@ func addSanitizeURLMiddleware(stack *middleware.Stack) {
 // another operation (e.g. it expects just '1234')
 func sanitizeHostedZoneIDInput(input interface{}) error {
 	switch i := input.(type) {
-	case *ChangeResourceRecordSetsInput:
+	case *AssociateVPCWithHostedZoneInput:
 		if i.HostedZoneId != nil {
-			values := strings.SplitN(*i.HostedZoneId, "/", 2)
+			values := strings.Split(*i.HostedZoneId, "/")
```
jasdel, line 210: General feedback not specific to this PR. We should consider generating individual functions per operation input instead of a large switch statement. Having a large switch statement on types like these probably isn't super cheap, and static functions per operation wouldn't add significant source code.

skotambkar: would take up a separate task to do this, as similar refactor would be needed in S3. 

Created https://github.com/aws/aws-sdk-go-v2/issues/847

jasdel (approved):

Issue #827, opened 2020-10-15
Making API operation call without configured AWS region results in odd error.


jasdel:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
Making an API operation call with a client that does not have a `Region` value configured results in an error because DNS lookup fails. The SDK should require the `Region` configuration, and return an error when the endpoint resolver does not receive an Region.

**Version of AWS SDK for Go?**
`v0.26.0`

**Version of Go (`go version`)?**
`go 1.15.2`

**To Reproduce (observed behavior)**
Steps to reproduce the behavior (please share code or minimal repo)

**Expected behavior**
Call API operation with a client that is not configured for a region.

**Additional context**

> 2020/10/15 16:00:20 operation error DynamoDB: Scan, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequsetID: , request send failed, Post "https://dynamodb..amazonaws.com/": dial tcp: lookup dynamodb..amazonaws.com: no such host

Issue #364, opened 2019-08-24
DisableHostPrefix doesn't change s3 client behavior


eriktate:
Related to #328 

### Version of AWS SDK for Go?
v0.10.0

### Version of Go (`go version`)?
go version go1.13rc1 linux/amd64

### What issue did you see?
When setting `DisableHostPrefix` to `true` in the `aws.Config` struct, the s3 client does not set `ForcePathStyle` to `true`. This causes problems when generating a config to target a local endpoint (localstack) for integration testing purposes.

If I am misunderstanding the purpose of the `DisableHostPrefix` flag, please let me know!

### Steps to reproduce
1. Create an `aws.Config` with `DisableHostPrefix` set to `true`.
2. Create a `s3.Client` with that config.
3. Check the `ForcePathStyle` flag on the `s3.Client`.

Issue #328, opened 2019-06-25
SDK's behavior for custom endpoint URL is not consistent with S3


jasdel:
The SDK's S3 API client ignores the fact a customer provided a custom endpoint URL when configuring the client, and will always prefix the bucket name to the URL's host unless the user also specifies the `S3ForcePathStyle` config option. This behavior isn't obvious, and will be surprising to users. 

the V2 SDK should consider changing/clarifying this behavior so that the behavior is more expected. The V2 SDK will also need to consider how it will handle future API operation where path style addressing may not be valid.

PR #836, opened 2020-10-20
Adds support for Accelerate and Dualstack for s3, s3control client

skotambkar:
#### S3:
* Adds support for accelerate and dualstack

#### S3Control:
* Adds support for dualstack

PR #848, opened 2020-10-21
codegen: Add flag to direct if SDK is allowed to modify a custom endpoint or not

jasdel:
Adds a new member to the `aws.Endpoint` struct, `HostnameImmutable`. This member directs if the SDK is allowed to modify the resolved endpoint to meet the requirements of the API.  If `HostnameImmutable` is true, the SDK will not attempt to modify the hostname via any customizations, or prefix behavior. Flag defaults to false.

- Fixes https://github.com/aws/aws-sdk-go-v2/issues/827 (SDK requiring region for endpoint resolution)
- Fixes https://github.com/aws/aws-sdk-go-v2/issues/328
- Fixes https://github.com/aws/aws-sdk-go-v2/issues/364
- Related to https://github.com/aws/aws-sdk-go-v2/pull/836 When Amazon S3 Accesspoint support is implemented.
- Related to https://github.com/aws/aws-sdk-go-v2/issues/479

Depends on https://github.com/awslabs/smithy-go/pull/222

### DIFF
--- a/example/service/s3/listObjects/go.mod
+++ b/example/service/s3/listObjects/go.mod
@@ -4,7 +4,7 @@ go 1.15
 
 require (
 	github.com/aws/aws-sdk-go-v2/config v0.2.1
-	github.com/aws/aws-sdk-go-v2/service/s3 v0.28.0
+	github.com/aws/aws-sdk-go-v2/service/s3 v0.28.1-0.20201027190912-8f6733991e52
 )
 
 replace github.com/aws/aws-sdk-go-v2/config => ../../../../config/
--- a/feature/s3/manager/go.mod
+++ b/feature/s3/manager/go.mod
@@ -5,7 +5,7 @@ go 1.15
 require (
 	github.com/aws/aws-sdk-go-v2 v0.28.1-0.20201027184009-8eb8fc303e7c
 	github.com/aws/aws-sdk-go-v2/config v0.2.1
-	github.com/aws/aws-sdk-go-v2/service/s3 v0.28.0
+	github.com/aws/aws-sdk-go-v2/service/s3 v0.28.1-0.20201027190912-8f6733991e52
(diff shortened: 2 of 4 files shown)


jasdel:
Fixes up a bug with the S3 endpoint resolve tests for virtual host path style that should resolve to path style for bucket names that are not valid hostname labels. This fallback is not used when accelerate is enabled.

Applies a couple fixes to s3 and s3control endpoint resolve tests to cover hostname immutable support.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/EndpointGenerator.java
```diff
@@ -387,6 +396,12 @@ private void generateInternalResolverImplementation(GoWriter writer) {
         writer.write("");
         writer.writeDocs("ResolveEndpoint resolves the service endpoint for the given region and options");
         writeInternalResolveEndpointImplementation(writer, resolverImplSymbol, "r", () -> {
+            // Currently all APIs require a region to derive the endpoint for that API. If there are ever a truly
+            // region-less API then this should be gated at codegen.
+            writer.addUseImports(AwsGoDependency.AWS_CORE);
```
skmcgrail, line 401: Any reason for doing that at this level rather then https://github.com/aws/aws-sdk-go-v2/blob/master/internal/endpoints/endpoints.go ?

jasdel: Primary reason to do it per service so that it would be easier for a service to opt out of this behavior if ever needed. With the idea that by default all API clients require a region.

skmcgrail (approved):

skotambkar (approved):

PR #862, opened 2020-10-29
fix invalid signature generation caused by unawareness of host prefix

moriyoshi:
This patch modifies autogenerated code too and I don't think it's appropriate to merge it as a whole, so please take it as just a reference.

## Background of the problem

Because V4 signer doesn't take it into account that the host prefix is propagated outside the regular `net/http` Request, it fails to build the valid signature string for the services that require a host prefix.

## See also

* https://github.com/awslabs/smithy-go/issues/230
  This addresses the additional problem regarding `Clone()`.

### DIFF
--- a/aws/signer/v4/middleware.go
+++ b/aws/signer/v4/middleware.go
@@ -219,7 +219,7 @@ func (s *SignHTTPRequestMiddleware) HandleFinalize(ctx context.Context, in middl
 		return out, metadata, &SigningError{Err: fmt.Errorf("failed to retrieve credentials: %w", err)}
 	}
 
-	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime())
+	err = s.signer.SignHTTP(ctx, credentials, req, payloadHash, signingName, signingRegion, sdk.NowTime())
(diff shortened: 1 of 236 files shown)


jasdel:
Thanks for reporting this issue @moriyoshi I created #863 to track this bug.

The request signer intentionally took a standard library `http.Request` so that it could be more generic. I think an alternative approach that should also be considered is to drop the `HostPrefix` member of smithyhttp.Request. and instead add the endpoint prefix similar to the SDK's Amazon S3 customization for bucket name and accesspoints.

moriyoshi:
That sounds like a cleaner approach. Where do you think the final host name would be resolved then? If the original intention of HostPrefix is to delay the resolution so it'll be done right before the actual request is made, inserting a service-specific  middleware that determines and sets the canonical host name to the request passed before the signer should do. However this might result in bloated code to generate.

jasdel:
Thanks for taking the time to report this issue and create this PR @moriyoshi. We've updated the SDK's code generation change in https://github.com/awslabs/smithy-go/pull/233 and relevant Amazon S3 Control API client change in https://github.com/aws/aws-sdk-go-v2/pull/876. Lets use those PRs for tracking this issue.

PR #866, opened 2020-10-29
Strip streaming operations

JordonPhillips:
This adds an integration that strips out streaming operations so that broken operations won't be generated. There were some upstream changes in smithy-go that impacted the generated code, so the code gen was run first without these changes and then a second time with. Each of these runs is in a different commit so that the impact of the change can be easily seen.


jasdel (approved):

jasdel:
Looks like protocol tests need updated for this change, looks like they are referencing the wrong name.

```
# github.com/aws/aws-sdk-go-v2/service/internal/benchmark/lexruntimeservice

vet: lexruntimeservice/put_session_test.go:119:28: MessageFormatTypePlain_text not declared by package types
```

PR #868, opened 2020-10-31
update Aws Go Dependency to 0.29.0 for codegen

skotambkar:
master branch - `AwsGoDependency` wasn't updated to `0.29.0`; this causes codegen to revert back to pseudo v0.28.x-xx versions.


jasdel:
Should the customizations module references be updated as well?

skotambkar:
No changes were introduced for s3shared module for release v0.29.0 . Thus, they don't need to be updated. 

If changes are present for those modules, then they should be updated to the latest tag.

jasdel (approved):

PR #870, opened 2020-11-02
Adds support for access-point and outposts for s3 / s3control client

skotambkar:
[x] Adds customizations for s3 to support accesspoint and outposts accesspoint arn formats
[x] Adds customizations for s3control to support outposts accesspoint and outposts bucket arn formats

- Depends on awslabs/smithy-go#223

### DIFF
--- a/service/s3/api_op_AbortMultipartUpload.go
+++ b/service/s3/api_op_AbortMultipartUpload.go
@@ -209,10 +209,10 @@ func addAbortMultipartUploadUpdateEndpoint(stack *middleware.Stack, options Opti
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getAbortMultipartUploadBucketMember,
-			SupportsAccelerate: true,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      true,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_CompleteMultipartUpload.go
+++ b/service/s3/api_op_CompleteMultipartUpload.go
@@ -300,10 +300,10 @@ func addCompleteMultipartUploadUpdateEndpoint(stack *middleware.Stack, options O
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getCompleteMultipartUploadBucketMember,
-			SupportsAccelerate: true,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      true,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_CopyObject.go
+++ b/service/s3/api_op_CopyObject.go
@@ -522,10 +522,10 @@ func addCopyObjectUpdateEndpoint(stack *middleware.Stack, options Options) error
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getCopyObjectBucketMember,
-			SupportsAccelerate: true,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      true,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_CreateBucket.go
+++ b/service/s3/api_op_CreateBucket.go
@@ -253,10 +253,10 @@ func addCreateBucketUpdateEndpoint(stack *middleware.Stack, options Options) err
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getCreateBucketBucketMember,
-			SupportsAccelerate: false,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      false,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_CreateMultipartUpload.go
+++ b/service/s3/api_op_CreateMultipartUpload.go
@@ -549,10 +549,10 @@ func addCreateMultipartUploadUpdateEndpoint(stack *middleware.Stack, options Opt
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getCreateMultipartUploadBucketMember,
-			SupportsAccelerate: true,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      true,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_DeleteBucket.go
+++ b/service/s3/api_op_DeleteBucket.go
@@ -145,10 +145,10 @@ func addDeleteBucketUpdateEndpoint(stack *middleware.Stack, options Options) err
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getDeleteBucketBucketMember,
-			SupportsAccelerate: false,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      false,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
--- a/service/s3/api_op_DeleteBucketAnalyticsConfiguration.go
+++ b/service/s3/api_op_DeleteBucketAnalyticsConfiguration.go
@@ -165,10 +165,10 @@ func addDeleteBucketAnalyticsConfigurationUpdateEndpoint(stack *middleware.Stack
 	return s3cust.UpdateEndpoint(stack, s3cust.UpdateEndpointOptions{
 		Accessor: s3cust.UpdateEndpointParameterAccessor{
 			GetBucketFromInput: getDeleteBucketAnalyticsConfigurationBucketMember,
-			SupportsAccelerate: true,
 		},
 		UsePathStyle:            options.UsePathStyle,
 		UseAccelerate:           options.UseAccelerate,
+		SupportsAccelerate:      true,
 		EndpointResolver:        options.EndpointResolver,
 		EndpointResolverOptions: options.EndpointOptions,
 		UseDualstack:            options.UseDualstack,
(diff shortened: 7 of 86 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -63,6 +65,23 @@
...
+    // List of operations that use Accesspoint field as ARN input source.
+    private  static final Set<String> LIST_ACCESSPOINT_ARN_INPUT = SetUtils.of(
+            "GetAccessPoint", "DeleteAccessPoint", "PutAccessPointPolicy",
+            "GetAccessPointPolicy", "DeleteAccessPointPolicy"
+    );
+
+    // List of operations that use OutpostID to resolve endpoint
+    private  static final Set<String> LIST_OUTPOST_ID_INPUT = SetUtils.of(
+      "CreateBucket","ListRegionalBuckets"
+    );
```
skmcgrail, line 83: nit: formatting is inconsistent

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -102,6 +134,175 @@ public void writeAdditionalFiles(
...
+     *       return iv.backfillAccountID(v)
+     *  }
+     *
+     */
+    private void writeBackfillAccountIDHelper(GoWriter writer, Model model,SymbolProvider symbolProvider, ServiceShape service) {
+     writer.write("func $L(in interface{}, v string) (interface{},error) { ", BACKFILL_ACCOUNT_ID);
+     writer.write("iv, ok := in.(interface{ $L(string) (interface{}, error) })", BACKFILL_ACCOUNT_ID);
+     writer.write("if !ok  { return in, nil } // as account id is not even supported");
+     writer.write("return iv.$L(v) }", BACKFILL_ACCOUNT_ID);
+    }
```
skmcgrail, line 157: nit: formatting

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -102,6 +134,175 @@ public void writeAdditionalFiles(
...
 
+
+
+
+    /**
+     * writes BackfillAccountID Helper function for s3 api client
+     *
+     *  Generates code:
+     *  === api_client.go===
+     *  func backfillAccountID(in interface, v string) (interface{},error) {
```
skmcgrail, line 145: I believe our designs did not want to generate methods on the types themselves. I would propose we generate a single `backfillAccountID` that handles the type cast and setting the value appropriately.

```go
func backfillAccountID(i interface{}, v string) error {
    switch in := i.(type) {
    case *OperationInput:
       if in.AccountId != nil {
		if !strings.EqualFold(*in.AccountId, v) {
			return fmt.Errorf("error backfilling account id")
		}
		return nil
	}
	in.AccountId = &v
	return nil
    default:
        return fmt.Errorf("unsupported operation")
    }
}
```

Similar feedback for the other member update functions.

skmcgrail (changes requested):

service/internal/s3shared/arn_lookup.go
```diff
@@ -0,0 +1,62 @@
...
+// middleware context. This middleware must be executed before input validation step or any other
+// arn processing middleware.
+type InitARNLookupMiddleware struct {
+
+	// GetARNValue takes in a input interface and returns a ptr to string and a bool
+	GetARNValue func(interface{}) (*string, bool)
+}
+
+// ID for the middleware
+func (m *InitARNLookupMiddleware) ID() string {
```
skmcgrail, line 23: nit: Avoid naming middleware with the name middleware in the type.

skmcgrail: Additionally no need to have "Middleware" in the ID itself.

service/s3control/internal/customizations/update_endpoint.go
```diff
@@ -3,20 +3,108 @@ package customizations
...
+	if err != nil {
+		return err
+	}
+
+	// outpostID middleware
+	err = stack.Serialize.Insert(&processOutpostIDMiddleware{
+		GetOutpostID: options.Accessor.GetOutpostIDInput,
 		UseDualstack: options.UseDualstack,
-		ServiceID:    "s3-control",
+	}, (&processARNResource{}).ID(), middleware.Before)
```
skmcgrail, line 96: nit: `(*processARNResource)(nil).ID()`

service/s3control/internal/customizations/update_endpoint_test.go
```diff
@@ -140,8 +143,549 @@ func TestUpdateEndpointBuild(t *testing.T) {
...
+		// 	options: s3control.Options{
+		// 		Region:       "us-west-2",
+		// 		UseARNRegion: true,
+		// 	},
+		// 	expectedReqURL:             "https://s3-outposts.us-east-1.amazonaws.com/v20180820/bucket/mybucket",
+		// 	expectedSigningName:        "s3-outposts",
+		// 	expectedSigningRegion:      "us-east-1",
+		// 	expectedHeaderForOutpostID: "op-01234567890123456",
+		// 	expectedHeaderForAccountID: true,
+		// },
```
skmcgrail, line 317: Should these be re-enabled?

skmcgrail (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3UpdateEndpoint.java
```diff
@@ -84,170 +113,543 @@ public void writeAdditionalFiles(
...
+            writer.insertTrailingNewline();
+        }
+
+        private void writeInputGetter(
+                GoWriter writer, Model model, SymbolProvider symbolProvider, ServiceShape service
+        ) {
+            writer.writeDocs(
+                    "getBucketFromInput returns a boolean indicating if the input has a modeled bucket name, " +
+                            " and a pointer to string denoting a provided bucket member value");
+            writer.openBlock("func getBucketFromInput(input interface{}) (*string, bool) {", "}", () -> {
```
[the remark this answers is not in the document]
skotambkar: The old switch pattern is being used in other services too, I want to decouple and view it as a separate change which also addresses other service customizations that use old pattern.

service/s3control/api_client.go
```diff
@@ -237,8 +239,17 @@ func addMetadataRetrieverMiddleware(stack *middleware.Stack) error {
...
+func getOutpostIDFromInput(input interface{}) (*string, bool) {
+	switch i := input.(type) {
+	case *CreateBucketInput:
+		return i.OutpostId, true
+	case *ListRegionalBucketsInput:
+		return i.OutpostId, true
+	default:
+		return nil, false
+	}
 }
```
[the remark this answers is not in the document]
skotambkar: This is limited to two operation cases. Separating them doesn't really provide much performance benefit, but can do it for consistency

service/s3control/api_op_CreateJob.go
```diff
@@ -270,3 +271,43 @@ func newServiceMetadataMiddleware_opCreateJob(region string) *awsmiddleware.Regi
...
+	}
+	cpy := *input
+	return &cpy, nil
+}
+func getCreateJobARNMember(input interface{}) (*string, bool) {
+	return nil, false
+}
+func setCreateJobARNMember(input interface{}, v string) error {
+	return nil
+}
```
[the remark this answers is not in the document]
skotambkar: That's the downside of accessor pattern. Not generating these would require code generation to have additional checks. But my concern was more towards customization always having to check for nil for accessors, it seemed more error prone when I wrote this.

service/s3control/internal/customizations/backfill_input.go
```diff
@@ -0,0 +1,56 @@
...
+	av, ok := s3shared.GetARNResourceFromContext(ctx)
+	if !ok {
+		return next.HandleInitialize(ctx, in)
+	}
+
+	// create a copy of input, and assign it on params
+	in.Parameters, err = m.CopyInput(in.Parameters)
+	if err != nil {
+		return out, metadata, fmt.Errorf("error creating a copy of input")
+	}
```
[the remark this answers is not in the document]
skotambkar: For s3control we mutate the input further down the middleware stack if accesspoint/outpost arn. But yeah I agree visibility is lost if paired only with this middleware. (Idea was to do it here, as this is one of the most initial middlewares to be executed).

May be we should have a key on context for this too if we want to conditionally copy. Without the context key, we would end up copying an input multiple times in different stack steps.

service/s3control/internal/customizations/process_arn_resource.go
```diff
@@ -0,0 +1,398 @@
...
+		// update the arnable field with access point name
+		err = m.UpdateARNField(in.Parameters, tv.AccessPointName)
+		if err != nil {
+			return out, metadata, fmt.Errorf("error updating arnable field while serializing")
+		}
+
+		// check if request region is FIPS and ARN region usage is not allowed
+		if resourceRequest.UseFips() && !m.UseARNRegion {
+			return out, metadata, s3shared.NewInvalidARNWithFIPSError(tv, nil)
+		}
```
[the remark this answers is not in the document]
skotambkar: Fips is not supported for outposts. The `UseFIPS and !UseARNRegion` combination checks if a request is configured for a fips region and  use arn region is not set. In this case, SDK will attempt to use the fips region, which is not valid for outposts.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/EndpointGenerator.java
```diff
@@ -505,11 +548,70 @@ private void writeEndpoint(GoWriter writer, ObjectNode node) {
         });
     }
 
+    private static class ResolveConfigField extends ConfigField {
```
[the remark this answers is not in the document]
skotambkar: This is just IntelliJ's Java formatter, which reshuffled few things as per its rules.  Not sure what ordering it follows

service/internal/s3shared/endpoint_error.go
```diff
@@ -20,27 +20,27 @@ type InvalidARNError struct {
 	origErr  error
 }
 
-// Error returns the InvalidARNError
+// Code returns the invalid ARN error code
+func (e InvalidARNError) Code() string {
+	return invalidARNErrorErrCode
+}
+
```
skmcgrail, line 27: We don't want to expose this concept of a Code on non service API errors. This should only be `Error` and `Unwrap` and optionally any sort of "getter" type accessors if the error is wrapping and wanting to provide additional context.

service/s3control/internal/customizations/process_arn_resource.go
```diff
@@ -110,6 +113,15 @@ func (m *processARNResource) HandleSerialize(
...
 		}
 
+		if !s3shared.IsClonedInput(ctx) {
+			in.Parameters, err = m.CopyInput(in.Parameters)
+			if err != nil {
+				return out, metadata, fmt.Errorf("error creating a copy of input while processing arn")
+			}
+			// set copy input key on context
+			ctx = s3shared.SetClonedInputKey(ctx, true)
+		}
```
skmcgrail, line 123: nit: You could move the clone action to immediately before the switch statement. At worst if the user provided an invalid ARN resource we would have made an unnecessary copy. But it results in not needing to duplicate this between resource types.

skmcgrail (changes requested):
Just a couple minor comments otherwise looks good

service/s3control/internal/customizations/process_arn_resource.go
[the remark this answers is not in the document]
skmcgrail: Ah ok so `UseARNRegion` and `UseFIPS` act slightly differently for outposts ARNs. Cause for accesspoints I believe the logic is different in that a FIPS reuqest with UseARNRegion is not allowed.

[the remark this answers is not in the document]
skotambkar: yeah outposts has a different behavior and doesn't support fips at all.

service/s3/internal/customizations/update_endpoint.go
```diff
@@ -7,68 +7,120 @@ import (
...
+type UpdateEndpointParameterAccessor struct {
 	// functional pointer to fetch bucket name from provided input.
 	// The function is intended to take an input value, and
 	// return a string pointer to value of string, and bool if
 	// input has no bucket member.
 	GetBucketFromInput func(interface{}) (*string, bool)
 
+	// indicates if an operation supports s3 transfer acceleration.
+	SupportsAccelerate bool
+}
```
jasdel, line 38: Should `SupportsAccelerate` be a member of the `UpdateEndpointOptions`? Seems out of place in the Accessor's group.

skmcgrail (approved):

PR #872, opened 2020-11-03
AWS Client Loggers and Client Log Mode Support

skmcgrail:
Adds support for the smithy-go client logger, and introduces an AWS ClientLogMode bit group. This group is used when constructing middlewares to determine whether certain classes of logging events should be sent to the logger.


aws/logging.go
```diff
@@ -0,0 +1,68 @@
+// Code generated by aws/logging_generate.go DO NOT EDIT.
+package aws
+
+// ClientLogMode represents the logging mode of SDK clients. The client logging mode is a bit-field where
+// each bit is a flag that describes the logging behavior for one or more client components.
+// The entire 64-bit group is reserved for later expansion by the SDK.
+//
+// Example
+//  clientLogMode := aws.LogRetries | aws.LogRequest
+type ClientLogMode uint64
```
jasdel, line 10: minor nit, maybe also include example with adding a log level to an existing `clientLogMode` value.

```go
clientLogMode |= aws.LogRequest
```

aws/logging_generate.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"log"
+	"os"
+	"text/template"
+)
+
+var config = struct {
+	ModeBits []string
+}{
+	// Items should be appended only to keep bit-flag positions table
```
jasdel, line 14: nit `table` -> `stable`

aws/logging_generate.go
```diff
@@ -0,0 +1,71 @@
...
+{{ range $_, $field := .ModeBits }}
+func (m ClientLogMode) Is{{- $field }}() bool {
+	return m&{{- (symbolName $field) }} != 0
+}
+{{ end }}
+
+{{ range $_, $field := .ModeBits }}
+func (m *ClientLogMode) Clear{{- $field }}() {
+	*m &^= {{- (symbolName $field) }}
+}
```
jasdel, line 56: nit should add godocs to these methods.

aws/retry/middleware.go
```diff
@@ -30,13 +31,19 @@ type retryMetadataKey struct{}
 // AttemptMiddleware is a Smithy FinalizeMiddleware that handles retry attempts using the provided
 // Retryer implementation
 type AttemptMiddleware struct {
+	LogAttempts bool
```
jasdel, line 34: Nit would be helpful to add docs for this member describing what it does.

aws/retry/middleware.go
```diff
@@ -74,6 +84,10 @@ func (r AttemptMiddleware) HandleFinalize(ctx context.Context, in smithymiddle.F
 					return out, metadata, fmt.Errorf("failed to rewind transport stream for retry, %w", err)
 				}
 			}
+
+			if r.LogAttempts {
+				logger.Logf(logging.Debug, "retrying request %s/%s, attempt %d", service, operation, attemptNum)
+			}
```
jasdel, line 90: Should there also be an log for an failed attempt with information about if it will or wont be retried?

aws/retry/middleware.go
```diff
@@ -30,13 +31,19 @@ type retryMetadataKey struct{}
...
 // NewAttemptMiddleware returns a new AttemptMiddleware
-func NewAttemptMiddleware(retryer Retryer, requestCloner RequestCloner) AttemptMiddleware {
-	return AttemptMiddleware{retryer: retryer, requestCloner: requestCloner}
+func NewAttemptMiddleware(retryer Retryer, requestCloner RequestCloner, optFns ...func(*AttemptMiddleware)) AttemptMiddleware {
+	m := AttemptMiddleware{retryer: retryer, requestCloner: requestCloner}
+	for _, fn := range optFns {
+		fn(&m)
+	}
+	return m
 }
```
jasdel, line 47: Does this `AttemptMiddleware` and its constructor need to be exported? Looks like its added via the `AddRetryMiddlewares`

aws/signer/v4/v4.go
```diff
@@ -92,6 +94,12 @@ type Signer struct {
 	// http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
 	DisableURIPathEscaping bool
 
+	// The logger to send log message to
+	Logger logging.Logger
+
+	// Enable logging of signed requests
+	LogSigning bool
```
jasdel, line 101: Would be helpful to clarify the docs on what will be logged, e.g. string to sign, and canonical signature, etc

aws/signer/v4/v4.go
```diff
@@ -230,6 +238,15 @@ func buildAuthorizationHeader(credentialStr, signedHeadersStr, signingSignature
 	return parts.String()
 }
 
+func (v4 Signer) getLogger(ctx context.Context) (logger logging.Logger) {
+	if v4.Logger == nil {
+		logger = logging.Noop{}
+	} else {
+		logger = logging.WithContext(ctx, v4.Logger)
+	}
+	return logger
```
jasdel, line 247: nit style suggestion for early exit vs if/else block

```go
if v4.Logger == nil {
    return logging.Noop{}
}

return logging.WithContext(ctx, v4.Logger)
```

aws/signer/v4/v4.go
```diff
@@ -230,6 +238,15 @@ func buildAuthorizationHeader(credentialStr, signedHeadersStr, signingSignature
 	return parts.String()
 }
 
+func (v4 Signer) getLogger(ctx context.Context) (logger logging.Logger) {
+	if v4.Logger == nil {
+		logger = logging.Noop{}
```
jasdel, line 243: We need to standardize on our nomenclature for no-op. e.g. `smithyhttp.NopClient` vs `logging.Noop`.

Created https://github.com/aws/aws-sdk-go-v2/issues/878

config/provider.go
```diff
@@ -508,7 +509,102 @@ func getEndpointResolver(configs configs) (f aws.EndpointResolver, found bool, e
...
+// WithLogConfigurationErrors implements a LogConfigurationErrorsProvider and returns the wrapped boolean value.
+type WithLogConfigurationErrors bool
+
+// GetLogConfigurationErrors returns the wrapped boolean.
+func (w WithLogConfigurationErrors) GetLogConfigurationErrors() (bool, bool, error) {
+	return bool(w), true, nil
+}
+
+func getLogConfigurationErrors(configs configs) (v bool, found bool, err error) {
+	for _, c := range configs {
```
jasdel, line 598: Should this be name be more generic log than specific to errors? errors implies failures that may be blocking. whereas the config will ignore some of these cases such as profile not exists.

feature/s3/manager/download.go
```diff
@@ -166,6 +173,13 @@ func (d Downloader) Download(ctx context.Context, w io.WriterAt, input *s3.GetOb
 		option(&impl.cfg)
 	}
 
+	// Ensures we don't need nil checks later on
+	if impl.cfg.Logger == nil {
+		impl.cfg.Logger = logging.Noop{}
+	} else {
+		impl.cfg.Logger = logging.WithContext(ctx, impl.cfg.Logger)
+	}
```
jasdel, line 181: Can this use the generic `logging.WithContext`?  As in if the logger passed in is nil, return a no-op logger? Would prevent duplicating this block between s3 manager and signer.

aws/config.go
```diff
@@ -43,6 +46,13 @@ type Config struct {
 	// client requests will be handled. This is useful for adding additional
 	// tracing data to a request, or changing behavior of the SDK's client.
 	APIOptions []func(*middleware.Stack) error
+
+	// The logger writer interface to write logging messages to. Defaults to
+	// standard error.
+	Logger logging.Logger
+
+	// Configures the events that will be sent to the configured logger.
+	ClientLogMode ClientLogMode
```
jasdel, line 55: would be good to callout documentation of these and how they will be used by the client potentially.

aws/logging_generate.go
```diff
@@ -0,0 +1,71 @@
+// +build clientlogmode
```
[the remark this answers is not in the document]
skmcgrail: This would create a conflict in this package since there would be two generator files with a main function within the main package.

aws/retry/middleware.go
skmcgrail: Let's track this separately as it does not pertain to this PR.

config/provider.go
skmcgrail: Can change this to warnings, but warnings has the opposite implication as it does not communicate whether the item it warns about are skipped versus proceeded with. This is meant to communicate more environmental configuration errors like misconfigured shared config etc.

feature/s3/manager/download.go
skmcgrail: I personally find that a little strange as WithContext is meant to either assert the context interface and give the opportunity to make the logger context aware or return the original logger unmodified. I feel like this overloads the concerns for this function.

aws/retry/middleware.go
skmcgrail: I think it makes sense to log whether we won't attempt to retry if we encounter an un-retryable, but otherwise I think it's sufficient to log the retries as shown. I'm cautious to make it to verbose, and am not sure if we should print out the full error message, or just report we encountered an unretryable error.

aws/logging_generate.go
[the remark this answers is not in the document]
jasdel: Ah didn't realize that issue would occur. good to know.

jasdel (approved):

PR #873, opened 2020-11-03
Draft: support s3control using different xml error format than s3

skotambkar:
previously we assumed s3 and s3control had same xml error response format. The assumption is incorrect. 

This pr fixes the issue, to correctly parse s3control xml error which has an extra <Error> wrapper around error code and message which needs to be deserialized.


skotambkar:
Need to rebase and remove commits from accesspoint/seaport pr

skotambkar:
replaced by #875

PR #874, opened 2020-11-03
Migrates integ test for s3, s3 control

skotambkar:
Migrates custom integration test for s3, s3 control. 

The tests setup might be unstable due to absence of waiters. 

S3Control tests depends on #875


skotambkar:
Replaced by #877

PR #877, opened 2020-11-03
Migrates s3 and s3-control integration tests

skotambkar:
Migrates custom integration test for s3, s3 control.

The tests setup might be unstable due to absence of waiters.

S3Control tests depends on #875 and this pr currently includes those changes. 

The PR also includes s3 access-point support test, which need #870 (Haven't rebased with those change, so test for access-point will fail until 870 is merged in). 

This pr replaces #874

### DIFF
--- a/service/internal/integrationtest/s3shared/integ_test_setup.go
+++ b/service/internal/integrationtest/s3shared/integ_test_setup.go
@@ -31,7 +31,7 @@ func SetupBucket(ctx context.Context, svc *s3.Client, bucketName string) (err er
 		},
 	})
 	if err != nil {
-		fmt.Errorf("failed to create bucket %s, %v", bucketName, err)
+		return fmt.Errorf("failed to create bucket %s, %v", bucketName, err)
 	}
 
 	// TODO: change this to use waiter to wait until BucketExists instead of loop
@@ -53,7 +53,7 @@ pt:
 			goto pt
 		}
 		// fail if not succeed after 10 attempts
-		fmt.Errorf("failed to determine if a bucket %s exists and you have permission to access it", bucketName)
+		return fmt.Errorf("failed to determine if a bucket %s exists and you have permission to access it", bucketName)
 	}
 
 	return nil


service/internal/integrationtest/s3shared/integ_test_setup.go
```diff
@@ -0,0 +1,134 @@
...
+	_, err = svc.CreateBucket(ctx, &s3.CreateBucketInput{
+		Bucket: &bucketName,
+	})
+	if err != nil {
+		fmt.Errorf("failed to create bucket %s, %v", bucketName, err)
+	}
+
+	// TODO: change this to use waiter to wait until BucketExists instead of sleep timer
+	// 	svc.WaitUntilBucketExists(HeadBucketInput)
+	time.Sleep(1 * time.Minute)
```
jasdel, line 36: Instead of sleeping this should just do a loop for `HeadBucket` until the bucket exists.  Similar for other waiters that were removed. This will prevent extra waited time and probably more stable tests.

service/internal/integrationtest/s3control/api_test.go
```diff
@@ -0,0 +1,44 @@
+// +build integration
+
+package s3control
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/awslabs/smithy-go/ptr"
```
jasdel, line 10: Not significant, but for consistency should this use the SDK's ptr methods?

service/internal/integrationtest/s3control/setup_test.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+var (
+	svc                            *s3control.Client
+	s3ControlEndpoint, stsEndpoint string
+	accountID                      string
+	insecureTLS, useDualstack      bool
+)
+
+var region = "us-east-1"
```
jasdel, line 27: Does this still need to default to `us-east-1` can it use `us-west-2`. thats what most of the other integration tests default to.

jasdel (changes requested):

jasdel (approved):

PR #879, opened 2020-11-05
fix: sample code

int-tt:
The sample code has been fixed because the following error occurs.
```
cannot use aws.CredentialsCache literal (type aws.CredentialsCache) as type aws.CredentialsProvider in assignment:
        aws.CredentialsCache does not implement aws.CredentialsProvider (Retrieve method has pointer receiver)
```


skmcgrail (approved):

skmcgrail:
Thank you for contributing this fix @int-tt, the change looks good.

Issue #820, opened 2020-10-14
0.26.0 broke all the things


skyzyx:
`<arghh>`

Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**

![shits broke yo](https://user-images.githubusercontent.com/39447/96033270-bb555a80-0e14-11eb-9c76-e6434004dae0.jpg)

**Version of AWS SDK for Go?**

v0.26.0

**Version of Go (`go version`)?**

```
go version go1.15.2 darwin/amd64
```

**To Reproduce (observed behavior)**

Quite honestly I don't even know where to begin. Things got refactored, moved, removed, etc. Go modules broke. There is no upgrade guide. There is no updated documentation about how things are supposed to work now. Some discussion on the Gitter channel (the one in the README, that nobody from AWS participates in anymore) about how this version broke all the things. What little documentation there is, is impenetrable.

I have been patient. I have been nice. I have tickets that have been open for more than 2 years. At this specific moment in time, I am out of patience.

_WTAF?_

**Expected behavior**

* **DOC. U. MEN. TATION.**
* Upgrade guides when shit breaks, yo.
* Notes in the CHANGELOG and Release Notes about breaking changes.

**Additional context**

Reviewed the following, looking for a hint of what do to.

* https://docs.aws.amazon.com/sdk-for-go/v2/api/
* https://docs.aws.amazon.com/sdk-for-go/v2/api/aws/external/
* https://pkg.go.dev/mod/github.com/aws/aws-sdk-go-v2/
* https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.26.0
* https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.26.0?tab=overview
* https://gitter.im/aws/aws-sdk-go-v2

This is not a thankless open-source gig. You get paid (adequate) money in exchange for supporting the SDKs you build. All I'm asking is that you _actually support it_. And if you're not going to, then be up-front about it and say so in the README or move it over to `awslabs`.

`</arghh>`

[...]

PR #885, opened 2020-11-09
Regenerated Clients with Paginator Support

### LINKED ISSUE #439: Refactoring API client Paginators for better refactored API Client

skmcgrail:
Adds support for paginators.

Related #820 
Related https://github.com/aws/aws-sdk-go/issues/3513
Closes #439

Depends on:
https://github.com/awslabs/smithy-go/pull/234
https://github.com/awslabs/smithy/pull/628


internal/protocoltest/awsrestjson/types/enums.go
```diff
@@ -6,11 +6,11 @@ type FooEnum string
...
-	FooEnumBaz FooEnum = "Baz"
-	FooEnumBar FooEnum = "Bar"
-	FooEnum1   FooEnum = "1"
-	FooEnum0   FooEnum = "0"
+	FooEnumFoo  FooEnum = "Foo"
+	FooEnumBaz  FooEnum = "Baz"
+	FooEnumBar  FooEnum = "Bar"
+	FooEnumOne  FooEnum = "1"
+	FooEnumZero FooEnum = "0"
 )
```
jasdel, line 14: Guessing this is a upstream smithy-go change that didn't make it in before?

service/accessanalyzer/paginators.go
```diff
@@ -0,0 +1,307 @@
...
+	// The maximum number of results to return in the response.
+	Limit *int32
+
+	// Set to true if pagination should stop if the service returns the same taken
+	// passed to it
+	StopOnDuplicateToken bool
+}
+
+// ListFindingsPaginator is a paginator for
+// github.com/aws/aws-sdk-go-v2/service/accessanalyzer.ListFindings
```
jasdel, line 28: Assuming this should be the name of the operation, not the import path :)

service/accessanalyzer/paginators.go
```diff
@@ -0,0 +1,307 @@
...
+		return false
+	}
+	return p.firstPage || (p.nextToken != nil && len(*p.nextToken) > 0)
+}
+
+// NextPage retrieves the next ListFindings page.
+func (p *ListFindingsPaginator) NextPage(ctx context.Context, optFns ...func(*Options)) (*ListFindingsOutput, error) {
+	if !p.HasMorePages() {
+		return nil, fmt.Errorf("no more pages available")
+	}
```
jasdel, line 64: Is there any value in this being a typed error?, Its a terminal error that should only happen if there is a bad behavior in the calling application, so probably not too actionable.

service/accessanalyzer/paginators.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return nil, err
+	}
+	p.firstPage = false
+	p.nextToken = result.NextToken
+	if p.options.StopOnDuplicateToken && params.NextToken != nil && p.nextToken != nil && *params.NextToken == *p.nextToken {
+		p.done = true
+	} else {
+		p.done = false
+	}
```
jasdel, line 79: With the terminal error case above `if !p.HasMorePages` is the `else` case here possible?  If using `nextToken` of `nil` in the case of duplicate tokens would replicate the behavior of service returning no next token, and would remove the need for the `done` member.

internal/protocoltest/awsrestjson/types/enums.go
skmcgrail: This probably is due to having to use the tip of smithy to incorporate fixes for pagination token paths.

service/accessanalyzer/paginators.go
skmcgrail: Agreed that was my thought which is why I didn't produce a typed error. User should be using `HasMorePages` to indicate that they shouldn't be calling NextPages.

service/accessanalyzer/api_op_ListArchiveRules.go
```diff
@@ -113,3 +114,78 @@ func addOperationListArchiveRulesMiddlewares(stack *middleware.Stack, options Op
...
+var _ ListArchiveRulesAPIClient = (*Client)(nil)
+
+// ListArchiveRulesPaginatorOptions is the paginator options for ListArchiveRules
+type ListArchiveRulesPaginatorOptions struct {
+	// The maximum number of results to return in the request.
+	Limit *int32
+
+	// Set to true if pagination should stop if the service returns a pagination token
+	// that matches the most recent token provided to the service.
+	StopOnDuplicateToken bool
```
jasdel, line 133: It would be great to have a way to limit which operations this is generated based on if the service returns duplicate tokens or not. But i doubt there is any sane way to keep track of this and for it to be correct.

I could see this as a confusing option for users, unsure when they should set this field.  Maybe we can improve this by expanding the documentation here giving hint when to use this option.

jasdel (approved):

internal/protocoltest/awsrestjson/api_op_JsonMaps.go
```diff
@@ -27,11 +27,11 @@ func (c *Client) JsonMaps(ctx context.Context, params *JsonMapsInput, optFns ...
 }
 
 type JsonMapsInput struct {
-	MyMap map[string]types.GreetingStruct
+	MyMap map[string]*types.GreetingStruct
 }
 
 type JsonMapsOutput struct {
-	MyMap map[string]types.GreetingStruct
+	MyMap map[string]*types.GreetingStruct
```
jasdel, line 35: Looks like this change also includes changes to the smithy protocol test as well. Were these maps/list updated to be `@sparse`?

service/alexaforbusiness/api_op_ListSkills.go
```diff
@@ -118,6 +119,85 @@ func addOperationListSkillsMiddlewares(stack *middleware.Stack, options Options)
...
+}
+
+var _ ListSkillsAPIClient = (*Client)(nil)
+
+// ListSkillsPaginatorOptions is the paginator options for ListSkills
+type ListSkillsPaginatorOptions struct {
+	// The maximum number of results to include in the response. If more results exist
+	// than the specified MaxResults value, a token is included in the response so that
+	// the remaining results can be retrieved.
+	Limit *int32
```
jasdel, line 134: Even though this member is modeled as boxed, should the paginator `Limit` option also treat it as boxed? Or would it be better to force the `Limit` member to value types, if zero do nothing.

A boxed `Limit` member suggests the user could request zero length pages which makes no sense.

service/s3/api_op_ListObjectsV2.go
```diff
@@ -266,6 +267,85 @@ func addOperationListObjectsV2Middlewares(stack *middleware.Stack, options Optio
...
+
+// NextPage retrieves the next ListObjectsV2 page.
+func (p *ListObjectsV2Paginator) NextPage(ctx context.Context, optFns ...func(*Options)) (*ListObjectsV2Output, error) {
+	if !p.HasMorePages() {
+		return nil, fmt.Errorf("no more pages available")
+	}
+
+	params := *p.params
+	params.ContinuationToken = p.nextToken
+	params.MaxKeys = p.options.Limit
```
jasdel, line 331: This should only happen if `Limit` is a non-zero value.

jasdel (changes requested):

internal/protocoltest/awsrestjson/api_op_JsonMaps.go
skmcgrail: Yup, looks like master tests were updated to have sparse trait.

jasdel:
What do you think about the following as a way we can simplify some of the logic around that `Limit` option value.

The Paginator should auto fill `Limit` option based on the the value in the Input struct. So if a user does use the func option for setting the Limit option it would start off in what ever value was in the Input parameter if there was one.

This strategy highlights a question: What if user wants to clear the page size limit via the Limit option? How would this be done?  Would Limit option value of 0 be translated to be the zero value of a boxed limit input member?

```go
// For boxed PageSize
var options PaginatorOptions
options = ptr.ToInt32(input.PageSize)


// For unboxed PageSize
var options PaginatorOptions
options.Limit = input.PageSize
```

```go
// if PageSize is boxed
if p.options.Limit == 0 {
     input.PageSize = nil
} else {
    input.PageSize = ptr.Int32(p.options.Limit)
}

// if PageSize is unboxed
input.PageSize = p.options.Limit
```

The above has one notable downside it is impossible to send a page size of 0 with this strategy. Arguably that probably is OK, because the behavior isn't really well defined if 0 were to be sent, Could be failure be service, could be infinite loop, service could ignore page size and use a default.

jasdel (approved):

skotambkar (approved):

PR #887, opened 2020-11-11
codegen: Add support for list and maps generated as value members instead of pointer

jasdel:
Updates the SDK code generation to make use of the new smithy `NullableIndex` and smithy-go's `GoPointableIndex`. These utilities allow the SDK's code generation to be aware of API shapes and members that are nullable, and can be rendered as values by the code generation instead of pointers.

This update updates the SDK's code generation types to be value instead of pointer in most cases.
* Numbers (int8, int16, int32, int64, float32, float64)
* bool
* lists - e.g. `[]string` instead of `[]*string`
* maps - e.g. `map[string]string` instead of `map[string]*string`.

Number and boolean types that are not boxed (not pointer) will not serialize their zero value unless they are members of a list or map.

This change also significantly refactors how the SDK represents nullability and pointability when generation types and members from an API model. Utilities were added in the smithy-go change that help ensure more consistent logic through the sdk's code generation.

Depends on https://github.com/awslabs/smithy-go/pull/232
Fixes https://github.com/aws/aws-sdk-go-v2/issues/896


skmcgrail (approved):

skmcgrail:
Have you ran `make integration` to verify these changes?

jasdel:
Yeah verified that all integration tests pass successfully.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -48,14 +53,16 @@
...
+            boolean isXmlAttributeMember
     ) {
         this.context = context;
+        this.member = member;
         this.dataDest = dataDest;
         this.timestampFormat = timestampFormat;
         this.isXmlAttributeMember = isXmlAttributeMember;
-        this.isFlattened = isFlattened;
+        this.isFlattened = member.hasTrait(XmlFlattenedTrait.ID);
+        ;
```
skotambkar: accidental?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
```diff
@@ -69,21 +71,21 @@ public XmlShapeDeserVisitor(GenerationContext context, Predicate<MemberShape> me
...
 
     // generates code to define and initialize output variable for an aggregate shape
-    private void generatesIntializerForOutputVariable(GenerationContext context, Shape shape) {
+    private void generatesInitializerForOutputVariable(GenerationContext context, Shape shape) {
         GoWriter writer = context.getWriter();
         Symbol shapeSymbol = context.getSymbolProvider().toSymbol(shape);
 
         writer.write("var sv $P", shapeSymbol);
+        // TODO [denseListMap] need to reference if pointable type. Assumes containers are always double pointers.
+        // TODO this probably isn't true for union types.
```
skotambkar, line 88: still todo's?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
```diff
@@ -48,14 +53,16 @@
     private final boolean isFlattened;
 
     public XmlMemberDeserVisitor(
-            GenerationContext context, String dataDest, Format timestampFormat,
-            boolean isXmlAttributeMember, boolean isFlattened
+            GenerationContext context, MemberShape member, String dataDest, Format timestampFormat,
+            boolean isXmlAttributeMember
```
skotambkar, line 57: Nit: one on each line

skotambkar (approved):

skotambkar:
Looks like smithy-go dep version is not correct? CI for Java fails as it doesn't find methods from awslabs/smithy-go#232

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
jasdel: I thought I removed all of those, must of missed one, thanks!

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlMemberDeserVisitor.java
jasdel: Yeah typo

jasdel:
That's probably result of bad merge. There where no Go changes in the smithy-go change. I won't be able to have CI succeed on this PR until the smithy change is merged because it depends on Java changes in that pr.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeDeserVisitor.java
jasdel: This comment was wrong, union de-serialization will always be a struct that satisfy the union member interface. And all types in this context will be pointers like list/maps, or double pointers like structs.

PR #888, opened 2020-11-11
Add presign-url client support

### LINKED ISSUE #891: Refactor SDK's presign URL feature
The v2 SDK should provide an easy way to create a presigned URL for an operation similar to the concept in the v1 SDK. There were a couple issues with the v1 SDK's implementation that the v2 SDK's should avoid.

1.) Presign was available for all API operations regardless if it was actually supported.
2.) Presign was supported for all HTTP methods, even though it was nearly impossible to use non-GET and non-streaming operations.
3.) Presign exposed an `Expires` behavior for all operations, when this was only valid for Amazon S3 operations.

To address these issues the v2 SDK's presigner design should be generated independently of the API client. And the presigner should be generated with a select subset of supported operations that are known to work, and where it makes sense to have a presigner for.

The work with autofill presign, #799 provided the ground work for generated presign clients for an API. This change also broke out the `Expires` concept into a conditionally generated option available on the presign clients for APIs that support that behavior.

The PR #888 works to refactor the presign client generator implemented in #799 and export it for for APIs like Amazon S3.

The presign client should be based off of the API client, providing flexable options to initialize it, with optional parameters as needed.

```go
// PresignOptions represents the presign client options
type PresignOptions struct {
	// ClientOptions are list of functional options to mutate client options used by
	// presign client
	ClientOptions []func(*Options)
	// Presigner is the presigner used by the presign url client
	Presigner v4.HTTPPresigner
}

// PresignClient represents the presign url client
type PresignClient struct { /*... */}

// NewPresignClient generates a presign client using provided Client options and
// presign options
func NewPresignClient(options Options, optFns ...func(*PresignOptions)) *PresignClient { /*...*/}

// NewPresignClientWrapper generates a presign client using provided API Client and
// presign options
func NewPresignClientWrapper(c *Client, optFns ...func(*PresignOptions)) *PresignClient {/*...*/}

// NewPresignClientFromConfig generates a presign client using provided AWS config
// and presign options
func NewPresignClientFromConfig(cfg aws.Config, optFns ...func(*PresignOptions)) *PresignClient {/*...*/}

// PresignCopyDBClusterSnapshot will create a AWS sigv4 request presigned URL for the CopyDBClusterSnapshot operation.
func (c *PresignClient) PresignCopyDBClusterSnapshot(ctx context.Context, params *CopyDBClusterSnapshotInput, optFns ...func(*PresignOptions)) (req *v4.PresignedHTTPRequest, err error) {
```

skotambkar:
[x] Generates presign url clients for s3, rds and ec2 client.
[x] Modifies existing rds autofill customization and test runner.
[x] Generates presign operation functions for a handful of operations. 
[x] Write test for s3 presign url operation and client. 

Fixes #891

### DIFF
--- a/service/internal/integrationtest/s3/presign_test.go
+++ b/service/internal/integrationtest/s3/presign_test.go
@@ -36,13 +36,12 @@ func TestInteg_PresignURL(t *testing.T) {
 		"nil-body": {
 			expectedSignedHeader: http.Header{},
 		},
-		// TODO: fix PutObject with an empty body returning 501
-		// "empty-body": {
-		// 	body: bytes.NewReader([]byte("")),
-		// 	expectedSignedHeader: http.Header{
-		// 		"content-type": {"application/octet-stream"},
-		// 	},
-		// },
+		"empty-body": {
+			body: bytes.NewReader([]byte("")),
+			expectedSignedHeader: http.Header{
+				"content-type": {"application/octet-stream"},
+			},
+		},
 	}
 
 	for name, c := range cases {
@@ -72,7 +71,7 @@ func TestInteg_PresignURL(t *testing.T) {
 
 			presignRequest, err := presignerClient.PresignPutObject(ctx, putObjectInput)
 			if err != nil {
-				t.Errorf("expect no error, got %v", err)
+				t.Fatalf("expect no error, got %v", err)
 			}
 
 			for k, v := range c.expectedSignedHeader {
@@ -146,6 +145,9 @@ func sendHTTPRequest(presignRequest *v4.PresignedHTTPRequest, body io.Reader) (*
 	// assign the request body if not nil
 	if body != nil {
 		req.Body = ioutil.NopCloser(body)
+		if req.ContentLength == 0 {
+			req.Body = nil
+		}
 	}
 
 	// Upload the object to S3.
--- a/service/internal/integrationtest/go.mod
+++ b/service/internal/integrationtest/go.mod
@@ -87,6 +87,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/wafv2 v0.29.0
 	github.com/aws/aws-sdk-go-v2/service/workspaces v0.29.0
 	github.com/awslabs/smithy-go v0.3.1-0.20201120010914-c02b9493fe20
+	github.com/google/go-cmp v0.5.2
 )
 
 go 1.15


service/internal/presigned-url/middleware_test.go
```diff
@@ -136,11 +139,14 @@ type mockURLPresignInput struct {
...
+func presignURL(ctx context.Context, client interface{}, srcRegion string, params interface{}) (
+	req *v4.PresignedHTTPRequest, err error,
 ) {
 	in := params.(*mockURLPresignInput)
 
-	return "https://example." + srcRegion + ".amazonaws.com/?DestinationRegion=" + in.DestinationRegion,
-		http.Header{}, nil
+	return &v4.PresignedHTTPRequest{
+		URL:          "https://example." + srcRegion + ".amazonaws.com/?DestinationRegion=" + in.DestinationRegion,
+		Method:       "POST",
```
jasdel, line 149: not too important, but this should be `GET` HTTP method.

service/s3/internal/customizations/presigned_expires.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/awslabs/smithy-go/middleware"
+	smithyhttp "github.com/awslabs/smithy-go/transport/http"
+)
+
+// AddExpiresOnPresignedURL  represents a build middleware used to assign
+// expiration on a presigned URL
+type AddExpiresOnPresignedURL struct {
+	// Expires is time.Duration within which presigned url should be expired
+	Expires time.Duration
```
jasdel, line 17: Should this callout that seconds precision will be used?

service/s3/api_client.go
```diff
@@ -452,3 +452,125 @@ func addRequestResponseLogging(stack *middleware.Stack, o Options) error {
...
 	}, middleware.After)
 }
+
+// PresignOptions represents the presign client options
+type PresignOptions struct {
+	// ClientOptions are list of functional options to mutate client options used by
+	// presign client
+	ClientOptions []func(*Options)
+	// Presigner is the presigner used by the presign url client
+	Presigner v4.HTTPPresigner
```
jasdel, line 462: Should this be an interface similar to the Signer for the API client?

```
	// Signature Version 4 (SigV4) Signer
	HTTPSignerV4 HTTPSignerV4
```

jasdel, line 462: minor formatting nit, adding new lines between members since they have their own docs.

service/s3/api_client.go
```diff
@@ -452,3 +452,125 @@ func addRequestResponseLogging(stack *middleware.Stack, o Options) error {
...
+	var presigner v4.HTTPPresigner
+	if presignOptions.Presigner != nil {
+		presigner = presignOptions.Presigner
+	} else {
+		presigner = v4.NewSigner()
+	}
+	return &PresignClient{
+		client: client, presigner: presigner,
+		expires: presignOptions.Expires,
+	}
```
jasdel, line 535: minor nit, but the `presigner` variable isn't really needed and can be simplified.

```go
if presignerOptions.Presigner == nil {
     presignerOptions.Presigner = v4.NewSigner()
}

return &PresignClient{
	client: client,
        presigner: presignerOptions.Presigner,
	expires: presignOptions.Expires,
}
```

service/s3/api_client.go
```diff
@@ -452,3 +452,125 @@ func addRequestResponseLogging(stack *middleware.Stack, o Options) error {
...
+}
+
+// NewPresignClientFromConfig generates a presign client using provided AWS config
+// and presign options
+func NewPresignClientFromConfig(cfg aws.Config, optFns ...func(*PresignOptions)) *PresignClient {
```
jasdel, line 540: These three constructors duplicate a lot of their behavior. It would be nice to merge these together into a single function that does all the work, and two others that change how configuration is passed down.

service/s3/api_client.go
```diff
@@ -452,3 +452,125 @@ func addRequestResponseLogging(stack *middleware.Stack, o Options) error {
...
+		return err
+	}
+	if c.expires != 0 {
+		// add middleware to set expiration for s3 presigned url
+		err = stack.Build.Add(&s3cust.AddExpiresOnPresignedURL{Expires: c.expires}, middleware.After)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
jasdel, line 575: I'm not 100% positive about this, but the `X-Amz-Expires` header may be required for S3 presigned operations. We should validate this. From the Amazon S3 docs on presigned URLs it not really clear, other than saying the minimum time is 1 second. 

https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

service/s3/api_client.go
```diff
@@ -452,3 +452,125 @@ func addRequestResponseLogging(stack *middleware.Stack, o Options) error {
...
+
+// PresignOptions represents the presign client options
+type PresignOptions struct {
+	// ClientOptions are list of functional options to mutate client options used by
+	// presign client
+	ClientOptions []func(*Options)
+	// Presigner is the presigner used by the presign url client
+	Presigner v4.HTTPPresigner
+	// Expires sets the expiration duration for the generated presign url
+	Expires time.Duration
```
jasdel, line 464: This should call out that expires has a precision of seconds. Also if the field is not required, and the SDK shouldn't set a default, it should call out the default behavior if unset.

jasdel (changes requested):

service/s3/api_client.go
skotambkar: `v4.HTTPPresigner` is an interface defined for v4 signer.

jasdel: Yeah i think the intention here is to have the client define its own interface similar to the way it does for `HTTPSignerV4`

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
```diff
@@ -166,138 +304,254 @@ public void writePresignClientType(GoWriter writer) {
...
+
+            // s3 service has an additional Expires options
+            if (isS3ServiceShape(model, serviceShape)) {
+                writer.write("");
+                writer.writeDocs(
+                        String.format("Expires sets the expiration duration for the generated presign url. This should "
+                                + "be the duration in seconds the presigned URL should be considered valid for."
+                        )
+                );
+                writer.write("Expires time.Duration");
```
jasdel, line 505: Should call out the default value if unset. Since it seems like this header is required.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/PresignURLAutoFill.java
```diff
@@ -71,6 +66,122 @@
...
+            StructureShape input
+    ) {
+        Symbol operationSymbol = symbolprovider.toSymbol(operation);
+        Symbol inputSymbol = symbolprovider.toSymbol(input);
+
+        Symbol removeMiddleware = SymbolUtils.createValueSymbolBuilder("RemoveMiddleware",
+                AwsCustomGoDependency.PRESIGNEDURL_CUSTOMIZATION).build();
+
+        writer.openBlock(
+                "func $L(ctx context.Context, client interface{}, region string, params interface{}) "
```
jasdel, line 134: Does the `client` need to be an empty interface? I'd think this would be specific to the `*PresignClient`.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/PresignURLAutoFill.java
```diff
@@ -71,6 +66,122 @@
...
+        return String.format("set%s%s", operationName, memberName);
+    }
+
+    private static String copyInputFuncName(String inputName) {
+        return String.format("copy%sForPresign", inputName);
+    }
+
+    private static String presignFuncName(String operationName, boolean exported) {
+        return exported ? String.format("Presign%s", operationName) : String.format("presign%s", operationName);
+    }
```
jasdel, line 184: These probably don't need to be exported, since they should only be used directly by the customization middleware.

service/internal/presigned-url/middleware.go
```diff
@@ -3,35 +3,43 @@ package presignedurl
...
-type URLPresigner interface {
-	PresignURL(ctx context.Context, srcRegion string, params interface{}) (
-		presignedURL string, signedHeader http.Header, err error,
-	)
+	// SetPresignedURL accessor points to a function that sets presigned url on api input struct
+	SetPresignedURL func(interface{}, string) error
+
+	// PresignOperation is the presign function accessor
+	PresignOperation func(ctx context.Context, client interface{},
+		srcRegion string, params interface{}) (*v4.PresignedHTTPRequest, error)
```
jasdel, line 33: Down side of putting this as a function on the accessory instead of an interface on the middleware's `Options` is it creates the opportunity for a runtime error because types are not matched up correctly.

What do you think about leaving in the unexported type for wrapping a API client's `PresignClient` type as the `PresignURL` interface defined in this middleware?  That provides stronger type association and reduces the need for empty interfaces.

service/internal/presigned-url/middleware.go
```diff
@@ -3,35 +3,43 @@ package presignedurl
...
 
 // Options provides the set of options needed by the presigned URL middleware.
 type Options struct {
-	Accessor  ParameterAccessor
-	Presigner URLPresigner
+	// Accessor are the parameter accessors used by this middleware
+	Accessor ParameterAccessor
+
+	// PresignClient is the presigner client used to presign an operation request.
+	PresignClient interface{}
```
jasdel, line 42: Making this an empty interface leaves a lot of room for type miss-matches to happen vs the old `PresignURL`

service/rds/api_op_CopyDBClusterSnapshot.go
```diff
@@ -385,27 +352,12 @@ func addCopyDBClusterSnapshotPresignURLMiddleware(stack *middleware.Stack, optio
 			CopyInput:            copyCopyDBClusterSnapshotInputForPresign,
 			SetDestinationRegion: setCopyDBClusterSnapshotdestinationRegion,
 			SetPresignedURL:      setCopyDBClusterSnapshotPreSignedUrl,
+			PresignOperation:     presignCopyDBClusterSnapshot,
```
jasdel, line 355: This doesn't really need to be a function using empty interfaces. If it were a method on a type that wrapped the presigned client, there is no need to pass the Presign client down via multiple levels of empty interfaces.

jasdel (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
```diff
@@ -17,123 +17,261 @@
...
+    ) {
+        Symbol operationSymbol = symbolProvider.toSymbol(operationShape);
+
+        Shape operationInputShape = model.expectShape(operationShape.getInput().get());
+        Symbol operationInputSymbol = symbolProvider.toSymbol(operationInputShape);
+
+        writer.writeDocs(
+          String.format("Presign%s is used to generate a presigned HTTP Request which contains presigned URL, signed headers "
+                  + "and HTTP method used.", operationSymbol.getName())
+        );
```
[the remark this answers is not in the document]
skotambkar: it doesn't :(

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
```diff
@@ -121,19 +247,76 @@ public void writePresignClientType(GoWriter writer) {
...
+        if (operationInputShape.members().stream().noneMatch(memberShape -> {
+            return memberShape.getMemberTrait(model, StreamingTrait.class).isPresent();
+        })) { return; }
+
+        writer.openBlock("func $L(stack $P, options Options) error {", "}",
+                addAsUnsignedPayloadName(operationSymbol.getName()),
+                SymbolUtils.createPointableSymbolBuilder("Stack", SmithyGoDependency.SMITHY_MIDDLEWARE).build(),
+                () -> {
+                    writer.write("return $T(stack)", SymbolUtils.createValueSymbolBuilder(
+                            "AddUnsignedPayloadMiddleware", AwsGoDependency.AWS_SIGNER_V4).build());
```
jasdel, line 305: Does the compute sha middleware needed to be removed as well?  Or does the unsigned Payload middleware get added before that compute sha is run?

service/internal/featuretest/s3/presign_test.go
```diff
@@ -0,0 +1,180 @@
+package unit_test
```
jasdel, line 1: I think this would be better suited as a unit test inside of the relevant service, e.g. `service/s3/internal/customizaions/presign_test.go`

Each one of the presign unit tests that we add will need to be their own file anyways, so might as well keep it with the service instead of a separate module.

service/s3/api_op_PutObject.go
```diff
@@ -399,3 +400,42 @@ func newServiceMetadataMiddleware_opPutObject(region string) *awsmiddleware.Regi
...
+	if err != nil {
+		return nil, err
+	}
+
+	out := result.(*v4.PresignedHTTPRequest)
+	return out, nil
+}
+
+func addPutObjectPayloadAsUnsigned(stack *middleware.Stack, options Options) error {
+	v4.RemoveContentSHA256HeaderMiddleware(stack)
```
jasdel, line 438: This header middleware is probably still useful. just don't need the compute sha. Doesn't the unsigned payload middleware expect the content sha header middleware to be present to set the header?

jasdel (changes requested):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
jasdel: looks like this was already fixed,

service/s3/api_op_PutObject.go
jasdel: Looks like i was wrong on this one, it should be fine to remove this middleware.

service/internal/integrationtest/s3/presign_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/internal/integrationtest"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+func TestInteg_PresignURL_PutObject(t *testing.T) {
```
jasdel, line 19: Nit would be good to have a PresignURL_PutObject integ test that creates the presign URL without an explicit body.  This also means that the content-length header would not be signed.

```go
params := &s3.PutObjectInput{
     Bucket: &setupMetadata.Buckets.Source.Name,
     Key: &key,
}
```

Looks like this test is almost all setup for it, just need to add the case.

service/s3/internal/customizations/presign_test.go
```diff
@@ -0,0 +1,180 @@
...
+		expectError             string
+	}{
+		"standard case": {
+			input: s3.PutObjectInput{
+				Bucket: aws.String("mock-bucket"),
+				Key:    aws.String("mockkey"),
+				Body:   strings.NewReader("hello-world"),
+			},
+			expectPresignedURLHost: "https://mock-bucket.s3.us-west-2.amazonaws.com/mockkey?",
+			expectedRequestURILabel: []string{
```
jasdel, line 34: Nit, `expectRequestURIQuery` :)  label would be in the path.

service/s3/internal/customizations/presign_test.go
```diff
@@ -0,0 +1,180 @@
...
+	cases := map[string]struct {
+		input                   s3.PutObjectInput
+		options                 s3.PresignOptions
+		expectPresignedURLHost  string
+		expectedRequestURILabel []string
+		expectSignedHeader      http.Header
+		expectMethod            string
+		expectError             string
+	}{
+		"standard case": {
```
jasdel, line 27: One additional testcase that would be nice to have would be one where the user explicitly specifies a `X-Amz-Content-Sha256` header value the request should be presigned with.

jasdel: That value should be used for the presigned URL, and would be a signed header that must be sent by the client using the presigned URL.

service/internal/integrationtest/s3/presign_test.go
jasdel: Could also have a presigned URL integration test case where the X-Amz-Content-Sha256 header is also provided manually by the user, which would become a signed header.

jasdel: Adding a check against expected set of signed headers would be helpful as well.

service/internal/integrationtest/s3/presign_test.go
```diff
@@ -5,49 +5,128 @@ package s3
...
+		"nil-body": {
+			expectedSignedHeader: http.Header{},
+		},
+		// TODO: fix PutObject with an empty body returning 501
+		// "empty-body": {
+		// 	body: bytes.NewReader([]byte("")),
+		// 	expectedSignedHeader: http.Header{
+		// 		"content-type": {"application/octet-stream"},
+		// 	},
+		// },
```
jasdel, line 45: Probably don't want to generate PutObject presign url support until this item is fixed.   PutObject with nil body is the common way of using presign PutObject

jasdel (commented):
everything looks good with the exception of probably want to exclude putObject until signing issue is fixed.

service/internal/integrationtest/s3/presign_test.go
jasdel: Does nil `body` have the same issue as empty body?

service/internal/integrationtest/s3/presign_test.go
```diff
@@ -0,0 +1,156 @@
...
+	cases := map[string]struct {
+		body                 io.Reader
+		expires              time.Duration
+		sha256Header         string
+		expectedSignedHeader http.Header
+	}{
+		"standard": {
+			body: bytes.NewReader([]byte("Hello-world")),
+			expectedSignedHeader: http.Header{
+				"content-type":   {"application/octet-stream"},
```
jasdel, line 32: This header probably shouldn't be set by default for put objects.  This header being set requires users of the Presign URL to use the signed header in addition to the URL.

jasdel: We should have a release candidate issue to track if the `content-type` header auto population should be removed being set for streams by default. Having this header hear make the presign URL more difficult to use because the map of headers have to be passed to the downstream client using the presigned URL.

service/internal/integrationtest/s3/presign_test.go
```diff
@@ -0,0 +1,156 @@
...
+			},
+		},
+		"nil-body": {
+			expectedSignedHeader: http.Header{},
+		},
+		"empty-body": {
+			body: bytes.NewReader([]byte("")),
+			expectedSignedHeader: http.Header{
+				"content-type": {"application/octet-stream"},
+			},
```
jasdel, line 43: With regard to the content-type, we should consider if empty body has the same behavior as nil body.

jasdel (approved):

service/internal/integrationtest/s3/presign_test.go
skotambkar: Created https://github.com/aws/aws-sdk-go-v2/issues/923 to track this

PR #898, opened 2020-11-12
 aws/transport/http: Move aws.BuildableHTTPClient to http transport package

### LINKED ISSUE #730: aws.BuildableHTTPClient with options method pattern is awkward to use.
Using the client configuration options with the BuildableHTTPClient is fairly awkward to use. If multiple options are desired like Transport and Dialer the return value of the option method must be recasted back to the BuildableHTTPClient before the new option can be applied.

e.g. in the updated EC2 IMDS client custom configuration of the Client's timeouts requires this cast.

```go
client := aws.NewBuildableHTTPClient()

// Use a custom Dial timeout for the EC2 Metadata service to account
// for the possibility the application might not be running in an
// environment with the service present. The client should fail fast in
// this case.
client = c.WithDialerOptions(func(d *net.Dialer) {
	d.Timeout = defaultDialerTimeout
})

// Cast back to the BuildableHTTPClient to add additional options.
c = client.(*aws.BuildableHTTPClient)

// Use a custom Transport timeout for the EC2 Metadata service to
// account for the possibility that the application might be running in
// a container, and EC2Metadata service drops the connection after a
// single IP Hop. The client should fail fast in this case.
client = c.WithTransportOptions(func(tr *http.Transport) {
	tr.ResponseHeaderTimeout = defaultResponseHeaderTimeout
})
```

jasdel:
Moves the BuildableHTTPClient from the SDK's `aws` package to the `aws/transport/http` package as `BuildableClient` to with other HTTP specific utilities.

Updates the BuildableClient's method return type to be the BuildableClient type to make chaining `With` statements more  natural.

Fixes #730

### DIFF
--- a/aws/transport/http/client.go
+++ b/aws/transport/http/client.go
@@ -69,7 +69,7 @@ func (b *BuildableClient) Do(req *http.Request) (*http.Response, error) {
 }
 
 func (b *BuildableClient) build() {
-	b.client = wrapWithoutRedirect(&http.Client{
+	b.client = wrapWithLimitedRedirect(&http.Client{
 		Timeout:   b.clientTimeout,
 		Transport: b.GetTransport(),
 	})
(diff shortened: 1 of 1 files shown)


skmcgrail (approved):

aws/transport/http/client.go
```diff
@@ -222,3 +215,60 @@ func shallowCopyStruct(src interface{}) interface{} {
 
 	return dstVal.Interface()
 }
+
+func wrapWithoutRedirect(c *http.Client) *http.Client {
```
skotambkar, line 219: Should this be exported as a utility?

aws/transport/http/client.go
```diff
@@ -222,3 +215,60 @@ func shallowCopyStruct(src interface{}) interface{} {
...
+	cc := *c
+	cc.CheckRedirect = limitedRedirect
+	cc.Transport = stubBadHTTPRedirectTransport{
+		tr: tr,
+	}
+
+	return &cc
+}
+
+func limitedRedirect(r *http.Request, via []*http.Request) error {
```
skotambkar, line 234: Nice to have a comment about what this does or when it limits?

aws/transport/http/client.go
```diff
@@ -222,3 +215,60 @@ func shallowCopyStruct(src interface{}) interface{} {
...
+
+const stubBadHTTPRedirectLocation = `https://amazonaws.com/badhttpredirectlocation`
+
+func (t stubBadHTTPRedirectTransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	resp, err := t.tr.RoundTrip(r)
+	if err != nil {
+		return resp, err
+	}
+
+	// TODO S3 is the only known service to return 301 without location header.
```
skotambkar, line 264: S3 would need a separate Buildable Client then?  Or Can this behavior be a BuildableClient option?

ec2imds/api_client.go
```diff
@@ -233,28 +234,25 @@ const (
...
+				// environment with the service present. The client should fail fast in
+				// this case.
+				d.Timeout = defaultDialerTimeout
+			}).
+			WithTransportOptions(func(tr *http.Transport) {
+				// Use a custom Transport timeout for the EC2 Metadata service to
+				// account for the possibility that the application might be running in
+				// a container, and EC2Metadata service drops the connection after a
+				// single IP Hop. The client should fail fast in this case.
+				tr.ResponseHeaderTimeout = defaultResponseHeaderTimeout
```
skotambkar, line 254: Nit: formatting looks a little off to me .

aws/transport/http/client.go
jasdel: Would prefer not to, unless there is a demand for it.

ec2imds/api_client.go
jasdel: This is because the `c.` on the top. Put that on its own line so that the `With` method following it are aligned with each other.

aws/transport/http/client.go
jasdel: In general this is block helps prevent the underlying HTTP client returning errors when a `301` HTTP response is received that doesn't contain the location tag. The comment about S3 is background for the reason this even exists, but this logic would be triggered by any 301 API response that doesn't contain a location header.

The Go HTTP client returns a error if an 301 HTTP response is received without a location header, this change prevents that error.

skotambkar (approved):

PR #912, opened 2020-11-19
config: Fix AWS_CA_BUNDLE usage while loading default config

jasdel:
Fixes the LoadDefaultConfig's configuration provider order to correctly load a custom HTTP client prior to configuring the client for AWS_CA_BUNDLE.

Fixes the configuration of AWS_CA_BUNDLE to create a default HTTP client if one has not already been created.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,19 +17,46 @@ var defaultLoaders = []loader{
 //
 // This will setup the AWS configuration's Region,
 var defaultAWSConfigResolvers = []awsConfigResolver{
+	// Resolves the default configuration the SDK's aws.Config will be
+	// initialized with.
 	resolveDefaultAWSConfig,
+
+	// Sets the logger to be used. Could be user provided logger, and client
+	// logging mode.
 	resolveLogger,
 	resolveClientLogMode,
+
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -20,8 +20,8 @@ var defaultAWSConfigResolvers = []awsConfigResolver{
 	resolveDefaultAWSConfig,
```
skotambkar, line 20: So this is the order in which default config is loaded? If so, a comment would be helpful. 

Also how do we verify this is the order is consistent across SDKs

skmcgrail (approved):

config/config.go
jasdel: This order is for how the sdk sources configuration from the user mostly then falls back to it's own default resolution behavior. E.g user provides custom http client then sdk augments that client with custom ca bundle.

But yeah docs would be helpful for us. This is internal behavior we're not exposing yet.  There isn't really a portable translation between SDKs for most of this.

jasdel: Though to call out this doesn't drive the default credentials configuration loading, as that is all handled by the `resolveCredentials` at the bottom of the list.

skotambkar (approved):

Issue #940, opened 2020-12-01
credentials/ec2rolecreds: New should not require an explict Options value to be passed to the function.


skmcgrail:
[ec2rolecreds.New](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials@v0.1.5/ec2rolecreds#New) requires that an `Options` value be passed to the constructor function. Since the constructor does not require any explicit arguments to be populated in options this function should be updated to only take a list of functions that mutate the options. This keeps the design consistent with other credential providers like [endpointcreds.New](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials@v0.1.5/endpointcreds#New) which take the explicit arguments required by the provider, and then the functions that mutate the options.

PR #946, opened 2020-12-03
Add ExpiryWindow and ExpiryWindowJitterFrac to CredentialsCache

skmcgrail:
Refactors the credential providers to remove ExpiryWindow behavior to the CredentialsCache. Adds a ExpiryWindowJitterFrac capability to the CredentialsCache to support the addition of jitter the configured refresh window.

Fixes: https://github.com/aws/aws-sdk-go-v2/issues/940

### DIFF
--- a/config/resolve_credentials.go
+++ b/config/resolve_credentials.go
@@ -24,7 +24,6 @@ const (
 
 var (
 	ecsContainerEndpoint = "http://169.254.170.2" // not constant to allow for swapping during unit-testing
-
 )
 
 // resolveCredentials extracts a credential provider from slice of config sources.
@@ -66,7 +65,7 @@ func resolveCredentialProvider(cfg *aws.Config, cfgs configs) (bool, error) {
 		return false, nil
 	}
 
-	cfg.Credentials = &aws.CredentialsCache{Provider: credProvider}
+	cfg.Credentials = wrapWithCredentialsCache(credProvider)
 
 	return true, nil
 }
@@ -100,7 +99,7 @@ func resolveCredentialChain(cfg *aws.Config, configs configs) (err error) {
 	}
 
 	// Wrap the resolved provider in a cache so the SDK will cache credentials.
-	cfg.Credentials = &aws.CredentialsCache{Provider: cfg.Credentials}
+	cfg.Credentials = wrapWithCredentialsCache(cfg.Credentials)
 
 	return nil
 }
@@ -198,7 +197,6 @@ func resolveLocalHTTPCredProvider(cfg *aws.Config, endpointURL, authToken string
 func resolveHTTPCredProvider(cfg *aws.Config, url, authToken string, configs configs) error {
 	optFns := []func(*endpointcreds.Options){
 		func(options *endpointcreds.Options) {
-			options.ExpiryWindow = 5 * time.Minute
 			if len(authToken) != 0 {
 				options.AuthorizationToken = authToken
 			}
@@ -217,7 +215,9 @@ func resolveHTTPCredProvider(cfg *aws.Config, url, authToken string, configs con
 
 	provider := endpointcreds.New(url, optFns...)
 
-	cfg.Credentials = provider
+	cfg.Credentials = wrapWithCredentialsCache(provider, func(options *aws.CredentialsCacheOptions) {
+		options.ExpiryWindow = 5 * time.Minute
+	})
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


aws/credential_cache.go
```diff
@@ -3,17 +3,74 @@ package aws
...
+	"time"
 
+	sdkrand "github.com/aws/aws-sdk-go-v2/internal/rand"
 	"github.com/aws/aws-sdk-go-v2/internal/sync/singleflight"
 )
 
+// CredentialsCacheOptions are the options
+type CredentialsCacheOptions struct {
+	// Provider is the CredentialProvider implementation to be wrapped by the CredentialCache.
+	Provider CredentialsProvider
```
jasdel, line 15: Should `Provider` be an option since its a required parameter in the constructor? Having it as an option seems like a way to get unexpected behavior having it in both places. 

I think it would be better to be a unexported member on the `CredentialsCache` type instead a part of the options.

aws/credential_cache.go
```diff
@@ -41,8 +98,17 @@ func (p *CredentialsCache) singleRetrieve() (interface{}, error) {
 		return *creds, nil
 	}
 
-	creds, err := p.Provider.Retrieve(context.TODO())
+	creds, err := p.options.Provider.Retrieve(context.TODO())
```
jasdel, line 101: This probably should be similar to v1 where it passes in the context that suppresses cancel.

aws/credential_cache.go
```diff
@@ -41,8 +98,17 @@ func (p *CredentialsCache) singleRetrieve() (interface{}, error) {
...
+		if creds.CanExpire {
+			randFloat64, err := sdkrand.CryptoRandFloat64()
+			if err != nil {
+				return Credentials{}, err
+			}
+			jitter := time.Duration(randFloat64 * p.options.ExpiryWindowJitterFrac * float64(p.options.ExpiryWindow))
+			creds.Expires = creds.Expires.Add(-(p.options.ExpiryWindow - jitter))
+		}
+
 		p.creds.Store(&creds)
```
jasdel, line 112: Is there anyway for this to log, or report, when the a code bug sets ExpiryWindow is so large a value that it causes the credentials to immediately expire? Incorrect expiry windows being set seem to be a good way for apps to get stuck in a refresh, or credential expire loop without a clear indication why.

jasdel (changes requested):

aws/credential_cache.go
skmcgrail: There are two issues I would foresee with this:
1. this condition can't be verified up front during construction of the provider cache in a reliable fashion. This is due to the credential expiration time not being known until the credentials are retrieved. Even so, subsequent refreshes of the credential can still cause the situation. So even if you were to pre-warm the cache on construction and verify this condition does not occur, it would not hold true for subsequent retrievals.
2. Logging would require the cache provider to be configured with the logger, and subsequently could make the situation worse in that you would have a log storm of messages about this condition despite a still functioning application (a bit with unnecessary latency due to need to refresh credentials before a call).

jasdel (approved):

innovia:
@skmcgrail 

i'm having a super hard time, trying to get this to work, with this new change

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds#hdr-Assume_Role

can you please share here an example of how to assume role with the service client?

skotambkar:
Hi @innovia, could you describe the issue are you facing? 

Our migration guide has an example on how to assume a role.  Please take a look - https://aws.github.io/aws-sdk-go-v2/docs/migrating/#assumerole .

innovia:
@skmcgrail thanks for the link, 

I am using okta was and have different roles to assume per environment, I'm gonna try the link you referenced

innovia:
@skmcgrail I couldn't get it to work from the docs, they are very unclear...

this code  (which i found by digging into github open source codes) works but is undocumented

```go
if assumeRoleArn != "" {
    stsClient := sts.Client(t)
    creds := stscreds.NewAssumeRoleProvider(stsClient, assumeRoleArn)
    awsConfig.Credentials = creds
  }

  client := ec2.NewFromConfig(awsConfig)
```

but according to the doc this code should be used. but its unclear:

```go
  if assumeRoleArn != "" {
    stsClient := sts.Client(t)
    creds := stscreds.NewAssumeRoleProvider(stsClient, assumeRoleArn)
    value, crErr := creds.Retrieve(ctx)
    if crErr != nil {
      return nil, fmt.Errorf("error retrieving credentials from sts: %+v", crErr)
    }
    awsConfig.Credentials = value
  }
```

it's very confusing because if I check the type of `awsConfig.Credentials` it's `CredentialsProvider` which is an interface to `Retrieve` but the result is this error that takes me in circles

`cannot use value (type aws.Credentials) as type aws.CredentialsProvider `

PR #948, opened 2020-12-04
dynamodb/attributevalue: Add utility to (un)marshal Attribtue Values

### LINKED ISSUE #115: dynamodb/attributevalue: Consider changing default behavior for empty maps and slices
aws/aws-sdk-go#682 raised an issue that the SDK's dynamodbattribute.Marshal methods will marshal empty slices and maps as `NULL`. This makes is cumbersome to intentionally have the values marshaled as empty list or map.

Consider if it makes sense to change the logic, so that a non-nil slice/map will marshal to a empty list `av.L = []dynamodb.AttributeValue{}` or map. `av.M = map[string]dynamodb.AttributeValue{}`. `nil` maps or slices would still be marshaled as `av.NULL = true`.

Issue identified in [gitter](https://gitter.im/aws/aws-sdk-go?at=5a723b66b3c4a0d376c52384) where using an empty list with the `expressions` Update caused unexpected errors.

### LINKED ISSUE #895: dynamodb/attributevalue: Add AttributeValue marshaler utilities
The DynamodB attribute value marshaler utilities in the v1 SDK's `/service/dynamodb/dynamoattributevalue` package were removed in the v0.25.0 update. These utilities need to be brought back, and reintroduced to the updated SDK, and refactored for the SDK's list, map, and union representations.

#### Blocking issues
* [ ] Smithy model for DynamoDB's API needs to be updated to backfill [tagged union](https://awslabs.github.io/smithy/spec/core.html#union) to the API and similar types.
* [x] codegen: SDK use API model to render parameters as value instead of pointers types #896

#### Related issues
* [ ] Consider changing default behavior for empty maps and slices #115
* [ ]  Support more than string types for map keys. #411
* [ ] Support Text(Un)marshaler interfaces to avoid using special marshalers in simple cases #645
* [ ] Encode map[string]struct{} as StringSet, not just []string #646
* [ ] Marshal does not support json.Number to Numeric data type #929

#### Related Concerns
* [ ] How should attribute value marshaler handle DynamoDB streams vs regular Dynamodb attribute values types?

### LINKED ISSUE #929: dynamodb/attributevalue: Marshal does not support json.Number to Numeric data type
**Is your feature request related to a problem? Please describe.**
The deprecated `dynamodbattribute.ConvertTo(...)` has broader functionality than the new `dynamdbattribute.Marshal(...)`.

**Describe the solution you'd like**
Ideally, the new function should be able to convert json.Number type to the numeric dynamodb data type N or the sdk should have another _non-deprecated_ function for making that conversion.

**Additional context**
Refer to V1 issue [#3643](https://github.com/aws/aws-sdk-go/issues/3643) for code samples and explanation.

jasdel:
Adds utilities for mapping Amazon DynamoDB API AttributeValue to and from Go types.

* Adds conversion util for DynamoDB  -> DynamoDBStreams AttributeValue.

* Fixes #895 -  Adds back marshaler/unmarshalers for attribute value to/from go types.
* Fixes #929 - Adds generic support for `json.Number` like types that are aliases of `string`s.
* Fixes #115 - Simplifies the rules of `null` vs skipped vs zero value
  for all types. 
  * All nil pointers, map, slice members are serialized as NULL AttributeValue.
  * `omitempty` struct tag skips zero value of members with the struct tag. `omitemptyelem` skips elements of list/map that have zero value. 
  * `nullempty` struct tag serializes zero value of members with the struct tag as NULL AttributeValue. `nullemptyelem` does same for elements of list/map that have zero value.
  * Empty and Nil Sets (NS, BS, SS) are serialized to null by default unless `NullEmptySets` EncoderOptions is set to false. True by default. Nil sets are always serialized as NULL, unless the`omitempty` struct tag is used.
  * Adds `nullempty` and `nullemptyelem` struct tags to direct if the encoder should marshal the member as a AttributeValue NULL if the member's value is the type's zero value, (e.g. "" for string, 0 for number, nil for pointer/map/slice, false for bool)

TODO:
* [x] Duplicate behavior for DynamoDB Streams
* [x] Update docs
* [x] add changelog

### DIFF
--- a/feature/dynamodb/attributevalue/marshaler_examples_test.go
+++ b/feature/dynamodb/attributevalue/marshaler_examples_test.go
@@ -5,8 +5,8 @@ import (
 	"reflect"
 
 	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/internal/awsutil"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
-	"github.com/aws/aws-sdk-go/aws/awsutil"
 )
 
 func ExampleMarshal() {
--- a/feature/dynamodb/attributevalue/go.mod
+++ b/feature/dynamodb/attributevalue/go.mod
@@ -3,7 +3,6 @@ module github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go v1.35.37
 	github.com/aws/aws-sdk-go-v2 v0.30.0
 	github.com/aws/aws-sdk-go-v2/service/dynamodb v0.30.0
 	github.com/aws/aws-sdk-go-v2/service/dynamodbstreams v0.30.0
--- a/feature/dynamodbstreams/attributevalue/go.mod
+++ b/feature/dynamodbstreams/attributevalue/go.mod
@@ -3,7 +3,6 @@ module github.com/aws/aws-sdk-go-v2/feature/dynamodbstreams/attributevalue
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go v1.35.37
 	github.com/aws/aws-sdk-go-v2 v0.30.0
 	github.com/aws/aws-sdk-go-v2/service/dynamodb v0.30.0
 	github.com/aws/aws-sdk-go-v2/service/dynamodbstreams v0.30.0
--- a/feature/dynamodbstreams/attributevalue/marshaler_examples_test.go
+++ b/feature/dynamodbstreams/attributevalue/marshaler_examples_test.go
@@ -5,8 +5,8 @@ import (
 	"reflect"
 
 	"github.com/aws/aws-sdk-go-v2/feature/dynamodbstreams/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/internal/awsutil"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodbstreams/types"
-	"github.com/aws/aws-sdk-go/aws/awsutil"
 )
 
 func ExampleMarshal() {


feature/dynamodb/attributevalue/decode.go
```diff
@@ -0,0 +1,711 @@
...
+			}
+		}
+		v.Set(reflect.ValueOf(set))
+		return nil
+	default:
+		return &UnmarshalTypeError{Value: "string set", Type: v.Type()}
+	}
+
+	for i := 0; i < v.Cap() && i < len(ss); i++ {
+		v.SetLen(i + 1)
```
skmcgrail, line 574: V1 had a condition here explicitly if the type was `reflect.Array` determined by the switch statement above. https://github.com/aws/aws-sdk-go/blob/d7b5372b5b7583d5a30a3d7440d9d684aba581a9/service/dynamodb/dynamodbattribute/decode.go#L588 From the docs v.SetLen will panic if v is not a slice.

feature/dynamodb/attributevalue/decode.go
```diff
@@ -0,0 +1,711 @@
...
+			return err
+		}
+		v.Set(reflect.ValueOf(t).Convert(v.Type()))
+		return nil
+	}
+
+	switch v.Kind() {
+	case reflect.String:
+		v.SetString(s)
+	case reflect.Interface:
```
skmcgrail, line 541: V1 had another case here for `reflect.Slice` for base64 encoded binary data. This was to maintain compatibility from the ConvertFrom family. should that be included here? I noticed you kept the ConvertFrom behavior of times above, so curious on this particular removal.

It's also interesting that we didn't support this in V1 on the encoder path though, making this feature asymmetric.

feature/dynamodb/attributevalue/decode.go
```diff
@@ -0,0 +1,711 @@
...
+		msg = fmt.Sprintf("cannot unmarshal to nil value, %s", e.Type.String())
+	}
+
+	return fmt.Sprintf("unmarshal failed, %s", msg)
+}
+
+// An UnmarshalError wraps an error that occurred while unmarshaling a
+// AttributeValue element into a Go type. This is different from
+// UnmarshalTypeError in that it wraps the underlying error that occurred.
+type UnmarshalError struct {
```
skmcgrail, line 700: Should this implement `Unwrap`?

jasdel: Yes all of these with ntested errors, thanks!

feature/dynamodb/attributevalue/decode.go
jasdel: reference original issue from v1, https://github.com/aws/aws-sdk-go/issues/1594

feature/dynamodb/attributevalue/decode.go
jasdel: Ref v1 change: https://github.com/aws/aws-sdk-go/commit/b608d29f7aa2b2d05e6bfd1c4ed160557c02b8f7#diff-ce8c08c3b104c532c9c755ce29719c4fbb9b666570edadbf4937ff94fd8210c9

jasdel: In discussion decided to leave out this change for now, since it was a backwards compatibility with the deprecated Convert utilities within v1.

feature/dynamodb/attributevalue/decode.go
```diff
@@ -0,0 +1,748 @@
...
+
+var stringInterfaceMapType = reflect.TypeOf(map[string]interface{}(nil))
+var byteSliceType = reflect.TypeOf([]byte(nil))
+var byteSliceSliceType = reflect.TypeOf([][]byte(nil))
+var timeType = reflect.TypeOf(time.Time{})
+
+func (d *Decoder) decode(av types.AttributeValue, v reflect.Value, fieldTag tag) error {
+	var u Unmarshaler
+	_, isNull := av.(*types.AttributeValueMemberNULL)
+	if av == nil || isNull {
```
skmcgrail, line 171: If av is not nil and the type is `AttributeValueMemberNULL` should this check that `av.Value == true`?

skmcgrail: It would be off if it was set to false, but it's the more correct check.

skmcgrail (approved):

feature/dynamodb/attributevalue/marshaler_examples_test.go
```diff
@@ -0,0 +1,94 @@
+package attributevalue_test
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"github.com/aws/aws-sdk-go/aws/awsutil"
```
skmcgrail, line 9: This example needs to be updated since this is still using the V1 utility.

feature/dynamodb/attributevalue/decode.go
jasdel: We could check if `Value` is false. I'll check if Dynamo looks at this value. I had a feeling it ignored or always set the value to true on its side.  A value of false doesn't really have any clear meaning. (Other than potentially error)

skmcgrail (approved):

Issue #926, opened 2020-11-23
config: Design of LoadDefaultConfig options is very generic, allows invalid loading options


jasdel:
The v2 SDK's design for `LoadDefaultConfig` options are very generic, and will accept any value since `Config` is an alias for `interface{}`. This means there are no direct links between the configuration loading utility and the options a user can provide. Other than looking for `With...` types/functions in the `config` module.  In addition, since `Config` is an empty interface, users could provide any value here regardless if it is valid and usable or not.

To address these issues, I propose the `config.LoadDefaultConfig` function signature be changed to take function options for a `LoadOptions` struct instead of abstract `Config` empty interface type. Instead of LoadDefaultConfig taking an abstract set of configuration provider getters, LoadDefaultConfig would take functional options of a discrete set of options that are valid for loading the configuration. Helpers like `WithRegion`, `WithSharedConfigProfile`, etc would be functional options that set fields on the `LoadOptions` struct instead of value getters.

Usage:
```go
cfg, err := config.LoadDefaultConfig(config.WithRegion("us-west-2"),
     func(o *config.LoadOptions) {
          o.SharedConfigProfile = "foo"
     })
```

config Package type definitions:
```go
type LoadOptions struct {
     aws.Config
     
     SharedConfigProfile string
     CustomCABundle io.Reader // or string?
}

func LoadDefaultConfig(optFns ...func(*LoadOptions)) (aws.Config, error) {}
```

The embedded `aws.Config` allows configuration of default configuration options such as Region, and EndpointResolver. the value of the embedded `aws.Config` would be initialized with the config's current `resolveDefaultConfig` function before the user provided option functions are invoked. This design also removes the needed for resolvers like `resolveEndpointResolvers`, `resolveRegion`, and `resolveHTTPClient`.

The `LoadOptions` type would implement the provider getters that are used by the config loading resolvers.

jasdel:
`resolveRegion` resolver would still need to exist, since region can come from additional sources, such as environment variables and shared config. In addition to ec2 region if the SDK ever supports that.

skotambkar:
Fixed in #951

Issue #819, opened 2020-10-14
config: LoadDefaultConfig should require passing context.Context during config loading.


skmcgrail:
Update `config.LoadDefaultConfig` to require `context.Context` during configuration loading.

skotambkar:
Fixed in #951

PR #951, opened 2020-12-07
config: refactor `LoadDefaultConfig` to take in context and concrete options

skotambkar:
* Refactors config's `LoadDefaultConfig` util to take in a context and LoadOptions functional options, to replace usage of empty interface{} 

* address https://github.com/aws/aws-sdk-go-v2/issues/926 and https://github.com/aws/aws-sdk-go-v2/issues/819


config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
 // The default configuration sources are:
 // * Environment Variables
 // * Shared Configuration and Shared Credentials files.
-func LoadDefaultConfig(cfgs ...Config) (aws.Config, error) {
-	var cfgCpy configs
-	cfgCpy = append(cfgCpy, cfgs...)
+func LoadDefaultConfig(ctx context.Context, optFns ...func(*LoadOptions) error) (aws.Config, error) {
+	var cfg aws.Config
+	options := LoadOptions{}
+
```
jasdel, line 229: styling nit, but prefer `var options LoadOptions` similar style for API clients, and `aws.Config` above. 

Could also do named returns for `cfg` instead of var here but no preference either way.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+// Getter functions to satisfy interfaces
+
+// getRegion returns Region from config's LoadOptions
+func (o LoadOptions) getRegion(context.Context) (string, error) {
+	if len(o.Region) != 0 {
+		return o.Region, nil
+	}
+
+	return o.DefaultRegion, nil
+}
```
jasdel, line 259: returning `DefaultRegion` probably is not correct here.  Since region can be provided by other sources such as env and shared config.

Resolving the `DefaultRegion` should be its own explicit action and resolver, or it will stomp on top of other resolvers.

jasdel: In addition, `getRegion` probably should have the same return signature as `getDefaultRegion`. If the bool is needed in one it would be needed in both.

Since its not really needed by `getRegion` and the caller will do a check of `len() == 0` then the bool probably isn't needed by getDefaultRegion.

Thoughts?

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+	if len(o.DefaultRegion) == 0 {
+		return o.DefaultRegion, false, nil
+	}
+
+	return o.DefaultRegion, true, nil
+}
+
+// WithDefaultRegion is a helper function to construct functional options
+// that sets a DefaultRegion on config's LoadOptions.
+func WithDefaultRegion(v string) func(*LoadOptions) error {
```
jasdel, line 281: should also call out that setting a default region to empty string will result in the default region value being ignored.

Similar for other helpers.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+}
+
+// getDefaultRegion returns DefaultRegion from config's LoadOptions
+func (o LoadOptions) getDefaultRegion(ctx context.Context) (string, bool, error) {
+	if len(o.DefaultRegion) == 0 {
+		return o.DefaultRegion, false, nil
+	}
+
+	return o.DefaultRegion, true, nil
+}
```
jasdel, line 277: minor styling nit, the implementation for default region is inverse of getRegion.  == 0 vs != 0.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+
+// getWebIdentityRoleCredentialOptions returns the wrapped function
+func (o LoadOptions) getWebIdentityRoleCredentialOptions(context.Context) (func(*stscreds.WebIdentityRoleOptions), bool, error) {
+	if o.WebIdentityRoleCredentialOptions == nil {
+		return func(*stscreds.WebIdentityRoleOptions) {}, false, nil
+	}
+
+	return o.WebIdentityRoleCredentialOptions, true, nil
+}
```
jasdel, line 420: Nit inconsistency in these function if they are doing set vs not set as the first. Would be helpful for readability if they were all consistent.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+		o.WebIdentityRoleCredentialOptions = v
+		return nil
+	}
+}
+
+// getAssumeRoleCredentialOptions returns AssumeRoleCredentialOptions from LoadOptions
+func (o LoadOptions) getAssumeRoleCredentialOptions(context.Context) (func(options *stscreds.AssumeRoleOptions), bool, error) {
+	if o.AssumeRoleCredentialOptions == nil {
+		return func(*stscreds.AssumeRoleOptions) {}, false, nil
+	}
```
jasdel, line 435: Since they are returning `false` for is present, these should return `nil` for the function pointer, Returning a value is a easy way for bugs to make their way in that will not check for `false` correctly.

jasdel: Also since these are functional parameters and `nil` has meaning, the `bool` parameter isn't really even needed.

The bool has meaning when the returned value of the option cannot distinguish the difference of set vs unset based on its value.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+}
+
+func (o LoadOptions) getAPIOptions(ctx context.Context) ([]func(*middleware.Stack) error, bool, error) {
+	if o.APIOptions == nil {
+		return []func(stack *middleware.Stack) error{}, false, nil
+	}
+
+	return o.APIOptions, true, nil
+}
```
jasdel, line 472: should not initialize values for unset options.  Opens room for bugs because code didn't check the bool correctly.

config/config.go
```diff
@@ -166,14 +223,327 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+	}
+}
+
+func (o LoadOptions) getClientLogMode(ctx context.Context) (aws.ClientLogMode, bool, error) {
+	if o.ClientLogMode == 0 {
+		return 0, false, nil
+	}
+
+	return o.ClientLogMode, true, nil
+}
```
jasdel, line 540: This would make it impossible for user to set ClientLogMode to a zero value via a functional option.  I think we should take a step back and reconsider the value of these `bool` vs value parameters for the option.

Arguably in this case the bool doesn't really have a value, and actually creates complexity.  Without the bool, the value of the `ClientLogMode` is always returned.

This is the reason for the LoadDefaultConfig function initializing the `aws.Config` embedded field of `LoadOptions` **before** processing the functional options provided by the caller, this ensures that the SDK has initialized all config defaults appropriately, before applications functional options are applied.

config/provider.go
```diff
@@ -85,31 +66,22 @@ func getSharedConfigFiles(configs configs) ([]string, bool, error) {
 	return nil, false, nil
 }
```
jasdel, line 67: Did we intentionally squash SharedCredentialFiles into a single field `SharedConfigFiles`?  I thought those were two different fields at one point. I may of missed that.

config/provider.go
```diff
@@ -118,26 +90,26 @@ func getCustomCABundle(configs configs) ([]byte, bool, error) {
...
+// // interface.
+// type WithRegion string
+//
+// // GetRegion returns the region string.
+// func (v WithRegion) GetRegion(ctx context.Context) (string, error) {
+// 	return string(v), nil
+// }
 
-// getRegion searchds the configs for a RegionProvider and returns the value
+// getRegion searchds the configs for a regionProvider and returns the value
```
jasdel, line 107: nit whitespace / extra comments.

config/provider.go
```diff
@@ -182,38 +150,19 @@ func (p WithEC2IMDSRegion) GetRegion() (string, error) {
 	return result.Region, nil
 }
```
jasdel, line 151: Intentionally remove the `WithEC2IMDSRegion` functional option helper? Maybe this should be `WithUseEC2IMDSRegion`? to imply its behavior?

config/provider.go
```diff
@@ -348,25 +263,17 @@ func getEndpointCredentialProviderOptions(configs configs) (f func(*endpointcred
...
-}
-
-// WithWebIdentityRoleCredentialOptions wraps a function and satisfies the EC2RoleCredentialOptionsProvider interface
-type WithWebIdentityRoleCredentialOptions func(*stscreds.WebIdentityRoleOptions)
-
-// GetWebIdentityRoleCredentialOptions returns the wrapped function
-func (w WithWebIdentityRoleCredentialOptions) GetWebIdentityRoleCredentialOptions() (func(*stscreds.WebIdentityRoleOptions), bool, error) {
-	return w, true, nil
+type webIdentityRoleCredentialOptionsProvider interface {
+	getWebIdentityRoleCredentialOptions(ctx context.Context) (func(*stscreds.WebIdentityRoleOptions), bool, error)
```
jasdel, line 269: Where the `With...` helpers intentional removed from these?

config/provider.go
```diff
@@ -580,57 +420,15 @@ func getClientLogMode(configs configs) (m aws.ClientLogMode, found bool, err err
...
-
-// WithLogConfigurationWarnings implements a LogConfigurationWarningsProvider and returns the wrapped boolean value.
-type WithLogConfigurationWarnings bool
-
-// GetLogConfigurationWarnings returns the wrapped boolean.
-func (w WithLogConfigurationWarnings) GetLogConfigurationWarnings() (bool, bool, error) {
-	return bool(w), true, nil
-}
-
-func getLogConfigurationWarnings(configs configs) (v bool, found bool, err error) {
```
jasdel, line 597: Probably still want to keep this as a load option paramater, and its functional option helper. In addition probably want to keep all of the `With...` helpers.

config/resolve.go
```diff
@@ -1,23 +1,26 @@
...
-	"github.com/awslabs/smithy-go/logging"
 )
 
 // resolveDefaultAWSConfig will write default configuration values into the cfg
 // value. It will write the default values, overwriting any previous value.
 //
 // This should be used as the first resolver in the slice of resolvers when
 // resolving external configuration.
-func resolveDefaultAWSConfig(cfg *aws.Config, cfgs configs) error {
+func resolveDefaultAWSConfig(ctx context.Context, cfg *aws.Config, cfgs configs) error {
```
jasdel, line 23: This resolve probably should be hoisted out of the resolvers and called directly by `LoadDefaultConfig` before the caller's passed in functional options are processed.

config/resolve.go
```diff
@@ -93,12 +102,12 @@ func resolveRegion(cfg *aws.Config, configs configs) error {
...
 // resolveDefaultRegion extracts the first instance of a default region and sets `aws.Config.Region` to the default
 // region if region had not been resolved from other sources.
-func resolveDefaultRegion(cfg *aws.Config, configs configs) error {
+func resolveDefaultRegion(ctx context.Context, cfg *aws.Config, configs configs) error {
 	if len(cfg.Region) > 0 {
 		return nil
 	}
 
-	region, found, err := getDefaultRegion(configs)
+	region, found, err := getDefaultRegion(ctx, configs)
```
jasdel, line 110: Example where we should take a look at the value of the `bool` parameter. In this case it doesn't add any value, and actually increases complexity.

config/resolve_assume_role_test.go
```diff
@@ -31,11 +31,13 @@ func TestAssumeRole(t *testing.T) {
...
 
-	config, err := LoadDefaultConfig(WithHTTPClient(client))
+	config, err := LoadDefaultConfig(context.Background(), WithHTTPClient(client))
 	if err != nil {
 		t.Fatalf("expect no error, got %v", err)
 	}
 
+	// config.HTTPClient = client
+
 	creds, err := config.Credentials.Retrieve(context.Background())
```
jasdel, line 41: is this comment still needed?

config/resolve_credentials_test.go
```diff
@@ -201,9 +201,9 @@ func TestSharedConfigCredentialSource(t *testing.T) {
 
 			var credChain []string
 
-			configSources := []Config{
+			configSources := []func(*LoadOptions) error{
```
jasdel, line 204: nit not `configSources` anymore but load options

credentials/processcreds/doc.go
```diff
@@ -32,7 +32,7 @@
 // using, the credential process will not be used.
 //
 //     // Initialize a session to load credentials.
-// 	cfg, _ := config.LoadDefaultConfig()
+// 	cfg, _ := config.LoadDefaultConfig(context.Background())
```
jasdel, line 35: nit would be better to use `context.TODO` for these examples.

jasdel (changes requested):

config/resolve.go
skotambkar: resolveDefaultAWSConfig was not hoisted as, Loaders need to be processed before this resolve occurs.

If this resolve function is hoisted the loader resolve functions would not resolve credentials correctly from env config, shared config coz anonymous credentials are already set by `resolveDefaultAWSConfig` function.

config/provider.go
skotambkar: Don't know the history behind this change. It is currently squashed.  They can be separate, but the current usage doesn't require them to be.

config/provider.go
skotambkar: The `With..` helpers were moved to config.go. This was to help improve visibility in code, and bundle them with `LoadOptions` which are defined in config.go .

config/config.go
```diff
@@ -166,14 +260,384 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+
+// getDefaultRegion returns DefaultRegion from config's LoadOptions
+func (o LoadOptions) getDefaultRegion(ctx context.Context) (string, bool, error) {
+	if len(o.DefaultRegion) == 0 {
+		return "", false, nil
+	}
+
+	return o.DefaultRegion, true, nil
+}
```
jasdel, line 311: Should `getDefaultRegion` be used to resolve the EC2 IMDS Region, if no other region is found, `UseEC2IMDSRegion` option if its not nil?

**OR** Is there somewhere else in the "chain" of configuration sources where `UseEC2IMDSRegion` option should be used?

Probably will want to document cleary on that member when in the order of configuration sources the SDK will decide to get the region from the EC2 instance.


I might be missing it but i'm not seeing how the `UseEC2IMDSRegion` option is used to derive a region.

config/shared_config.go
```diff
@@ -128,6 +129,9 @@ type SharedConfig struct {
 }
 
 // GetEnableEndpointDiscovery returns whether to enable service endpoint discovery
+// TODO: These probably should be unexported and
+// 	EnableEndpointDiscovery should be an option on LoadOptions?
+//	Will also need a resolver (qq: do these need to be resolved before other resolver?)
 func (c *SharedConfig) GetEnableEndpointDiscovery() (value, ok bool, err error) {
 	if c.EnableEndpointDiscovery == nil {
 		return false, false, nil
```
jasdel, line 137: Comment still valid? These probably shouldn't be exported.

config/shared_config.go
```diff
@@ -138,6 +142,9 @@ func (c *SharedConfig) GetEnableEndpointDiscovery() (value, ok bool, err error)
 
 // GetS3UseARNRegion returns if the S3 service should allow ARNs to direct the region
 // the client's requests are sent to.
+// TODO: GetS3UseARNRegion probably should be unexported and
+//	S3UseARNRegion should be an option on LoadOptions?
+//	Will also need a resolver (qq: do these need to be resolved before other resolver? probably not)
 func (c *SharedConfig) GetS3UseARNRegion() (value, ok bool, err error) {
 	if c.S3UseARNRegion == nil {
```
jasdel, line 149: Comment still valid? These probably shouldn't be exported.

credentials/processcreds/doc.go
```diff
@@ -32,7 +32,7 @@
 // using, the credential process will not be used.
 //
 //     // Initialize a session to load credentials.
-// 	cfg, _ := config.LoadDefaultConfig()
+// 		cfg, _ := config.LoadDefaultConfig(context.TODO())
 //
 //     // Create S3 service client to use the credentials.
 //     svc := s3.NewFromConfig(cfg)
```
jasdel, line 38: spacing?

feature/s3/manager/bucket_region.go
```diff
@@ -25,7 +25,7 @@ const bucketRegionHeader = "X-Amz-Bucket-Region"
 // For example to get the region of a bucket which exists in "eu-central-1"
 // you could provide a region hint of "us-west-2".
 //
-//	cfg, err := config.LoadDefaultConfig()
+//	cfg, err := config.LoadDefaultConfig(context.Background())
```
jasdel, line 28: Nit TODO :)

jasdel (changes requested):

config/config.go
```diff
@@ -147,6 +157,90 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+	// service and region Please see the `aws.EndpointResolver` documentation on usage.
+	EndpointResolver aws.EndpointResolver
+
+	// Retryer guides how HTTP requests should be retried in case of
+	// recoverable failures.
+	Retryer aws.Retryer
+
+	// ConfigSources are the sources that were used to construct the Config.
+	// Allows for additional configuration to be loaded by clients.
+	ConfigSources []interface{}
```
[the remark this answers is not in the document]
skotambkar: > This field was more for the SDK's load config to provide values service specific from the shared config for API clients without creating a direct coupling/dependency between the two modules. 

Not following the above reasoning. Are ConfigSources only resolved via shared config or env config?

config/config.go
```diff
@@ -166,14 +260,384 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
...
+
+// WithProcessCredentialOptions is a helper function to construct functional options
+// that sets a function to use processcreds.Options on config's LoadOptions. If process
+// credential options is set to nil, the process credential value will be ignored.
+func WithProcessCredentialOptions(v func(*processcreds.Options)) LoadOptionsFunc {
+	return func(o *LoadOptions) error {
+		o.ProcessCredentialOptions = v
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
skotambkar: Yeah this is a behavior change, and should have been called out. I thought I added a TODO somewhere, as this affects one of the tests.

config/env_config.go
```diff
@@ -254,19 +264,29 @@ func (c EnvConfig) GetSharedConfigFiles() ([]string, error) {
...
+	return bytes.NewReader(b), true, nil
 }
 
 // GetEnableEndpointDiscovery returns whether to enable service endpoint discovery
+// TODO: These probably should be unexported and
+// 	EnableEndpointDiscovery should be an option on LoadOptions?
+//	Will also need a resolver (qq: do these need to be resolved before other resolver?)
 func (c EnvConfig) GetEnableEndpointDiscovery() (value, ok bool, err error) {
 	if c.EnableEndpointDiscovery == nil {
 		return false, false, nil
```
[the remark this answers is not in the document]
skotambkar: yes, this is an open question, should these be a part of load options? no resolver calls these yet. (same for the next todo around s3UseARNRegion)

config/config.go
skotambkar: UseEC2IMDSRegion isn't resolved at the moment. There was a previously exisiting TODO around that. 

This PR didn't address that TODO. I think we should let another resolver resolve EC2IMDSRegion instead of having it here as a by product of getRegion. That way if EC2IMDSRegion can be enabled by shared config, all we have to do is implement the interface. Also helps stay consistent. 

We can change this later as all is unexported, so fine either way.  (Same reasoning for  above comment )

jasdel (approved):

skotambkar:
This PR will need to b paired with doc/example change

skotambkar:
Fixed in https://github.com/aws/aws-sdk-go-v2/pull/951

Issue #917, opened 2020-11-19
Unable to enable ClientLogMode for request signing on individual operation.


jasdel:
It is not currently possible to enable or disable request signing via `ClientLogMode` API Client `Options` on individual operations. This configuration can only be provided when the API Client is first initialized, and only if a custom `HTTPSignerV4` is not provided. (Its OK that the log option is ignored if custom signer is used.) 

#### Issues

There is also no way to enable `LogSigning` per operation call because the `addHTTPSignerV4Middleware` does not pass down logging level.  

The default v4 signer's `LogSigning` is only modified when  when API client is created without custom HTTPSignerV4.
https://github.com/aws/aws-sdk-go-v2/blob/3f8ff9a5595e4792a2167694dc678ce0ab75ef88/service/s3/api_client.go#L215-L226

When an Operation is invoked and the `addHTTPSignerV4Middleware` helper is run to add the signing middleware, the helper takes functional parameters instead of a structure. This would make it impossible to add additional parameters in the future without using a different middleware helper in the v4 package.
https://github.com/aws/aws-sdk-go-v2/blob/3f8ff9a5595e4792a2167694dc678ce0ab75ef88/service/s3/api_client.go#L207-L209
https://github.com/aws/aws-sdk-go-v2/blob/3f8ff9a5595e4792a2167694dc678ce0ab75ef88/aws/signer/v4/middleware.go#L204-L207

The `HTTPSignerV4` API client generated interface does not provide functional options for additional configurations.
https://github.com/aws/aws-sdk-go-v2/blob/3f8ff9a5595e4792a2167694dc678ce0ab75ef88/service/s3/api_client.go#L211-L213

#### Proposal

Update `v4.NewSignHTTPRequestMiddleware` to take structure for options instead of individual function parameters. This will allow future options to be passed down from the operation's adding the middleware.

```go
type SignHTTPRequestMiddlewareOptions struct {
     Credentials aws.CredentialsProvider
     Signer         HTTPSigner
     LogSigning bool
}

func NewSignHTTPRequestMiddleware(options SignHTTPRequestMiddlewareOptions) *SignHTTPRequest {
     return &SignHTTPRequest{
          credentialsProvider: options.Credentials,
          signer: options.Signer,
          logSigning: options.logSigning,
     }
}
```
[...]

PR #955, opened 2020-12-09
Update SigV4 Signer Design

skmcgrail:
Updates the SigV4 signer design to allow for overriding of signer configuration when invoking SignHTTP or PresignHTTP.

Fixes: https://github.com/aws/aws-sdk-go-v2/issues/917
Fixes: #960
Fixes: #958

### DIFF
--- a/aws/signer/v4/middleware.go
+++ b/aws/signer/v4/middleware.go
@@ -195,36 +195,36 @@ func (m *contentSHA256Header) HandleBuild(
 	return next.HandleBuild(ctx, in)
 }
 
-// SignHTTPRequestMiddlewareOptions is the configuration options for the SignHTTPRequest middleware.
+// SignHTTPRequestMiddlewareOptions is the configuration options for the SignHTTPRequestMiddleware middleware.
 type SignHTTPRequestMiddlewareOptions struct {
 	CredentialsProvider aws.CredentialsProvider
 	Signer              HTTPSigner
 	LogSigning          bool
 }
 
-// SignHTTPRequest is a `FinalizeMiddleware` implementation for SigV4 HTTP Signing
-type SignHTTPRequest struct {
+// SignHTTPRequestMiddleware is a `FinalizeMiddleware` implementation for SigV4 HTTP Signing
+type SignHTTPRequestMiddleware struct {
 	credentialsProvider aws.CredentialsProvider
 	signer              HTTPSigner
 	logSigning          bool
 }
 
-// NewSignHTTPRequestMiddleware constructs a SignHTTPRequest using the given Signer for signing requests
-func NewSignHTTPRequestMiddleware(options SignHTTPRequestMiddlewareOptions) *SignHTTPRequest {
-	return &SignHTTPRequest{
+// NewSignHTTPRequestMiddleware constructs a SignHTTPRequestMiddleware using the given Signer for signing requests
+func NewSignHTTPRequestMiddleware(options SignHTTPRequestMiddlewareOptions) *SignHTTPRequestMiddleware {
+	return &SignHTTPRequestMiddleware{
 		credentialsProvider: options.CredentialsProvider,
 		signer:              options.Signer,
 		logSigning:          options.LogSigning,
 	}
 }
 
-// ID is the SignHTTPRequest identifier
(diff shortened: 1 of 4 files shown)


aws/signer/v4/middleware.go
```diff
@@ -195,15 +195,27 @@ func (m *contentSHA256Header) HandleBuild(
...
 	credentialsProvider aws.CredentialsProvider
 	signer              HTTPSigner
+	logSigning          bool
 }
 
 // NewSignHTTPRequestMiddleware constructs a SignHTTPRequest using the given Signer for signing requests
-func NewSignHTTPRequestMiddleware(credentialsProvider aws.CredentialsProvider, signer HTTPSigner) *SignHTTPRequest {
-	return &SignHTTPRequest{credentialsProvider: credentialsProvider, signer: signer}
+func NewSignHTTPRequestMiddleware(options SignHTTPRequestMiddlewareOptions) *SignHTTPRequest {
+	return &SignHTTPRequest{
```
jasdel, line 214: should `SignHTTPRequest` be renamed to `SignHTTPRequestMiddleware` to prevent confusion with `HTTPSigner` and `Signer`  types?

aws/signer/v4/middleware.go
```diff
@@ -235,7 +247,9 @@ func (s *SignHTTPRequest) HandleFinalize(ctx context.Context, in middleware.Fina
 		return out, metadata, &SigningError{Err: fmt.Errorf("failed to retrieve credentials: %w", err)}
 	}
 
-	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime())
+	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(), func(options *SignerOptions) {
+		options.LogSigning = s.logSigning
+	})
```
jasdel, line 252: super minor, but you could avoid the additional function closure being allocated by creating a method on `SignHTTPRequest` that sets the logging, and can be passed in as a functional option.

```go
func (s *SignHTTPRequest) addLogSignerOptions(o *SignerOptions) {
     o.LogSigning = s.logSigning
)
```

```go
err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(),
     s.addLogSignerOptions,
)
```

aws/signer/v4/presign_middleware.go
```diff
@@ -30,33 +31,42 @@ type PresignedHTTPRequest struct {
...
+	LogSigning          bool
+}
+
+// PresignHTTPRequest provides the Finalize middleware for creating a
 // presigned URL for an HTTP request.
 //
 // Will short circuit the middleware stack and not forward onto the next
 // Finalize handler.
-type PresignHTTPRequestMiddleware struct {
+type PresignHTTPRequest struct {
```
jasdel, line 46: Similar to SignHTTPRequest. Since these types are in same package as `Signer` and `Presigner` they might get confusing without the `Middleware` suffix on their name. Also since their constructors have that on their name.

aws/signer/v4/presign_middleware.go
```diff
@@ -96,7 +106,9 @@ func (s *PresignHTTPRequestMiddleware) HandleFinalize(
 	}
 
 	u, h, err := s.presigner.PresignHTTP(ctx, credentials,
-		httpReq, payloadHash, signingName, signingRegion, sdk.NowTime())
+		httpReq, payloadHash, signingName, signingRegion, sdk.NowTime(), func(options *SignerOptions) {
+			options.LogSigning = s.logSigning
+		})
```
jasdel, line 111: similar a method on `PresignHTTPRequest` would prevent the need for a closure.

jasdel (commented):
I think we probably want to included the suffix `Middleware` on the `SignHTTPRequest` and `PresignHTTPRequest` types since they are in the same package as the signing and presign implementations. If they were in their own package separate from the Signer/Presigner, and HTTPSigner/HTTPPresigner types then probably wouldn't need the suffix, but since they are together, I can the type names being confusing.

jasdel (approved):

PR #964, opened 2020-12-10
aws/signer/v4: Ensure logger option passed down to sign method

jasdel:
Ensures the middleware's Logger option value is passed down to the signer's sign method via functional options.

### DIFF
--- a/aws/signer/v4/middleware_test.go
+++ b/aws/signer/v4/middleware_test.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
@@ -93,7 +93,6 @@ func TestSignHTTPRequestMiddleware(t *testing.T) {
 	cases := map[string]struct {
 		creds       aws.CredentialsProvider
 		hash        string
-		logger      logging.Logger
 		logSigning  bool
 		expectedErr error
 	}{
@@ -112,15 +111,9 @@ func TestSignHTTPRequestMiddleware(t *testing.T) {
 		"nil creds": {
 			creds: nil,
 		},
-		"with logger": {
-			creds:  unit.StubCredentialsProvider{},
-			hash:   "0123456789abcdef",
-			logger: logging.NewStandardLogger(os.Stdout),
-		},
 		"with log signing": {
 			creds:      unit.StubCredentialsProvider{},
 			hash:       "0123456789abcdef",
-			logger:     logging.NewStandardLogger(os.Stdout),
 			logSigning: true,
 		},
 	}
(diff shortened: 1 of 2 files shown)


aws/signer/v4/middleware.go
```diff
@@ -247,18 +247,18 @@ func (s *SignHTTPRequestMiddleware) HandleFinalize(ctx context.Context, in middl
 		return out, metadata, &SigningError{Err: fmt.Errorf("failed to retrieve credentials: %w", err)}
 	}
 
-	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(), s.addSignerOptions)
+	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(),
+		func(o *SignerOptions) {
```
skotambkar, line 251: what do you think about this functional option returning an error?

skmcgrail (approved):

aws/signer/v4/middleware_test.go
```diff
@@ -108,6 +112,17 @@ func TestSignHTTPRequestMiddleware(t *testing.T) {
...
 		},
+		"with logger": {
+			creds:  unit.StubCredentialsProvider{},
+			hash:   "0123456789abcdef",
+			logger: logging.NewStandardLogger(os.Stdout),
+		},
+		"with log signing": {
+			creds:      unit.StubCredentialsProvider{},
+			hash:       "0123456789abcdef",
+			logger:     logging.NewStandardLogger(os.Stdout),
```
skotambkar, line 123: How does this asset that it correctly logged signed requests?

aws/signer/v4/middleware.go
skmcgrail: I think it made sense in LoadOptions case, but I'm not sure if I agree with signing having an error for an option setter function.

aws/signer/v4/middleware_test.go
jasdel: yeah this doesn't assert anything particular is logged since the Signer is mocked out there is no actual behavior that is being checked. 

Can update the logic so that the signer does use the logger if the flag is set, and check some expected content.

aws/signer/v4/middleware.go
jasdel: we probably want to common pattern to be no error for these functional options, since their intention is to only modify configuration of a thing as it is invoked.  Potentially this could be applied to LoadOptions since, its functional options purpose are similar. The error value there may be a bit helpful for colliding options that aren't compatible. In the case of SignerOptions, the usage of the options is owned by the Signer, and not a separate entity like. The Signer it self can perform any validation as needed to determine that the options set are valid.

skmcgrail (approved):

skotambkar (approved):

PR #966, opened 2020-12-11
Make presigner use unsigned payload for presigning

skotambkar:
S3 operations need presigners to generate request with unsigned payload. 

Previously we were incorrectly only converting presigner to be unsigned for streaming operations.  This conversion to unsigned payload request, should apply on each s3 presigned url request .


jasdel:
Looks like both this Pr and https://github.com/aws/aws-sdk-go-v2/pull/967 touch the same files and sections. Maybe would be better to rebase this Pr on top of #967?

skotambkar:
closed in favor of https://github.com/aws/aws-sdk-go-v2/pull/967

PR #967, opened 2020-12-11
`presigner`: Remove content-type header if empty/nil body for streaming operation

skotambkar:
1. Presigner clients should not send content-type header if a streaming operation has an empty or nil payload. This is to reduce burden on customers as they have to sign content-type header. 

2. S3 operations need presigners to generate request with unsigned payload. Previously we were incorrectly only converting presigner to be unsigned for streaming operations. This conversion to unsigned payload request, should apply on each s3 presigned url request .


aws/middleware/content_type.go
```diff
@@ -0,0 +1,42 @@
+package middleware
```
jasdel, line 1: suggest move this to `aws/transport/http` pkg.

jasdel (approved):

Issue #952, opened 2020-12-07
SignatureDoesNotMatch error with special characters


asterikx:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone through the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
I'm generating pre-signed put URLs and hitting `SignatureDoesNotMatch` errors on special characters.

SignatureDoesNotMatch error:
```
Object Key:  1/o-2bWeAMR/some-filename_(1).pdf
Request URL: https://bucket-name.eu-central-1.amazonaws.com/1/o-2bWeAMR/some-filename_%281%29.pdf?...
```

No error:
```
Object Key:  1/o-2bWeAMR/some-filename.pdf
Request URL: https://bucket-name.eu-central-1.amazonaws.com/1/o-2bWeAMR/some-filename.pdf?...
```

So `(` and/or `)` seems to cause the error in this example.

EDIT: Spaces and Umlauts also lead to this error.

**Version of AWS SDK for Go?**
`v0.30.0`

**Version of Go (`go version`)?**
`1.15`

**To Reproduce (observed behavior)**
```go
req, err := c.PresignPutObject(
  ctx,
  &s3.PutObjectInput{
    Bucket: aws.String("bucket-name"),
    Key:    aws.String("1/o-2bWeAMR/some-filename_(1).pdf"),
  },
  s3.WithPresignExpires(in.ExpiresIn),
)
```

**Expected behavior**
Pre-sign client handles special characters correctly (as `aws-sdk-go` `v1` did)

asterikx:
@jasdel can we fix this somehow for the time being? If you can give me some pointers to what needs to be done, I might be able to contribute.

jasdel:
HI @asterikx I'm investigating this issue, but don't have a solid workaround for this issue yet. It looks like logging for presigned SDK computed signature isn't working. Ive created https://github.com/aws/aws-sdk-go-v2/issues/958 to track request signing not working. We'll most likely need to resolve why presign logging isn't working with PresignClient in order to resolve the issue the request signature.

jasdel:
[...]

PR #969, opened 2020-12-11
 codegen: Update PresignClient to default use signer and options

jasdel:
Updates the PresignClient generation to use the same Signer initialization as the API client. In addition, the PresignClient was updated to ensure the PresignOptions are passed down and used at each level.

Change updates the PresignClient and API client to use the same `v4.Signer` initialization if a (pre)signer has not been configured. Also updates the PresignClient to have PresignOption as a member instead of separate fields similar to API client's design. This allows the presign client to more easily pass down its configuration and be modified on a per operation base.

All changes to `aws/signer/v4` are because this change is based on, https://github.com/aws/aws-sdk-go-v2/pull/964, and will rebased once that that PR is merged. 

* [x] Will need to be rebased with https://github.com/aws/aws-sdk-go-v2/pull/967 after that PR is merged.

Fixes https://github.com/aws/aws-sdk-go-v2/issues/952

### DIFF
--- a/.changes/next-release/service.s3-bugfix-1607970870724256000.json
+++ b/.changes/next-release/service.s3-bugfix-1607970870724256000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.s3-bugfix-1607970870724256000",
+ "SchemaVersion": 1,
+ "Module": "service/s3",
+ "Type": "bugfix",
+ "Description": "Fixes PresignClient S3 signing behavior for special characters, [#969](https://github.com/aws/aws-sdk-go-v2/pull/969)",
+ "MinVersion": "",
+ "AffectedModules": null
+}


aws/signer/v4/middleware.go
```diff
@@ -247,18 +247,18 @@ func (s *SignHTTPRequestMiddleware) HandleFinalize(ctx context.Context, in middl
 		return out, metadata, &SigningError{Err: fmt.Errorf("failed to retrieve credentials: %w", err)}
 	}
 
-	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(), s.addSignerOptions)
+	err = s.signer.SignHTTP(ctx, credentials, req.Request, payloadHash, signingName, signingRegion, sdk.NowTime(),
+		func(o *SignerOptions) {
+			o.Logger = middleware.GetLogger(ctx)
+			o.LogSigning = s.logSigning
+		})
```
skmcgrail, line 254: If you wanted to avoid the function enclosure, we could switch HandleFinalize to be a value receiver. Then you could override the logger on that copy of the middleware with the context value.

jasdel: Sounds good can update to use that pattern.

jasdel: hmm actually this would require adding `Logger` member to the middleware type, which is probably out of place since its not a value that would only be present for this method call.  Might be better to just keep the closure at that point.

service/s3/api_client.go
```diff
@@ -359,7 +371,7 @@ func (c *PresignClient) convertToPresignMiddleware(stack *middleware.Stack, opti
 	}
 	// add middleware to set expiration for s3 presigned url, if expiration is set to
 	// 0, this middleware sets a default expiration of 900 seconds
-	err = stack.Build.Add(&s3cust.AddExpiresOnPresignedURL{Expires: c.expires}, middleware.After)
+	err = stack.Build.Add(&s3cust.AddExpiresOnPresignedURL{Expires: c.Expires}, middleware.After)
 	if err != nil {
 		return err
 	}
```
skotambkar, line 377: learnings from waiter implementation: since time.Duration is alias for int64.

Here c.Expires can be negative. The `AddExpiresOnPresignedURL` middleware does not handle the negative expiration case.  Should return an error.

skotambkar (approved):
Looks good to me. Could you also update the missed case of handling negative expiration input for s3 presigner?

skotambkar (approved):

PR #979, opened 2020-12-14
codegen: fix union serialization cases

### LINKED ISSUE #978: serializers: use of illegal fallthrough statement in a type switch
We use fallthrough in a type switch statement in union serializers. Needs a simple fix to not use fallthrough for unknown types, andjust use the default statement.

jasdel:
Fixes the cases of generated union serializer functions to use the correct types for union members.

Fixes #978

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
@@ -11,6 +11,7 @@ import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.go.codegen.GoValueAccessUtils;
 import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SymbolUtils;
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
```diff
@@ -183,17 +184,18 @@ protected void serializeUnion(GenerationContext context, UnionShape shape) {
...
-                String exportedMemberName = symbol.getName() + symbolProvider.toMemberName(member);
+                Symbol memberSymbol = SymbolUtils.createValueSymbolBuilder(
+                        symbolProvider.toMemberName(member),
+                        symbol.getNamespace()
+                ).build();
 
-                writer.openBlock("case *$L:", "", exportedMemberName, () -> {
+                writer.openBlock("case *$T:", "", memberSymbol, () -> {
                     writer.write("av := value.MemberElement($S)", memberSymbol.getName());
                     target.accept(getMemberSerVisitor(member, "uv.Value()", "av"));
```
skotambkar, line 194: Another fix is required for `MemberElement` usage on line 190. Instead of passing in the string, have to construct a xml start element and pass it in.

skotambkar (approved):

skmcgrail (approved):

Issue #919, opened 2020-11-19
S3: Generate ResponseError interface for retrieving RequestID and HostID


skmcgrail:
Due to the [ResponseError](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/internal/s3shared#ResponseError) being in the s3shared package, users wanting to use `errors.As` to retrieve the `RequestID` and `HostID` together would need to define their own interface definition. We should generate a duplicate interface definition in the client package to allow better access to this wrapped error metadata.

Issue #918, opened 2020-11-19
S3: Unable to retrieve response HostID from ResponseMetadata member.


skmcgrail:
Users are not able to retrieve the S3 response `HostID` from the `ResponseMetadata` member on the operation output shape. This us due to the accessor method being placed in an the internal s3shared package.

See https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/internal/s3shared#GetHostIDMetadata

PR #983, opened 2020-12-16
Export s3shared ResponseError and GetHostIDMetadata in S3 and S3 Control client packages.

skmcgrail:
Exports an interface definition in the S3 and S3 Control client packages for the `s3shared.ResponseError` type. Exports a wrapper around the `s3shared.GetHostIDMetadata`.

Fixes https://github.com/aws/aws-sdk-go-v2/issues/919
Fixes https://github.com/aws/aws-sdk-go-v2/issues/918


service/s3/api_client.go
```diff
@@ -259,6 +259,22 @@ func disableAcceptEncodingGzip(stack *middleware.Stack) error {
 	return acceptencodingcust.AddAcceptEncodingGzip(stack, acceptencodingcust.AddAcceptEncodingGzipOptions{})
 }
 
+// ResponseError provides the HTTP centric error type wrapping the underlying error
+// with the HTTP response value and the deserialized RequestID.
+type ResponseError interface {
+	error
+
+	ServiceHostID() string
+	ServiceRequestID() string
```
skotambkar, line 268: Nit: Just curious if the following could have been an option to expose this. 

Mostly thinking about additions to metadata would break interface. Though that's rare and probably can just introduce an extra interface in that case.  So this option can be ruled out but still like to know your thoughts. 

```
type ResponseError interface {
   error
    
   ServiceMetadata(key interface{}) (interface{}, error)
}

key being S3.HostIDKey{} or aws.RequestIDKey{} or any other metadata type added in future for S3. 
```

This adds the overhead of specifying a key type and return of a loose value. This does pair up with how context works though.

service/s3control/api_client.go
```diff
@@ -276,6 +276,22 @@ func addResponseErrorMiddleware(stack *middleware.Stack) error {
...
+	ServiceHostID() string
+	ServiceRequestID() string
+}
+
+// GetHostIDMetadata retrieves the host id from middleware metadata returns host id
+// as string along with a boolean indicating presence of hostId on middleware
+// metadata.
+func GetHostIDMetadata(metadata middleware.Metadata) (string, bool) {
+	return s3shared.GetHostIDMetadata(metadata)
+}
```
skotambkar, line 293: In case of an error, do we not return this `middleware.Metadata`? If not, should we?

skotambkar (commented):
few thoughts/questions

service/s3control/api_client.go
jasdel: The metadata is primarily for the response since the errors can use layering and dont use the metadata. Though we could consider expanding the `smithy.OperationFailure` error to include metadata

skotambkar: https://github.com/awslabs/smithy-go/issues/243

service/s3/api_client.go
jasdel: ServiceResponseID is a member of the `aws/transport/http#ResponseError` so in order to expand on that  method the specific method would be needed. This is a part of the core difference between errors and responses. Errors can be layered with interfaces matching specific layers and metadata. Whereas output cannot be layered and only option is via metadata accessors.

This does make error and response slightly asymmetric, but the purpose fits in their common use cases better.  If only metadata were used on the errors, it wouldn't be possible to use strong typing to layer errors the way that is being done between, smithy's ResponseError, aws's ResponseError, and S3's ResponseError.

jasdel (approved):

PR #984, opened 2020-12-16
feature/ec2/imds: Move the ec2imds module

jasdel:
Moves the `ec2imds` module from the root of the SDK to a nested module under the `feature` folder. This improvement brings the EC2 IMDS client inline with other higher level features such as `s3/manager` and `dynamodb/attributevalue`

### DIFF
--- a/.changes/next-release/config-feature-1608146507258154000.json
+++ b/.changes/next-release/config-feature-1608146507258154000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "config-feature-1608146507258154000",
+ "SchemaVersion": 1,
+ "Module": "config",
+ "Type": "feature",
+ "Description": "Update to use moved feature/ec2/imds module",
+ "MinVersion": "",
+ "AffectedModules": null
+}


feature/ec2/imds/api_client.go
```diff
@@ -1,4 +1,4 @@
-package ec2imds
```
skotambkar, line 1: is this change required? Kinda like `ec2imds` parity with v1 sdk.

skotambkar (approved):
Would want documentation update to be paired with this change.

feature/ec2/imds/api_client.go
jasdel: good point I'll look at the doc branch and update these.

skotambkar: Wait, was your ^ comment for documentation update ask,  OR on my comment to not change package name from `ec2imds` to `imds`?

jasdel: https://github.com/aws/aws-sdk-go-v2/pull/986

PR #990, opened 2020-12-19
refactor shared config Loading

skotambkar:
PR for Shared Config Loading

Changes:
* Separates shared config and shared credentials files.
* Updates shared config handling to be consistent with AWS SDKs and CLI
* Updates test 

TODO:

[ x ] More TESTING
[ x ] Error handling. 
[ x ] Logs things that are dropped. 
[ x ] Documentation update
[ x ] Config SourceFile should be a list of all filenames that were used to get Config

### DIFF
--- a/config/shared_config.go
+++ b/config/shared_config.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/awslabs/smithy-go/logging"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -13,6 +12,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/internal/ini"
+	"github.com/aws/smithy-go/logging"
 )
 
 const (
--- a/config/shared_config_test.go
+++ b/config/shared_config_test.go
@@ -4,15 +4,16 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/internal/ini"
-	"github.com/awslabs/smithy-go/logging"
-	"github.com/awslabs/smithy-go/ptr"
 	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
 	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/ini"
+	"github.com/aws/smithy-go/logging"
+	"github.com/aws/smithy-go/ptr"
 )
 
 var _ regionProvider = (*SharedConfig)(nil)
--- a/config/example_test.go
+++ b/config/example_test.go
@@ -3,8 +3,6 @@ package config_test
 import (
 	"context"
 	"fmt"
-	"github.com/awslabs/smithy-go/middleware"
-	smithyhttp "github.com/awslabs/smithy-go/transport/http"
 	"log"
 	"net/http"
 	"path/filepath"
--- a/config/resolve_credentials_test.go
+++ b/config/resolve_credentials_test.go
@@ -3,8 +3,6 @@ package config
 import (
 	"context"
 	"fmt"
-	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
-	"github.com/awslabs/smithy-go/middleware"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -16,7 +14,9 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/smithy-go/middleware"
 )
 
 func swapECSContainerURI(path string) func() {
--- a/config/resolve_test.go
+++ b/config/resolve_test.go
@@ -3,7 +3,6 @@ package config
 import (
 	"bytes"
 	"context"
-	"github.com/awslabs/smithy-go/logging"
 	"io/ioutil"
 	"net/http"
 	"testing"
@@ -13,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting/unit"
+	"github.com/aws/smithy-go/logging"
 )
 
 func TestResolveCustomCABundle(t *testing.T) {


config/go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v0.0.0-20201216182737-05d6a8e2a8df
 	github.com/aws/aws-sdk-go-v2/service/sts v0.30.0
 	github.com/awslabs/smithy-go v0.4.1-0.20201216214517-20e212c92831
+	github.com/go-delve/delve v1.5.1
```
skotambkar, line 11: Remove this

internal/ini/visitor.go
```diff
@@ -125,8 +164,27 @@ func (t Sections) List() []string {
...
+	Name string
+
+	// values are the values within parsed profile
 	values values
+
+	// Errors is the list of errors
+	Errors []error
+
+	// Logs is the list of logs
+	Logs []string
```
skotambkar, line 177: This Logs field is never used as of now. The ideas was to stash logs, and only call them if related profile is referenced. But that doesn't work well for profiles that are dropped without a similar profile used in place of it.

internal/ini/visitor.go
```diff
@@ -125,8 +164,27 @@ func (t Sections) List() []string {
...
+
+	// SourceFile is the INI Source file from where this section
+	// was retrieved.
+	SourceFile string
```
skotambkar, line 181: This probably should be a list of strings. But not clear how that will work. Coz if a reference is no longer relevant after merging multiple files, we will have to remove it?

config/shared_config_test.go
```diff
@@ -264,64 +258,64 @@ func TestLoadSharedConfigFromFile(t *testing.T) {
...
 		},
 		6: {
-			Profile: "full_profile",
+			Profile: "profile full_profile",
 			Expected: SharedConfig{
 				Credentials: aws.Credentials{
 					AccessKeyID:     "full_profile_akid",
 					SecretAccessKey: "full_profile_secret",
-					Source:          fmt.Sprintf("SharedConfigCredentials: %s", testConfigFilename),
+					Source:          fmt.Sprintf("SharedConfigCredentials: %s", ""),
```
skotambkar, line 299: We don't have data for source name when we loadConfigFromSection.  But this test is not really needed.

config/shared_config.go
```diff
@@ -192,15 +200,56 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+		logger, found, err = getLogger(ctx, configs)
+		if err != nil {
+			return SharedConfig{}, err
+		}
+		if !found {
+			logger = logging.NewStandardLogger(os.Stderr)
+		}
+	}
+
+	return NewSharedConfig(ctx, profile, sharedConfigsLoadInfo{
```
jasdel, line 235: Input parameter not exported, `sharedConfigsLoadInfo`.  Consider changing `sharedConfigsLoadInfo` to functional inputs, e.g. `LoadSharedConfigProfileOptions`, and change `NewSharedConfig` to `LoadSharedConfigProfile`

In addition, the `CredentialsFiles` and `ConfigFiles` are optional parameters and `credentialsFiles = DefaultSharedCredentialsFiles` is done internally to the `LoadSharedConfigProfile` function if not specified.

config/shared_config.go
```diff
@@ -192,15 +200,56 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+		CredentialsFiles:   credentialsFiles,
+		ConfigurationFiles: configurationFiles,
+		Logger:             logger,
+	})
+}
+
+// sharedConfigsLoadInfo struct contains Values that can be used to load the config.
+type sharedConfigsLoadInfo struct {
```
jasdel, line 243: this would become `LoadSharedConfigOptions` with the above

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
 //
 // For example, given two files A and B. Both define credentials. If the order
 // of the files are A then B, B's credential values will be used instead of A's.
-func NewSharedConfig(profile string, filenames []string) (SharedConfig, error) {
-	if len(filenames) == 0 {
-		return SharedConfig{}, fmt.Errorf("no shared config files provided")
+func NewSharedConfig(ctx context.Context, profile string, options sharedConfigsLoadInfo) (SharedConfig, error) {
+	if len(options.ConfigurationFiles) == 0 && len(options.CredentialsFiles) == 0 {
+		return SharedConfig{}, fmt.Errorf("no shared config or shared credentials options provided")
```
jasdel, line 264: this would become `LoadSharedConfigProfile` with the above

config/shared_config.go
jasdel: And used as functional options.

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+			sections.SetSection(section, v)
+		}
+	}
+	return nil
+}
+
+func processCredentialsSections(ctx context.Context, sections ini.Sections, logger logging.Logger) error {
+
+	prefix := "profile"
+	for _, section := range sections.List() {
```
jasdel, line 359: This would capture `profileBar`, not just `profile bar`.  Probably want to move this to a const shared by both config and credentials files.

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+			}
+		}
+	}
+	return nil
+}
+
+type parsedINIFile struct {
 	Filename string
 	IniData  ini.Sections
 }
```
jasdel, line 380: can delete if not used.

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+						sectionName, s3UseARNRegionKey, dstSection.SourceFile[s3UseARNRegionKey],
+						s3UseARNRegionKey, srcSection.SourceFile[s3UseARNRegionKey]))
+			}
+
+			dstSection.UpdateValue(s3UseARNRegionKey, val)
+			dstSection.UpdateSourceFile(s3UseARNRegionKey, srcSection.SourceFile[s3UseARNRegionKey])
+		}
+
+		// role duration seconds key update
+		if srcSection.Has(roleDurationSecondsKey) {
```
jasdel, line 692: Nit suggest grouping near the role arn key, etc blocks above.

config/shared_config.go
jasdel, line 359: These checks for profile prefix may need to be a bit more robust, So that it handles other kind of whitespace like tabs.

e.g. `[profile\tfoo]`  would not be caught by the prefix check. Maybe the ini parser squashes `\t` into ` `?

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+			}
+		}
+	}
+
+	// rename sections to remove `profile ` prefixing to match with credentials file.
+	// if default is already present, it will be dropped.
+	for _, section := range sections.List() {
+		if strings.HasPrefix(section, prefix) {
+			v, ok := sections.GetSection(section)
+			if !ok {
```
jasdel, line 332: Also probably need a check that section name isn't empty after `profile ` prefix is stripped off.

config/shared_config.go
```diff
@@ -210,78 +259,478 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+
+			// assign non-prefixed name to section
+			v.Name = section
+			sections.SetSection(section, v)
+		}
+	}
+	return nil
+}
+
+func processCredentialsSections(ctx context.Context, sections ini.Sections, logger logging.Logger) error {
```
jasdel, line 356: What happens if section is `[profile]` ?   Assuming this would be considered a valid name. Probably want special case test for credentials file with that name, to ensure its not broken later.

config/shared_config.go
```diff
@@ -337,26 +793,34 @@ func (c *SharedConfig) setFromIniFiles(profiles map[string]struct{}, profile str
...
+		for _, v := range section.SourceFile {
+			sources = append(sources, v)
+		}
+
+		return SharedConfigProfileNotExistError{
+			Filename: sources,
+			Profile:  profile,
+			Err:      nil,
+		}
+	}
```
jasdel, line 816: probably want a different error since this case isn't expected to happen due to the ini section processing above.

config/shared_config_test.go
```diff
@@ -527,3 +503,248 @@ func TestLoadSharedConfig(t *testing.T) {
...
+		LoadFn        func(context.Context, configs) (Config, error)
+		Expect        SharedConfig
+		ExpectLog     string
+		Err           string
+	}{
+		"duplicate profiles in the configuration files": {
+			LoadOptionFns: []func(*LoadOptions) error{
+				WithSharedConfigProfile("duplicate-profile"),
+				WithSharedConfigFiles([]string{"testdata/load_config"}),
+				WithSharedCredentialsFiles([]string{"testdata/empty_creds_config"}),
```
jasdel, line 523: nit filepath.Join

config/shared_config_test.go
```diff
@@ -527,3 +503,248 @@ func TestLoadSharedConfig(t *testing.T) {
...
+		"profile prefix used in credentials files": {
+			LoadOptionFns: []func(*LoadOptions) error{
+				WithSharedConfigProfile("unused-profile"),
+				WithSharedConfigFiles([]string{"testdata/empty_creds_config"}),
+				WithSharedCredentialsFiles([]string{"testdata/load_credentials"}),
+				WithLogConfigurationWarnings(true),
+				WithLogger(logger),
+			},
+			LoadFn:    loadSharedConfig,
+			ExpectLog: "profile unused-profile is ignored. A profile with the \"profile \" prefix is invalid for the shared credentials file",
```
jasdel, line 590: nit could use ` characters instead of using \"

config/testdata/load_config
```diff
@@ -0,0 +1,19 @@
...
+region = ap-south-1
+
+[profile default]
+region = ap-north-1
+
+[profile duplicate-profile]
+region = us-east-1
+
+[profile duplicate-profile]
+region = us-west-2
```
jasdel, line 11: If not already present, would be helpful to unit test that cover validating that sections that should be dropped are correctly done so.

```ini
# Config file
[default]
[profile default]

[profile foo]
[foo]
```
```ini
# credentials file
[profile foo]
[foo]
```

internal/ini/ini.go
```diff
@@ -34,13 +34,13 @@ func Parse(f io.Reader) (Sections, error) {
...
 // ParseBytes will parse the given bytes and return the parsed sections.
-func ParseBytes(b []byte) (Sections, error) {
+func ParseBytes(b []byte, path string) (Sections, error) {
 	tree, err := ParseASTBytes(b)
 	if err != nil {
 		return Sections{}, err
 	}
 
-	v := NewDefaultVisitor()
+	v := NewDefaultVisitor(path)
```
jasdel, line 43: suggest changing `path` here to empty string and not add extra parameter to `ParseBytes`

config/shared_config.go
```diff
@@ -232,58 +229,75 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+// LoadSharedConfigOptions struct contains optional values that can be used to load the config.
+type LoadSharedConfigOptions struct {
 
 	// CredentialsFiles are the shared credentials files
 	CredentialsFiles []string
 
-	// ConfigurationFiles are the shared config files
-	ConfigurationFiles []string
+	// ConfigFiles are the shared config files
+	ConfigFiles []string
```
jasdel: what to clarify behavior of default if fields are nil vs empty vs set.

config/shared_config.go
```diff
@@ -232,58 +229,75 @@ func loadSharedConfig(ctx context.Context, configs configs) (Config, error) {
...
+		fn(&option)
+	}
+
+	if len(option.ConfigFiles) == 0 {
+		option.ConfigFiles = DefaultSharedConfigFiles
+	}
+
+	if len(option.CredentialsFiles) == 0 {
+		option.CredentialsFiles = DefaultSharedCredentialsFiles
 	}
```
jasdel: suggest doing nil check here instead of len so caller could specify an empty list for the `ConfigFiles`/`CredentialsFiles`

jasdel (approved):

Issue #968, opened 2020-12-11
S3 Client Must Load Client Specific Options from aws.Config#ConfigSources


skmcgrail:
See `v0.24.0` behavior: https://github.com/aws/aws-sdk-go-v2/blob/v0.24.0/service/s3/api_client_config.go#L9

PR #991, opened 2020-12-21
Adds resolver function for loading S3 client specific config

skotambkar:
* Adds resolver function for loading client specific config

Fixes https://github.com/aws/aws-sdk-go-v2/issues/968

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/ResolveClientConfig.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/ResolveClientConfig.java
@@ -56,8 +56,8 @@ public class ResolveClientConfig implements GoIntegration {
             writer.addUseImports(AwsGoDependency.AWS_CORE);
             writer.openBlock("func $L(cfg aws.Config, o *Options) error {", "}",
                     CLIENT_CONFIG_RESOLVER_FUNC, () -> {
-                writer.openBlock("if cfg.$L == nil || len(cfg.$L) == 0 {", "}",
-                        CONFIG_SOURCE_CONFIG_NAME, CONFIG_SOURCE_CONFIG_NAME,
+                writer.openBlock("if len(cfg.$L) == 0 {", "}",
+                        CONFIG_SOURCE_CONFIG_NAME,
                         () -> writer.write("return nil")
                 );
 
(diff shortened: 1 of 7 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/ResolveClientConfig.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.smithy.go.codegen.SymbolUtils;
+import software.amazon.smithy.go.codegen.integration.GoIntegration;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.utils.ListUtils;
+
+/**
+ * Registers additional client specific configuration fields
+ */
+public class ResolveClientConfig implements GoIntegration {
```
jasdel, line 20: This class should probably be named specific to S3 since the option being added are specific to those services. Also since the internals of this class aren't really generic, and pretty specific to the S3 options.

service/s3/api_client.go
```diff
@@ -241,6 +243,21 @@ func addRetryMiddlewares(stack *middleware.Stack, o Options) error {
 	return retry.AddRetryMiddlewares(stack, mo)
 }
 
+// resolves client config
+func resolveClientConfig(cfg aws.Config, o *Options) error {
+	if cfg.ConfigSources == nil || len(cfg.ConfigSources) == 0 {
+		return nil
+	}
```
jasdel, line 250: Nit could be simplified to `len(cfg.ConfigSources) == 0` since that returns 0 if the slice is nil.  Or remove the check all together and let `s3sharedconfig.ResolveUseARNRegion` handle it.

jasdel (commented):
Should update the gohash's for s3 and s3control modules to depend on the update with service/internal/s3shared having the config load behavior.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/ResolveClientConfig.java
skotambkar: We can extend the class easily to be used by other services. I would want to leave the naming as is.  

It seems tightly coupled as we have predicates written for s3shared at the moment.

jasdel (approved):

PR #1017, opened 2021-01-05
 aws/signer/v4: Fix presign hoisting for x-amz-tagging header

### LINKED ISSUE #1016: service/s3: presigned PutObject with Tagging set does not add tagging to object in S3 bucket.
  
**Describe the bug**
Objects uploaded to S3  ith presign Put Object URL with`Tagging` parameter set, do not have the tagging metadata. The SDK hoists the `x-amz-tagging` header to the query string, which is ignored by S3. S3 requires the `x-amz-tagging` to be a signed header.

v0.31.0`

v1.15.x

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"strings"

	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
)

func main() {
	bucket := os.Args[1]
	key := os.Args[2]
	fileToPut := os.Args[3]

	presigned, err := getPresignedPut(context.TODO(), bucket, key)
	if err != nil {
		log.Fatalf("failed to get presigned URL, %v", err)
	}

	var headerStr strings.Builder
	for k, vs := range presigned.SignedHeader {
		for _, v := range vs {
			fmt.Fprintf(&headerStr, "-H \"%s: %s\" \\\n", k, v)
		}
	}
	fmt.Printf("curl -X %s \\\n%s-d @%s \\\n%q\n",
		presigned.Method, headerStr.String(), fileToPut, presigned.URL)
}

func getPresignedPut(ctx context.Context, bucket, key string) (*v4.PresignedHTTPRequest, error) {
	if err != nil {
		return nil, fmt.Errorf("failed to load config, %w", err)
	}

	client := s3.NewPresignClient(s3.NewFromConfig(cfg))

	presigned, err := client.PresignPutObject(ctx, &s3.PutObjectInput{
		Bucket:  &bucket,
		Key:     &key,
		Tagging: aws.String("mytagname=mytagvalue"),
	})
	if err != nil {
		return nil, fmt.Errorf("failed to get PutObject presigned URL, %w", err)
	}

	return presigned, nil
}
```

**Expected behavior**
To be able to put the object to S3 with the tagged header.

**Additional context**

jasdel:
The `x-amz-tagging` header should not be hoisted to the query string and must be sent as a signed header. If the value is hoisted to the query string in a presigned Put, it is ignored by Amazon S3.

Fixes #1016

### DIFF
--- a/config/go.mod
+++ b/config/go.mod
@@ -3,11 +3,11 @@ module github.com/aws/aws-sdk-go-v2/config
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2 v0.31.0
+	github.com/aws/aws-sdk-go-v2 v0.31.1-0.20210105194811-58b543144e2a
 	github.com/aws/aws-sdk-go-v2/credentials v0.2.0
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v0.1.0
 	github.com/aws/aws-sdk-go-v2/service/sts v0.31.0
-	github.com/aws/smithy-go v0.5.0
+	github.com/aws/smithy-go v0.5.1-0.20210104190327-c7045c94c1ec
 	github.com/google/go-cmp v0.5.4
 )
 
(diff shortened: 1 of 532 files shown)


aws/signer/internal/v4/headers.go
```diff
@@ -53,6 +53,7 @@ var RequiredSignedHeaders = Rules{
 			"X-Amz-Storage-Class":                                         struct{}{},
 			"X-Amz-Website-Redirect-Location":                             struct{}{},
 			"X-Amz-Content-Sha256":                                        struct{}{},
+			"X-Amz-Tagging":                                               struct{}{},
```
skotambkar, line 56: 1. Is this a valid change for any other rest based service apart from `s3` which use this Header?

skotambkar (approved):

skmcgrail (approved):

aws/signer/internal/v4/headers.go
jasdel: Yeah this would be valid for all other services, as not hoisting headers is valid approach. In addition S3 is the only service that sends this header for a presigned url.

Issue #908, opened 2020-11-17
LocationRestraint always empty on GetBucketLocation call


jcarter3:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
`GetBucketLocation` returns empty LocationConstraint

**Version of AWS SDK for Go?**
Example: v0.29.0

**Version of Go (`go version`)?**
go version go1.15.5 linux/amd64

**To Reproduce (observed behavior)**
```go
cfg, err := config.LoadDefaultConfig()
// check err
svc := s3.NewFromConfig(cfg)
bucketLocation, err := svc.GetBucketLocation(context.Background(), &s3.GetBucketLocationInput{Bucket: aws.String("somebucket")})
// check err
fmt.Printf("bucket location: %v\n", bucketLocation.LocationConstraint)
// prints empty location for all buckets in all regions, (even those outside us-east-1)
```
**Expected behavior**
For the region that the bucket exists in to be returned (only nil/empty for buckets in us-east-1, per the documentation).

**Additional context**
I've validated that the proper region is returned via the aws cli

LumpKim:
Is any solution for this? I had same issue about `s3.client.GetBucketLocationOutput.LocationCostraint` is empty.

jasdel:
Thanks for reporting this issue @jcarter3  and @LumpKim  we're investigating what the cause of this issue is.

Chandrian:
I'm having the same issue, I tried with an assume role using the assume role workaround and getting the same empty result.
This was working in older v2 version. I'm using RC now


ctx := context.TODO()
cfg, err = config.LoadDefaultConfig(ctx,
			//config.WithClientLogMode(aws.LogSigning|aws.LogRequestWithBody|aws.LogResponseWithBody),
			// This can be remove when the following bug has been fixed and dependencies updated to latest
			// https://github.com/aws/aws-sdk-go-v2/issues/914
			//config.WithClientLogMode(aws.LogSigning|aws.LogRequestWithBody|aws.LogResponseWithBody),
			config.WithAPIOptions([]func(stack *middleware.Stack) error{
[...]

PR #1027, opened 2021-01-08
S3: adds custom deserializer for get-bucket-location operation

skotambkar:
* Code generates a custom deserializer for s3: get-bucket-location api operation. This implementation differs from how v1 sdk handles custom deserialization.  V1 uses regex matching on response. In this change, we make use of already generated deserializer functions. 

* Adds unit test to validate the change.

* Fixes aws/aws-sdk-go-v2#908

### DIFF
--- a/service/s3/api_op_GetBucketLocation.go
+++ b/service/s3/api_op_GetBucketLocation.go
@@ -190,7 +190,7 @@ func (m *awsRestxml_deserializeOpGetBucketLocation_custom) HandleDeserialize(ctx
 	return out, metadata, err
 }
 
-// $L is helper to swap in a custom deserializer
+// Helper to swap in a custom deserializer
 func swapDeserializerHelper(stack *middleware.Stack) error {
 	_, err := stack.Deserialize.Swap("OperationDeserializer", &awsRestxml_deserializeOpGetBucketLocation_custom{})
 	if err != nil {


service/s3/api_op_GetBucketLocation.go
```diff
@@ -133,6 +143,62 @@ func addOperationGetBucketLocationMiddlewares(stack *middleware.Stack, options O
...
+			Err:      fmt.Errorf("failed to decode response body, %w", err),
+			Snapshot: snapshot.Bytes(),
+		}
+	}
+
+	return out, metadata, err
+}
+
+// $L is helper to swap in a custom deserializer
+func swapDeserializerHelper(stack *middleware.Stack) error {
```
jasdel, line 194: nil `$L` here. probably should be `%s` in codegen.

service/s3/api_op_GetBucketLocation.go
```diff
@@ -133,6 +143,62 @@ func addOperationGetBucketLocationMiddlewares(stack *middleware.Stack, options O
...
+	_, err := stack.Deserialize.Swap("OperationDeserializer", &awsRestxml_deserializeOpGetBucketLocation_custom{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
jasdel, line 200: is there any reasonable way this can be done in the code generator?  Replace at codegen time vs runtime is something we should look at how the codegen can be updated to support custom (de)serializers.

Don't have to do that here, but we should be looking at ways the codegen can be updated to support that.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3GetBucketLocation.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param model          is the generation model.
+     * @param symbolProvider is the symbol provider.
+     * @param operation      is the operation shape for which custom deserializer is written.
+     */
+    private void writeCustomDeserializer(
+            GoWriter goWriter,
+            Model model,
+            SymbolProvider symbolProvider,
+            OperationShape operation
+    ) {
```
jasdel, line 139: is there anyway this could reuse the existing deserializer generator? This duplicates a lot of behavior from the generic deserializer. I think we need to investigate how the deserializers can be extended and augmented with custom behavior. Because the only thing that changes here is how the response body document is deserialized.

Ideally codegen generates everything the same, but the shape GetBucketLocationOutput deserializer func has custom behavior instead of duplicating the operation deserializer as well.

jasdel: Keeping this deserializer and regular deserializer in parity adds an extra burden that could lead to unexpected divergences unless they share the same API.

service/s3/api_op_GetBucketLocation.go
skotambkar: Since middlewares are added at runtime, replacing them at runtime made sense. Customizing the code generator is a larger change, but not needed.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3GetBucketLocation.java
skotambkar: We reuse the deserializers for member shapes. The top level deserializer function is generated via smithy-go. This custom deserializer generates a custom top level deserializer function, and uses the existing member deserializer function. 

Best case would be able to override the deserializer function method, and be able to override it for a specific operation shape. But seems unnecessary for a handful of operations that do not follow the protocol pattern.

service/s3/api_op_GetBucketLocation.go
jasdel: One benefit of code generator taking care of this would be to prevent runtime allocations churn that never would be needed, and dead code that would never be used. Basically cleaner and less overhead for something that will never be used.

Also agree, right now this is probably too big of a change to take today, but we should be thinking about this long term.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3GetBucketLocation.java
jasdel: Yeah this change is an artifact of the (de)serializer generators not being too rigid in its capabilities. This is something we should be looking at how we can improve in the long term. Creating duplicate code is definitely not a long term solution, and will seriously hurt maintainability in the future.

service/s3/api_op_GetBucketLocation.go
skotambkar: I fixed this before, but didn't regenerate the  service. Will do. Thanks

jasdel (approved):

Chandrian:
probably related, but just in case, putting it here. To work around this issue, I was trying to use HeadBucket and get the value of the "X-Amz-Bucket-Region" header, but getting the following error message: 

failed to call service: S3, operation: HeadBucket, error: https response error StatusCode: 301, RequestID: xxx, HostID: xxx, api error MovedPermanently: Moved Permanently

jasdel:
Hi @Chandrian I suggest using the SDK's [feature/s3/manager#GetBucketRegion](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#GetBucketRegion) utility instead of the API call directly. The `HeadBucket` and `GetBucketLocation` operations have some unexpected behavior depending on how they are called. The SDK's `GetBucketRegion` utility takes care of that and should always be able to return the region a bucket was created in.

Chandrian:
> Hi @Chandrian I suggest using the SDK's [feature/s3/manager#GetBucketRegion](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#GetBucketRegion) utility instead of the API call directly. The `HeadBucket` and `GetBucketLocation` operations have some unexpected behavior depending on how they are called. The SDK's `GetBucketRegion` utility takes care of that and should always be able to return the region a bucket was created in.

Thank you, GetBucketRegion is working perfectly

Issue #1021, opened 2021-01-06
STS Presign Client


enderv:
To use aws-iam-authenticator need a [signed GetCallerIdentity request](https://github.com/kubernetes-sigs/aws-iam-authenticator#api-authorization-from-outside-a-cluster) 

Currently only see that it was added for s3, rds and ec2. Looks like wiould need the  same NewPresignClient functionality added to sts.

PR #1030, opened 2021-01-08
Add support for presigning sts:GetCallerIdentity

skmcgrail:
Add support for presigning sts:GetCallerIdentity.

Fix: https://github.com/aws/aws-sdk-go-v2/issues/1021

Validated with the following code:
```go
package main

import (
	"context"
	"fmt"
	"io/ioutil"
	"net/http"
	"net/url"

)

func main() {
	if err != nil {
		panic(err)
	}

	presignClient := sts.NewPresignClient(sts.NewFromConfig(cfg))

	getCallerIdentity, err := presignClient.PresignGetCallerIdentity(context.Background(), &sts.GetCallerIdentityInput{})
	if err != nil {
		panic(err)
	}

	req := &http.Request{
		Method: getCallerIdentity.Method,
		URL: func() *url.URL {
			parse, err := url.Parse(getCallerIdentity.URL)
			if err != nil {
				panic(err)
			}
			return parse
		}(),
		Header: getCallerIdentity.SignedHeader,
	}

	response, err := http.DefaultClient.Do(req)
	if err != nil {
		panic(err)
	}

	body, err := ioutil.ReadAll(response.Body)
	if err != nil {
		panic(err)
	}

	fmt.Printf("%s\n", body)
}
```

### DIFF
--- a/config/go.mod
+++ b/config/go.mod
@@ -17,3 +17,5 @@ replace (
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds => ../feature/ec2/imds/
 	github.com/aws/aws-sdk-go-v2/service/sts => ../service/sts/
 )
+
+replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../service/internal/presigned-url/
--- a/credentials/go.mod
+++ b/credentials/go.mod
@@ -14,3 +14,5 @@ replace (
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds => ../feature/ec2/imds/
 	github.com/aws/aws-sdk-go-v2/service/sts => ../service/sts/
 )
+
+replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../service/internal/presigned-url/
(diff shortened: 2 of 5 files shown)


aws/signer/v4/v4.go
```diff
@@ -344,7 +345,14 @@ func (s *Signer) PresignHTTP(
 
 	logSigningInfo(ctx, options, &signedRequest, true)
 
-	return signedRequest.Request.URL.String(), signedRequest.SignedHeaders, nil
+	signedHeaders = make(http.Header)
+
+	for k, v := range signedRequest.SignedHeaders {
+		key := textproto.CanonicalMIMEHeaderKey(k)
+		signedHeaders[key] = append(signedHeaders[key], v...)
+	}
```
jasdel, line 353: would be helpful to have a comment here calling out why this is done, vs just the request's original-cased header keys.

jasdel:
Looks like compile issue with `service/s3`. Also looks like the `eachmodule` tool has issue and doesn't correct output stdout or stderr in case of a failure correctly.

skmcgrail:
> Looks like compile issue with `service/s3`. Also looks like the `eachmodule` tool has issue and doesn't correct output stdout or stderr in case of a failure correctly.

Fixed didn't notice the S3 presign test cases beforehand.

jasdel (approved):

PR #1031, opened 2021-01-09
`retry`: adds ability to return metadata for each request attempt

skotambkar:
Adds requestAttempt and requestAttempts struct to handle  metadata for all request attempts.

TODO 

- [x]  needs tests


aws/retry/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+	Attempts []requestAttempt
+}
+
+type requestAttempt struct {
+
+	// Response is raw response if received for the request attempt.
+	Response interface{}
+
+	// Error is the error if received for the request attempt.
+	Error error
```
jasdel, line 22: Nit `Err` is the common naming for an error member

aws/retry/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+
+type requestAttempts struct {
+
+	// Attempts is a slice consisting metadata from all request attempts.
+	// Attempts are stored in last in first order i.e. the last attempt
+	// would be at the top.
+	Attempts []requestAttempt
+}
+
+type requestAttempt struct {
```
jasdel, line 16: Should these two types be exported?

aws/retry/middleware.go
```diff
@@ -87,10 +90,19 @@ func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInp
 			AttemptClockSkew: attemptClockSkew,
 		})
 
+		var attemptMetadata requestAttempt
+		defer func() {
+			allAttemptMetadata.Attempts = append(allAttemptMetadata.Attempts, attemptMetadata)
+		}()
+
```
jasdel, line 97: defer's inside of for loops are generally not a good idea as they can lead to an unexpected behavior when they are unrolled.  If using defer here the content of the `for` probably should be moved to a function so that there are no unexpected behaviors with the defer.

aws/retry/middleware.go
```diff
@@ -87,10 +90,19 @@ func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInp
...
 		if attemptNum > 1 {
+			attemptMetadata.Retried = true
+
```
jasdel, line 100: Won't this refer to the previous attempt begin retried not this current attempt being retried?  I'd expect the `Retried` member to be set on the previous attempt not the current attempt.

consider moving setting `Retried` until after the check if will retry or not, e.g. line 140.

aws/retry/middleware.go
```diff
@@ -139,6 +155,7 @@ func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInp
 
 		responseMetadata := awsmiddle.GetResponseMetadata(metadata)
 		attemptClockSkew = responseMetadata.AttemptSkew
+		attemptMetadata.Retryable = true
```
jasdel, line 158: `Retryable` should be set higher up probably, e.g. result of line 126. This is where the `Retried` member should be set instead, since by fact of reaching this point the AttemptMiddleware will loop again.

This helps distinguish the difference between `Retryable` => `May be retried`, and `Retried` => `Was retried`.

aws/retry/middleware_test.go
```diff
@@ -193,6 +198,20 @@ func TestAttemptMiddleware(t *testing.T) {
...
+				{
+					Error:     mockRetryableError{b: true},
+					Retryable: true,
+					Retried:   true,
+				},
+				{
+					Error:     mockRetryableError{b: true},
+					Retryable: true,
+				},
+			}},
```
jasdel, line 214: Are these in reverse order? I'd expect the order to be [first, second, thrid] but this looks to be [thrid, second, first].  This might be an artifact of the `defer` within the for loop in middleware.go

aws/retry/middleware_test.go
```diff
@@ -131,10 +132,11 @@ func TestAttemptMiddleware(t *testing.T) {
...
-		Request testRequest
-		Next    func(retries *[]retryMetadata) middleware.FinalizeHandler
-		Expect  []retryMetadata
-		Err     error
+		Request        testRequest
+		Next           func(retries *[]retryMetadata) middleware.FinalizeHandler
+		Expect         []retryMetadata
+		Err            error
+		ExpectMetadata requestAttempts
 	}{
```
jasdel, line 140: It would also be helpful to have a test case that asserts the `Response` value is valid and for an attempt metadata.

jasdel (changes requested):

aws/retry/middleware_test.go
jasdel: Also didn't see it but should, if not already, should have a test case for nested `middleware.Metadata`

aws/retry/metadata.go
skotambkar: If the intention is to let customers access these values programmatically, then yes.

aws/retry/middleware_test.go
skotambkar: yeah they are in reverse order. Thought the last request is most important in most cases, and thus used defer.

aws/retry/middleware_test.go
```diff
@@ -153,6 +155,9 @@ func TestAttemptMiddleware(t *testing.T) {
 					MaxAttempts: 3,
 				},
 			},
+			ExpectResults: AttemptResults{Results: []AttemptResult{
+				{},
```
jasdel, line 159: even for a single attempt, no error, shouldn't an `AttemptResult` be present with at minimum its Response member populated?

Though it looks like the nested metadata item covers that a little. Maybe have test where the `next` does have a `Response` value, and on that doesn't.

aws/retry/middleware_test.go
```diff
@@ -252,6 +299,66 @@ func TestAttemptMiddleware(t *testing.T) {
...
+						err = fmt.Errorf("more requests then expected")
+					} else {
+						err = reqsErrs[num]
+						num++
+					}
+
+					if err != nil {
+						metadata.Set("testKey", "testValue")
+					} else {
+						out.Result = "mockResponse"
```
jasdel, line 330: `out.Result` will be the deserialized output not the HTTP response right?  I think the `AttemptResult.Response` should be the deserialized transport (e.g. HTTP) raw response message.  This would be needed in order to provide access to the headers of the raw response.

aws/retry/middleware.go
```diff
@@ -64,19 +64,15 @@ func (r Attempt) logf(logger logging.Logger, classification logging.Classificati
 func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInput, next smithymiddle.FinalizeHandler) (
 	out smithymiddle.FinalizeOutput, metadata smithymiddle.Metadata, err error,
 ) {
-	var attemptNum, retryCount int
+	var attemptNum int
 	var attemptClockSkew time.Duration
+	var shouldRetry bool
```
jasdel, line 69: Looks like `shouldRetry` is only used in side of the for lop.  It probably should be moved into that scope.  Is there a reason it should be persisted between loops?

aws/retry/middleware.go
```diff
@@ -87,61 +83,103 @@ func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInp
...
-			}
+		var attemptResult AttemptResult
 
-			r.logf(logger, logging.Debug, "retrying request %s/%s, attempt %d", service, operation, attemptNum)
-		}
+		out, attemptResult, err = r.handleAttempt(attemptCtx, attemptInput, next)
 
-		out, metadata, reqErr := next.HandleFinalize(attemptCtx, attemptInput)
+		responseMetadata := awsmiddle.GetResponseMetadata(attemptResult.ResponseMetadata)
+		attemptClockSkew = responseMetadata.AttemptSkew
```
jasdel, line 91: Should `GetResponseMetadata` be remand to something more specific, like `GetResponseClockSckew` ?  `ResponseMetadata` seems out of place for what it is doing, and combined with the `AttemptResult(s)`

aws/retry/middleware.go
```diff
@@ -87,61 +83,103 @@ func (r Attempt) HandleFinalize(ctx context.Context, in smithymiddle.FinalizeInp
...
+	return out, metadata, err
+}
+
+// handleAttempt handles an individual request attempt.
+func (r Attempt) handleAttempt(ctx context.Context, in smithymiddle.FinalizeInput, next smithymiddle.FinalizeHandler) (
+	out smithymiddle.FinalizeOutput, attemptResult AttemptResult, err error,
+) {
+	defer func() {
+		attemptResult.Err = err
+		attemptResult.Response = out.Result
```
jasdel, line 112: `out.Result` will be the deserialized output shape of the operation call, not the Raw response message. In order to get things like status code and headers RawResponse is needed.

I think another middleware is needed to add `RawResponse` to the `middleware.Metadata` as the most `before` in the `Deserialize` step.

jasdel: If this pattern is used the explict `Response` member isn't needed and instead could have a method on `AttemptResult` that attempts to retrieve the RawResponse from the nested `ResponseMetadata.

jasdel (changes requested):

jasdel (approved):

PR #1039, opened 2021-01-13
Tweak wording on Migration guide

jasdel:
Updates SDK's v1 => v2 Migration guide in

### DIFF
--- a/content/en/docs/migrating/_index.md
+++ b/content/en/docs/migrating/_index.md
@@ -475,7 +475,7 @@ go get github.com/aws/aws-sdk-go-v2/service/s3
 
 You can construct clients in the {{% alias sdk-go %}} using either the `New` or `NewFromConfig` constructor functions in the client's package.
 When migrating from the AWS SDK for Go we recommend that you use the `NewFromConfig` variant, which will return a new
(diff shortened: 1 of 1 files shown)


content/en/docs/migrating/_index.md
```diff
@@ -431,21 +465,26 @@ Service Name | V1 Import Path | V2 Import Path
...
+```
 
 ### Client Construction
 
-You can construct clients in the {{% alias sdk-go %}} using either the `New` or `NewFromConfig` constructor functions.
-When migrating from the AWS SDK for Go we recommend that you use the `NewFromConfig` variant, which enables construction
-of a service client using the `aws.Config` created while loading the SDK shared configuration using
+You can construct clients in the {{% alias sdk-go %}} using either the `New` or `NewFromConfig` constructor functions in the client's package.
+When migrating from the AWS SDK for Go we recommend that you use the `NewFromConfig` variant, which will return a new
+service client value using a `aws.Config` value. The `aws.Config` value will of been created while loading the SDK shared configuration using
```
skmcgrail, line 478: Maybe "using values from a`aws.Config`"?

skmcgrail (approved):

PR #1044, opened 2021-01-15
Ensure UserAgent is added, regenerate clients to include unix-time behavior changes

skmcgrail:
Ensure UserAgent is added, regenerate clients to include unix-time behavior changes.


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsClientUserAgent.java
```diff
@@ -46,9 +46,8 @@ public void writeAdditionalFiles(
             writer.openBlock("func $L(stack $P) error {", "}", MIDDLEWARE_RESOLVER, SymbolUtils.createPointableSymbolBuilder("Stack",
                     SmithyGoDependency.SMITHY_MIDDLEWARE).build(), () -> {
                 // TODO: This should include the module version
```
skotambkar, line 48: TODO still valid?

skotambkar (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsClientUserAgent.java
skmcgrail: Yes it is.

PR #1048, opened 2021-01-15
Fix xml attribute serialization

JordonPhillips:
This fixes an issue where xml attributes were not being serialized except at the root level.


JordonPhillips:
btw this is still failing because the updated go hash is still not getting picked up

JordonPhillips:
need to address using an enum, which isn't currently handled

JordonPhillips:
Turns out the test jasdel submitted was off, there's a pr to fix it: https://github.com/awslabs/smithy/pull/694

jasdel (approved):

PR #1050, opened 2021-01-15
codegen: Updates SDK's generated serde for unmodeled operation input/output

### LINKED ISSUE #1047: sqs.Client.TagQueue returns an error but still succeessfully tags a resource

**Describe the bug**

Attempts to tag an SQS queue with [sqs.Client.TagQueue](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/sqs@v0.31.0?#Client.TagQueue) call fails with the following error:

	operation error SQS: TagQueue, https response error StatusCode: 200, RequestID: 551b5c11-ecbd-55de-8d46-046a7bfc9c2d, deserialization failed, failed to decode response body, TagQueueResult node not found

API call actually succeeds, as provided tags appear on a resource.

0.31.0

go version go1.16beta1 darwin/arm64

Steps to reproduce the behavior (please share code or minimal repo)

go.mod contents:

```
module sdk-bug

go 1.16

require (
	github.com/aws/aws-sdk-go-v2 v0.31.0
	github.com/aws/aws-sdk-go-v2/config v0.4.0
	github.com/aws/aws-sdk-go-v2/service/sqs v0.31.0
)
```

main.go contents:

```go
package main

import (
	"context"
	"errors"
	"flag"
	"log"

	"github.com/aws/aws-sdk-go-v2/service/sqs"
)

func main() {
	log.SetFlags(0)
	flag.Parse()
	if err := run(context.Background(), flag.Arg(0)); err != nil {
		log.Fatal(err)
	}
}

func run(ctx context.Context, queueURL string) error {
	if queueURL == "" {
		return errors.New("provide a queue url to tag")
	}
	log.Println("SDK version:", aws.SDKVersion)
	if err != nil {
		return err
	}
	svc := sqs.NewFromConfig(cfg)
	_, err = svc.TagQueue(ctx, &sqs.TagQueueInput{
		QueueUrl: &queueURL,
		Tags:     map[string]string{"Environment": "staging"},
	})
	return err
}
```

Build and run this program providing an SQS queue URL as a first positional argument.

**Expected behavior**

Program prints SDK version on stderr and finishes with 0 exit status. Resource is tagged.

**Actual behavior**

Program prints SDK version on stderr and finishes with non-zero exit status, printing an error:

    operation error SQS: TagQueue, https response error StatusCode: 200, RequestID: 551b5c11-ecbd-55de-8d46-046a7bfc9c2d, deserialization failed, failed to decode response body, TagQueueResult node not found

Resource *is* tagged.

**Additional context**

Similar error happens with [rds.Client.AddTagsToResource](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/rds@v0.31.0#Client.AddTagsToResource) calls.

jasdel:
Updates the SDK's generate serializers and deserializers for operations without modeled input and output types. This fixes #1047 where the SDK was attempting to deserialize a response from a service for an operation that had no modeled output. The SDK should of ignored the response instead of trying to deserialize it.

This update will not attempt to serialize inputs for operations without modeled inputs, and will discard responses from services of operations with unmodeled outputs.

Related to aws/smithy-go#261
Fixes #1047

### DIFF
--- a/service/internal/integrationtest/s3/presign_test.go
+++ b/service/internal/integrationtest/s3/presign_test.go
@@ -28,11 +28,8 @@ func TestInteg_PresignURL(t *testing.T) {
 		expectedSignedHeader http.Header
 	}{
 		"standard": {
-			body: bytes.NewReader([]byte("Hello-world")),
-			expectedSignedHeader: http.Header{
-				"content-type":   {"application/octet-stream"},
-				"content-length": {"11"},
-			},
+			body:                 bytes.NewReader([]byte("Hello-world")),
+			expectedSignedHeader: http.Header{},
 		},
 		"special characters": {
 			key: "some_value_(1).foo",


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlProtocolUtils.java
```diff
@@ -237,7 +237,9 @@ private static String getSerializedXMLShapeName(ProtocolGenerator.GenerationCont
         Optional<SyntheticClone> clone = shape.getTrait(SyntheticClone.class);
         if (clone.isPresent()) {
             SyntheticClone cl = clone.get();
-            shapeName = cl.getArchetype().getName();
+            if (cl.getArchetype().isPresent()) {
+                shapeName = cl.getArchetype().get().getName();
+            }
```
skotambkar, line 242: I thought the only purpose of clone was to stub a top-level shape if not present in model. 

This code-block suggests we rename shapes using synthetic shape, even when they are present in the model. Is that true?

skotambkar (approved):

skmcgrail (approved):

PR #1069, opened 2021-01-22
Add PresignHeadObject to S3 PresignClient

milosgajdos:
This PR adds `PresignHeadObject` method to `s3` service's `PresignClient` which was missing in V2 SDK, but is available in V1 via `S3.HeadObjectRequest.Presign`

### DIFF
--- a/service/s3/api_op_HeadObject.go
+++ b/service/s3/api_op_HeadObject.go
@@ -6,8 +6,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
+
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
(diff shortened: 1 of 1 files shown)


jasdel (changes requested):
Thanks for taking the time to create this PR @milosgajdos. This file is generated, and any changes made to it directly will be lost next time its generated. This change needs to be made to the [SDK's Presign generator](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java#L82-L90).

milosgajdos:
Oh, I'm a moron. Ignore me :-) I just spotted `// Code generated by smithy-go-codegen DO NOT EDIT.` line. Ok, let me close this PR.

PR #1070, opened 2021-01-22
aws: Add endpoint source property

skotambkar:
`aws/endpoints`: Adds `EndpointSource` property to `Endpoint` struct. This allows customers to provide an endpoint source, which can then be used to give SDK more context on handling custom Endpoints.

### DIFF
--- a/.changes/next-release/aws-feature-1611348426115463000.json
+++ b/.changes/next-release/aws-feature-1611348426115463000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "aws-feature-1611348426115463000",
+ "SchemaVersion": 1,
+ "Module": "aws",
+ "Type": "feature",
+ "Description": "updates aws module to include endpoint source for aws Endpoint",
+ "MinVersion": "",
+ "AffectedModules": null
+}


aws/endpoints.go
```diff
@@ -54,8 +54,27 @@ type Endpoint struct {
 	// The signing method that should be used for signing the requests to the
 	// endpoint.
 	SigningMethod string
+
+	// The source of the Endpoint. By default, this will be ServiceMetadata.
+	// When providing a custom endpoint, you should set the source as CustomSource.
+	// If source is not provided when providing a custom endpoint, the SDK may not
+	// perform required host mutations correctly. EndpointSource should be used along with
+	// HostnameImmutable property as per the usage requirement.
+	EndpointSource EndpointSource
```
jasdel: `EndpointSource` as the parameter name is a bit redundant since the type is `Endpoint`. e.g. `Endpoint.EndpointSource`  Could name this member just `Source`.

internal/protocoltest/awsrestjson/endpoints.go
```diff
@@ -45,6 +45,26 @@ func resolveDefaultEndpointConfiguration(o *Options) {
...
+	return EndpointResolverFunc(
+		func(region string, options EndpointResolverOptions) (aws.Endpoint, error) {
+			if len(e.SigningRegion) == 0 {
+				e.SigningRegion = region
+			}
+			return e, nil
+		},
+	)
+}
+
```
jasdel: Generation of these should be updated for the Endpoint's `Source` member being set to `Custom`

jasdel (approved):
Looks good, but needs change notes.

aws/endpoints.go
```diff
@@ -54,8 +54,27 @@ type Endpoint struct {
...
+
+const (
+	// ServiceMetadata denotes service provided endpoint metadata used as Endpoint Source.
+	ServiceMetadata EndpointSource = iota
+
+	// CustomSource denotes endpoint is a custom endpoint.
+	// This source should be used when providing a custom endpoint to be used by the SDK.
+	CustomSource
+)
+
```
jasdel, line 77: Forgot to mention these probably should be prefixed With `EndpointSource` since this is in the pretty large namespace of `aws` packag.e

jasdel (approved):

PR #1071, opened 2021-01-22
codegen: Fix SDK's serialization of operations with empty inputs

jasdel:
Fixes the JSON RPC API clients to serialize empty JSON objects for stub synthetic cloned inputs. Also this ensure empty JSON modeled inputs are correctly serialized. Previously the SDK would incorrectly skip serialization for JSON RPC input shapes with no members. Whereas JSON RPC requires operations with input modeled serialize empty JSON object.

### DIFF
--- a/.changes/next-release/internal.protocoltest.jsonrpc10-bugfix-1611357209709451000.json
+++ b/.changes/next-release/internal.protocoltest.jsonrpc10-bugfix-1611357209709451000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "internal.protocoltest.jsonrpc10-bugfix-1611357209709451000",
+ "SchemaVersion": 1,
+ "Module": "internal/protocoltest/jsonrpc10",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/.changes/next-release/service.alexaforbusiness-bugfix-1611357209739908000.json
+++ b/.changes/next-release/service.alexaforbusiness-bugfix-1611357209739908000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.alexaforbusiness-bugfix-1611357209739908000",
+ "SchemaVersion": 1,
+ "Module": "service/alexaforbusiness",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/.changes/next-release/service.applicationdiscoveryservice-bugfix-1611357209770363000.json
+++ b/.changes/next-release/service.applicationdiscoveryservice-bugfix-1611357209770363000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.applicationdiscoveryservice-bugfix-1611357209770363000",
+ "SchemaVersion": 1,
+ "Module": "service/applicationdiscoveryservice",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/.changes/next-release/service.appstream-bugfix-1611357209799313000.json
+++ b/.changes/next-release/service.appstream-bugfix-1611357209799313000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.appstream-bugfix-1611357209799313000",
+ "SchemaVersion": 1,
+ "Module": "service/appstream",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/.changes/next-release/service.cloudhsm-bugfix-1611357209833690000.json
+++ b/.changes/next-release/service.cloudhsm-bugfix-1611357209833690000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.cloudhsm-bugfix-1611357209833690000",
+ "SchemaVersion": 1,
+ "Module": "service/cloudhsm",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/.changes/next-release/service.codebuild-bugfix-1611357209867885000.json
+++ b/.changes/next-release/service.codebuild-bugfix-1611357209867885000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.codebuild-bugfix-1611357209867885000",
+ "SchemaVersion": 1,
+ "Module": "service/codebuild",
+ "Type": "bugfix",
+ "Description": "fix serialize input",
+ "MinVersion": "",
+ "AffectedModules": null
+}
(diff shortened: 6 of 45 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonRpcProtocolGenerator.java
```diff
@@ -63,10 +63,13 @@ protected void serializeInputDocument(GenerationContext context, OperationShape
...
-        writer.addUseImports(SmithyGoDependency.SMITHY_JSON);
 
-        // If there are no members then there's nothing to serialize
-        if (input.members().size() == 0) {
+        if (CodegenUtils.isStubSyntheticClone(ProtocolUtils.expectInput(context.getModel(), operation))) {
+            writer.addUseImports(SmithyGoDependency.STRINGS);
+            writer.openBlock("if request, err = request.SetStream(strings.NewReader(`{}`)); err != nil {",
+                    "}", () -> {
+                        writer.write("return out, metadata, &smithy.SerializationError{Err: err}");
+                    });
```
skotambkar, line 72: Would this affect rest-json too? Assuming empty json body must always be set to `{}`

service/lightsail/serializers.go
```diff
@@ -656,6 +656,15 @@ func (m *awsAwsjson11_serializeOpCreateContainerServiceRegistryLogin) HandleSeri
...
 
+	jsonEncoder := smithyjson.NewEncoder()
+	if err := awsAwsjson11_serializeOpDocumentCreateContainerServiceRegistryLoginInput(input, jsonEncoder.Value); err != nil {
+		return out, metadata, &smithy.SerializationError{Err: err}
+	}
+
+	if request, err = request.SetStream(bytes.NewReader(jsonEncoder.Bytes())); err != nil {
+		return out, metadata, &smithy.SerializationError{Err: err}
+	}
+
```
skotambkar, line 667: If there is no input, why do we need to generate these helpers?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonRpcProtocolGenerator.java
```diff
@@ -63,10 +63,13 @@ protected void serializeInputDocument(GenerationContext context, OperationShape
         GoWriter writer = context.getWriter();
         StructureShape input = ProtocolUtils.expectInput(context.getModel(), operation);
         String functionName = ProtocolGenerator.getDocumentSerializerFunctionName(input, getProtocolName());
-        writer.addUseImports(SmithyGoDependency.SMITHY_JSON);
 
-        // If there are no members then there's nothing to serialize
-        if (input.members().size() == 0) {
```
skotambkar, line 69: We might still need this check. Why generate serializers if no input members?

skotambkar (commented):
Few questions/comments

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonRpcProtocolGenerator.java
jasdel: No this would only be for JSON RPC.

service/lightsail/serializers.go
jasdel: Posted PR too soon, I'm going back and regenerating the clients to fix this. The rule should do nothing if stub synthetic clone, else do what ever serialization is needed. Don't have short cut behavior of `{}`.   Will post update shortly fixing this.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonRpcProtocolGenerator.java
jasdel: Will update PR shortly that addresses this issue.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonRpcProtocolGenerator.java
```diff
@@ -63,10 +63,19 @@ protected void serializeInputDocument(GenerationContext context, OperationShape
         GoWriter writer = context.getWriter();
         StructureShape input = ProtocolUtils.expectInput(context.getModel(), operation);
         String functionName = ProtocolGenerator.getDocumentSerializerFunctionName(input, getProtocolName());
-        writer.addUseImports(SmithyGoDependency.SMITHY_JSON);
 
-        // If there are no members then there's nothing to serialize
+        // Stub synthetic clone inputs mean there never was an input modeled, so nothing should be serialized.
+        if (CodegenUtils.isStubSyntheticClone(ProtocolUtils.expectInput(context.getModel(), operation))) {
+            return;
```
skotambkar, line 69: If input shape is not modeled, we don't need to send the `{}` ? This seems inconsistent. I would imagine we always send an empty json object regardless, and leave it up to service to ignore it?

jasdel: This is not clear. I'm reaching out to the Smithy team to clarify this.

jasdel: Have verified that empty JSON object `{}`  should be serialized for all empty modeled input. Its not clear if this should be done for unmodeled input. Will research that separately.

service/gamelift/serializers.go
```diff
@@ -2395,6 +2395,15 @@ func (m *awsAwsjson11_serializeOpDescribeVpcPeeringAuthorizations) HandleSeriali
...
 
+	jsonEncoder := smithyjson.NewEncoder()
+	if err := awsAwsjson11_serializeOpDocumentDescribeVpcPeeringAuthorizationsInput(input, jsonEncoder.Value); err != nil {
+		return out, metadata, &smithy.SerializationError{Err: err}
+	}
+
+	if request, err = request.SetStream(bytes.NewReader(jsonEncoder.Bytes())); err != nil {
+		return out, metadata, &smithy.SerializationError{Err: err}
+	}
+
```
skotambkar, line 2406: Is this intentional? Asking as on a previous comment, you responded you would be removing these. 

I am fine with having them for now.  Think they can be replaced with 

```
	if request, err = request.SetStream(strings.NewReader(`{}`)); err != nil {
		return out, metadata, &smithy.SerializationError{Err: err}
	}
```

If we make the change, would also want to consider not generating dead code.

jasdel: So I left using the generated serializer function if there was a modeled input. This latest updates the generator to only have the short cut path if there was nothing modeled. This also has consistent behavior across all of the protocols too, (e.g. REST-JSON/REST-XML). The reasoning behind this was more for consistent behavior across the generator. As in, all protocols that have stub synethentic clones (aka unmodeled input/output) the generated code will short cut and not render document serializers.

To not generate document (de)serializers for empty input/output shapes is a bit more complicated as that is spread across both smithy-go and sdk's codegen.

skotambkar (approved):

skotambkar (approved):

PR #1072, opened 2021-01-23
AWS Single Sign-On (SSO) Provider Support

skmcgrail:
Adds support for the AWS Single Sign-On (SSO) service by supporting the ability to utilize tokens cached in `~/.aws/sso/cache`.

Adds support for the AWS SSO provider in the default credential loading chain.

### DIFF
--- a/config/resolve_credentials.go
+++ b/config/resolve_credentials.go
@@ -110,9 +110,6 @@ func resolveCredentialChain(ctx context.Context, cfg *aws.Config, configs config
 func resolveCredsFromProfile(ctx context.Context, cfg *aws.Config, envConfig *EnvConfig, sharedConfig *SharedConfig, configs configs) (err error) {
 
 	switch {
-	case sharedConfig.hasSSOConfiguration():
-		err = resolveSSOCredentials(ctx, cfg, sharedConfig, configs)
-
 	case sharedConfig.Source != nil:
 		// Assume IAM role with credentials source from a different profile.
 		err = resolveCredsFromProfile(ctx, cfg, envConfig, sharedConfig.Source, configs)
@@ -122,6 +119,9 @@ func resolveCredsFromProfile(ctx context.Context, cfg *aws.Config, envConfig *En
 		cfg.Credentials = credentials.StaticCredentialsProvider{
 			Value: sharedConfig.Credentials,
 		}
+
+	case sharedConfig.hasSSOConfiguration():
+		err = resolveSSOCredentials(ctx, cfg, sharedConfig, configs)
 
 	case len(sharedConfig.CredentialProcess) != 0:
 		// Get credentials from CredentialProcess
@@ -166,7 +166,7 @@ func resolveSSOCredentials(ctx context.Context, cfg *aws.Config, sharedConfig *S
 		options = append(options, v)
 	}
 
-	cfgCopy := *cfg
+	cfgCopy := cfg.Copy()
 	cfgCopy.Region = sharedConfig.SSORegion
 
 	cfg.Credentials = ssocreds.New(sso.NewFromConfig(cfgCopy), sharedConfig.SSOAccountID, sharedConfig.SSORoleName, sharedConfig.SSOStartURL, options...)
--- a/config/resolve_credentials_test.go
+++ b/config/resolve_credentials_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"os"
@@ -15,6 +16,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
+	"github.com/aws/aws-sdk-go-v2/service/sso"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/aws/smithy-go/middleware"
 )
@@ -66,12 +68,24 @@ func setupCredentialsEndpoints(t *testing.T) (aws.EndpointResolver, func()) {
 					Format("2006-01-02T15:04:05Z"))))
 		}))
 
+	ssoServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(fmt.Sprintf(
+			getRoleCredentialsResponse,
+			time.Now().
+				Add(15*time.Minute).
+				UnixNano()/int64(time.Millisecond))))
+	}))
+
 	resolver := aws.EndpointResolverFunc(
 		func(service, region string) (aws.Endpoint, error) {
 			switch service {
 			case sts.ServiceID:
 				return aws.Endpoint{
 					URL: stsServer.URL,
+				}, nil
+			case sso.ServiceID:
+				return aws.Endpoint{
+					URL: ssoServer.URL,
 				}, nil
 			default:
 				return aws.Endpoint{},
@@ -83,8 +97,47 @@ func setupCredentialsEndpoints(t *testing.T) (aws.EndpointResolver, func()) {
 		resetECSEndpoint()
 		ecsMetadataServer.Close()
 		ec2MetadataServer.Close()
+		ssoServer.Close()
 		stsServer.Close()
 	}
+}
+
+func ssoTestSetup() (func(), error) {
+	dir, err := ioutil.TempDir("", "sso-test")
+	if err != nil {
+		return nil, err
+	}
+
+	cacheDir := filepath.Join(dir, ".aws", "sso", "cache")
+	err = os.MkdirAll(cacheDir, 0750)
+	if err != nil {
+		os.RemoveAll(dir)
+		return nil, err
+	}
+
+	tokenFile, err := os.Create(filepath.Join(cacheDir, "eb5e43e71ce87dd92ec58903d76debd8ee42aefd.json"))
+	if err != nil {
+		os.RemoveAll(dir)
+		return nil, err
+	}
+	defer tokenFile.Close()
+
+	_, err = tokenFile.WriteString(fmt.Sprintf(ssoTokenCacheFile, time.Now().
+		Add(15*time.Minute).
+		Format(time.RFC3339)))
+	if err != nil {
+		os.RemoveAll(dir)
+		return nil, err
+	}
+
+	if runtime.GOOS == "windows" {
+		os.Setenv("USERPROFILE", dir)
+	} else {
+		os.Setenv("HOME", dir)
+	}
+
+	return func() {
+	}, nil
 }
 
 func TestSharedConfigCredentialSource(t *testing.T) {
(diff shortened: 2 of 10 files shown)


credentials/ssocreds/provider.go
```diff
@@ -0,0 +1,176 @@
...
+	StartURL string
+}
+
+// Provider is an AWS credential provider that retrieves temporary AWS credentials by exchanging an SSO login token.
+type Provider struct {
+	options Options
+}
+
+// New returns a new AWS Signle Sign-On (AWS SSO) credential proivder.
+func New(client GetRoleCredentialsAPIClient, accountID, region, roleName, startURL string, optFns ...func(options *Options)) *Provider {
```
skotambkar, line 51: Are accountID, region, ... all required, and thus explicit params? 

Can we diverge from the pattern of constructors not returning an error, and have validation here for missing required members instead of explicit params?

Wouldn't this constructor break if SSO spec adds a new Option in the future, and that is made a required param? (Though that may never happen)

skmcgrail: Yes these are all required parameters, and match the ones that must be provided in the `[profile]` section for the SSO configuration. Our pattern for the providers has been to have the required parameters declared on the type constructor function, and then optional values would be provided via the function options passed in.

I don't think new required parameters could be added to the profile section as to do so would be a backwards incompatible change. Likewise any new parameters would have to be optional and thus this constructor design would be fine.

Also you could make this statement about any service like STS where we have providers that have a series of required and optional parameters so not sure what additional thought paths you were going for.

jasdel: Agree using function parameters to describe required-ness is a more obvious means of communicating that concept. It is possible that future versions of SSO come out that don't need all of these parameters, but in that case an alternative function would probably be the best answer. I'd not expect that to happen often.

config/load_options.go
```diff
@@ -592,3 +597,27 @@ func WithS3UseARNRegion(v bool) LoadOptionsFunc {
 		return nil
 	}
 }
+
+
+// getAssumeRoleCredentialOptions returns AssumeRoleCredentialOptions from LoadOptions
+func (o LoadOptions) getSSOProviderOptions(context.Context) (func(options *ssocreds.Options), bool, error) {
```
jasdel, line 603: nit doc copy-paste  issue

config/resolve_credentials.go
```diff
@@ -151,6 +156,20 @@ func resolveCredsFromProfile(ctx context.Context, cfg *aws.Config, envConfig *En
...
+	var options []func(*ssocreds.Options)
+	v, found, err := getSSOProviderOptions(ctx, configs)
+	if err != nil {
+		return err
+	}
+	if found {
+		options = append(options, v)
+	}
+
+	cfg.Credentials = ssocreds.New(sso.NewFromConfig(cfg.Copy()), sharedConfig.SSOAccountID, sharedConfig.SSORegion, sharedConfig.SSORoleName, sharedConfig.SSOStartURL, options...)
```
jasdel, line 169: Do the SDK's client's `NewFromConfig` functions not already perform a copy internally on the `Config` type?  If not they probably should all be doing that internally. Because if not any modifications to the config's list or pointer member values could pollute the client if the passed in `Config` value is later modified.

config/shared_config.go
```diff
@@ -750,13 +761,13 @@ func (c *SharedConfig) setFromIniSections(profiles map[string]struct{}, profile
...
 		// the ARN is also set.
-		if err := c.validateCredentialsRequireARN(profile); err != nil {
+		if err := c.validateCredentialsConfig(profile); err != nil {
 			return err
 		}
 	}
 
 	// if not top level profile and has credentials, return with credentials.
-	if len(profiles) != 0 && c.Credentials.HasKeys() {
+	if len(profiles) != 0 && (c.Credentials.HasKeys()) {
```
jasdel, line 770: not important but, wouldn't this be a gofmt fixup?  I'd not expect the wrapping `(`, `)` to be needed.

credentials/ssocreds/doc.go
```diff
@@ -0,0 +1,57 @@
+// Package provides a credential provider for retrieving temporary AWS credentials using an SSO access token.
+//
+// IMPORTANT: The provider in this package does not initiate or perform the AWS SSO login flow. The SDK provider
```
jasdel, line 3: I think this will be a lint error unless this line was `Package ssocreds ...`

credentials/ssocreds/os.go
```diff
@@ -0,0 +1,9 @@
+// +build !windows
+
+package ssocreds
+
+import "os"
+
+func getHomeDirectory() string {
+	return os.Getenv("HOME")
+}
```
jasdel, line 9: Does this duplicate the behavior used by the shared config file loading for the default file name?  I think there are a few additional places the Windows home can be. Should these be considered for the sso creds file as well?

credentials/ssocreds/provider.go
```diff
@@ -0,0 +1,176 @@
...
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/sdk"
+	"github.com/aws/aws-sdk-go-v2/service/sso"
+)
+
+const ProviderName = "SSOProvider"
```
jasdel, line 19: nit godoc

credentials/ssocreds/provider.go
skotambkar: Right, I didn't think about the usage with STS.  Optional params that are added later, can be used with functional options, or any WithHelpers we expose.

credentials/ssocreds/doc.go
```diff
@@ -0,0 +1,57 @@
...
+//
+//  credentials, err := provider.Retrieve(context.TODO())
+//  if err != nil {
+//      return err
+//  }
+//
+// It is important that you wrap the Provider with aws.CredentialsCache if you are programmatically constructing the
+// provider directly. This prevents your application from accessing the cached access token and requesting new
+// credentials each time the credentials are used.
+package ssocreds
```
jasdel, line 57: is there an AWS CLI or other AWS documentation page we can link to with more information on getting started and using SSO?

credentials/ssocreds/provider.go
```diff
@@ -0,0 +1,176 @@
...
+	Err error
+}
+
+func (i *InvalidTokenError) Unwrap() error {
+	return i.Err
+}
+
+func (i *InvalidTokenError) Error() string {
+	return "the SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile."
+}
```
jasdel, line 150: Should this error message include the nested error as well? Also this is a really long message :)

jasdel (changes requested):
Looks good couple suggestions, mainly with gofmt/go lint.

config/resolve_credentials.go
skmcgrail: Yes they do copy already, will remove this unnecessary step.

credentials/ssocreds/os.go
skmcgrail: Yes this logic is the same, in the `os_windows.go` file the implementation is to use the `USERPROFILE` environment variable. I opted for the build file strategy rather then the `runtime.GOOS` check that the shared_config currently usses.

credentials/ssocreds/doc.go
skmcgrail: Yes, will update with this information.

credentials/ssocreds/provider.go
skmcgrail: This was the error message from the proposal, but we could decrease the wordiness in the `Error` return string, and add the details to the type documentation of the error.

config/shared_config.go
```diff
@@ -750,13 +761,13 @@ func (c *SharedConfig) setFromIniSections(profiles map[string]struct{}, profile
 		// First time a profile has been seen, It must either be a assume role
 		// or credentials. Assert if the credential type requires a role ARN,
 		// the ARN is also set.
```
skotambkar, line 763: Nit: comment should be updated as well

config/shared_config.go
```diff
@@ -890,8 +919,39 @@ func (c *SharedConfig) validateCredentialType() error {
...
+		missing = append(missing, ssoRoleNameKey)
+	}
+
+	if len(c.SSOStartURL) == 0 {
+		missing = append(missing, ssoStartURL)
+	}
+
+	if len(missing) > 0 {
+		return fmt.Errorf("profile %q is configured to use SSO but is missing required configuration: %s",
+			profile, strings.Join(missing, ","))
```
skotambkar, line 954: Nit: add space after `,` .  ```strings.Join(missing, ", "))```

config/shared_config_test.go
```diff
@@ -218,6 +218,33 @@ func TestNewSharedConfig(t *testing.T) {
...
+				SourceProfileName: "sso_creds",
+				Source: &SharedConfig{
+					Profile:      "sso_creds",
+					SSOAccountID: "012345678901",
+					SSORegion:    "us-west-2",
+					SSORoleName:  "TestRole",
+					SSOStartURL:  "https://127.0.0.1/start",
+				},
+			},
+		},
```
skotambkar, line 247: Might be redundant, but helpful to add test cases for profiles with:

1.  Chained Assume role with AWS SSO Credentials. 
2. Profile with SSO creds modeled together with static credentials.
3. Partial SSO creds is covered in a test below, so may or may not be added here.

credentials/ssocreds/os_windows.go
```diff
@@ -0,0 +1,7 @@
+package ssocreds
+
+import "os"
+
+func getHomeDirectory() string {
+	return os.Getenv("USERPROFILE")
+}
```
skotambkar, line 7: Assuming this does not require  ```// +build windows``` at the top, as other file has  ```// +build !windows```

credentials/ssocreds/provider_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func swapNowTime(referenceTime time.Time) func() {
+	original := sdk.NowTime
+	sdk.NowTime = func() time.Time {
+		return referenceTime
+	}
+	return func() {
+		sdk.NowTime = original
+	}
+}
```
skotambkar, line 72: We can move this to our internal/awstesting lib

skotambkar (approved):

credentials/ssocreds/os_windows.go
jasdel: I think the `windows` part is implicit because of the filename. it doesn't hurt to have `// +build windows`, but i think the Go compiler is supossed to figure it out in these cases. Though linux tests shoudl fail if this was not the case.

credentials/ssocreds/os.go
jasdel: +1 build tag instead of runtime.

credentials/ssocreds/os_windows.go
skmcgrail: Correct the _GOOS at the tail end of the file before the extension is an implicit build flag.

config/shared_config_test.go
skmcgrail: 1. Is covered by the `source_sso_creds` profile test which uses a profile with a role_arn and a source profile that is SSO credentials.
2. Good call out, since this should favor the SSO credentials,
3. Can add this to just to ensure that the error surfaces correctly.

skotambkar: 1. I was thinking of multi-level chaining with assume role. Since credentials take precedence starting the second level. 

Basically looking for what's the behavior if we go through the following flow : 

```assume role -> assume role + sso creds -> assume role + creds ```

File: config/resolve_credentials_test.go [hunk 15]
```diff
@@ -87,6 +101,44 @@ func setupCredentialsEndpoints(t *testing.T) (aws.EndpointResolver, func()) {
...
+	}
+
+	cacheDir := filepath.Join(dir, ".aws", "sso", "cache")
+	err = os.MkdirAll(cacheDir, 0750)
+	if err != nil {
+		os.RemoveAll(dir)
+		return nil, err
+	}
+
+	tokenFile, err := os.Create(filepath.Join(cacheDir, "eb5e43e71ce87dd92ec58903d76debd8ee42aefd.json"))
```
skotambkar, line 117: qq: Should `eb5e43e71ce87dd92ec58903d76debd8ee42aefd` be a randomized string instead?

config/resolve_credentials_test.go
```diff
@@ -278,7 +370,15 @@ func TestSharedConfigCredentialSource(t *testing.T) {
...
-				t.Errorf("expected %v, but received %v", e, a)
+				t.Errorf("expect %v, but received %v", e, a)
+			}
+
+			if e, a := c.expectedSessionToken, creds.SessionToken; e != a {
+				t.Errorf("expect %v, got %v", e, a)
+			}
+
+			if cleanup != nil {
+				cleanup()
```
skotambkar, line 381: Should we use a `defer cleanup()` instead? we have few `t.Fatalf` calls above.

config/shared_config_test.go
skotambkar: I was unclear with `assume role` expectation, I should have written `assume role with source profile`.

config/resolve_credentials_test.go
skmcgrail: This needs to match the startURL content of the profile, in this case this is that sha1 of `https://127.0.0.1/start`

config/resolve_credentials_test.go
skmcgrail: good catch

skotambkar (approved):
Looks good to me!

jasdel (approved):

PR #1099, opened 2021-02-01
middleware: fix date header parse error log

danielhochman:
This log message in `middleware.go` is not properly formatted because the format is being passed as the classification to `Logf(classification Classification, format string, v ...interface{})`.

Example output:
`
SDK 2021/02/01 22:20:44 failed to parse response Date Header value, got %!v(MISSING) parsing time "Mon, 1 Feb 2021 22:20:44 GMT" as "Mon Jan _2 15:04:05 2006": cannot parse ", 1 Feb 2021 22:20:44 GMT" as " "
`

Looking at the error and associated parsing code, it looks like an issue with some AWS endpoint not returning compliant timestamps that causes this warning (should technically be `01 Feb` not `1 Feb`). I can provide a fix for that in a separate PR with slightly more permissive Date header parsing if needed.

### DIFF
--- a/aws/middleware/middleware.go
+++ b/aws/middleware/middleware.go
@@ -71,7 +71,7 @@ func (a RecordResponseTiming) HandleDeserialize(ctx context.Context, in middlewa
 		serverTime, parseErr = http.ParseTime(respDateHeader)
 		if parseErr != nil {
 			logger := middleware.GetLogger(ctx)
-			logger.Logf(logging.Warn, "failed to parse response 	Date header value, got %v",
+			logger.Logf(logging.Warn, "failed to parse response Date header value, got %v",
 				parseErr.Error())
 			break
 		}


skmcgrail (approved):

skmcgrail:
@danielhochman out of curiosity, are you seeing this with a specific service or is this message getting logged for all service responses?

danielhochman:
@skmcgrail looks like it's coming from the Kinesis API on `ListStreams` call though I can't be 100% sure

danielhochman:
@skmcgrail I tested in three regions, `us-east-1`, `us-west-1`, and `us-west-2`. `ListStreams` in `us-west-2` is reproducible and returns non-compliant timestamps.

skmcgrail (approved):

skmcgrail:
Thank you for this PR @danielhochman , I've cut a tracking issue so that we can follow-up on supporting the datetime format Kinesis is sending package in the HTTP `Date` header. Thank you for providing the investigation details.

PR #1100, opened 2021-02-02
s3: Fix ListPartsPaginator never stopping

stefansundin:
Currently, `ListPartsPaginator` never stops. `NextPartNumberMarker` is always set so we need to also check `IsTruncated`. When `IsTruncated` is `true` and `NextPartNumberMarker` is greater than the actual number of parts, the results of that request is an empty page with `NextPartNumberMarker` set to `"0"`. After that the listing restarts from the beginning.

Easy workaround:

```go
for p.HasMorePages() {
	page, err := p.NextPage(context.TODO())
	for _, part := range page.Parts {
		// ...
	}
	if !page.IsTruncated {
		break
	}
}
```

I think the logic can be slightly refactored, but for simplicity I left it like this. It could also be changed to:

```go
	if result.IsTruncated {
		if p.options.StopOnDuplicateToken && prevToken != nil && p.nextToken != nil && *prevToken == *p.nextToken {
			p.nextToken = nil
		} else {
			p.nextToken = result.NextPartNumberMarker
		}
	} else {
		p.nextToken = nil
	}
```

Or something else entirely. Let me know what you want.

Thanks!

### DIFF
--- a/service/s3/api_op_ListParts.go
+++ b/service/s3/api_op_ListParts.go
@@ -334,7 +334,11 @@ func (p *ListPartsPaginator) NextPage(ctx context.Context, optFns ...func(*Optio
 	p.firstPage = false
 
 	prevToken := p.nextToken
-	p.nextToken = result.NextPartNumberMarker
+	if result.IsTruncated {
+		p.nextToken = result.NextPartNumberMarker
+	} else {
+		p.nextToken = nil
+	}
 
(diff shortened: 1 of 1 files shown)


skmcgrail:
Thank you for bringing this to our attention @stefansundin, since this change required changes to our code-generation I have created a separate PR: https://github.com/aws/aws-sdk-go-v2/pull/1141

I am going to close this PR and we can track the status of this fix in that PR.

PR #1103, opened 2021-02-02
Fix SSO Validation of Incomplete Profiles

skmcgrail:
* Ports the fix from V1 [#3769](https://github.com/aws/aws-sdk-go/pull/3769) around SSO profile validation
* Fixes an uncovered bug where environment credentials were not correctly taking precedence over profile credentials

### DIFF
--- a/config/resolve_credentials.go
+++ b/config/resolve_credentials.go
@@ -80,12 +80,15 @@ func resolveCredentialProvider(ctx context.Context, cfg *aws.Config, cfgs config
 // credentials are only refreshed when needed. This also protects the
 // credential provider to be used concurrently.
 func resolveCredentialChain(ctx context.Context, cfg *aws.Config, configs configs) (err error) {
-	_, sharedProfileSet, err := getSharedConfigProfile(ctx, configs)
+	envConfig, sharedConfig, other := getAWSConfigSources(configs)
+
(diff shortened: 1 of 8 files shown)


jasdel (approved):

config/resolve_credentials.go
```diff
@@ -80,13 +80,16 @@ func resolveCredentialProvider(ctx context.Context, cfg *aws.Config, cfgs config
 // credentials are only refreshed when needed. This also protects the
 // credential provider to be used concurrently.
 func resolveCredentialChain(ctx context.Context, cfg *aws.Config, configs configs) (err error) {
-	_, sharedProfileSet, err := getSharedConfigProfile(ctx, configs)
+	envConfig, sharedConfig, other := getAWSConfigSources(configs)
+
+	// When checking if a profile was specified programmatically we should only consider the "other"
+	// configuration sources that have been provided. This ensures we correctly honor the expected credential
+	// hierarchy.
+	_, sharedProfileSet, err := getSharedConfigProfile(ctx, other)
```
skotambkar, line 88: should we add test for this behavior, so that we don't change the hierarchy accidentally?

skotambkar (approved):

config/resolve_credentials.go
skotambkar: looks like we have it already.

PR #1138, opened 2021-02-17
New Config Field Resolver Design

skmcgrail:
This redesign config field resolver to allow them to apply to two locations: the `New` client constructor or `invokeOperation` when an operation is invoked. Each resolver can then target either the initial set of client field options prior to user mutation, or a resolver can target the final mutated version of options (should be used sparingly).

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsSignatureVersion4.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsSignatureVersion4.java
@@ -26,6 +26,7 @@ import software.amazon.smithy.go.codegen.GoWriter;
 import software.amazon.smithy.go.codegen.SmithyGoDependency;
 import software.amazon.smithy.go.codegen.SymbolUtils;
 import software.amazon.smithy.go.codegen.integration.ConfigField;
+import software.amazon.smithy.go.codegen.integration.ConfigFieldResolver;
 import software.amazon.smithy.go.codegen.integration.GoIntegration;
 import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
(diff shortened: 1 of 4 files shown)


internal/protocoltest/awsrestjson/api_client.go
```diff
@@ -117,6 +117,7 @@ func (c *Client) invokeOperation(ctx context.Context, opID string, params interf
 	ctx = middleware.ClearStackValues(ctx)
 	stack := middleware.NewStack(opID, smithyhttp.NewStackRequest)
 	options := c.options.Copy()
+
 	for _, fn := range optFns {
```
jasdel, line 121: Really minor but are we able to "fix" this new line begin added when not real change is needed?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsSignatureVersion4.java
```diff
@@ -120,7 +122,12 @@ private void writeNewV4SignerFunc(GoWriter writer, ServiceShape serviceShape) {
                         .type(SymbolUtils.createValueSymbolBuilder(SIGNER_INTERFACE_NAME).build())
                         .documentation("Signature Version 4 (SigV4) Signer")
                         .build())
-                .resolveFunction(SymbolUtils.createValueSymbolBuilder(SIGNER_RESOLVER).build())
+                .addconfigFieldResolver(
```
jasdel, line 125: minor nit should this be "add`C`onfigFieldResolver" ?  looks like the plural version is correct. `addConfigFieldResolvers`

skmcgrail: Yup this was a miss on my part!

jasdel (approved):

Issue #1140, opened 2021-02-18
ListPartsPaginator paginates forever


morsecodist:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
When paginating through pages of multipart upload parts with `ListPartsPaginator` the iterator loops forever.

**Version of AWS SDK for Go?**
`1.2.0`

**Version of Go (`go version`)?**

`go version go1.16 linux/amd64`

```
paginator := s3.NewListPartsPaginator(client, params)
for paginator.HasMorePages() {
	parts, err := paginator.NextPage(context.TODO())
	if err != nil {
		return nil, err
	}
}
```

**Expected behavior**
I expect `HasMorePages` to return `false` when the last page has been fetched and the iterator to stop.

**Additional context**
This is happening because the `nextToken` is `&"0"` on the last page and not `nil`. The paginator seems to expect `nil`.

I don't really understand how the auto-generated stuff works or I would fix it myself. I fixed this temporarily in my fork by making the following change to `service/s3/api_op_ListParts.go`:

```go
// NextPage retrieves the next ListParts page.
func (p *ListPartsPaginator) NextPage(ctx context.Context, optFns ...func(*Options)) (*ListPartsOutput, error) {
	if !p.HasMorePages() {
		return nil, fmt.Errorf("no more pages available")
	}

	params := *p.params
	params.PartNumberMarker = p.nextToken

	params.MaxParts = p.options.Limit

	result, err := p.client.ListParts(ctx, &params, optFns...)
	if err != nil {
		return nil, err
	}
	p.firstPage = false

	prevToken := p.nextToken
	p.nextToken = result.NextPartNumberMarker

	if p.options.StopOnDuplicateToken && prevToken != nil && p.nextToken != nil && *prevToken == *p.nextToken {
		p.nextToken = nil
	}

	// start of fix
	if p.nextToken != nil && *p.nextToken == "0" {
		p.nextToken = nil
	}
	// end of fix

	return result, nil
}
```

[...]

Issue #1120, opened 2021-02-05
Add S3 customizations documentation


skotambkar:
We should have documentation around S3 customizations and how they work in Developer documentation. For example, private-link, access-points etc.

PR #1141, opened 2021-02-19
Add S3 Customization for IsTruncated paginator check

skmcgrail:
Adds an S3 customization that ensures that the `isTruncated` member is checked before setting the next pagination token.

DependsOn: https://github.com/aws/smithy-go/pull/274
Fixes: https://github.com/aws/aws-sdk-go-v2/issues/1140

### DIFF
--- a/.changes/next-release/service.s3-bugfix-1614016162967995000.json
+++ b/.changes/next-release/service.s3-bugfix-1614016162967995000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.s3-bugfix-1614016162967995000",
+ "SchemaVersion": 1,
+ "Module": "service/s3",
+ "Type": "bugfix",
+ "Description": "Fixed an issue where ListObjectsV2 and ListParts paginators could loop infinitely",
+ "MinVersion": "",
+ "AffectedModules": null
+}


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3PaginationExtensions.java
```diff
@@ -0,0 +1,83 @@
...
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
+
+public class S3PaginationExtensions implements GoIntegration {
+    @Override
+    public byte getOrder() {
+        return -127;
+    }
```
skotambkar, line 37: Nit: Don't have to change it for this, but in general if we are not changing the order, we don't need this override. It can be informational though.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3PaginationExtensions.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public Model preprocessModel(
+            Model model, GoSettings settings
+    ) {
+        ServiceShape service = settings.getService(model);
+        if (!service.expectTrait(ServiceTrait.class).getSdkId().equalsIgnoreCase("S3")) {
```
skotambkar, line 44: This check for S3 is used so often, wondering if it makes sense to declare a generic util that does does it.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3PaginationExtensions.java
```diff
@@ -0,0 +1,83 @@
...
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.PaginatedIndex;
+import software.amazon.smithy.model.knowledge.PaginationInfo;
+import software.amazon.smithy.model.shapes.MemberShape;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+import software.amazon.smithy.model.shapes.StructureShape;
+
+public class S3PaginationExtensions implements GoIntegration {
```
skotambkar, line 33: Should we have a doc.go for why this extension is needed? I did have a doc.go for S3 customization but that is probably out of sync now ! We do have a github issue to document this customizations. May be add a comment on that issue https://github.com/aws/aws-sdk-go-v2/issues/1120 and link this PR. 


Edit: on second thought you should have a commnent on top of this class describing the need of this extension. Ideally, Java CI should fail since this is a public class without Java doc comment.

jasdel (approved):

skotambkar (approved):

service/s3/api_op_ListParts.go
```diff
@@ -334,8 +334,10 @@ func (p *ListPartsPaginator) NextPage(ctx context.Context, optFns ...func(*Optio
 	p.firstPage = false
 
 	prevToken := p.nextToken
-	p.nextToken = result.NextPartNumberMarker
+	if result.IsTruncated {
+		p.nextToken = result.NextPartNumberMarker
 
+	}
```
stefansundin, line 340: I am pretty sure we also need to set `p.nextToken` to `nil`, otherwise won't it just retain the token of the last page? And then not stop just as before.

The current fix will only work if there is only a single page returned.

PR #1160, opened 2021-03-07
Fix nil pointer dereference bug in NewDescribeInstancesPaginator

nnamzlas:
There is a bug in the NewDescribeInstancesPaginator function due to the nil check being after dereferencing the argument. This PR will move the nil check up to prevent this from happening.

The same bug seems to be in other `New_Paginator` functions as well, e.g. https://github.com/aws/aws-sdk-go-v2/blob/main/service/ec2/api_op_DescribeVpcs.go#L211-L223


jasdel:
Thanks for creating this PR @nnamzlas, we've created #1163 that fixes this issue at the SDK's code generator level to ensure this issue is corrected across all API clients.

PR #1161, opened 2021-03-10
codegen: Update Query Protocol generation to always serialized unboxed

### LINKED ISSUE #1107: service/ec2: ModifyNetworkInterfaceAttribute fails when Value: false but works when Value: true
Confirm by changing [ ] to [x] below:
- [x] I've gone though the [API reference](https://github.com/aws/aws-sdk-go-v2/issues)

**Describe the question**

Using v1.1.0 of the sdk, I am unable to disable SourceDestCheck on ENIs with the following code

```
		input := &ec2.ModifyNetworkInterfaceAttributeInput{
			NetworkInterfaceId: aws.String(eni),
			SourceDestCheck: &types.AttributeBooleanValue{
				Value: false,
			},
		}

		_, err := client.ModifyNetworkInterfaceAttribute(context.Background(), input)
```

This returns...

`operation error EC2: ModifyNetworkInterfaceAttribute, https response error StatusCode: 400, RequestID: 947bd2b9-...-8e276816081c, api error InvalidParameterCombination: No attributes specified.`

However when I change Value to true (enable SourceDestCheck) the code works just fine and will enable SourceDestCheck if it isn't already enabled on the ENI

I feel like this is a bug but I could be wrong.

jasdel:
Updates the EC2Query Protocol serializer generation to always serialize unboxed member values to the querystring. This prevents issues with members values never being serialized to the request.

Fixes #1107

### DIFF
--- a/.changes/next-release/service.ec2-bugfix-1615341031066265000.json
+++ b/.changes/next-release/service.ec2-bugfix-1615341031066265000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "service.ec2-bugfix-1615341031066265000",
+ "SchemaVersion": 1,
+ "Module": "service/ec2",
+ "Type": "bugfix",
+ "Description": "Fixes Amazon EC2 serialization of unboxed members, to serialize zero values in request messages.",
+ "MinVersion": "",
+ "AffectedModules": null
+}


JordonPhillips (commented):
Implementation wise this looks fine, but it seems potentially dangerous to be always serializing zero values. What other operations / parameters have you tested this with?

jasdel:
This is the rough part about how SDK's handle `unboxed` members. The API is not supposed to distinguish the difference between unset and zero value (e.g. `false). But in this case with the EC2Query protocol, the API is either treating the value as boxed, which means its modeled wrong, or the field is actually required, but not modeled as so.

In the case of https://github.com/aws/aws-sdk-go-v2/issues/1107 not modeled as required when it should be seems to be the issue.  Whereas [InstanceNetworkInterfaceSpecification.DeviceIndex](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2/types#InstanceNetworkInterfaceSpecification.DeviceIndex) is modeled as an unboxed number but treated as boxed by the API.

```go
	// The position of the network interface in the attachment order. A primary network
	// interface has a device index of 0. If you specify a network interface when
	// launching an instance, you must specify the device index.
	DeviceIndex int32
```

jasdel:
So going to close this PR because this solution is not valid for EC2Query since the API is treating these fields as boxed, and returns errors if zero values are sent in request when they are unexpected.

Issue #1108, opened 2021-02-04
Body in PutObjectInput


wunderbarb:
**Describe the issue with documentation**
in service/s3, `PutObjectInput.Body` is declared as `io.Reader` although it seems to request an `io.ReadSeeker`

**To Reproduce (observed behavior)**
Use a pure `io.Reader` without Seeker when calling `PutObject` generates **operation error S3: PutObject, failed to compute payload hash: failed to seek body to start, request stream is not seekable**
When using a type that supports Reader, Closer and Seeker, no more error.   In V1, it was declared as an io.ReadCloseSeeker.
S3
**Expected behavior**
As declared, I would have expected that `Body` only requires `io.Reader` interface support.

**Screenshots**

**Additional context**

billyzaelani:
I've also tried it. Confirmed that the `PutObjectInput.Body` is `io.ReadSeeker`.

If they want to continue with their current major version, fix the documentation in `PutObjectInput.Body`. But still unpleasant because we miss the compile-time checking.

If they change `PutObjectInput.Body` to `io.ReadSeeker`, they need to increment the major version. I don't know about this solution. Maybe fix the documentation is the way to go.

jasdel:
Thanks for reporting this issue @billyzaelani . A workaround for this is to specify the `UNSIGNED-PAYLOAD` for the object to be upload. This will prevent the SDK from attempting to compute the SHA256 hash of the body. Using the [v4.AddUnsignedPayloadMiddleware](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/signer/v4#AddUnsignedPayloadMiddleware) and [v4.RemoveComputePayloadSHA256Middleware](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/signer/v4#RemoveComputePayloadSHA256Middleware) middleware helpers.

```go
resp, err := client.PutObject(context.TODO(), &s3.PutObjectInput{
    // Params
}, s3.WithAPIOptions(
    v4.AddUnsignedPayloadMiddleware,
    v4.RemoveComputePayloadSHA256Middleware,
))
```

This will instruct the SDK to not compute the SHA256 hash of the body, and instead use the `UNSIGNED-PAYLOAD` value instead. Note PutObject operation call still will require the `ContentLenth` parameter to be set, because The S3 API requires that parameter for all object PUTs.

dectl:
This issue still persists, cannot pass bytes.Buffer to Body.

[...]

PR #1175, opened 2021-03-16
Add documentation for using Amazon S3 with unseekable body

jasdel:
Adds documentation for Using Amazon S3 PutObject operations with values
that do not implement the `io.Seeker` interface.

PR #1176, opened 2021-03-17
service/s3: Add documentation for using unseekable body PutObject and UploadPart 

jasdel:
Adds documentation to the PutObject and UploadPart operations Body member how to upload unseekable objects to an Amazon S3 Bucket.

Fixes: #1108
Depends on: #1175

### DIFF
--- a/aws/signer/v4/middleware.go
+++ b/aws/signer/v4/middleware.go
@@ -156,6 +156,16 @@ func (m *computePayloadSHA256) HandleBuild(
 	return next.HandleBuild(ctx, in)
 }
 
+// SwapComputePayloadSHA256ForUnsignedPayloadMiddleware replaces the
+// ComputePayloadSHA256 middleware with the UnsignedPayload middleware.
+//
+// Use this to disable computing the Payload SHA256 checksum and instead use
+// UNSIGNED-PAYLOAD for the SHA256 value.
+func SwapComputePayloadSHA256ForUnsignedPayloadMiddleware(stack *middleware.Stack) error {
+	_, err := stack.Build.Swap(computePayloadHashMiddlewareID, &unsignedPayload{})
+	return err
+}
+
 // contentSHA256Header sets the X-Amz-Content-Sha256 header value to
 // the Payload hash stored in the context.
 type contentSHA256Header struct{}
--- a/.changes/next-release/sdk-feature-1615946574682127000.json
+++ b/.changes/next-release/sdk-feature-1615946574682127000.json
@@ -1,0 +1,9 @@
+{
+ "ID": "sdk-feature-1615946574682127000",
+ "SchemaVersion": 1,
+ "Module": "/",
+ "Type": "feature",
+ "Description": "Add helper to V4 signer package to swap compute payload hash middleware with unsigned payload middleware",
+ "MinVersion": "",
+ "AffectedModules": null
+}
--- a/aws/signer/v4/middleware_test.go
+++ b/aws/signer/v4/middleware_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/aws/smithy-go/logging"
 	"github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestComputePayloadHashMiddleware(t *testing.T) {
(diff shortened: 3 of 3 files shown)


.changes/next-release/service.s3-bugfix-1615945295096775000.json
```diff
@@ -0,0 +1,9 @@
+{
+ "ID": "service.s3-bugfix-1615945295096775000",
+ "SchemaVersion": 1,
+ "Module": "service/s3",
+ "Type": "bugfix",
```
skotambkar, line 5: Is this bugFix type intentional?

aws/signer/v4/middleware.go
```diff
@@ -156,6 +156,16 @@ func (m *computePayloadSHA256) HandleBuild(
...
 
+// SwapComputePayloadSHA256ForUnsignedPayloadMiddleware replaces the
+// ComputePayloadSHA256 middleware with the UnsignedPayload middleware.
+//
+// Use this to disable computing the Payload SHA256 checksum and instead use
+// UNSIGNED-PAYLOAD for the SHA256 value.
+func SwapComputePayloadSHA256ForUnsignedPayloadMiddleware(stack *middleware.Stack) error {
+	_, err := stack.Build.Swap(computePayloadHashMiddlewareID, &unsignedPayload{})
+	return err
+}
```
skotambkar, line 167: Don't think we should export this as standard util in the SDK. Unsigned payload doesn't work with all services, and is not a security best practice. Why is this required?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3AddPutObjectUnseekableBodyDoc.java
```diff
@@ -0,0 +1,88 @@
...
+            ShapeId.from("com.amazonaws.s3#AmazonS3"), SetUtils.of(
+                    new Pair(ShapeId.from("com.amazonaws.s3#PutObjectRequest"), "Body"),
+                    new Pair(ShapeId.from("com.amazonaws.s3#UploadPartRequest"), "Body")
+            )
+    );
+
+    @Override
+    public byte getOrder() {
+        // This integration should happen before other integrations that rely on the presence of this trait
+        return -60;
```
skotambkar, line 33: Wouldn't pre-process model always run prior to usage of model? Don't think this ordering is required.

jasdel: Can remove. The ordering is always present, will inherit from parent if not defined here. But yeah I don't think order maters in this case. Other customizations where order model is pre-processed is where this will be relevant.

.changes/next-release/service.s3-bugfix-1615945295096775000.json
jasdel: Yeah, the changetool doesn't have `documentation` type, `announcement` doesn't really fit well, but could use that instead of `bugfix`.

aws/signer/v4/middleware.go
jasdel: This could be moved into S3 as a custom middleware generated there, but while its usage is specific to S3, its behavior could be used in other services that do support unsigned payload. Not sure S3.

skotambkar: We can keep this for S3 and generate for services that use Unsignedpayload trait?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3AddPutObjectUnseekableBodyDoc.java
skotambkar: >  Other customizations where order model is pre-processed is where this will be relevant.

Can't think of a use case where order of pre-processing documentation would matter.

aws/signer/v4/middleware.go
jasdel: Moving this to be generated by the relevant services is a option. 

This utility needs to be present in some form because the payload hash middleware ID nor any of the concrete middleware themselves.

.changes/next-release/service.s3-bugfix-1615945295096775000.json
jasdel: Yeah annoucement isn't a fit either since it does not have a version bump associated with it.

given this `bugfix` is still the best type unless we also add `documentation` type.

skotambkar (approved):

skmcgrail (approved):

Thomasdezeeuw:
It seems the documentation change was undone by https://github.com/aws/aws-sdk-go-v2/pull/787, which resulted in me hitting https://github.com/aws/aws-sdk-go-v2/issues/1108 again.

Issue #1192, opened 2021-03-24
ParseInt Error with role duration seconds key


Zeouterlimits:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
Unable to use configured roles due to parsing error of duration_seconds

When instiating an aws client (s3 in my case), the aws go v2 sdk is trying to parse my local roles duration_seconds (28800) and throwing a parse int error exception: 
`error merging role duration seconds key, strconv.ParseInt: parsing "炀": invalid syntax`

Checking out the SDK locally and running the tests reproduces this issue:
```
> cd config
> go build .
> go test .
2021/03/25 11:01:53 http: TLS handshake error from 127.0.0.1:49463: remote error: tls: bad certificate
--- FAIL: TestResolveLogger (0.00s)
    resolve_test.go:206: expect no error, got error merging role duration seconds key, strconv.ParseInt: parsing "炀": invalid syntax
FAIL
FAIL    github.com/aws/aws-sdk-go-v2/config     0.805s
FAIL
> go version
go version go1.15.2 darwin/amd64
```

**Version of AWS SDK for Go?**
aws-sdk-go-v2 1.3.0

**Version of Go (`go version`)?**
1.15.2
Update: Occurs on go1.15.10 too

**To Reproduce (observed behavior)**
Have a configured local aws profile with duration_seconds e.g:
```
[staging]
role_arn = arn:aws:iam::WHATEVER
mfa_serial = arn:aws:iam::WHATEVER
source_profile = aSourceProfile
duration_seconds = 28800
```
Run the sdk unit tests

**Expected behavior**
No error occurs, duration_seconds correctly applied to client config

**Additional context**

jasdel:
Thanks for reporting this issue @Zeouterlimits I added comment in the PR, about compatibility of this change with other SDKs. I don't think the trailing `s` notation is supported by any of the other SDKs, as this field is documented as integer seconds.

Because of this we couldn't take this change, because it changes the behavior of this field, and would not be compatible with the other SDKs.

[...]

PR #1193, opened 2021-03-25
#1192 - Fix shared config loading of Duration Seconds from roles to use raw instead of int64

Zeouterlimits:
For https://github.com/aws/aws-sdk-go-v2/issues/1192

### DIFF
--- a/internal/ini/literal_tokens.go
+++ b/internal/ini/literal_tokens.go
@@ -203,7 +203,7 @@ func NewIntValue(i int64) (Value, error) {
 	return newValue(IntegerType, 10, []rune{rune(i)})
 }
 
-// NewRawIntValue returns a Value type generated using an int64 input.
+// NewRawIntValue returns an Integer Value type generated using a rune input.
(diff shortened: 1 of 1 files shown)


jasdel:
Thank you for reaching out and creating this PR @Zeouterlimits. After reviewing this parameter I don't think it would be appropriate for the SDK to take this change. The field is documented as only taking an integer of seconds. Parsing this field as a `time.Duration` would create the risk of compatibility issues between the v2 SDK and other AWS SDKs. Since the other SDKs only support integer seconds for this field.

Zeouterlimits:
(Reopened per further discussion in issue #1192)

jasdel:
I've not been able to reproduce this issue, and created PR https://github.com/aws/aws-sdk-go-v2/pull/1568 exercising this. That Pr can replace this one until can reproduce the error. Lets continue discussion in [#](https://github.com/aws/aws-sdk-go-v2/issues/1192)

PR #1196, opened 2021-03-29
Typo in file name

sciarrilli:
Change from v1 to v2 in header in the doc


jasdel:
Thanks for fixing this typo @sciarrilli. This file is generated from another repository. We'll need to update the source the file is pulled from in order to fix this.

Issue #1191, opened 2021-03-23
API Gateway request dropping stage from endpoint


laverboy:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
When using a required custom resolver with apigatewaymanagmentapi service it appears that the stage is being dropped from the url used.

**Version of AWS SDK for Go?**
v1.2.1

**Version of Go (`go version`)?**
1.15

**To Reproduce (observed behavior)**
We are using the following custom resolver
```
customResolver := aws.EndpointResolverFunc(func(service, region string) (aws.Endpoint, error) {
		if service == apigatewaymanagementapi.ServiceID && region == "eu-west-1" {
			return aws.Endpoint{
				PartitionID:       "aws",
				URL:                 "https://xxxxxx.execute-api.eu-west-1.amazonaws.com/LATEST",
				SigningRegion:     "eu-west-1",
				HostnameImmutable: true,
				Source:            aws.EndpointSourceCustom,
			}, nil
		}
		return aws.Endpoint{}, fmt.Errorf("unknown endpoint requested")
	})

	cfg, err := config.LoadDefaultConfig(context.Background(), config.WithEndpointResolver(customResolver), config.WithClientLogMode(aws.LogRetries|aws.LogRequest))
```

And we are making the request like so:
```
_, err := wsc.client.GetConnection(ctx, &apigatewaymanagementapi.GetConnectionInput{
		ConnectionId: aws.String("xxxxxxxxxx=x"),
	})
```

See logs below

**Expected behavior**
We are expecting to get a successful response from the api.

**Additional context**
We are seeing the following error logs, it appears that the url no longer includes *LATEST* which is the name of our stage

```
SDK 2021/03/23 15:44:54 DEBUG Request
--
GET /@connections/cpV6NdyODoECInw%3D HTTP/1.1
Host: xxxxxxxx.execute-api.eu-west-1.amazonaws.com
User-Agent: aws-sdk-go-v2/1.2.1
[...]

PR #1198, opened 2021-03-30
Fix API clients ignoring resolved endpoint's URL Path and RawQuery

jasdel:
Fixes the API clients to not ignore the resolved endpoint's URL Path and RawQuery parameters. Instead of ignoring the values the SDK will append the operation's modeled Path and Query String.

Fixes https://github.com/aws/aws-sdk-go-v2/issues/1191

Depends on https://github.com/aws/smithy-go/pull/280


jasdel:
closing this in favor of https://github.com/aws/aws-sdk-go-v2/pull/1199

PR #1210, opened 2021-04-06
codegen: Add make task to easily pull in smithy-go codegen modules

jasdel:
Updates SDK's makefile to provide a command that can be chained with other
make tasks for getting the latest smithy-go codegen into maven local.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,11 @@ smithy-build-%: gen-repo-mod-replace
 smithy-clean:
 	cd codegen && ./gradlew clean
 
+smithy-go-publish-local:
+	rm -rf /tmp/smithy-go-local
+	git clone https://github.com/aws/smithy-go /tmp/smithy-go-local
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -64,6 +64,11 @@ smithy-build-%: gen-repo-mod-replace
 smithy-clean:
 	cd codegen && ./gradlew clean
 
+smithy-go-publish-local:
+	rm -rf /tmp/smithy-go-local
+	git clone https://github.com/aws/smithy-go /tmp/smithy-go-local
+	make -C /tmp/smithy-go-local smithy-publish-local
```
skotambkar, line 70: should we do `smithy-clean` before `publish-local` ?

jasdel: Shouldn't be needed since the `smithy-go` repo is freshly created before the make command is run. But it shouldn't hurt if there is some caching that gits stuck out there.

skotambkar (approved):

PR #1213, opened 2021-04-07
sync api models and endpoint models

skotambkar:
Sync smithy models


skotambkar:
closed in favor of #1219

PR #1215, opened 2021-04-08
aws/signer/v4: Keep Object-Lock headers a header in presigning signing requests

manucorporat:
Object-lock is not working for presigned URLs


jasdel:
Thanks for taking the time to create this PR @manucorporat I took this change and created PR https://github.com/aws/aws-sdk-go-v2/pull/1307 so a pattern of headers are not hoisted to the query string.

Issue #1169, opened 2021-03-12
rdsutils.BuildAuthToken missing in SDK v2


szymon3:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
func BuildAuthToken for RDS Proxy (from rdsutils module), known from SDK v1, is missing in SDK v2.
Is there any alternative way to retrieve token using SDK v2?

**Version of AWS SDK for Go?**
1.2.0

**Version of Go (`go version`)?**
go version go1.16 darwin/amd64

**To Reproduce (observed behavior)**
-

**Expected behavior**
I expect to have a function which will return me a token for RDS Proxy.

**Additional context**
I saw that functionality was copied from SDK v1, but later was removed in one of commits (https://github.com/aws/aws-sdk-go-v2/commit/eecb706f5d1e3ca44aafca5c042ea275f4050764#diff-457ec6738454cb66ee5a04f7b14c84ecf31f37cb2f42f428cc28dc099970f8cd).

skmcgrail:
Thank you for bringing this to our attention @szymon3 it looks like we missed pulling this feature back over after we performed the client refresh last year.

haochunchang:
For those who got here before reading v2 developer guide,
When I want to use `rdsutils.BuildAuthToken` for IAM database authentication, I got the import error when using:
```go
import "github.com/aws/aws-sdk-go-v2/service/rds/rdsutils"
```
Since the [documentation](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/rds/rdsutils?utm_source=gopls#pkg-overview) path is: [service](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service) > [rds](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/rds) > [rdsutils](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.24.0/service/rds/rdsutils)

However, the import statement in v2 should be:
```go
[...]

PR #1230, opened 2021-04-16
Add Utilities for RDS IAM Authentication

skmcgrail:
Ports V1 utilities for constructing RDS database connection strings using IAM authentication.

Closes: #1169

### DIFF
--- a/feature/rds/auth/connect.go
+++ b/feature/rds/auth/connect.go
@@ -1,0 +1,96 @@
+package auth
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
+)
+
+const (
+	signingID        = "rds-db"
+	emptyPayloadHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+)
+
+// BuildAuthTokenOptions is the optional set of configuration properties for BuildAuthToken
+type BuildAuthTokenOptions struct{}
+
+// BuildAuthToken will return an authorization token used as the password for a DB
+// connection.
+//
+// * endpoint - Endpoint consists of the port needed to connect to the DB. <host>:<port>
+// * region - Region is the location of where the DB is
+// * dbUser - User account within the database to sign in with
+// * creds - Credentials to be signed with
+//
+// The following example shows how to use BuildAuthToken to create an authentication
+// token for connecting to a MySQL database in RDS.
+//
+//   authToken, err := BuildAuthToken(dbEndpoint, awsRegion, dbUser, awsCreds)
+//
+//   // Create the MySQL DNS string for the DB connection
+//   // user:password@protocol(endpoint)/dbname?<params>
+//   connectStr = fmt.Sprintf("%s:%s@tcp(%s)/%s?allowCleartextPasswords=true&tls=rds",
+//      dbUser, authToken, dbEndpoint, dbName,
+//   )
+//
+//   // Use db to perform SQL operations on database
+//   db, err := sql.Open("mysql", connectStr)
+//
+// See http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html
+// for more information on using IAM database authentication with RDS.
+func BuildAuthToken(ctx context.Context, endpoint, region, dbUser string, creds aws.CredentialsProvider, optFns ...func(options *BuildAuthTokenOptions)) (string, error) {
+	o := BuildAuthTokenOptions{}
+
+	for _, fn := range optFns {
+		fn(&o)
+	}
+
+	if creds == nil {
+		return "", fmt.Errorf("credetials provider must not ne nil")
+	}
+
+	// the scheme is arbitrary and is only needed because validation of the URL requires one.
+	if !(strings.HasPrefix(endpoint, "http://") || strings.HasPrefix(endpoint, "https://")) {
+		endpoint = "https://" + endpoint
+	}
+
+	req, err := http.NewRequest("GET", endpoint, nil)
+	if err != nil {
+		return "", err
+	}
+	values := req.URL.Query()
+	values.Set("Action", "connect")
+	values.Set("DBUser", dbUser)
+	req.URL.RawQuery = values.Encode()
+
+	signer := v4.NewSigner()
+
+	credentials, err := creds.Retrieve(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	// Expire Time: 15 minute
+	query := req.URL.Query()
+	query.Set("X-Amz-Expires", "900")
+	req.URL.RawQuery = query.Encode()
+
+	signedURI, _, err := signer.PresignHTTP(ctx, credentials, req, emptyPayloadHash, signingID, region, time.Now().UTC())
+	if err != nil {
+		return "", err
+	}
+
+	url := signedURI
+	if strings.HasPrefix(url, "http://") {
+		url = url[len("http://"):]
+	} else if strings.HasPrefix(url, "https://") {
+		url = url[len("https://"):]
+	}
+
+	return url, nil
+}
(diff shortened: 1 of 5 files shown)


skmcgrail:
I'm on the fence of whether we want to keep `ConnectionStringBuilder` in the V2 implementation. The external examples from the RDS service docuemtnation has only ever documented usage around `BuildAuthToken`. Reasons for not keeping `ConnectionStringBuilder` is that the connection string produced is database driver specific and differs between MySQL and PostgreSQL. Removing this would leave it up to the user to define the connection string format that is appropriate for their database driver.

feature/rds/auth/builder.go
```diff
@@ -0,0 +1,123 @@
...
+// WithFormat will return a builder with the given connection format
+func (b ConnectionStringBuilder) WithFormat(f ConnectionFormat) ConnectionStringBuilder {
+	b.connectFormat = f
+	return b
+}
+
+// WithTCPFormat will set the format to TCP and return the modified builder
+func (b ConnectionStringBuilder) WithTCPFormat() ConnectionStringBuilder {
+	return b.WithFormat(TCPFormat)
+}
```
skotambkar, line 92: Won't these `With<Member>` helpers return a copy of builder each time.  Don't think these helpers are needed.

feature/rds/auth/builder.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	connectionStr := fmt.Sprintf("%s:%s@%s(%s)/%s",
+		b.user, authToken, string(b.connectFormat), b.endpoint, b.dbName,
+	)
+
+	if len(b.params) > 0 {
+		connectionStr = fmt.Sprintf("%s?%s", connectionStr, b.params.Encode())
+	}
+	return connectionStr, nil
```
skotambkar, line 122: Nit: can be shortened

```
if len(b.params) > 0 {
    return fmt.Sprintf("%s?%s", connectionStr, b.params.Encode()), nil
}

return fmt.Sprintf("%s:%s@%s(%s)/%s", 
    b.user, authToken, string(b.connectFormat), b.endpoint, b.dbName), nil
```

feature/rds/auth/builder.go
```diff
@@ -0,0 +1,123 @@
...
+//	connectStr, err := b.WithTCPFormat().Build(context.TODO())
+//	if err != nil {
+//		panic(err)
+//	}
+//	const dbType = "mysql"
+//	db, err := sql.Open(dbType, connectStr)
+func (b ConnectionStringBuilder) Build(ctx context.Context) (string, error) {
+	if b.connectFormat == NoConnectionFormat {
+		return "", fmt.Errorf("no connection format was specified")
+	}
```
skotambkar, line 108: If `NoConnectionFormat` isn't a valid choice, why have it as an enum const?

feature/rds/auth/builder_test.go
```diff
@@ -0,0 +1,59 @@
...
+		{
+			user:                 "foo",
+			endpoint:             "foo.bar",
+			region:               "region",
+			dbName:               "name",
+			format:               auth.NoConnectionFormat,
+			creds:                &staticCredentials{AccessKey: "AKID", SecretKey: "SECRET", Session: "SESSION"},
+			expectErr:            true,
+			expectedConnectRegex: "",
+		},
```
skotambkar, line 35: This test will be removed/change if `NoConnectionFormat` is not kept as a valid enum

skotambkar (changes requested):

feature/rds/auth/builder.go
skmcgrail: See general question I asked about whether we want to provide this type.

feature/rds/auth/builder.go
```diff
@@ -0,0 +1,123 @@
...
+// NewConnectionStringBuilder will return an ConnectionStringBuilder
+func NewConnectionStringBuilder(endpoint, region, dbUser, dbName string, creds aws.CredentialsProvider) ConnectionStringBuilder {
+	return ConnectionStringBuilder{
+		dbName:   dbName,
+		endpoint: endpoint,
+		region:   region,
+		user:     dbUser,
+		creds:    creds,
+	}
+}
```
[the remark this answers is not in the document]
skmcgrail: It's an explicit builder type, so the intent is that you don't use struct initialization with it.

feature/rds/auth/builder.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+)
+
+// ConnectionFormat is the type of connection that will be
+// used to connect to the database
+type ConnectionFormat string
+
+// ConnectionFormat enums
+const (
+	NoConnectionFormat ConnectionFormat = ""
```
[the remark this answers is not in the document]
skmcgrail: No I don't think that makes sense to apply generically across the whole SDK, and it's not something we've done in other non-generated code. That pattern exists in codegen to reduce the chances of creating conflicting types.

feature/rds/auth/builder.go
skmcgrail: Since it is a builder pattern you need to have a sentinel value to know whether the value was set to something else.

feature/rds/auth/builder.go
skmcgrail: That would not be valid, as that results in a different connection string.

feature/rds/auth/connect.go
```diff
@@ -0,0 +1,82 @@
...
+		return "", err
+	}
+	values := req.URL.Query()
+	values.Set("Action", "connect")
+	values.Set("DBUser", dbUser)
+	req.URL.RawQuery = values.Encode()
+
+	signer := v4.NewSigner()
+
+	credentials, err := creds.Retrieve(ctx)
```
skotambkar, line 59: qq: can creds be nil?

feature/rds/auth/connect_test.go
```diff
@@ -0,0 +1,55 @@
+package auth_test
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"regexp"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
```
skotambkar, line 9: nit: import formating

feature/rds/auth/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/aws/aws-sdk-go-v2/feature/rds/auth
+
+go 1.15
```
skotambkar, line 3: can we make this util have min version go 1.16?

skotambkar (approved):

feature/rds/auth/connect.go
skmcgrail: The returned credentials type is a value type. But I will added a check to verify that the `CredentialsProvider` itself is not nil.

feature/rds/auth/go.mod
skmcgrail: I don't think we want to manage user accessible modules in that manner. Otherwise it's a pretty confusing experience on communicating what versions of the Go compiler we support. For modules that are not internal I think it's best we are consistent with what the minimum version is across the board. Obvious exceptions to that statement when we start major versioning components and "deprecating" old versions etc.

feature/rds/auth/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package auth is used to generate authentication tokens used to
+// connect to a givent Amazon Relational Database Service (RDS) database.
+//
+// Before using the authentication please visit the docs here to ensure
+// the database has the proper policies to allow for IAM token authentication.
+// https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html#UsingWithRDS.IAMDBAuth.Availability
```
skotambkar, line 6: Nit: we can point to our developer docs too - since they are more helpful wrt usage in SDK .

skotambkar (approved):

feature/rds/auth/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package auth is used to generate authentication tokens used to
+// connect to a givent Amazon Relational Database Service (RDS) database.
```
skotambkar, line 2: Nit: givent ->given

PR #1238, opened 2021-04-22
Sync API Models

skmcgrail:
* Sync API Models
* Added support for `documentation` change type.
* Added change metadata for next release.


skotambkar:
The miscellanous cleanup commit has a change to sts model . We are missing a corresponding changelog entry for sts.

jasdel (approved):

skotambkar (approved):

PR #1240, opened 2021-04-23
Migrate to smithy 1.7 and generate APIGateway service

skotambkar:
Depends on https://github.com/aws/smithy-go/pull/289

* Generates latest APIGateway model 

* Updates codegen to use rename pattern i.e. shapeId.getName(service) when retrieving shape name

TODO:

[ x ] generate and validate protocol tests

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestXmlProtocolGenerator.java
@@ -378,7 +378,7 @@ abstract class RestXmlProtocolGenerator extends HttpBindingProtocolGenerator {
         });
     }
 
-    // Write middleware that delegates to deserializers for shapes that have implicit payload
+    // Writes middleware that delegates to deserializers for shapes that have implicit payload.
     private void writeMiddlewareDocumentBindingDeserializerDelegator(
             GenerationContext context,
             GoWriter writer,


Makefile
```diff
@@ -71,7 +71,9 @@ smithy-go-publish-local:
 
 gen-config-asserts:
 	@echo "Generating SDK config package implementor assertions"
-	cd config && go generate
+	cd config \
+	    && go mod tidy \
```
jasdel, line 75: does mod tidy need to be mixed into these tasks? mod tidy has a separate task that will be run by make generate.

service/apigateway/serializers.go
```diff
@@ -4371,20 +4369,6 @@ func awsRestjson1_serializeOpHttpBindingsGetExportInput(v *GetExportInput, encod
...
-func awsRestjson1_serializeOpDocumentGetExportInput(v *GetExportInput, value smithyjson.Value) error {
-	object := value.Object()
-	defer object.Close()
-
-	if v.Parameters != nil {
-		ok := object.Key("parameters")
-		if err := awsRestjson1_serializeDocumentMapOfStringToString(v.Parameters, ok); err != nil {
-			return err
-		}
-	}
```
jasdel, line 4383: Was this incorrectly being serialized to the payload body before this change?  Looks like so, because this is a GET operation.

jasdel (approved):

Makefile
skotambkar: We have to run go mod tidy, when we cd into Config. Or we get an error related to go sum entry not found.

service/apigateway/serializers.go
skotambkar: Yeah looks like it, this change should fix it.

Makefile
jasdel: Ah i see, could also set `GOFLAGS=-mod=mod` here as well potentially instead of running mod tidy.

PR #1249, opened 2021-05-03
Add dynamoDB examples

jrichard8:
Add some example of use of dynamoDB with AWS SDK v2

### DIFF
--- a/example/service/dynamodb/scanItems/scanItems.go
+++ b/example/service/dynamodb/scanItems/scanItems.go
@@ -20,6 +20,7 @@ func init() {
 	flag.StringVar(&region, "region", "", "The `region` of your AWS project.")
 }
 
+// Record holds info about the records returned by Scan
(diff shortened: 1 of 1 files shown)


skotambkar:
@Doug-AWS can you take a look at this and see if these should be added to our V2 examples repo?

jasdel (changes requested):
Examples look good, we'll want to update the dependencies to be SDK's latest versions, and update the examples to use the SDK's ddb attribute value (un)marshalers.

jrichard8:
I think I made requested changes, waiting for review.

jrichard8:
@jasdel can you take a look ?

jasdel (approved):
Thanks for taking the time to create these examples. I've kicked off the CI Actions, and the changes look good. When these pass we should be able to merge in these example modules.

PR #1255, opened 2021-05-13
config: Enable SSO provider to be mixed with other credential providers

### LINKED ISSUE #1204: Failed to load assume role, of profile, failed to get shared config profile
The SDK appears to have a bug that does not exist in the V1 SDK. Given the following config and credentials setup the SDK will return an unexpected error `failed to load assume role foofoofoofoofoofoofoofoofoofoofoofoofoofoo, of profile A, failed to get shared config profile, A`.

`~/.aws/config`:
```
[default]
region=us-west-2
external_id=FooBar
role_session_name=FooSessionName
role_arn=foofoofoofoofoofoofoofoofoofoofoofoofoofoo
duration_seconds = 43200
source_profile=A

[profile A]
external_id=AExternalID
role_session_name=ASessionName
role_arn=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
duration_seconds = 43210
source_profile=B
```

`~/.aws/credentials`:
```
[A]
aws_access_key_id=A_PROFILE_ACCESS_KEY_ID
aws_secret_access_key=A_PROFILE_SECRET_ACCESS_KEY
aws_session_token=A_PROFILE_SESSION_TOKEN

[B]
aws_access_key_id=B_PROFILE_ACCESS_KEY_ID
aws_secret_access_key=B_PROFILE_SECRET_ACCESS_KEY
aws_session_token=B_PROFILE_SESSION_TOKEN
```

skmcgrail:
Also fixes:

- Fixes #1204 
- Fixes incorrect assume role call with web identity token provider


config/testdata/gh1204_credentials
```diff
@@ -0,0 +1,9 @@
+[A]
```
skotambkar, line 1: Could you also add testcases for:
1. duplicate profiles with same case type within a single file. 
2. similar profiles with different case type (that match if normalized) within a single file.
3. duplicate profiles with same case type across config & cred file. 
4. similar profiles with different case type (that match if normalized) across config & cred file.

config/testdata/gh1204_config
```diff
@@ -0,0 +1,14 @@
+[default]
+region=us-west-2
+external_id=FooBar
+role_session_name=FooSessionName
+role_arn=foofoofoofoofoofoofoofoofoofoofoofoofoofoo
+duration_seconds = 43200
+source_profile=A
```
skotambkar, line 7: Just as an extra check, could you add two additional profiles in this file - one with mixed case & other lower case profile name that may potentially have conflict with the profile name used for source_profile. 

Could also add an error case to assert a descriptive error is thrown if an exact match isn't available for the profile specified as source_profile.

.changelog/016af652b95e472f960ded9a2e976e11.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "016af652-b95e-472f-960d-ed9a2e976e11",
+    "type": "bugfix",
+    "description": "`internal/ini`: Disable noramalization of config profile names",
+    "modules": [
+        "."
+    ]
```
skotambkar, line 7: Minor nit: s/noramalization/normalization 

Also do you think its more readable to use "core" than "." (not much important as this info is displayed externally)

skmcgrail: Fixed typo, as indicated in another PR `modules` is used to indicate paths relative to the repository source root where a module is located, and not the name of the module. This intentionally done to keep the code generic enough so that we can use it in the smithy-go repository without special logic. Thus why there is no special treatment or aliasing involved here.

skotambkar (approved):

config/resolve_credentials_test.go
```diff
@@ -61,11 +61,31 @@ func setupCredentialsEndpoints(t *testing.T) (aws.EndpointResolver, func()) {
...
+
+			form := r.Form
+
+			switch form.Get("Action") {
+			case "AssumeRole":
+				w.Write([]byte(fmt.Sprintf(
+					assumeRoleRespMsg,
+					time.Now().
+						Add(15*time.Minute).
+						Format("2006-01-02T15:04:05Z"))))
```
jasdel, line 77: similar to v1 using defined constant for these timestamp formats

jasdel (approved):

PR #1257, opened 2021-05-13
Fix REST-XML Codegen Bug, Move metrics to User-Agent header, Update SDKVersion to refer to goModuleVersion

skmcgrail:
* Make go1.16 available in Travis Java builds
* aws: Updated SDKVersion to refer to goModuleVersion constant.
* Add Changelog Annotation
* Regenerated Clients
* internal/repotools: Support specifying alternative package location for go_module_metadata.go
* Updated middleware to temporarily direct client metadata to user-agent
* Add serviceId and module version to user-agent header
* Fix deserialization of enum types marked with payload trait

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,11 @@ matrix:
       env:
         - JAVA_TOOL_OPTIONS="-Xmx1g"
       before_script:
+          - sudo apt-get -y install golang
+          - go get golang.org/dl/go1.16.4
+          - $(go env GOPATH)/bin/go1.16.4 download
+          - export PATH=$HOME/sdk/go1.16.4/bin:$PATH
+          - export GOROOT=$HOME/sdk/go1.16.4
           - rm -rf /tmp/smithy-go
           - git clone https://github.com/aws/smithy-go /tmp/smithy-go
           - pushd /tmp/smithy-go/codegen
@@ -61,6 +66,11 @@ matrix:
       env:
         - JAVA_TOOL_OPTIONS="-Xmx1g"
       before_script:
+          - sudo apt-get -y install golang
+          - go get golang.org/dl/go1.16.4
+          - $(go env GOPATH)/bin/go1.16.4 download
+          - export PATH=$HOME/sdk/go1.16.4/bin:$PATH
+          - export GOROOT=$HOME/sdk/go1.16.4
           - rm -rf /tmp/smithy-go
           - git clone https://github.com/aws/smithy-go /tmp/smithy-go
           - pushd /tmp/smithy-go/codegen


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsClientUserAgent.java
```diff
@@ -41,13 +41,18 @@ public void writeAdditionalFiles(
             GoDelegator goDelegator
     ) {
         ServiceTrait serviceTrait = settings.getService(model).expectTrait(ServiceTrait.class);
+        String servceId = serviceTrait.getSdkId().replace("-", "").replace(" ", "").toLowerCase();
```
skotambkar, line 44: Nit: Java replace takes a regex. Right now it will traverse that string thrice I think.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/RestJsonProtocolGenerator.java
```diff
@@ -379,7 +379,7 @@ private void writePayloadBindingDeserializer(
                                     } else { // string
                                         writer.addUseImports(SmithyGoDependency.SMITHY_PTR);
                                         if (targetShape.hasTrait(EnumTrait.class)) {
-                                            writer.write("v.$L = string(bs)", memberName);
+                                            writer.write("v.$L = $T(bs)", memberName, symbolProvider.toSymbol(targetShape));
```
skotambkar, line 382: Nit: can use similar suggestion here - https://github.com/aws/smithy-go/pull/296#discussion_r632050640

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
```diff
@@ -106,7 +106,7 @@ protected void serializeMap(GenerationContext context, MapShape shape) {
             writer.insertTrailingNewline();
 
             // Serialize zero values as empty values.
-            GoValueAccessUtils.writeIfZeroValue(context.getModel(), writer, shape.getValue(), "v[i]", () -> {
+            GoValueAccessUtils.writeIfZeroValue(context.getModel(), writer, shape.getValue(), "v[key]", () -> {
```
skotambkar, line 109: This makes me wonder if `queryParamsMap` serialization need something similar.  What is this changing exactly?

skotambkar (approved):
LGTM, most are minor comments / questions. Can merge this PR.

.travis.yml
```diff
@@ -48,6 +48,11 @@ matrix:
       env:
         - JAVA_TOOL_OPTIONS="-Xmx1g"
       before_script:
+          - sudo apt-get -y install golang
+          - go get golang.org/dl/go1.16.4
+          - $(go env GOPATH)/bin/go1.16.4 download
+          - export PATH=$HOME/sdk/go1.16.4/bin:$PATH
+          - export GOROOT=$HOME/sdk/go1.16.4
```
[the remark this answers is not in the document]
skmcgrail: Yeah we can use this strategy to replace our current gotip build process etc.

.changelog/679a53c14cfe4943867d8e23cc85abb9.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "679a53c1-4cfe-4943-867d-8e23cc85abb9",
+    "type": "feature",
+    "description": "`AddSDKAgentKey` and `AddSDKAgentKeyValue` in `aws/middleware` package have been updated to direct metadata to `User-Agent` HTTP header.",
+    "modules": [
+        "."
```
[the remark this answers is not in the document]
skmcgrail: The reason for not doing this is mainly to avoid having to deal with that condition in multiple spots, "." automatically plays nice with `path` and `filepath` as you would expect when you are joining this relative locations to the repository path on the filesystem.

internal/repotools/cmd/updatemodulemeta/main.go
```diff
@@ -175,7 +183,7 @@ func writeModuleMetadata(dir string, goPackage string, version string) (err erro
 
 	return metadataTemplate.Execute(f, metadata{
 		Package: goPackage,
-		Version: version,
+		Version: strings.TrimPrefix(version, "v"),
```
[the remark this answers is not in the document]
skmcgrail: Mainly since the SDK version constant has always been reported without it. Same with the product version that is sent in the user-agent, so opt-ed for consistency here.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/XmlShapeSerVisitor.java
skmcgrail: `i` was not a valid variable here, looks like this was some sort of copy paste error, but there was no code-gen that exercised this path before. the variable in scope that it should have been using was `key`. Probably was copied from the code that handled lists.

PR #1263, opened 2021-05-18
Endpoint discovery support

skotambkar:
Expect Java codegen CI to fail until smithy-go PR is merged.

Depends on https://github.com/aws/smithy-go/pull/301

Please review individual commits while reviewing this PR 

Changes per commit
---
1. Add configuration setting for enabling endpoint discovery.
2. Adds shared components for endpoint discovery endpoint - cache and middleware
3. Adds assertions for config sources
4. Adds java codegen logic to orchestrate endpoint discovery.  
5. Generated TimeStreamWrite client
6. Generated TimeStreamQuery client
7. Generated Dynamodb client
8. Adds integration test for timestreamwrite service. 
9. All other service clients and protocol tests
10. Repotool change to not fail if build artifact is absent. 

Revision 2 
---

Commits to review

11. `EnableEndpointDiscoveryState` enums are now typed. Shared config resolve does not fall back to `auto` for enable endpoint discovery setting. Renames `assert` package to `configtesting` as suggested. 
12. Java codegen to encapsulate endpoint discovery related options within `EndpointDiscoveryOptions` struct. 
13. Update endpoint discovery cache to try deletion before storing endpoint key if endpoint cache limit is reached. 
14. Generated Dynamodb client
15. Generated timestream clients

Revision 3
---
16. Removes stale code and fixes a comment.
17. Adds changelog entries.z
18. Rebased with main branch, and generated client change for add helpers.

### DIFF
--- a/service/kendra/api_op_BatchGetDocumentStatus.go
+++ b/service/kendra/api_op_BatchGetDocumentStatus.go
@@ -27,7 +27,7 @@ func (c *Client) BatchGetDocumentStatus(ctx context.Context, params *BatchGetDoc
 		params = &BatchGetDocumentStatusInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "BatchGetDocumentStatus", params, optFns, addOperationBatchGetDocumentStatusMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "BatchGetDocumentStatus", params, optFns, c.addOperationBatchGetDocumentStatusMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -69,7 +69,7 @@ type BatchGetDocumentStatusOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationBatchGetDocumentStatusMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationBatchGetDocumentStatusMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson11_serializeOpBatchGetDocumentStatus{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/pi/api_op_GetDimensionKeyDetails.go
+++ b/service/pi/api_op_GetDimensionKeyDetails.go
@@ -21,7 +21,7 @@ func (c *Client) GetDimensionKeyDetails(ctx context.Context, params *GetDimensio
 		params = &GetDimensionKeyDetailsInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "GetDimensionKeyDetails", params, optFns, addOperationGetDimensionKeyDetailsMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "GetDimensionKeyDetails", params, optFns, c.addOperationGetDimensionKeyDetailsMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -76,7 +76,7 @@ type GetDimensionKeyDetailsOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationGetDimensionKeyDetailsMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationGetDimensionKeyDetailsMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson11_serializeOpGetDimensionKeyDetails{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_AcceptEnvironmentAccountConnection.go
+++ b/service/proton/api_op_AcceptEnvironmentAccountConnection.go
@@ -21,7 +21,7 @@ func (c *Client) AcceptEnvironmentAccountConnection(ctx context.Context, params 
 		params = &AcceptEnvironmentAccountConnectionInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "AcceptEnvironmentAccountConnection", params, optFns, addOperationAcceptEnvironmentAccountConnectionMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "AcceptEnvironmentAccountConnection", params, optFns, c.addOperationAcceptEnvironmentAccountConnectionMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -50,7 +50,7 @@ type AcceptEnvironmentAccountConnectionOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationAcceptEnvironmentAccountConnectionMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationAcceptEnvironmentAccountConnectionMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson10_serializeOpAcceptEnvironmentAccountConnection{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_CancelEnvironmentDeployment.go
+++ b/service/proton/api_op_CancelEnvironmentDeployment.go
@@ -31,7 +31,7 @@ func (c *Client) CancelEnvironmentDeployment(ctx context.Context, params *Cancel
 		params = &CancelEnvironmentDeploymentInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "CancelEnvironmentDeployment", params, optFns, addOperationCancelEnvironmentDeploymentMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "CancelEnvironmentDeployment", params, optFns, c.addOperationCancelEnvironmentDeploymentMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -60,7 +60,7 @@ type CancelEnvironmentDeploymentOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationCancelEnvironmentDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationCancelEnvironmentDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson10_serializeOpCancelEnvironmentDeployment{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_CancelServiceInstanceDeployment.go
+++ b/service/proton/api_op_CancelServiceInstanceDeployment.go
@@ -34,7 +34,7 @@ func (c *Client) CancelServiceInstanceDeployment(ctx context.Context, params *Ca
 		params = &CancelServiceInstanceDeploymentInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "CancelServiceInstanceDeployment", params, optFns, addOperationCancelServiceInstanceDeploymentMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "CancelServiceInstanceDeployment", params, optFns, c.addOperationCancelServiceInstanceDeploymentMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -68,7 +68,7 @@ type CancelServiceInstanceDeploymentOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationCancelServiceInstanceDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationCancelServiceInstanceDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson10_serializeOpCancelServiceInstanceDeployment{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_CancelServicePipelineDeployment.go
+++ b/service/proton/api_op_CancelServicePipelineDeployment.go
@@ -34,7 +34,7 @@ func (c *Client) CancelServicePipelineDeployment(ctx context.Context, params *Ca
 		params = &CancelServicePipelineDeploymentInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "CancelServicePipelineDeployment", params, optFns, addOperationCancelServicePipelineDeploymentMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "CancelServicePipelineDeployment", params, optFns, c.addOperationCancelServicePipelineDeploymentMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -63,7 +63,7 @@ type CancelServicePipelineDeploymentOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationCancelServicePipelineDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationCancelServicePipelineDeploymentMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson10_serializeOpCancelServicePipelineDeployment{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_CreateEnvironment.go
+++ b/service/proton/api_op_CreateEnvironment.go
@@ -21,7 +21,7 @@ func (c *Client) CreateEnvironment(ctx context.Context, params *CreateEnvironmen
 		params = &CreateEnvironmentInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "CreateEnvironment", params, optFns, addOperationCreateEnvironmentMiddlewares)
+	result, metadata, err := c.invokeOperation(ctx, "CreateEnvironment", params, optFns, c.addOperationCreateEnvironmentMiddlewares)
 	if err != nil {
 		return nil, err
 	}
@@ -98,7 +98,7 @@ type CreateEnvironmentOutput struct {
 	ResultMetadata middleware.Metadata
 }
 
-func addOperationCreateEnvironmentMiddlewares(stack *middleware.Stack, options Options) (err error) {
+func (c *Client) addOperationCreateEnvironmentMiddlewares(stack *middleware.Stack, options Options) (err error) {
 	err = stack.Serialize.Add(&awsAwsjson10_serializeOpCreateEnvironment{}, middleware.After)
 	if err != nil {
 		return err
--- a/service/proton/api_op_CreateEnvironmentAccountConnection.go
+++ b/service/proton/api_op_CreateEnvironmentAccountConnection.go
@@ -22,7 +22,7 @@ func (c *Client) CreateEnvironmentAccountConnection(ctx context.Context, params 
 		params = &CreateEnvironmentAccountConnectionInput{}
 	}
 
-	result, metadata, err := c.invokeOperation(ctx, "CreateEnvironmentAccountConnection", params, optFns, addOperationCreateEnvironmentAccountConnectionMiddlewares)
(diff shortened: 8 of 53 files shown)


config/env_config.go
```diff
@@ -157,7 +157,7 @@ type EnvConfig struct {
 	// Enables endpoint discovery via environment variables.
 	//
 	//	AWS_ENABLE_ENDPOINT_DISCOVERY=true
-	EnableEndpointDiscovery *bool
+	EnableEndpointDiscovery *EndpointDiscoveryEnableState
```
skmcgrail, line 160: Should this be `EndpointDiscoveryState`? Then the constants are `EndpointDiscoveryAuto`, `EndpointDiscoveryEnabled`, `EndpointDiscoveryDisabled`?

Also what is your thought of making the first value (zero-value) as being `EndpointDiscoveryUnset`? Then we could distinguish the default from `auto`, `enabled`, `disabled` without having to use a pointer value.

service/internal/endpoint-discovery/middleware.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/smithy-go/logging"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+
+	"github.com/aws/smithy-go/middleware"
```
skmcgrail, line 12: The import grouping here is inconsistent.

config/env_config.go
skmcgrail: Ah, I failed to notice we exposed this setting on env_config despite not supported the feature at launch. That's unfortunate. Though this is still a breaking change regardless change from `*bool`.

jasdel: Yeah :/ this was missed before GA, since the parameter did nothing and never could be used we should consider doing a minor version bump to change this from an boolean to enum.

config/types.go
```diff
@@ -0,0 +1,22 @@
...
+// optionally use Endpoint Discovery will not use Endpoint Discovery
+// unless EndpointDiscovery is explicitly enabled.
+type EndpointDiscoveryEnableState uint
+
+// Enumeration values for EndpointDiscoveryEnableState
+const (
+	EndpointDiscoveryAuto     EndpointDiscoveryEnableState = iota // default state
+	EndpointDiscoveryDisabled                                     // endpoint discovery disabled
+	EndpointDiscoveryEnabled                                      // endpoint discovery enabled
+)
```
jasdel, line 17: Using these as string enums instead of int enums will simplify its overall usage and remove the need for the `ptr`. since the zero value of a string is `""` which is never a valid set value.

config/env_config.go
```diff
@@ -339,3 +339,30 @@ func setBoolPtrFromEnvVal(dst **bool, keys []string) error {
...
+			*dst = new(EndpointDiscoveryEnableState)
+		}
+
+		switch {
+		case strings.EqualFold(value, "false"):
+			**dst = EndpointDiscoveryDisabled
+		case strings.EqualFold(value, "true"):
+			**dst = EndpointDiscoveryEnabled
+		case strings.EqualFold(value, "auto"):
+			**dst = EndpointDiscoveryAuto
```
jasdel, line 360: These comparisons should be against the constant values of `false`, `true`, `auto`.

service/internal/endpoint-discovery/middleware.go
```diff
@@ -0,0 +1,101 @@
...
+	// Discover Endpoint operation.
+	Options []func(*DiscoverEndpointOptions)
+
+	// DiscoverOperation represents the endpoint discovery operation that
+	// returns an Endpoint or error.
+	DiscoverOperation func(ctx context.Context, key string, required bool, options ...func(*DiscoverEndpointOptions)) (WeightedAddress, error)
+
+	// EndpointDiscoveryEnableState represents the customer configuration for endpoint
+	// discovery feature.
+	EndpointDiscoveryEnableState config.EndpointDiscoveryEnableState
```
jasdel, line 45: `EndpointDiscoveryEnableState` should be moved to `aws` since nothing in the SDK may depend on the `config` module/pkg.

service/internal/endpoint-discovery/go.mod
```diff
@@ -0,0 +1,12 @@
...
+go 1.15
+
+require (
+	github.com/aws/aws-sdk-go-v2/config v1.1.7
+	github.com/aws/smithy-go v1.4.0
+)
+
+replace (
+	github.com/aws/aws-sdk-go-v2/config => ./../../../config
+)
```
jasdel, line 12: This runs the risk of creating circular dependencies.  For example if STS were to ever add endpoint discovery this would break.

`config` -> `credentials/stscreds` -> `service/sts` -> `service/internal/endpoint-discovery` -> `config`

Nothing in the SDK may depend on `config` module.

config/types.go
jasdel:
```go
type EndpointDiscoveryEnableState string

// Enumeration values for EndpointDiscoveryEnableState
const (
	EndpointDiscoveryAuto      =  EndpointDiscoveryEnableState("auto") // default state
	EndpointDiscoveryDisabled  =  EndpointDiscoveryEnableState("false") // endpoint discovery disabled
	EndpointDiscoveryEnabled   =  EndpointDiscoveryEnableState("true")  // endpoint discovery enabled
)
```

service/timestreamwrite/api_client.go
```diff
@@ -4,16 +4,21 @@ package timestreamwrite
...
 import (
 	"context"
+	"fmt"
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/aws/signer/v4"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
+	"github.com/aws/aws-sdk-go-v2/config"
+	discovery "github.com/aws/aws-sdk-go-v2/service/internal/endpoint-discovery"
```
jasdel, line 14: Cannot depend on `config` because risk of circular dependencies.

service/timestreamwrite/api_client.go
```diff
@@ -24,6 +29,8 @@ const ServiceAPIVersion = "2018-11-01"
 // Write.
 type Client struct {
 	options Options
+
+	endpointCache *discovery.EndpointCache
 }
```
jasdel, line 34: it doesn't look like there is any way to configure endpoint discovery from the `aws.Config` value. I'm not seeing where the `LoadDefaultConfig` getting the endpoint discovery flag from the env or shared config.


The LoadDefaultConfig probably needs to be updated to populate a `EndpointDiscovery` field on the `aws.Config` structure. And define the enum in the `aws` or similar package.

service/timestreamwrite/api_client.go
```diff
@@ -24,6 +29,8 @@ const ServiceAPIVersion = "2018-11-01"
 // Write.
 type Client struct {
 	options Options
+
+	endpointCache *discovery.EndpointCache
```
jasdel, line 33: We should consider making this `endpointCache` member an interface type not hard dependency on `discovery.EndpointCache` implementation.  A Default resolver could provide the value similar to HTTP client but this probably should be an interface for similar reasons as HTTP client is interface.

service/timestreamwrite/api_client.go
```diff
@@ -87,6 +97,16 @@ type Options struct {
 	// The HTTP client to invoke API calls with. Defaults to client's default HTTP
 	// implementation if nil.
 	HTTPClient HTTPClient
+
+	// The cache used to store discoevred endpoint. Defaults to 10.
+	CacheSize int64
```
jasdel, line 102: Nit this should be `EndpointDiscoveryCacheSize` to be more specific to its purpose.

service/timestreamwrite/api_client.go
```diff
@@ -87,6 +97,16 @@ type Options struct {
...
+
+	// EndpointUsedForDiscovery is the endpoint used internally by the operation
+	// with the endpoint discovery api request call.
+	EndpointUsedForDiscovery string
+
+	// EnableEndpointDiscovery indicates if endpoint discovery is enabled, auto or disabled.
+	EnableEndpointDiscovery config.EndpointDiscoveryEnableState
```
jasdel, line 109: We should consider making these a single interface member value that is auto populated by the client's constructor if unset by the user. 

This would allow the endpoint discovery interface to be mocked out or replaced similar to `HTTPSigner` and `HTTPClient`.

jasdel (changes requested):
Overall looks good, biggest feedback is dependencies on `config` module/package. Nothing other than leaf testing modules may depend on the `config` module/pkg, or it creates risks of circular dependencies.

Moving the enum to the `aws` or similar package will help  resolve this issue. `config`, service/internal/endpoint_discovery` and `service/<api>` packages would all depend on the common `aws` definition of the enum.  Could consider putting endpoint discover into its own package under `aws`.

config/env_config.go
skotambkar: I don't have a strong opinion on having `EndpointDiscoveryUnset` or not. I do feel having Unset and Auto as values will confuse customers as they are effectively the same in behavior.  

What do we gain by replacing the pointer value? Or is it an idiomatic change?

skotambkar: yes breaking but its non-functional  today so seems fine.

service/timestreamwrite/api_client.go
```diff
@@ -241,6 +261,84 @@ func addRetryMiddlewares(stack *middleware.Stack, o Options) error {
...
+	}
+	if found {
+		o.EnableEndpointDiscovery = value
+	}
+	return nil
+}
+
+// resolves endpoint cache on client
+func resolveEndpointCache(c *Client) {
+	c.endpointCache = internalEndpointDiscovery.NewEndpointCache(internalEndpointDiscovery.DefaultEndpointCacheSize)
```
skmcgrail: To my earlier comment, I think it might make sense to just generate the default value inline into this call. Not sure if it makes sense for the endpointdiscovery package to export a constant for a default that it doesn't directly reference itself. Thoughts?

service/timestreamwrite/api_client.go
```diff
@@ -241,6 +261,84 @@ func addRetryMiddlewares(stack *middleware.Stack, o Options) error {
...
+
+		cachedInMinutes := e.CachePeriodInMinutes
+		u, err := url.Parse(address)
+		if err != nil {
+			continue
+		}
+
+		addr := internalEndpointDiscovery.WeightedAddress{
+			URL:     u,
+			Expired: time.Now().Add(time.Duration(cachedInMinutes) * time.Minute),
```
skmcgrail: This probably a spot where we want to strip monotonic clock readings off to make this safe if an application is on a system that is hibernated and then resumed. By calling `t.Round(0)`

config/env_config.go
```diff
@@ -339,3 +339,31 @@ func setBoolPtrFromEnvVal(dst **bool, keys []string) error {
 
 	return nil
 }
+
+func setEndpointDiscoveryTypeFromEnvVal(dst *aws.EndpointDiscoveryEnableState, keys []string) error {
+	for _, k := range keys {
+		value := os.Getenv(k)
+		if len(value) == 0 {
+			continue // skip if empty
```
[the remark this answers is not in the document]
skotambkar: We should only fall back to "auto" if no other configuration sets the `EnableEndpointDiscovery` setting.

 If we interpret empty string as "auto" here, we would be ignoring configuration set in `SharedConfig` file.

service/internal/endpoint-discovery/cache.go
```diff
@@ -0,0 +1,98 @@
+package endpointdiscovery
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// DefaultCacheSize is the default endpoint discovery cache size that should be
+// used to construct endpoint cache, if not provided by customer
+const DefaultEndpointCacheSize int64 = 10
```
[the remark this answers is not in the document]
skotambkar: This is referred by the client to build a new cache. 

Not much opinion on this, as it is hard-coded any way. Keeping it here let's us keep value in go code, instead of some constant in java code

service/internal/endpoint-discovery/middleware.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/aws/smithy-go/logging"
+	"github.com/aws/smithy-go/middleware"
+)
+
+// invalidConfigErrorCode denotes type of config error
+const invalidConfigErrorCode = "Invalid configuration"
+
+type DiscoveryInterface interface {
+	discoverEndpoint()
+}
```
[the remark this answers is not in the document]
skotambkar: not used - will remove

service/internal/endpoint-discovery/middleware.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+
+	"github.com/aws/smithy-go/logging"
+	"github.com/aws/smithy-go/middleware"
+)
+
+// invalidConfigErrorCode denotes type of config error
+const invalidConfigErrorCode = "Invalid configuration"
```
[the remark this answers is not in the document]
skotambkar: not used - will remove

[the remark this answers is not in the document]
skotambkar: Used as a constant string in an error below. Will remove and use string directly

aws/config.go
```diff
@@ -86,3 +86,19 @@ func (c Config) Copy() Config {
...
+// optionally use Endpoint Discovery will not use Endpoint Discovery
+// unless EndpointDiscovery is explicitly enabled.
+type EndpointDiscoveryEnableState string
+
+// Enumeration values for EndpointDiscoveryEnableState
+const (
+	EndpointDiscoveryAuto     = EndpointDiscoveryEnableState("auto")  // default state
+	EndpointDiscoveryDisabled = EndpointDiscoveryEnableState("false") // endpoint discovery disabled
+	EndpointDiscoveryEnabled  = EndpointDiscoveryEnableState("true")  // endpoint discovery enabled
+)
```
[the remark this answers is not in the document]
skotambkar: I am fine with using ``string`` or ``uint``. I followed [feedback as per comment](https://github.com/aws/aws-sdk-go-v2/pull/1263#discussion_r634747199) which was suggested to avoid pointer value representation. 

With ``uint`` we would need the extra Unset enum, and have a minor optimization of using less bytes for ``uint`` vs ``string``.

service/timestreamwrite/api_client.go
```diff
@@ -65,12 +77,19 @@ type Options struct {
...
+
 	// The endpoint options to be used when attempting to resolve an endpoint.
 	EndpointOptions EndpointResolverOptions
 
 	// The service endpoint resolver.
 	EndpointResolver EndpointResolver
 
+	// Allows configuring an endpoint used when attempting request to the endpoint
+	// discovery service
+	EndpointUsedForDiscovery string
```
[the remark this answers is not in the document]
skotambkar: I like the idea but we do not use case around multi-region support. Can discuss offline.

service/timestreamwrite/api_client.go
skotambkar: I see - So if we want to change the default endpoint cache size we won't need to bump the endpoint-discovery module. 

My reasoning was to keep it as Go constant, instead of constant in java codegen.  Agree with your idea to avoid bumping endpoint-discovery module.

service/internal/endpoint-discovery/cache.go
```diff
@@ -0,0 +1,98 @@
...
+// Add is a concurrent safe operation that will allow new endpoints to be added
+// to the cache. If the cache is full, the number of endpoints equal endpointLimit,
+// then this will remove the oldest entry before adding the new endpoint.
+func (c *EndpointCache) Add(endpoint Endpoint) {
+	// de-dups multiple adds of an endpoint with a pre-existing key
+	if iface, ok := c.endpoints.Load(endpoint.Key); ok {
+		e := iface.(Endpoint)
+		if e.Len() > 0 {
+			return
+		}
```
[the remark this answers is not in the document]
skotambkar: Ported directly from V1, I think new endpoints are dropped if old are still valid to keep the implementation concurrent safe. Also this edge case should never actually trigger.

If endpoints are still valid for the endpoint key, we would never fetch newer endpoint for the same key.  Don't think we need to over optimize in this case.

aws/config.go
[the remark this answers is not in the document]
skotambkar: Thinking again,  using ``uint`` would need us to export both ``Unset`` and ``Auto`` as configuration values. However both have the same behavior. 

Keeping this string but open to debate/thoughts.

service/internal/endpoint-discovery/cache.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	c.endpoints.Store(endpoint.Key, endpoint)
+
+	size := atomic.AddInt64(&c.size, 1)
+	if size > 0 && size > c.endpointLimit {
+		c.deleteRandomKey()
+	}
```
skmcgrail: Should we delete a random key first if it were to overflow and then add in the new endpoint? Otherwise there is a risk of removing this endpoint correct? Really the whole random key deletion thing is not the ideal way to reduce the cache size.

skmcgrail: Though arguably there is always a chance that something else will delete our key anyways since deleteRandomKey can happen concurrently.

service/dynamodb/api_client.go
```diff
@@ -78,6 +90,9 @@ type Options struct {
 	// Disabled by default.
 	EnableAcceptEncodingGzip bool
 
+	// Allows configuring endpoint discovery
+	EnableEndpointDiscovery aws.EndpointDiscoveryEnableState
```
jasdel: how is the endpoint for endpoint discovery option specified? looks like that option is only available for APIs that require endpoint discovery. Does it make sense to always have the property?

service/timestreamquery/api_client.go
```diff
@@ -69,12 +81,19 @@ type Options struct {
...
 	// The endpoint options to be used when attempting to resolve an endpoint.
 	EndpointOptions EndpointResolverOptions
 
 	// The service endpoint resolver.
 	EndpointResolver EndpointResolver
 
+	// Allows configuring an endpoint resolver to use when attempting an endpoint
+	// discovery api request.
+	EndpointResolverUsedForDiscovery EndpointResolver
+
```
jasdel: are these only added for APIs where endpoint discovery is required?   I could seem the being useful where endpoint discovery is optional for same reason as APIs where its required.

service/dynamodb/api_client.go
skotambkar: We maintain a list of service ids in smithy codegen for which this option is available. This is not generated as per required/optional discovery. Only `timestream` services have the option for now. 

Adding services to the list is easy. I would like to keep it restricted until we have a feature request or requirement around it.

service/timestreamquery/api_client.go
skotambkar: answered in the comment above.

service/internal/endpoint-discovery/cache.go
skotambkar: As discussed offline, not fixing this since it is an edge case, and over optimization for randomness.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/EndpointDiscoveryGenerator.java
```diff
@@ -498,18 +537,18 @@ public void processFinalizedModel(GoSettings settings, Model model) {
...
+//                RuntimeClientPlugin.builder()
+//                        .servicePredicate(EndpointDiscoveryGenerator::serviceSupportsCustomDiscoveryEndpoint)
+//                        .addConfigField(ConfigField.builder()
+//                                .name(ENDPOINT_RESOLVER_USED_FOR_DISCOVERY)
+//                                .type(ENDPOINT_RESOLVER_INTERFACE_NAME)
+//                                .documentation("Allows configuring an endpoint resolver to use when attempting an"
+//                                        + " endpoint discovery api request.")
+//                                .build())
+//                        .build()
+//        );
```
jasdel, line 551: still needed?

jasdel (approved):

PR #1274, opened 2021-05-25
FIPS & DualStack Endpoint Resolver Support

skmcgrail:
Support has been added for an updated `endpoints.json` model which contains modeled endpoint data for service dual-stack and FIPS endpoints.

Additionally, this support removes the need for the dual-stack customization in place for S3 and S3 Control service clients, and simplifies FIPS logic.

Requires: https://github.com/aws/smithy-go/pull/303

### DIFF
--- a/config/shared_config.go
+++ b/config/shared_config.go
@@ -988,12 +988,8 @@ func (c *SharedConfig) setFromIniSection(profile string, section ini.Section) er
 	}
 	updateString(&c.EC2IMDSEndpoint, section, ec2MetadataServiceEndpointKey)
 
-	if err := updateUseDualStackEndpoint(&c.UseDualStackEndpoint, section, useDualStackEndpoint); err != nil {
-		return fmt.Errorf("failed to load %s from shared config, %v", useDualStackEndpoint, err)
-	}
-	if err := updateUseFIPSEndpoint(&c.UseFIPSEndpoint, section, useFIPSEndpointKey); err != nil {
-		return fmt.Errorf("failed to load %s from shared config, %v", useFIPSEndpointKey, err)
-	}
+	updateUseDualStackEndpoint(&c.UseDualStackEndpoint, section, useDualStackEndpoint)
+	updateUseFIPSEndpoint(&c.UseFIPSEndpoint, section, useFIPSEndpointKey)
 
 	// Shared Credentials
 	creds := aws.Credentials{
@@ -1283,14 +1279,9 @@ func updateEndpointDiscoveryType(dst *aws.EndpointDiscoveryEnableState, section 
 
 // updateEndpointDiscoveryType will only update the dst with the value in the section, if
 // a valid key and corresponding EndpointDiscoveryType is found.
-func updateUseDualStackEndpoint(dst *aws.DualStackEndpointState, section ini.Section, key string) error {
+func updateUseDualStackEndpoint(dst *aws.DualStackEndpointState, section ini.Section, key string) {
 	if !section.Has(key) {
-		return nil
-	}
-
-	valueType, _ := section.ValueType(key)
-	if valueType != ini.BoolType {
-		return fmt.Errorf("expected true or false, got %v", section.String(key))
+		return
 	}
 
 	if section.Bool(key) {
@@ -1299,19 +1290,14 @@ func updateUseDualStackEndpoint(dst *aws.DualStackEndpointState, section ini.Sec
 		*dst = aws.DualStackEndpointStateDisabled
 	}
 
-	return nil
+	return
 }
 
 // updateEndpointDiscoveryType will only update the dst with the value in the section, if
 // a valid key and corresponding EndpointDiscoveryType is found.
-func updateUseFIPSEndpoint(dst *aws.FIPSEndpointState, section ini.Section, key string) error {
+func updateUseFIPSEndpoint(dst *aws.FIPSEndpointState, section ini.Section, key string) {
 	if !section.Has(key) {
-		return nil
-	}
-
-	valueType, _ := section.ValueType(key)
-	if valueType != ini.BoolType {
-		return fmt.Errorf("expected true or false, got %v", section.String(key))
+		return
 	}
 
 	if section.Bool(key) {
@@ -1320,5 +1306,5 @@ func updateUseFIPSEndpoint(dst *aws.FIPSEndpointState, section ini.Section, key 
 		*dst = aws.FIPSEndpointStateDisabled
 	}
 
-	return nil
+	return
 }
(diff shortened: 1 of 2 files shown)


jasdel:
Pending updated models.

internal/endpoints/endpoints_test.go
```diff
@@ -1,6 +1,7 @@
 package endpoints
 
 import (
+	"github.com/google/go-cmp/cmp"
```
skotambkar, line 4: Nit: import order

internal/endpoints/endpoints.go
```diff
@@ -47,12 +51,14 @@ func (ps Partitions) ResolveEndpoint(region string, opts Options) (aws.Endpoint,
...
-	Defaults          Endpoint
-	Endpoints         Endpoints
+	ID                 string
+	RegionRegex        *regexp.Regexp
+	PartitionEndpoint  string
+	IsRegionalized     bool
+	Defaults           Endpoint
+	DualStackDefaults  Endpoint
+	Endpoints          Endpoints
+	DualStackEndpoints Endpoints
```
skotambkar, line 61: Just curious if we could have updated the `Endpoint` to include an `isDualStack` identifier? 

That way, we can continue using `Defaults` and `Endpoints` member, and not need the `DualStackDefaults` and `DualStackEndpoints` member. 

Might help in future to avoid duplicate members in case another property similar to DualStack is added for an endpoint.

config/load_options.go
```diff
@@ -122,6 +122,10 @@ type LoadOptions struct {
 	// S3UseARNRegion specifies if the S3 service should allow ARNs to direct
 	// the region, the client's requests are sent to.
 	S3UseARNRegion *bool
+
+	// Specifies that SDK clients must resolve a dual-stack endpoint for
+	// services.
+	UseDualStack *bool
```
skotambkar, line 128: why `*bool` over an Enum?

internal/configsources/assert/config_test.go
```diff
@@ -0,0 +1,12 @@
+package assert
+
+import (
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/internal/configsources"
+)
+
+// UseARNRegionProvider Assertions
```
skotambkar, line 8: nit: stale comment

service/s3/internal/customizations/process_arn_resource.go
```diff
@@ -250,8 +251,15 @@ func buildAccessPointRequest(ctx context.Context, options accesspointOptions) (c
 
 	resolveService := tv.Service
 
+	eo := options.EndpointResolverOptions
+	if options.UseDualstack {
+		eo.UseDualStack = internalendpoints.EnableDualStackEndpoint
+	} else {
+		eo.UseDualStack = internalendpoints.DisableDualStackEndpoint
```
skotambkar, line 258: why disable instead of unset? what's the difference in unset vs disable?

internal/endpoints/endpoints.go
skmcgrail: This is constraint of the modeling choices for `endpoints.json` where the pseudo-region key in the `map[string]Endpoint` can overlap between dual-stack and non dual-stack.

config/load_options.go
skmcgrail: Mainly due to the fact that while we could make this an enum, the clients wouldn't be able to make depend on the declared constants in this package. Alternative would be to declare this an a string enum constant type, and have the resolver interface expose this as just a string value. This would require the logic of interpreting the string values to also be generated into the clients which is one of the downsides.

internal/endpoints/endpoints.go
```diff
@@ -66,18 +72,25 @@ func (p Partition) ResolveEndpoint(region string, options Options) (resolved aws
...
-func (p Partition) endpointForRegion(region string) (Endpoint, bool) {
-	if e, ok := p.Endpoints[region]; ok {
+func (p Partition) endpointForRegion(region string, endpoints Endpoints) (Endpoint, bool) {
+	if e, ok := endpoints[region]; ok {
 		return e, true
 	}
 
 	if !p.IsRegionalized {
-		return p.Endpoints[p.PartitionEndpoint], region == p.PartitionEndpoint
+		return endpoints[p.PartitionEndpoint], region == p.PartitionEndpoint
```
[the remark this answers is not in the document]
skmcgrail: I believe this is some legacy cruft from when the code was ported over from V1, and to my knowledge this boolean return is not used anymore in the V2 implementation. Once upon a time it was used to support the `StrictMatching` option that we opted to no longer support.

service/s3/internal/customizations/process_arn_resource.go
skmcgrail: `Unset` effectively maps to disabled, but `Unset` serves the purpose to allow us to see if the option was explicitly distinguish whether the option was set by the user rather then using a pointer.

skmcgrail:
Requires https://github.com/aws/smithy-go/pull/303

skmcgrail:
Changes that were included in the last force push:
* `UseDualStack` option on `EndpointResolverOptions` has been renamed to `DualStackEndpoint`
* Constants have been renamed to `DualStackEndpointUnset`, `DualStackEndpointEnabled`, `DualStackEndpointDisabled`.
* The config module refers to the setting a `UseDualStackEndpoint` to match the shared configuration name and better match  with the endpoint options configuration name.

skmcgrail:
Need to ensure the GetDNSSuffix functions support the dualstack option.

skmcgrail:
Updated PR to have S3 DNS Suffix functions for endpoints to support returning dualstack suffixs

config/env_config.go
```diff
@@ -363,6 +385,27 @@ func setEndpointDiscoveryTypeFromEnvVal(dst *aws.EndpointDiscoveryEnableState, k
...
+		}
+
+		switch {
+		case strings.EqualFold(value, "true"):
+			*dst = aws.DualStackEndpointEnabled
+		case strings.EqualFold(value, "false"):
+			*dst = aws.DualStackEndpointDisabled
+		default:
+			return fmt.Errorf(
+				"invalid value for environment variable, %s=%s, need true, false or auto",
```
skotambkar, line 402: Doesn't seem like we support `auto`, but error suggests we do

config/env_config.go
```diff
@@ -180,6 +182,12 @@ type EnvConfig struct {
 	//
 	// AWS_S3_USE_ARN_REGION=true
 	S3UseARNRegion *bool
+
+	// Specifies that SDK clients must resolve a dual-stack endpoint for
+	// services.
+	//
+	// AWS_USE_DUALSTACK_ENDPOINT=true
+	UseDualStackEndpoint aws.DualStackEndpoint
```
skotambkar, line 190: Had similar comment else where. 

IMO `Use + <Option>` seems boolean specific which this is. But since we use enums - may be we should name it agnostic. I think  in general `<option> + State` would be a better pattern to follow. 

`DualStackEndpointState` would be a better name, which can then be set as enabled/disabled or anything else.

config/env_config.go
skmcgrail: I think this was a copy/paste error from another error string. Will fix.

skotambkar (approved):

skotambkar (approved):

PR #1276, opened 2021-05-25
API module updates

jasdel:
Updates the SDK API clients to the latest API version.

* Deprecates `service/commander` module. This API client was incorrect named and is replaced with the AWS Systems Manager Incident Manager `service/ssmincidents` module.


skmcgrail (approved):

skmcgrail (approved):

codegen/sdk-codegen/aws-models/route53.2013-04-01.json
```diff
@@ -3104,23 +3104,6 @@
                     "method": "GET",
                     "uri": "/2013-04-01/change/{Id}",
                     "code": 200
-                },
```
skotambkar, line 3107: This should not be updated.

skotambkar (approved):

PR #1278, opened 2021-05-26
Update S3 ARN Handling for AccessPoints and Outpost ARNs

skmcgrail:
Update S3 ARN Handling for AccessPoints and Outpost ARNs.

### DIFF
--- a/service/s3/internal/customizations/update_endpoint_test.go
+++ b/service/s3/internal/customizations/update_endpoint_test.go
@@ -442,31 +442,28 @@ func TestEndpointWithARN(t *testing.T) {
 			expectedSigningName:   "s3-outposts",
 			expectedSigningRegion: "us-gov-east-1",
 		},
-		"Outpost AccessPoint Fips region": {
+		"Outpost AccessPoint FIPS cross-region": {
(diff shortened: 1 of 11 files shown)


service/s3/internal/customizations/update_endpoint_test.go
```diff
@@ -442,31 +442,28 @@ func TestEndpointWithARN(t *testing.T) {
...
+			bucket: "arn:aws-us-gov:s3-outposts:us-gov-east-1:123456789012:outpost:op-01234567890123456:accesspoint:myaccesspoint",
 			options: s3.Options{
-				Region:          "fips-us-gov-west-1",
-				EndpointOptions: endpoints.Options{DisableHTTPS: true},
-				UseARNRegion:    true,
+				Region:       "fips-us-gov-west-1",
+				UseARNRegion: true,
 			},
-			expectedErr: "InvalidARNError : resource ARN not supported for FIPS region",
+			expectedErr: "use of ARN is not supported when client or request is configured for FIPS",
```
skotambkar, line 458: These two tests seem similar enough but the error returned is different.

skmcgrail: This is because the check to see if `UseARNRegion` is enabled/disabled is before the FIPS check for outpost ARNs

skotambkar (approved):

PR #1283, opened 2021-06-03
awsQueryError trait support

skmcgrail:
Adds support for [awsQueryError](https://awslabs.github.io/smithy/1.0/spec/aws/aws-query-protocol.html#aws-protocols-awsqueryerror-trait) trait.

Requires https://github.com/aws/smithy-go/pull/304

### DIFF
--- a/internal/protocoltest/query/api_op_GreetingWithErrors_test.go
+++ b/internal/protocoltest/query/api_op_GreetingWithErrors_test.go
@@ -207,7 +207,6 @@ func TestClient_GreetingWithErrors_InvalidGreeting_awsAwsqueryDeserialize(t *tes
 }
 
 func TestClient_GreetingWithErrors_CustomCodeError_awsAwsqueryDeserialize(t *testing.T) {
-	t.Skip("skipping for release-2021-05-20")
 	cases := map[string]struct {
 		StatusCode    int
 		Header        http.Header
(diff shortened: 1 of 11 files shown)


skmcgrail:
FYI the order of a handful of switch cases have changed in the regeneration, these are expected since this moved from a TreeSet on the ShapeId to a TreeMap where the stable sorting is on the key string now.

mtdowling (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsQuery.java
```diff
@@ -152,4 +155,23 @@ protected void writeErrorMessageCodeDeserializer(GenerationContext context) {
...
+
+            Optional<AwsQueryErrorTrait> errorShapeTrait = errorShape.getTrait(AwsQueryErrorTrait.class);
+            if (errorShapeTrait.isPresent()) {
+                errors.put(errorShapeTrait.get().getCode(), shapeId);
+            } else {
+                errors.put(errorName, shapeId);
+            }
+        });
+
+        return errors;
```
skotambkar, line 175: Should this map key, value pair be reversed?  

I would have expected shapeId to be the key, while errorName or code is the value. Is errorName or errorCode guaranteed to be unique for each shape?

skmcgrail: Error name has to be unique cause it is the shapeId name by default. For query alias’s and whether those conflict that would be a for whether smithy validates that constraint.

skotambkar (approved):

PR #1285, opened 2021-06-04
internal/repotools: Tombstone Module Support

skmcgrail:
Support proper detection of modules that have been "tombstoned" (once released, but source code no longer present in the repository), and mark their paths as ignored when calculating module changes. This avoids unnecessary situation which would cause the parent module version to be bumped due to the removal of Go source code.

### DIFF
--- a/internal/repotools/release/calculate.go
+++ b/internal/repotools/release/calculate.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"path"
 	"path/filepath"
+	"sort"
 )
 
 // ModuleFinder is a type that
@@ -36,6 +37,13 @@ func Calculate(finder ModuleFinder, tags git.ModuleTags, config repotools.Config
 	}
 
 	modules := make(map[string]*Module)
+	var repositoryModuleTombstonePaths []string
+
+	for moduleDir := range tags {
+		if _, ok := repositoryModules[moduleDir]; !ok {
+			repositoryModuleTombstonePaths = append(repositoryModuleTombstonePaths, moduleDir)
+		}
+	}
 
 	for moduleDir := range repositoryModules {
 		moduleFile, err := gomod.LoadModuleFile(filepath.Join(rootDir, moduleDir), nil, true)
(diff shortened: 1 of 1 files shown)


internal/repotools/release/calculate.go
```diff
@@ -63,14 +71,25 @@ func Calculate(finder ModuleFinder, tags git.ModuleTags, config repotools.Config
...
+			subModulePaths := repositoryModules[moduleDir]
+
+			ignoredModulePaths := make([]string, 0, len(subModulePaths)+len(repositoryModuleTombstonePaths))
+			ignoredModulePaths = append(ignoredModulePaths, subModulePaths...)
+
+			if len(repositoryModuleTombstonePaths) > 0 {
+				ignoredModulePaths = append(ignoredModulePaths, repositoryModuleTombstonePaths...)
+				// IsModuleChanged expects the provided list of ignored modules paths to be sorted
+				sort.Strings(ignoredModulePaths)
+			}
```
skotambkar, line 83: are subModulePaths already sorted?

internal/repotools/release/calculate.go
```diff
@@ -63,14 +71,25 @@ func Calculate(finder ModuleFinder, tags git.ModuleTags, config repotools.Config
 				log.Fatalf("failed to get git changes: %v", err)
 			}
 
-			hasChanges, err = gomod.IsModuleChanged(moduleDir, repositoryModules[moduleDir], changes)
+			subModulePaths := repositoryModules[moduleDir]
```
skotambkar, line 74: Can you give an example of what can be considered a subModulePath?

skotambkar (approved):

internal/repotools/release/calculate.go
skmcgrail: Yes

internal/repotools/release/calculate.go
skmcgrail: For example when you look up the root module, all other modules in the repository like each service client path etc is listed in this slice so that it is ignored when looking at changes.

PR #1291, opened 2021-06-07
internal/ini: Add fuzzing using experimental Go1.18 fuzzer

jasdel:
Experiment adding fuzzing to the `internal/ini` module using the Go 1.18 beta fuzzer feature.


internal/ini/fuzz_test.go
```diff
@@ -1,29 +1,57 @@
...
-// TestFuzz is used to test for crashes and not validity of the input
-func TestFuzz(t *testing.T) {
-	paths, err := filepath.Glob("testdata/fuzz/*")
+func FuzzParse(f *testing.F) {
+	corpus, err := loadCorpus(
+		// Also loads "testdata/corpus/FuzzParse" automatically,
+		// for previous known panics if present and contains any files
+		filepath.Join("testdata", "valid", "*"),
+		filepath.Join("testdata", "invalid", "*"),
+	)
```
skotambkar, line 19: How are these files used? Do we need to prefill cases which are run randomly?

jasdel: This reuses the existing tests that the rest of the ini parser's unit tests use.Theese are used to "seed" the corpus. It seems better to reuse these existing tests instead of duplicating them into the `testdata/corpus/FuzzParse`. folder.

PR #1307, opened 2021-06-18
aws/signer/v4: Keep Object-Lock headers a header in presigning signing requests

jasdel:
Updates the Sigv4 request signer to use keep `X-Amz-Object-Lock-*` headers as headers, and not hoisted to the query string for presigned URLs.

Revision of #1215

### DIFF
--- a/.changelog/4b776031a69541caa111950fd7dbe4fa.json
+++ b/.changelog/4b776031a69541caa111950fd7dbe4fa.json
@@ -1,0 +1,8 @@
+{
+    "id": "4b776031-a695-41ca-a111-950fd7dbe4fa",
+    "type": "bugfix",
+    "description": "Keep Object-Lock headers a header when presigning Sigv4 signing requests",
+    "modules": [
+        "."
+    ]
+}


skmcgrail (approved):

aws/signer/internal/v4/headers.go
```diff
@@ -56,10 +56,11 @@ var RequiredSignedHeaders = Rules{
...
 	},
+	Patterns{"X-Amz-Object-Lock-"},
 	Patterns{"X-Amz-Meta-"},
 }
 
-// AllowedQueryHoisting is a whitelist for Build query headers. The boolean value
+// AllowedQueryHoisting is a allowed list for Build query headers. The boolean value
 // represents whether or not it is a pattern.
 var AllowedQueryHoisting = InclusiveRules{
 	Blacklist{RequiredSignedHeaders},
```
skotambkar, line 66: can we switch to using `denyList `? Same comment for other occurrences in this file.

skotambkar (approved):

skotambkar (approved):

PR #1316, opened 2021-06-25
Regenerated Documentation

skotambkar:
Regenerated documentation


skmcgrail:
Did you regenerate this using `builddocs.sh`? Seems like you used a local version of Hugo for the generation and thus an older version was used.

skmcgrail (changes requested):
See earlier comment

skotambkar:
I didn't reinstall hugo ..

skmcgrail (approved):

PR #1320, opened 2021-06-28
Smithy Document Shape Support

skmcgrail:
Adds support for Smithy Document Shape Types for JSON protocols.

Depends on: https://github.com/aws/smithy-go/pull/310


feature/dynamodb/attributevalue/empty_collections_test.go
```diff
@@ -913,7 +915,7 @@ func TestEmptyCollectionsSpecialCases(t *testing.T) {
 	if err != nil {
 		t.Fatalf("expected no err got %v", err)
 	}
-	if diff := cmp.Diff(expectedEncode, actualEncode); len(diff) != 0 {
+	if diff := cmp.Diff(expectedEncode, actualEncode, cmpopts.IgnoreTypes(smithydocument.NoSerde{})); len(diff) != 0 {
```
jasdel, line 918: Will having the embedded `smithydocument.NoSerde{}` in the SDK types cause similar issues with customer applicaiton's doing compares on SDK generated types?

jasdel (commented):
We should consider the impact to downstream applications that might have behavior using utilities like `gocmp` that caused us to add `cmpopts.IgnoreTypes(smithydocument.NoSerde{})` to many places.

feature/dynamodb/attributevalue/empty_collections_test.go
skmcgrail: Discussed offline, it was determined that this was likely the best course of action compared to other alternatives.

modman.toml
```diff
@@ -1,5 +1,5 @@
 [dependencies]
-"github.com/aws/smithy-go" = "v1.5.1-0.20210713172319-36596d2cf389"
+"github.com/aws/smithy-go" = "v1.5.1-0.20210713214849-a87ef15659d4"
```
jasdel, line 2: This is probably out of date now.

jasdel (approved):

skmcgrail:
Rebased on `main`

JordonPhillips (approved):

PR #1321, opened 2021-06-28
Add Dependabot CI integration

jasdel:
Adds CI integration with dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: /
```
skotambkar, line 4: Wondering if `/` needs quotes in yml?

skotambkar (approved):

.github/dependabot.yml
jasdel: I think you're right, fixed that.

skmcgrail (approved):

skotambkar (approved):

skmcgrail (commented):
Does this work with our automation tooling? Currently `modman.toml` is the authoritative source for what versions modules need to be at for third party dependencies. How do we integrate the two, or should we integrate the two? What value will this provide outside of an alert (since we still need to do a release).

Also it's not clear from the documents if this enumerates all modules in the repository either.

jasdel:
I think `dependabot`'s relationship with `modman` will be more informative alerting us to dependencies outside of the SDK that have been updated, and prompting us that we can update the dependencies. 

One way to prevent tracking every upstream release is to specify the [versioning-strategy](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#versioning-strategy) property similar to the AWS SDK for Javascript v3 does, https://github.com/aws/aws-sdk-js-v3/blob/main/.github/dependabot.yml#L9  This will reduce the excessive noise of each version update that's not a security related issue.

jasdel:
Need more research on this tool to verify that it works with the SDK's multi Go module structure. Along with investigating how it can integrate with the `modman`'s tooling behavior for code generation and SDK release tooling since that tooling is used to generate code generated module version dependencies

jasdel:
Closing this PR since the SDK won't be able to use Dependable CI's PR directly due to how the SDK manages dependencies for code generation. The SDK will still receive notifications when dependencies have issues.

PR #1324, opened 2021-06-30
Support Op and Sep Tokens in Expression Values

skmcgrail:
Allows for `:`, `=`, `[`, and `]` to be present in expression values.

Related: https://github.com/aws/aws-sdk-go/issues/3958


internal/ini/testdata/valid/op_sep_in_values
```diff
@@ -0,0 +1,6 @@
+[default]
+sepInValue = =:[foo]]bar[
+output = json
+[assumerole]
+sepInValue==:[foo]]bar[
+output = json
```
skotambkar, line 6: The difference in values for `sepInValue` is too subtle to notice :) . Can we add cases with diverse set of values?

skotambkar: Could you add more valid/invalid cases like

1.
```
[assumerole]
sepInValue = []
```

2. 
```
[assumerole]
sepInValue = [value] x=a
```
3. 
```
[assumerole]
key[id] = value
```

4.  Want to confirm about nested values additions in future are not impacted in anyway.
```
[assumerole]
   [assumerole2]
        key=value
```
5. 
```
[=:assumerole]
key=value
```
6. 
```
[assumerole] 
key : value
```

7. Nested without spacing
```
[assumerole]
[assumerole2]
key=value
```

For my understanding, most of these are invalid (except 1 and 7 possibly?), is that correct?

internal/ini/visitor.go
```diff
@@ -63,7 +63,7 @@ func (v *DefaultVisitor) VisitExpr(expr AST) error {
 
 			rhs := children[1]
 
-			if rhs.Root.Type() != TokenLit {
+			if !(rhs.Root.Type() == TokenLit || rhs.Root.Type() == TokenOp || rhs.Root.Type() == TokenSep) {
```
skotambkar, line 66: Hard to follow exactly what we now support. Could you add a comment, and update the ini/doc.go

skotambkar (changes requested):
Minor testing and documentation call out

internal/ini/testdata/valid/op_sep_in_values
skmcgrail: Case number 5 is not valid FYI as `:`, `=`, `[`, or `]` can't appear in the section statement names.

skotambkar: Yeah I was guessing most of those are invalid. 

Yes, case 4 was about service specific nesting support.

skotambkar (approved):

internal/ini/ini_parser.go
```diff
@@ -242,7 +247,7 @@ loop:
 				}
 
 				children[len(children)-1] = rhs
-				k.SetChildren(children)
+				root.SetChildren(children)
```
skotambkar, line 250: I am curious about what this change does.

Was this previously setting children on incorrect parent?

skotambkar: Nit: Does the comment at https://github.com/aws/aws-sdk-go-v2/pull/1324/files#diff-42e49a0dfb77920f8deb63d9b2902c1c276fa955fb8089128a242adccf7d76feR210-R224 need an update too?

I am fine with just asking to look at doc.go

skotambkar (approved):

internal/ini/ini_parser.go
skmcgrail: Yeah I'd vote to remove this and not spread the grammar all about.

> Was this previously setting children on incorrect parent?

I took this as a bug (this code wasn't really exercised by the existing test cases before cause they would never hit this block). I was seeing that the children were also being added to the AST identified by K which didn't seem to make sense to me. The intention of this block seemed to be that you get the RHS from the children list of `root`, modify the RHS and set it back on `children` and ensure that the reference on root to it's children was still correct. Where as this was updating K with root's children.

skotambkar (approved):

PR #981, opened 2020-12-15
DynamoDB Expression Package

skmcgrail:
This PR re-introduces the `feature/dynamodb/expression` package.

Currently this branch is based on https://github.com/aws/aws-sdk-go-v2/pull/975 thus the diff includes those changes. Will rebase this PR after that is merged in. For now leaving this as a draft.

Fixes: https://github.com/aws/aws-sdk-go-v2/issues/894
Fixes: https://github.com/aws/aws-sdk-go-v2/issues/713

Related: https://github.com/aws/aws-sdk-go-v2/issues/115

PR #1329, opened 2021-07-01
feat: (feature/dynamodb/expression) re-add IsSet methods

zaquestion:
These changes were present in an older version of the expression package before it was moved, originally added in https://github.com/aws/aws-sdk-go-v2/pull/494, when the expression library was reintroduce in https://github.com/aws/aws-sdk-go-v2/pull/981 these changes werent included for some reason, but the need for checking if a condition is set or not still exists

### DIFF
--- a/.changelog/8e261b4b5633485cad821d38e2677e46.json
+++ b/.changelog/8e261b4b5633485cad821d38e2677e46.json
@@ -1,0 +1,8 @@
+{
+    "id": "8e261b4b-5633-485c-ad82-1d38e2677e46",
+    "type": "feature",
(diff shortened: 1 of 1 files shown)


zaquestion:
@jasdel pinging you since you originally approved this changeset (#494) before it was rehomed

zaquestion:
@skmcgrail @jasdel Any reason this can't come in? Seems non-controversial, but if you've got a different vision in mind for doing this check happy to code up a different solution.

skmcgrail (approved):

skmcgrail:
Sorry for the delay in reviewing this change @zaquestion and thank you for bring this to our attention that this was mistakenly removed when we changed the location of the package. I've pushed an update to your branch to include a changelog annotation for our next release, otherwise the change looks good.

skmcgrail (approved):

PR #1335, opened 2021-07-12
Smithy 1.9 Support

skmcgrail:
Adds support for Smithy 1.9 which introduces support for `NaN`, `Infinity`, and `-Infinity`.

Depends on: https://github.com/aws/smithy-go/pull/316


skmcgrail:
Codegen is expected to fail due to smithy-go dependencies.

jasdel (approved):

JordonPhillips (approved):
I don't see any explicit support for XML deserialization of non-numeric floats. Is that something you get for free already?

skmcgrail:
> I don't see any explicit support for XML deserialization of non-numeric floats. Is that something you get for free already?

For serialization both JSON and XML protocols share the same document serialization logic in `DocumentMemberSerVisitor`. For JSON we had to make deserialization changes due to the standard library json unmarshaler returning this as strings since they are not `json.Number` types that it understands. For XML the values are just being passed to `strings.ParseFloat` which already supports these values.

PR #1340, opened 2021-07-14
Update API Models

skmcgrail:
Update API models for upcoming release.

### DIFF
--- a/.changelog/e691a51d20364875a2f95475e9c3b35a.json
+++ b/.changelog/e691a51d20364875a2f95475e9c3b35a.json
@@ -1,6 +1,6 @@
 {
     "id": "e691a51d-2036-4875-a2f9-5475e9c3b35a",
-    "type": "bugfix",
+    "type": "feature",
     "description": "The ErrorCode method on generated service error types has been corrected to match the API model.",
     "modules": [
         "credentials",
--- a/.changelog/72fc48ecb3864e36a1d00d4e82e81f9e.json
+++ b/.changelog/72fc48ecb3864e36a1d00d4e82e81f9e.json
@@ -11,7 +11,6 @@
         "service/pricing",
         "service/sns",
         "service/ssm",
-        "service/ssmcontacts",
         "service/sts"
     ]
 }


.changelog/e691a51d20364875a2f95475e9c3b35a.json
```diff
@@ -0,0 +1,25 @@
+{
+    "id": "e691a51d-2036-4875-a2f9-5475e9c3b35a",
+    "type": "bugfix",
+    "description": "The ErrorCode method for on service error types has been corrected to match the API model.",
+    "modules": [
```
skotambkar, line 5: Thanks, I missed adding this in my change. 

Is bugfix the correct type? Since this is a minor breaking change.

skmcgrail: Potentially break-change, but was never really correct in the first place as it did not match service documentation. Kind of why I chose to keep it marked as a bugfix, but am open for changing that if we feel that it would be better to be under a feature version.

skotambkar: There are few recent versions which correctly resolved the error to the error type. If someone started using the error code for those, they would see a change in behavior. 

This is a bug-fix which may result in change in behavior under some cases. To be safe, we should make these feature version

.changelog/f4263acfa9964f27a95087734e6aed0d.json
```diff
@@ -0,0 +1,40 @@
+{
+    "id": "f4263acf-a996-4f27-a950-87734e6aed0d",
+    "type": "feature",
+    "description": "Updated service model to latest version.",
```
skotambkar, line 4: I see multiple .changelog file  for updating service model to latest version.  Intentional? 

I see `service/ssm` in both.

skotambkar (approved):

.changelog/f4263acfa9964f27a95087734e6aed0d.json
skmcgrail: Good eye, SSM should not be in both. The intention was to break out the documentation changes (patch bumps) from feature changes (minor bumps). Seems like there is a bug in this case in the git detection logic, as `ssm` should be marked for a feature bump and `ssmcontacts` should be just a patch bump. My likely hunch is this is getting incorrectly detected in both because of both having the same matching prefix. Will create a backlog item to investigate. This won't block the release though.

jasdel (approved):

PR #1354, opened 2021-07-25
s3: make streaming operations use unsigned payload by default

skotambkar:
Adds `SwapPayloadSHA256ResolverMiddleware` that swaps computPayloadSHA256 middleware if an S3 operation uses streaming payload. 

For S3 operations with streaming payload, we use unsigned payload strategy if TLS is enabled.


skotambkar:
Need audit for WriteGetObjectResponse with unsigned payload trait

Issue #1373, opened 2021-08-11
CloudFront Sign creates invalid signature if URL contains &


joshwoodkx:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
After switching from the V1 SDK to the V2 SDK, signing a URL with more than 2 query parameters results in an invalid signature and cloudfront will return a 403. For a url with 1 or less query parameters, the V1 SDK and V2 SDK signatures are identical. But if you have a URL with more than 2 query parameters, the V2 SDK will return a different signature than the V1 SDK. This looks like a regression from the v1 SDK - an issue was raised and fixed in V1 but it looks like the fix is not included in the v2 SDK https://github.com/aws/aws-sdk-go/issues/2163

**Version of AWS SDK for Go?**
1.40.18

**Version of Go (`go version`)?**
go1.16.4 darwin/amd64

**To Reproduce (observed behavior)**
Code example copied from this issue: https://github.com/aws/aws-sdk-go/issues/2163
```
signer := sign.NewURLSigner(keyID, privKey)
signedURL, _ := signer.Sign("https://myuploaddomain.com/", time.Now().Add(1*time.Hour))
// signedURL VALID
signedURL, _ := signer.Sign("https://myuploaddomain.com?key=one", time.Now().Add(1*time.Hour))
// signedURL VALID
signedURL, _ := signer.Sign("https://myuploaddomain.com?key=one&value=two", time.Now().Add(1*time.Hour))
// signedURL INVALID returns AccessDenied when uploading

```

**Expected behavior**
Signing a URL containing two query string parameters should result in the same signature as the V1 SDK.

PR #1374, opened 2021-08-11
Do not Escape HTML when encoding the policy

joshwoodkx:
Fixes [#1373 ](https://github.com/aws/aws-sdk-go-v2/issues/1373)

This code was taken from a related PR on the V1 SDK: https://github.com/aws/aws-sdk-go/pull/2164.

Since the V2 SDK only supports go1.15 and above, we can just use the `encoder.SetEscapeHTML(false)` that the go json encoder provides (the PR for V1 had to also support v1.7, so that PR also had to implement a solution without using this)

### DIFF
--- a/feature/cloudfront/sign/policy.go
+++ b/feature/cloudfront/sign/policy.go
@@ -182,11 +182,13 @@ func NewCannedPolicy(resource string, expires time.Time) *Policy {
 
 // encodePolicy encodes the Policy as JSON and also base 64 encodes it.
 func encodePolicy(p *Policy) (b64Policy, jsonPolicy []byte, err error) {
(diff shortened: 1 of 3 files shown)


skmcgrail:
Thank you @joshwoodkx for bringing this to our attention, and putting together a PR to address this issue. This seems to have been an oversight when porting over the code from V1. It appears that the logic was not copied from https://github.com/aws/aws-sdk-go/blob/e2d6cb448883e4f4fcc5246650f89bde349041ec/service/cloudfront/sign/policy_json_1_7.go as expected. We will get this merged in for our next release.

skmcgrail (approved):

PR #1383, opened 2021-08-17
Fix interface usage

syndbg:
Use the already defined interface that's already utilized in default AWS resolvers.

Ideally, this should result in "easier to maintain" code and less ad-hoc types and/or containers (e.g slices) with definitions such as `[]func(*LoadOptions) error`.


skmcgrail:
Thanks for putting this PR together @syndbg. Unfortunately we will not be able to accept this change, as it has the potential to break certain use cases. As an example https://play.golang.org/p/sNB4hvG0oz5 is such a case where this change would not be backwards compatible.

PR #1389, opened 2021-08-20
docs: correct context usage with timeout

Elgarni:
The context `context.WithTimeout(..)` returns two params `(Context, CancelFunc) `, but it's being used directly in `client.GetObject(..)` which of course will not work. Also, when using `withTimeout` it's best practice to call `defer cancel()`.

This PR reflects those changes.


jasdel (approved):
Thanks for creating this PR @Elgarni and correcting the documentation example. The change looks good. We'll publish this in the SDK's next release.

Issue #1413, opened 2021-09-10
aws/retry: Retry token post request attempt not being cleared


jasdel:
Noticed this relRetryToken value being returned, but the value is never used again, nor released. This potentially could lead to token leaking preventing subsequent retry attempts being made. This needs more investigation, and cleanup to correctly handle the returned retry token for the attempt.

https://github.com/aws/aws-sdk-go-v2/blob/a1be7274d533f8612217a3facc79c0b1a902612c/aws/retry/middleware.go#L170-L173

Skarlso:
I'll take a look if people don't mind. :)

Skarlso:
Yeah, there are some unused things in the test too. Especially the mockhandler and retryProvider. So there is a gap in testing as well.

In fact the retry errors are not tested... I suspect there might be a bug in there. I'll increase the coverage and fix the unused releaser.

Skarlso:
Okay, so this is what I found.... 

This is a tricky situation and please correct me if I'm wrong, but... looking at this code:

```go
	if err == nil {
		return out, attemptResult, err
	}

	retryable := r.retryer.IsErrorRetryable(err)
	if !retryable {
		r.logf(logger, logging.Debug, "request failed with unretryable error %v", err)
		return out, attemptResult, err
	}

	// set retryable to true
	attemptResult.Retryable = true

	if maxAttempts > 0 && attemptNum >= maxAttempts {
		r.logf(logger, logging.Debug, "max retry attempts exhausted, max %d", maxAttempts)
		err = &MaxAttemptsError{
			Attempt: attemptNum,
			Err:     err,
		}
		return out, attemptResult, err
	}

	relRetryToken, reqErr := r.retryer.GetRetryToken(ctx, err)
	if reqErr != nil {
		return out, attemptResult, reqErr
	}
```

It looks like, this whole thing stops if `err == nil`. And the release function is a no-op in case of errors.

```go
func (f releaseToken) release(err error) error {
	if err != nil {
		return nil
	}

	return f()
}
```

Which means, If I see this function correctly, calling the release there after that getToken call basically won't do anything ever. Because it wouldn't even get to this code segment if there was no error. And in the begin, once this function starts again, the same release is called.

I'll write some more tests however to prove that there is some kind of problem. I wrote a small test already which fiddles with this code path. Still, I might be incorrect, but please double check.

Cheers!

Skarlso:
[...]

PR #1424, opened 2021-09-18
Fix unreleased capacity after retryToken is requested

Skarlso:
Closes https://github.com/aws/aws-sdk-go-v2/issues/1413

For further details, please read the attached issue.

One issue I could still see possibly is that this is always overwritten and only called ONCE after a success. I'm not familiar with the code enough that after a success the Attempt and the retry is re-created completely? Because then this could be fine.

### DIFF
--- a/aws/retry/middleware_test.go
+++ b/aws/retry/middleware_test.go
@@ -2,6 +2,7 @@ package retry
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -10,11 +11,13 @@ import (
 	"testing"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/internal/sdk"
 	"github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
 	"github.com/google/go-cmp/cmp"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/ratelimit"
+	"github.com/aws/aws-sdk-go-v2/internal/sdk"
 )
 
 func TestMetricsHeaderMiddleware(t *testing.T) {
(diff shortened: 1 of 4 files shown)


aws/retry/middleware_test.go
```diff
@@ -427,13 +430,27 @@ func TestAttemptMiddleware(t *testing.T) {
...
 			if !ok {
 				t.Fatalf("expected metadata to contain attempt results, got none")
 			}
-			if e, a := tt.ExpectResults, attemptResults; !reflect.DeepEqual(e, a) {
-				t.Fatalf("expected %v, got %v", e, a)
+			e, a := tt.ExpectResults, attemptResults
+			if len(e.Results) != len(a.Results) {
+				t.Fatalf("expected len %d got %d", len(e.Results), len(a.Results))
+			}
+			for i, eRes := range e.Results {
```
Skarlso, line 441: This is a side effect of altering the return value with adding a function into it. Can no longer be deepEqualed. If anyone has a nice idea on how to fix that, I'm open for suggestions.

Sadly, it's late and I can't think of anything better ATM. :)

Skarlso: I can experiment with a mock retry provider in these cases.

jasdel: We should be able to change how these values are compared to use [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp@v0.5.6/cmp#Diff). This is commonly used for other data types in the v2 SDK.

Skarlso: I can try that, sure.

Skarlso: Though I'm not sure how that works for functions.

jasdel: Thanks for clarifying that. Yes, you're correct cmp.Diff doens't work with functions, checking for not-nil is pretty much the only way to compare in this case.

jasdel:
Thanks for taking the time to create this PR @Skarlso. I've taken these changes and incorporated them in https://github.com/aws/aws-sdk-go-v2/pull/1560. That PR takes the idea of these changes, but instead of using a member on `AttemptResults` updates `handleAttempt` to take in the previous `releaseRetryToken`, and return the next token. With `nopRelease` (No-Op func) being return if there is no retry token.

Lets use #1560 to get this changed merged in, and released into the SDK.

Skarlso:
@jasdel Nice improvement. :)

PR #1453, opened 2021-10-09
Support use of json tag if available in dynamo encoder

sudopla:
I think it would be nice that when we call `attributevalue.Marshal()`, `attributevalue.MarshalList()`, `attributevalue.MarshalMap()` and the other similar `unmarshal` methods,  they all support `json` tags in the struct by default like in the previous version of the SDK.

### DIFF
--- a/feature/dynamodb/attributevalue/decode.go
+++ b/feature/dynamodb/attributevalue/decode.go
@@ -136,7 +136,9 @@ type Decoder struct {
 // NewDecoder creates a new Decoder with default configuration. Use
 // the `opts` functional options to override the default configuration.
 func NewDecoder(optFns ...func(*DecoderOptions)) *Decoder {
-	var options DecoderOptions
+	options := DecoderOptions{
+		TagKey: "json",
+	}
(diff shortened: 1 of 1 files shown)


kenneth488:
Great work @sudopla! 

Just added this feature request my self: https://github.com/aws/aws-sdk-go-v2/issues/1457. Hopefully this gets approved.

jasdel:
Thanks for taking the time to create this Pr @sudopla. The best way to change the SDK's behavior in your application is to create an encoder and specify the `Tagkey` that should be used for the key name of attribute value map keys.  The SDK will still use `dynamodbav` for all non key options such as `omitempty` and `omitnull`.  

For more than that lets continue this discussion over in #1457.

```go
ddbAvDecoder := attributevalue.NewDecoder(func(o *attributevalue.DecoderOptions) {
    o.TagKey = "json"
})

ddbAvEncoder := attributevalue.NewEncoder(func(o *attributevalue.EncoderOptions) {
    o.TagKey = "json"
})
```

PR #1459, opened 2021-10-14
Fix SDK's API model sync utility to ignore models without sdkId

jasdel:
Updates the SDK's API model sync utility to ignore modeled service
shapes that do not have a `sdkId` or `aws.api#service` trait.


skotambkar:
Should sync with Smithy on this to verify this is supported behavior.

Issue #1414, opened 2021-09-11
Batch UpdateComputeEnvironment will not update MinvCpus to zero


cgmartin:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [x] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
Batch UpdateComputeEnvironment will not update MinvCpus to zero.

**Version of AWS SDK for Go?**
AWS version: 1.8.1

**Version of Go (`go version`)?**
Go version:  go1.16.7

**To Reproduce (observed behavior)**

```go
input := &batch.UpdateComputeEnvironmentInput{
	ComputeEnvironment: computeEnv.ComputeEnvironmentArn,
	ComputeResources: &batchTypes.ComputeResourceUpdate{
		MinvCpus:     0,
		MaxvCpus:     16,
		DesiredvCpus: 4,
	},
	State: batchTypes.CEStateDisabled,
}
output, err := s.BatchClient.UpdateComputeEnvironment(ctx, input)

// Results in the following request body to AWS backend:
// {"computeEnvironment":"arn:aws:batch:us-east-1:...","computeResources":{"desiredvCpus":4,"maxvCpus":16},"state":"DISABLED"}
```
The "minvCpus" property is not sent.

**Expected behavior**

I expect to be able to change the minvCpus value to 0 with an UpdateComputeEnvironment call.

**Additional context**

Serializers are currently implemented to exclude zero values for MinvCpus, MaxvCpus and DesiredvCpus, see https://github.com/aws/aws-sdk-go-v2/blob/a1be7274d533f8612217a3facc79c0b1a902612c/service/batch/serializers.go#L1767-L1780

jasdel:
Thanks for reporting this bug @cgmartin. We're investigating the best way to handle this issue. We think this is similar to the modeling issue the SDK had with Amazon EC2's API model, e.g. https://github.com/aws/aws-sdk-go-v2/issues/1107. 

[...]

PR #1466, opened 2021-10-20
Fix batch service update compute environment min, max, desired values

cgmartin:
This PR answers #1414


jasdel:
Thanks for taking the time to create this PR @cgmartin. While we cannot take these changes because it modifies the generated code. We are investigating how this issue should be solved from the SDK's code generation side driven by the `batch` service's API model.   I've updated #1414 with more information, and we can use that issue to track this bug and associated fix.

PR #1495, opened 2021-11-13
i1494 passing opts to attributevalue marshalling

### LINKED ISSUE #1494: Pass KeyTag to dynamo attributevalue marshalling
### Describe the feature

As a AWS SDK go v2 user, I want to be able to set a custom `KeyTag` on `attributevalue.MarshalMap`, so I can keep the compatibility with previous v1 SDK.

### Is your Feature Request related to a problem?

Currently the `attributevalue.MarshalMap` doesn't accept the Encoder options, even though it is possible without breaking changes.

By default, `KeyTag` is set to `dynamoav`. Passing the Encoder options will allow setting up a KeyTag like `json`.

### Proposed Solution

_No response_

### Describe alternatives you've considered

_No response_

### Acknowledge

### AWS Go SDK version used

v2

### Go version used

1.16

radutopala:
Fixes #1494

### DIFF
--- a/.changelog/5ab32814eb5e456ea921d0f0a645ca5b.json
+++ b/.changelog/5ab32814eb5e456ea921d0f0a645ca5b.json
@@ -1,0 +1,8 @@
+{
+    "id": "5ab32814-eb5e-456e-a921-d0f0a645ca5b",
+    "type": "feature",
+    "description": "Adds new MarshalWithOptions and UnmarshalWithOptions helpers allowing Encoding and Decoding options to be specified when serializing AttributeValues. Addresses issue: https://github.com/aws/aws-sdk-go-v2/issues/1494",
+    "modules": [
+        "feature/dynamodb/attributevalue"
+    ]
+}


jasdel (changes requested):
Thanks for taking the time to create this PR @radutopala. Modifing the signatures of the `Marshal` methods would be a breaking change, and we could not release it. 

Alternatively new `Marshal` methods could be defined with different names, e.g. `MarshalWithOptions` to take the functional options.

Also we'd probably want similar new functions for the `Unmarshal` set of functions.

radutopala:
Adding `optFns ...func(*EncoderOptions)` as a variadic arg means no breaking changes, one can call Marshall with or without passing any optFns. But if I'm missing something, please let me know.

jasdel:
Thanks for the feedback. Adding a variadic arg to a function call will not break existing callers, but since in Go functions can be types and variables, any application that creates a variable or struct member type matching the `Marshal` function signature would break if variadic arg were added.

https://go.dev/play/p/c4yT8-DsHo9

radutopala:
@jasdel it's an extreme edge case, wondering who'd use this pattern with this specific func. But ok, better safe! I'll push some changes.

radutopala:
@jasdel branch updated, please CR / provide next steps.

jasdel (approved):
Thanks for taking the time to put this PR together, and make the requested updates. I've added changelog notes to the PR. We'll get this update merged in and released in the SDK's next update.

Issue #1496, opened 2021-11-16
Defined exported error which allows to determine that no more pages are available in ListObjectsV2Paginator.NextPage()


elideveloper:
### Describe the feature

When I use ListObjectsV2Paginator.NextPage() I want to have an error like 's3.ErrNoMorePages' to react and process this case in my code. For now this function just returns general error (fmt.Errorf("no more pages available"))

### Is your Feature Request related to a problem?

I cannot easily distinguish case when paginator listed all pages from all other error cases.  

### Proposed Solution

Need to define specific exported error for that case (like ErrNoMorePages)

### Acknowledge

- [X] I may be able to implement this feature request

### AWS Go SDK version used

1.16.1

### Go version used

1.16

jasdel:
Thanks or taking the time to create this issue @elideveloper I updated the associated PR with feedback. Having a typed error for Paginator's `no more pages available` error, and Waiters' [literal errors](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_op_HeadObject.go#L591-L601) as well.

This change will involve updating the SDK's code generation for returning these concrete error types. We'll also want to see if we can change the error type from the type returned by `fmt.Error` to some typed error.

elideveloper:
thank U @jasdel , will be great if literal errors will turn to defined errors in such places

jasdel:
@elideveloper I took another look at this feature request and PR and realized I did not ask initial questions about the usecase and how your ran into this issue. Could you provide some background how you ran into the `no more pages available` error?  Where you using the `HasMorePages()` method in a loop? This is the preferred way of detecting if there are probably more pages or not remaining.

The best practice pattern of using paginators is via a `for` loop calling both `HasMorePages` and `NextPage`. e.g. [NewListObjectsV2Paginator](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#ListObjectsV2Paginator)

```go
paginator := s3.NewListObjectsV2Paginator(client, params)

for paginator.HasMorePages() {
    page, err := paginator.NextPage(context.TODO())
    // Handle error and page
}
```
Using the above pattern, the only error you should encounter with `NextPage` is API, or connection errors. `HasMorePages` should ensure that `NextPage` won't be called unless there are more pages available. `HasMorePages` will always return `true` for the first iteration since its unknown if there are more pages or not yet.

PR #1497, opened 2021-11-16
1496: defined ErrNoMorePages and returning it from NextPage()

elideveloper:
Tried to fix in a simple way, please fix me
https://github.com/aws/aws-sdk-go-v2/issues/1496

### DIFF
--- a/service/s3/api_op_ListObjectsV2.go
+++ b/service/s3/api_op_ListObjectsV2.go
@@ -4,7 +4,7 @@ package s3
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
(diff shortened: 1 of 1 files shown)


jasdel (changes requested):
Thanks for taking the time to create this PR @elideveloper. This change modifies the SDK's generated files so cannot be taken directly as is. With that said though, having a typed error for Paginator's `no more pages available` error, and Waiters' [literal errors](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_op_HeadObject.go#L591-L601) as well.

This change will involve updating the SDK's code generation for returning these concrete error types. We'll also want to see if we can change the error type from the type returned by `fmt.Error` to some typed error.

jasdel:
Lets use https://github.com/aws/aws-sdk-go-v2/issues/1496 to track this issue and implementation.

PR #1500, opened 2021-11-17
Fixes for DescribePlaceIndex service.

jaanauati:
This fixes the following error:

```
5 operation error Location: DescribePlaceIndex, https response error StatusCode: 403, RequestID: 35f23c7d-0a75-44ec-b47e-e4f56f07dff5, api error MissingAuthenticationTokenException: Missing Authentication Token
```

After some little digging I realized that `DescribePlaceIndex` is currently not using/including the `addHTTPSignerV4Middleware`  which leads to the mentioned auth errors.

### DIFF
--- a/service/location/internal/endpoints/endpoints.go
+++ b/service/location/internal/endpoints/endpoints.go
@@ -127,7 +127,7 @@ var defaultPartitions = endpoints.Partitions{
 			{
 				Variant: 0,
 			}: {
-				Hostname:          "places.geo.{region}.amazonaws.com",
+				Hostname:          "geo.{region}.amazonaws.com",
 				Protocols:         []string{"https"},
 				SignatureVersions: []string{"v4"},
 			},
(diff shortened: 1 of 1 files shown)


service/location/internal/endpoints/endpoints.go
```diff
@@ -127,7 +127,7 @@ var defaultPartitions = endpoints.Partitions{
 			{
 				Variant: 0,
 			}: {
-				Hostname:          "geo.{region}.amazonaws.com",
```
jaanauati, line 130: got this one from aws-cli (using `--debug` flag), i'm not sure about the rest of the urls in this file though (geo-fips, geo.{region}.api, etc).

skmcgrail:
Closing in favor of https://github.com/aws/aws-sdk-go-v2/pull/1518 which resolves this at the code generation level. Thanks for the report!

PR #1509, opened 2021-11-24
doc: make tag scope clear

charles-d-burton:
The word tag is loaded for Docker, this makes more clear which resource it's affecting.


jasdel:
Thanks for reporting this issue, and creating the PR. The file being modified is a file generated from a model owned by the service team, not the SDK. All API operations, types, and documentation are generated into the SDK from these models. This is the reason the SDK is unable to take any changes modifying these generated files, or their backing models.

The best way to resolve this issue is to submit feedback on the [API Reference Guide for UntagResource Action](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_UntagResource.html)

Issue #1486, opened 2021-11-08
`json` annotations are not supported in attributevalue.MarshalMap


Cyberax:
Confirm by changing [ ] to [x] below to ensure that it's a bug:
- [X] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution
  
**Describe the bug**
`json` annotations are not supported in attributevalue.MarshalMap. The only supported annotations are `dynamodbav`. This is a regression compared to the V1 SDK.

**Version of AWS SDK for Go?**
Example: v1.11.0, DDB: 1.7.0

**Version of Go (`go version`)?**
1.16

**To Reproduce (observed behavior)**
```
type SomeStruct struct{
    Id string `json:"id"`
}
s := SomeStruct{Id: "123"}
values, err := attributevalue.MarshalMap(&s)
// values will contain: {Id -> "123"} - note the upper case
```

**Expected behavior**
Return:
```
id -> "123"
```

moogacs:
I have encountered the same bug which `json` tag is not respected  when marshalling a `struct` and it has inconsistent behaviour that it's sometimes respected and sometimes not

```
go version v1.16
dynamodb v1.9.0 
aws sdk v1.11.1

```

0DegreesKelvin:
Same bug for me after update to aws sdk v1.11.1

jasdel:
Thanks for reaching out. The issue you're experiencing is the intended behavior of V2 SDK's AttributeValue (un)marshalers. The best way to configure the (un)marshalers to use alternative struct tags is to documented in the package's [Overview](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue#hdr-Struct_tags)

There are a few ways you can configure the (un)marshaler for the serialized id to use.
* add `dynamodbav:"name"` struct tag to fields
* Implement [attributevalue.Marshaler](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue#Marshaler) with custom specification of conversion between Go type and AttributeValue.
[...]

PR #1520, opened 2021-12-03
1486: dynamo/attributevalue  allow json tag encoding/decoding

moogacs:
#1486 reads json tags on marshalling structs


jasdel:
Thanks for taking the time to create this PR @moogacs I updated #1486 with information how to use alternative tags with the `attributevalue` package's (un)marshalers.

In addition, adding support for `json` struct tag name by default could be a breaking change for some users where serialization of structure field names would change, and no longer be compatible with their data stored in their table.

Lets use #1486 and #1494 for discussion of this issue.

PR #1522, opened 2021-12-04
Credential_cache: Exposing Renew method

fpozzobon:
The aim of this PR is to expose a Renew function.
Consumers of credential_cache can rotate credentials without having to invalidate the cache first.

Use case:
- Renew the credential before it expires outside of the "hot path" to avoid extra latency (eg when presigning S3 objects)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## Module Highlights
 * `github.com/aws/aws-sdk-go-v2/service/batch`: [v1.17.0](service/batch/CHANGELOG.md#v1170-2022-03-25)
   * **Feature**: Bug Fix: Fixed a bug where shapes were marked as unboxed and were not serialized and sent over the wire, causing an API error from the service.
+  * This is a breaking change, and has been accepted due to the API operation not being usable due to the members modeled as unboxed (aka value) types. The update changes the members to boxed (aka pointer) types so that the zero value of the members can be handled correctly by the SDK and service. Your application will fail to compile with the updated module. To workaround this you'll need to update your application to use pointer types for the members impacted.
 * `github.com/aws/aws-sdk-go-v2/service/ec2`: [v1.33.0](service/ec2/CHANGELOG.md#v1330-2022-03-25)
   * **Feature**: This is release adds support for Amazon VPC Reachability Analyzer to analyze path through a Transit Gateway.
 * `github.com/aws/aws-sdk-go-v2/service/ssm`: [v1.24.0](service/ssm/CHANGELOG.md#v1240-2022-03-25)
--- a/service/batch/CHANGELOG.md
+++ b/service/batch/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v1.17.0 (2022-03-25)
 
 * **Feature**: Bug Fix: Fixed a bug where shapes were marked as unboxed and were not serialized and sent over the wire, causing an API error from the service.
+* This is a breaking change, and has been accepted due to the API operation not being usable due to the members modeled as unboxed (aka value) types. The update changes the members to boxed (aka pointer) types so that the zero value of the members can be handled correctly by the SDK and service. Your application will fail to compile with the updated module. To workaround this you'll need to update your application to use pointer types for the members impacted.
 
 # v1.16.2 (2022-03-24)
 


aws/credential_cache.go
```diff
@@ -101,6 +101,11 @@ func (p *CredentialsCache) singleRetrieve(ctx context.Context) (interface{}, err
 		return *creds, nil
 	}
 
+	return p.Renew(ctx)
+}
+
+// Renew will fetch a new credential from provider and cache it
+func (p *CredentialsCache) Renew(ctx context.Context) (interface{}, error) {
```
jasdel, line 108: I think this change would introduce race conditions with the underlying provider's Retrieve method. The CredentailsCache plays two roles.

1.) caching of credentials so they are not retrieved as often
2.) concurrency safe wrapper around a unsafe provider.

a `Renew` method like this would expose the potentially unsafe provider to race conditions

jasdel (changes requested):
Thanks for taking the time to create this PR @fpozzobon. 

I think that the CredentialsCache can be used in the way you're looking for without adding additional methods.

The CredentialsCache's [Invalidate](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#CredentialsCache.Invalidate) method will invalidate the cache so that the next call to [Retrieve](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#CredentialsCache.Retrieve) will refresh the credentials. This can be done outside of a hot path without modifying the `CredentialsCache` type.

```go
cfg, err := config.LoadDefaultConfig(context.TODO())
// ...

cfg.Credentials.Invalidate()
_, err := cfg.Credentials.Retrieve(context.TODO())
```

fpozzobon:
> Thanks for taking the time to create this PR @fpozzobon.
> 
> I think that the CredentialsCache can be used in the way you're looking for without adding additional methods.
> 
> The CredentialsCache's [Invalidate](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#CredentialsCache.Invalidate) method will invalidate the cache so that the next call to [Retrieve](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#CredentialsCache.Retrieve) will refresh the credentials. This can be done outside of a hot path without modifying the `CredentialsCache` type.
> 
> ```go
> cfg, err := config.LoadDefaultConfig(context.TODO())
> // ...
> 
> cfg.Credentials.Invalidate()
> _, err := cfg.Credentials.Retrieve(context.TODO())
> ```

@jasdel thanks for your quick feedbacks.
my concern on invalidating the cache before retrieving the credentials is that if a request from the "hot path" comes in it will be slowed by retrieving new credentials.
To address this risk, I created a cache layer around CredentialCache but it is a bit redundant ^^'

aws/credential_cache.go
fpozzobon: Fair point, do I understand correctly that "p.sf.DoChan" is protecting from race conditions by deduplicating?
what if we add a simple mutex here, would it be acceptable?

fpozzobon:
Let me explain more in details the rational, our service is having a number of requests > 2k/s and a SLO of up to 150ms (request is cancelled after that)

@jasdel my concern on your suggestion is the following scenario:

- renewer: Object responsible for rotating the credential
- service: Object responsible for presigning s3 url
- credential_cache is shared on those 2 objects

1.  Renewer invalidate the cache: `cfg.Credentials.Invalidate()`

2. (bad luck) A request comes in, service presign s3 url
2.1. presign s3 url calls `cfg.Credentials.Retrieve(ctx)`
2.2. as cache is invalidated, cache request new credential from provider
2.3. wait for new credentials

3. Renewer calls. cfg.Credentials.Retrieve(context.TODO())
3.1. deduplication from cache and waiting for credentials

4. Credentials are retrieved, cached

5. Request can proceed to presign (but it can be too late due to waiting for new credentials)

=> Ideally the renewal should be isolated from the cache (ie not invalidated too early) until the new credentials are ready so that it doesn't disrupt the "hot path"


Is this scenario correct or am I missing something?

Many thanks

jasdel:
Thanks for updating this PR, and feedback. I'm working on a feature for the `smithy-go` module, and I'm wondering if this concept could be applied this PR's usecase as well. Where it would be helpful to renew credentials asynchronously, or out of band of API client's calling `Retrieve`.


I'm working on a feature in`smithy-go` for support of Smithy's [Bearer Token authentication](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#httpbearerauth-trait). A part of this feature needs its own [TokenCache](https://github.com/aws/smithy-go/blob/jasdel/feat/BearerToken/auth/bearer/token_cache.go). This `TokenCache` defaults to asynchronous refresh of the bearer token 

I'm thinking we could port the concept of asynchronous refresh of the `TokenCache` to the SDK's `CredentialsCache`. Maybe create an `CredentialsAsyncCache`, but would prefer if this could be worked into `CredentialsCache instead. This would allow asynchronous background refreshes of the SDK's AWS Credentials before credentials actually expire. I think that would address the usecase/feature request of this PR.

fpozzobon:
> Thanks for updating this PR, and feedback. I'm working on a feature for the `smithy-go` module, and I'm wondering if this concept could be applied this PR's usecase as well. Where it would be helpful to renew credentials asynchronously, or out of band of API client's calling `Retrieve`.
> 
> The feature in`smithy-go` is for support of Smithy's [Bearer Token authentication](https://awslabs.github.io/smithy/1.0/spec/core/auth-traits.html#httpbearerauth-trait). A part of this feature needs its own Caching utility, [TokenCache](https://github.com/aws/smithy-go/blob/jasdel/feat/BearerToken/auth/bearer/token_cache.go). Unlike the `CredentialsCache` in the SDK, the `TokenCache` defaults to asynchronous refresh of the bearer token, and does not modify the retrieved token's Expiration.
> 
> I'm thinking we could port the concept of asynchronous refresh of the `TokenCache` to the SDK's `CredentialsCache`. Maybe create an `CredentialsAsyncCache`, but would prefer if this could be worked into `CredentialsCache instead. This would allow asynchronous background refreshes of the SDK's AWS Credentials before credentials actually expire. I think that would address the usecase/feature request of this PR.

I think it's a good idea, there is a challenge though on how to deduce if the expiration of the token is "real expiration" or "anticipated expiration" with the `ExpiryWindow option set` as this logic is hidden behind `Expired() function` and `AdjustExpiresBy`.
Closing this PR as it was stalled since some times and having conflicts.

PR #1528, opened 2021-12-10
Replaced use of legacy binary units with IEC units for firehose codegen

comrumino:
AWS documentation uses KiB and MiB per IEC units. Firehose should not be using "legacy" unit definitions.


skmcgrail:
Hey @comrumino, since these models are vended by the service team and their documentation writers we can't modify the models directly here. The best way to ensure this feedback gets routed to the team is to submit this feedback via the 'Provide feedback' button on the service's API reference guide. You can find the documentation in question [here](https://docs.aws.amazon.com/firehose/latest/APIReference/API_PutRecordBatch.html).

comrumino:
Done. Good to know. Thanks!

Issue #1503, opened 2021-11-18
ExclusiveStartKey on NewScanPaginator not working


Zeedinstein:
### Documentation

- [X] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution

### Describe the bug

It seems the ExclusiveStartKey on the ScanInput for NewScanPaginator doesn't affect where the can starts. I'm trying to create a simple paginated API but every time I pass the ExclusiveStartKey it returns the same items as the previous request.

### Expected behavior

The ExclusiveStartKey should be used as the start key for the scan.

### Current behavior

The ExclusiveStartKey is ignored 

### Steps to Reproduce

You'll need a local instance of Dynamodb running. I've created a test.go file that shows what is happening. 

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
)

type PaginateID struct {
	ID string `dynamodbav:"id" json:"id"`
}

func main() {
	lastKey := PaginateID{ID: "ae82a99d-486e-11ec-a7a7-0242ac110002"}

	key, err := attributevalue.MarshalMap(lastKey)
	if err != nil {
		fmt.Println(err)
		return
	}

	cfg, err := config.LoadDefaultConfig(context.TODO(), func(o *config.LoadOptions) error {
		o.Region = os.Getenv("REGION")
		return nil
	})
	if err != nil {
		fmt.Println(err)
		return
	}

	svc := dynamodb.NewFromConfig(cfg, func(o *dynamodb.Options) {
		o.EndpointResolver = dynamodb.EndpointResolverFromURL("http://localhost:8000")
	})

	input := &dynamodb.ScanInput{
		TableName:         aws.String("TABLE_NAME"),
		Limit:             aws.Int32(1),
		ExclusiveStartKey: key,
	}

	paginator := dynamodb.NewScanPaginator(svc, input)

[...]

Issue #1452, opened 2021-10-08
dynamodb: QueryPaginator constructor does not respect the ExclusiveStartKey param


fedonev:
The dynamodb `QueryPaginator` ignores `params.ExclusiveStartKey` in the constructor.

I believe a simple change to `NewQueryPaginator` will fix the issue:   need to initialise the `nextToken` attribute.

### Expected
The paginator's  initial `NextPage` call should return items beginning with the record after the `ExclusiveStartKey`.

### Observed
The first `NextPage` call returns items as if no start key had been passed.  (I am using the latest `dynamodb` package v1.5.1 in a lambda function)

### Diagnosis
I believe the root cause is that `nextToken` is not initialised  [in the QueryPaginator constructor](https://github.com/aws/aws-sdk-go-v2/blob/a930583ec3794253f8fb7bc5cc099e3cda8a355f/service/dynamodb/api_op_Query.go#L560-L566).  As such, `nextToken` is always nil on the first call to [NextPage](https://github.com/aws/aws-sdk-go-v2/blob/a930583ec3794253f8fb7bc5cc099e3cda8a355f/service/dynamodb/api_op_Query.go#L573-L580). 

### Fix
I believe the fix is a one-line addition to the constructor to initialise `nextToken` to `params.ExclusiveStartKey`:

```go
// NewQueryPaginator returns a new QueryPaginator
func NewQueryPaginator(client QueryAPIClient, params *QueryInput, optFns ...func(*QueryPaginatorOptions)) *QueryPaginator {
	if params == nil {
		params = &QueryInput{}
	}


	options := QueryPaginatorOptions{}
	if params.Limit != nil {
		options.Limit = *params.Limit
	}


	for _, fn := range optFns {
		fn(&options)
	}


	return &QueryPaginator{
		options:   options,
		client:    client,
		params:    params,
		firstPage: true,
                nextToken: params.ExclusiveStartKey  // <-- add this line to initialise nextToken
	}
}
```

marc-ostrow:
[...]

PR #1531, opened 2021-12-13
Paginator Improvements

### LINKED ISSUE #1529: Infinite loop in DescribeSpotPriceHistoryPaginator
### Documentation

- [X] I've gone though the [API reference](https://docs.aws.amazon.com/sdk-for-go/v2/api/)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for answers
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-go-v2/issues) and didn't find any solution

### Describe the bug

There's an infinite loop in `DescribeSpotPriceHistoryPaginator` because the paginator doesn't check for an empty token in

```
// HasMorePages returns a boolean indicating whether more pages are available
func (p *DescribeSpotPriceHistoryPaginator) HasMorePages() bool {
	return p.firstPage || p.nextToken != nil
}
```

### Expected behavior

The function should check for an empty token:

```
// HasMorePages returns a boolean indicating whether more pages are available
func (p *DescribeSpotPriceHistoryPaginator) HasMorePages() bool {
	return p.firstPage || (p.nextToken != nil && *p.nextToken != "")
}
```

### Current behavior

Infinite loop

### Steps to Reproduce

Just using the paginator results in an infinite loop

```
func TestLoop(t *testing.T) {
    cfg, _ := config.LoadDefaultConfig(context.Background())
    api := ec2.NewFromConfig(cfg)
    paginator := ec2.NewDescribeSpotPriceHistoryPaginator(api, &ec2.DescribeSpotPriceHistoryInput{
        InstanceTypes: []types.InstanceType{types.InstanceTypeR4Xlarge},
        StartTime:     aws.Time(time.Now()),
        EndTime:       aws.Time(time.Now()),
    })
    for paginator.HasMorePages() {
        output, err := paginator.NextPage(context.Background())
        if err != nil {
            return
        }

        for _, sp := range output.SpotPriceHistory {
            println(*sp.SpotPrice)
        }
    }

}
```

### Possible Solution

_No response_

### AWS Go SDK version used

latest

### Compiler and Version used

go version go1.17.3 darwin/arm64

Darwin

skmcgrail:
* Treat empty length string tokens as paginator last page condition.
* Honor initial starting token specified by the provided input parameters.

Fixes #1529, #1503, #1452

Requires https://github.com/aws/smithy-go/pull/345


jasdel (approved):

kkolur:
@skmcgrail Hi, I was looking through the paginators again and the HasMorePages() method is inconsistent across some of the files. In [api_op_Query.go](https://github.com/aws/aws-sdk-go-v2/blob/release-2021-12-21/service/dynamodb/api_op_Query.go#L571) the current condition is :

`return p.firstPage || p.nextToken != nil`. 

Here we do not check that the length of the nextToken is not 0, but in [api_op_ListContributorInsights.go](https://github.com/aws/aws-sdk-go-v2/blob/release-2021-12-21/service/dynamodb/api_op_ListContributorInsights.go#L181), we do correctly check for the length of the nextToken 

`return p.firstPage || (p.nextToken != nil && len(*p.nextToken) != 0)`

Can we update all of the HasMorePages() to the condition where we check for the length of the nextToken as well?

PR #1553, opened 2022-01-10
SDK DefaultsMode Configuration Implementation

skmcgrail:
Requires: ~~https://github.com/aws/smithy-go/pull/349~~ https://github.com/aws/smithy-go/pull/350

### DIFF
--- a/aws/protocol/eventstream/go.mod
+++ b/aws/protocol/eventstream/go.mod
@@ -3,7 +3,7 @@ module github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream
 go 1.15
 
 require (
-	github.com/aws/smithy-go v1.9.1
+	github.com/aws/smithy-go v1.9.2-0.20220113020543-dec09760da63
 	github.com/google/go-cmp v0.5.6 // indirect
 )
 
--- a/aws/protocol/eventstream/go.sum
+++ b/aws/protocol/eventstream/go.sum
@@ -1,5 +1,5 @@
-github.com/aws/smithy-go v1.9.1 h1:5vetTooLk4hPWV8q6ym6+lXKAT1Urnm49YkrRKo2J8o=
-github.com/aws/smithy-go v1.9.1/go.mod h1:SObp3lf9smib00L/v3U2eAKG8FyQ7iLrJnQiAmR5n+E=
+github.com/aws/smithy-go v1.9.2-0.20220113020543-dec09760da63 h1:sczjOTbX9viYhobSMDT/vzMs1ZqJsyZPXhlkwPxd9zY=
+github.com/aws/smithy-go v1.9.2-0.20220113020543-dec09760da63/go.mod h1:SObp3lf9smib00L/v3U2eAKG8FyQ7iLrJnQiAmR5n+E=
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
 github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
--- a/config/go.mod
+++ b/config/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.3.3
 	github.com/aws/aws-sdk-go-v2/service/sso v1.8.0
 	github.com/aws/aws-sdk-go-v2/service/sts v1.13.0
-	github.com/aws/smithy-go v1.9.1
+	github.com/aws/smithy-go v1.9.2-0.20220113020543-dec09760da63
 	github.com/google/go-cmp v0.5.6
 )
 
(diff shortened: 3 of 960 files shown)


aws/defaultsmode.go
```diff
@@ -0,0 +1,95 @@
...
+	// The SDK will attempt to discover the execution environment to determine the
+	// appropriate settings automatically.
+	//
+	// Note that the auto detection is heuristics-based and does not guarantee 100%
+	// accuracy. STANDARD mode will be used if the execution environment cannot
+	// be determined. The auto detection might query EC2 Instance Metadata service
+	// (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html),
+	// which might introduce latency. Therefore we recommend choosing an explicit
+	// defaults_mode instead if startup latency is critical to your application
+	AutoDefaultsMode DefaultsMode = "auto"
```
jasdel, line 24: Nit, the name of these constants is counter to the naming pattern used in the rest of the SDK, `<type><name>`.

This would be `DefaultsModeAuto` with this pattern.

config/resolve_test.go
```diff
@@ -3,8 +3,14 @@ package config
...
 	"context"
+	"encoding/json"
+	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	"github.com/google/go-cmp/cmp"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
+	"os"
+	"strconv"
 	"testing"
```
jasdel, line 14: Nit should we all be using `goimports` to sort our import statements?  this update groups the new non-stdlib imports into the stdlib group.

config/shared_config.go
```diff
@@ -193,6 +195,19 @@ type SharedConfig struct {
 	//
 	// use_fips_endpoint=true
 	UseFIPSEndpoint aws.FIPSEndpointState
+
+	// Specifies which defaults mode should be used by services.
+	//
+	// defaults_mode=auto
+	DefaultsMode aws.DefaultsMode
```
jasdel, line 202: nit should this documentation refer to `standard` instead of `auto` since auto is experimental?

config/shared_config_test.go
```diff
@@ -510,6 +510,30 @@ func TestNewSharedConfig(t *testing.T) {
...
 		},
+		"defaults mode legacy": {
+			ConfigFilenames:      []string{testConfigFilename},
+			CredentialsFilenames: []string{testCredentialsFilename},
+			Profile:              "autodefaultsmode",
+			Expected: SharedConfig{
+				Profile:      "autodefaultsmode",
+				DefaultsMode: aws.AutoDefaultsMode,
+			},
+		},
```
jasdel, line 521: nit test case is labeled as model `legacy`, but expecting `auto` model retrieved from config.

internal/codegen/cmd/defaultsconfig/v1.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// ModifierMap is a map of configuration keys to their respective modifier descriptor.
+type ModifierMap map[string]Modifier
+
+// Modifier is a union type of either a multiply, add, or override modification descriptor.
+type Modifier struct {
+	Multiply *json.Number
+	Add      *int64
+	Override interface{}
```
jasdel, line 54: Should all of these numbers be `json.Number` instead of `interface` so they are always known type and type casting isn't needed?  Could `Override` ever be a non-number type, e.g. bool?

jasdel (commented):
Looks good, just a couple comments.

internal/codegen/cmd/defaultsconfig/v1.go
skmcgrail: According the current specification `Add` will be an int64, but I can move this to a json.Number and enforce it via a conversion to ensure it wasn't overflowed or something.

`Override` is a special case since it can be applied be a string or numerical value depending on the key the modifier is targeting.

skmcgrail: Upon further inspection we are going for `add` I was already doing what I had stated.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -193,13 +213,78 @@ private void writeRetryerResolvers(GoWriter writer) {
...
+        writer.putContext("errorf", SymbolUtils.createPointableSymbolBuilder("Errorf",
+                SmithyGoDependency.FMT).build());
+
+        writer.write("""
+                     func $resolverName:L(o *Options) {
+                         var buildable $buildableType:P
+                         
+                         if o.$optionName:L != nil {
+                             var ok bool
+                             buildable, ok = o.$optionName:L.($buildableType:P)
```
jasdel, line 250: `putContext` makes this so much nicer for large blocks!

aws/defaults/configuration.go
```diff
@@ -0,0 +1,39 @@
...
+// GetTLSNegotiationTimeout returns the TLSNegotiationTimeout value, returns false if the value is not set.
+func (c *Configuration) GetTLSNegotiationTimeout() (time.Duration, bool) {
+	if c.TLSNegotiationTimeout == nil {
+		return 0, false
+	}
+	return *c.TLSNegotiationTimeout, true
+}
+
+func duration(duration time.Duration) *time.Duration {
+	return &duration
```
jasdel, line 38: are these still needed with the new `aws.Duration`/`aws.ToDuration` helpers?

jasdel (approved):
Looks good, minor comment about extra code might not be needed anymore.

PR #1558, opened 2022-01-13
regenerate API developer guide documentation

jasdel:


docs/404.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-<meta name="generator" content="Hugo 0.91.2" />
+<meta name="generator" content="Hugo 0.82.0" />
```
skmcgrail, line 6: Looks like you may have generated this locally rather then using the docker container since this downgrade the hugo version. You will want to use the `builddocs.sh` script at the root of the branch.

skmcgrail: Also I should have updated this on the README guide but never did. :)

skmcgrail (approved):

PR #1562, opened 2022-01-15
config: Return error from optFns in LoadDefaultConfig

pingleig:
Built in LoadOptions (e.g. WithRegion) all returns nil. However `LoadOptions` is public so user can define options that returns error. I find existing code does not ignore error from resolvers or loaders and feel `LoadDefaultConfig` should do the same.

https://github.com/aws/aws-sdk-go-v2/blob/e10c0d2c8db721bd0e3b16070f3a74f6bc7171de/config/config.go#L136-L141


jasdel (approved):
Thanks for fixing this bug @pingleig. I've added changelog entry for the fix. We'll get this merged in and included in the SDKs next release.

PR #1563, opened 2022-01-17
Rely on os.UserHomeDir()

skitt:
Instead of hard-coding Windows and *nix environment variables, rely on
os.UserHomeDir() to do the right thing. Ignore errors since those only
happen in situations we don't care about (and didn't handle
previously).

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/.changelog/4f325261c632499cb3f4464799763f58.json
+++ b/.changelog/4f325261c632499cb3f4464799763f58.json
@@ -1,0 +1,8 @@
+{
+    "id": "4f325261-c632-499c-b3f4-464799763f58",
+    "type": "bugfix",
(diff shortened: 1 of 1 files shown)


jasdel (approved):
Thanks for taking the time to create this PR @skitt. The change looks good. I made a minor tweak to remove unused import, and added changelog entry.

skitt:
> Thanks for taking the time to create this PR @skitt. The change looks good. I made a minor tweak to remove unused import, and added changelog entry.

Ah yes, I missed the no-longer-necessary import! Thanks for the fixes.

skmcgrail (approved):

PR #1576, opened 2022-01-31
Add suport for Adaptive retry mode

jasdel:
Adds support for the experimental `Adaptive` retry mode to the SDK. This mode can be configured in code directly via the `aws/retry#NewAdaptiveMode` constructor, or via the SDK's shared configuration and environment variables. The adaptive retryer expands on the exponential backoff retry strategies of the `Standard` mode, adding client attempt rate limiting based on throttle responses received from an API.

The `Adaptive` retry mode is experimental and its behavior is subject to change in the future.

* `RetryMaxAttempts` - Is used to set the maximum number of attempts an SDK should attempt for an operation before giving up. Can have a positive `int` value. A value of `0` is treated as unset, and will be ignored. Will be ignored if the API client or SDK's `config` has a custom `Retryer`.
* `RetryMode` - Is used to specify the retry mode the API client will use if no custom Retryer is specified. Can be `standard`, the default, or `adaptive`.

Environment Variables:
* `AWS_MAX_ATTEMPTS=3`
* `AWS_RETRY_MODE=standard`

Shared Config:
* `max_attempts = 3`
* `retry_mode = standard`

Depends on https://github.com/aws/smithy-go/pull/352 (merged)

Fixes: https://github.com/aws/aws-sdk-go-v2/issues/1584

### DIFF
--- a/aws/retry/middleware.go
+++ b/aws/retry/middleware.go
@@ -121,6 +121,15 @@ func (r *Attempt) handleAttempt(
 		attemptResult.Err = err
 	}()
 
+	// Short circuit if this attempt never can succeed because the context is
+	// canceled. This reduces the chance of token pools being modified for
+	// attempts that will not be made
+	select {
+	case <-ctx.Done():
+		return out, attemptResult, nopRelease, ctx.Err()
+	default:
+	}
+
 	//------------------------------
 	// Get Attempt Token
 	//------------------------------


jasdel:
Fixed broken codegen

aws/config.go
```diff
@@ -38,31 +38,52 @@ type Config struct {
...
+	// Config.Retryer member is not nil. This value will be ignored if
+	// Retryer is not nil.
+	RetryMaxAttempts int
+
+	// RetryMode specifies the retry model the API client will be created with.
+	//
+	// API Clients will only use this value to construct a retryer if the
+	// Config.Retryer member is not nil. This value will be ignored if
+	// Retryer is not nil.
+	RetryMode RetryMode
```
skmcgrail, line 73: Thoughts on whether these should be grouped into a `RetryOptions` structure or should this remain stand-alone? Also it looks like this doesn't contain the wiring up of using the retry mode specified by defaults configuration. We should make sure we add the plumbing to wire-up the retryer with the retry mode specified from the SDK's configured DefaultsMode when this `RetryMode` on `aws.Config` is not set.

aws/ratelimit/token_bucket.go
```diff
@@ -44,8 +47,50 @@ func (t *TokenBucket) Refund(amount uint) {
...
+	}
+	return b
+}
+
+func uintMax(a, b uint) uint {
+	if a > b {
+		return a
 	}
+	return b
 }
```
skmcgrail, line 96: Oh generics where are you?

aws/retry/adaptive.go
```diff
@@ -0,0 +1,173 @@
...
+var DefaultThrottles = []IsErrorThrottle{
+	ThrottleErrorCode{
+		Codes: DefaultThrottleErrorCodes,
+	},
+}
+
+// AdaptiveModeOptions provides the functional options for configuring the
+// adaptive retry mode, and delay behavior.
+type AdaptiveModeOptions struct {
+	FailOnNoAttemptTokens bool
```
skmcgrail, line 30: Is this option called out in the specification, should we just rely on context cancellation behavior instead?

aws/retry/adaptive.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// AdaptiveMode provides an experimental retry strategy that expands on the
+// Standard retry strategy, adding client send rate limits. The send rate limit
+// is initially unrestricted, but becomes restricted when the attempt fails
+// with for a throttle error. Eventually unrestricted send rate limit is
+// restored once attempts no longer are failing for throttle errors.
+type AdaptiveMode struct {
+	options   AdaptiveModeOptions
+	throttles IsErrorThrottles
```
skmcgrail, line 56: What's the intention for copying throttles off of options here?

aws/retry/adaptive_test.go
```diff
@@ -0,0 +1 @@
+package retry
```
skmcgrail, line 1: Empty file?

internal/protocoltest/query/api_client.go
```diff
@@ -253,7 +266,17 @@ func resolveRetryer(o *Options) {
...
+	switch o.RetryMode {
+	case aws.RetryModeAdaptive:
+		o.Retryer = retry.NewAdaptiveMode(func(oo *retry.AdaptiveModeOptions) {
+			oo.MaxAttempts = o.RetryMaxAttempts
+		})
+	default:
+		o.Retryer = retry.NewStandard(func(oo *retry.StandardOptions) {
+			oo.MaxAttempts = o.RetryMaxAttempts
+		})
+	}
```
skmcgrail, line 279: Would need to codegen using the defaults mode retry configuration here

skmcgrail (changes requested):

aws/retry/adaptive.go
```diff
@@ -0,0 +1,173 @@
...
+
+	retryer   aws.RetryerV2
+	rateLimit *adaptiveRateLimit
+}
+
+// NewAdaptiveMode returns an initialized AdaptiveMode retry strategy.
+func NewAdaptiveMode(optFns ...func(*AdaptiveModeOptions)) *AdaptiveMode {
+	adaptiveOptions := AdaptiveModeOptions{
+		RequestCost: DefaultRequestCost,
+		Throttles:   DefaultThrottles,
```
skmcgrail, line 66: Should this be a copy as to avoid potential for mutations of the package global slice?

aws/retry/adaptive.go
```diff
@@ -0,0 +1,173 @@
...
+
+		MaxAttempts: DefaultMaxAttempts,
+		MaxBackoff:  DefaultMaxBackoff,
+		Retryables:  DefaultRetryables,
```
skmcgrail, line 70: Ditto here as well

skmcgrail (changes requested):

aws/config.go
jasdel: Yeah keep these top level since `Retryer` was already here. If `Retryer` was within some `RetryOptions` would of used that.

Will look at what overlap there is for Default configurations with retry, and get those wired up. I think this should be at the API client resolve level, with any of these specific parameters overriding any DefaultConfig options.

aws/ratelimit/token_bucket.go
jasdel: sooooon

aws/retry/adaptive.go
jasdel: This option is explicitly called out as a SDK **should** provide option to immediately fail if no attempt rate limited tokens are available. This kind of behavior is different than Context cancel, since context is generally time based at a higher level. Whereas this option is action based, and will have no impact until the client is begin throttled, and the token bucket has been drained.

I think the context cancel is a good pattern for the operation TTL overall, but work as well to limit to per attempt dynamic behavior.

aws/retry/adaptive_test.go
jasdel: Thanks missed that.

aws/retry/adaptive.go
jasdel: Will fix that up. The `options.Throttles` will be copied within the constructor, but does leave room for the global var to be duplicated in the options.  Will moves these into some kind of copy.

aws/retry/adaptive.go
```diff
@@ -0,0 +1,156 @@
...
+// present to implement Retryer interface.
+func (a *AdaptiveMode) GetInitialToken() (releaseToken func(error) error) {
+	return nopRelease
+}
+
+// GetAttemptToken returns the attempt token that can be used to rate limit
+// attempt calls. Will be used by the SDK's retry package's Attempt
+// middleware to get a attempt token prior to calling the temp and releasing
+// the attempt token after the attempt has been made.
+func (a *AdaptiveMode) GetAttemptToken(ctx context.Context) (func(error) error, error) {
```
skmcgrail, line 132: Should this short-circuit before entering the loop and check if context is already cancelled?

jasdel: This have minor impact on the behavior by not deducting from the token pool if a throttle occurred, and the application canceled the operation's context. This is a subtle difference. This would change the HTTP request send handling of the canceled context to sometimes occur higher up in the Attempt middleware's call to `GetAttemptToken`. I say sometimes because this method for `Standard` is a no-op.

The only time the context would be considered in `Adaptive` mode is when the pool does not have enough tokens. 

This raises the question should other behavior that takes `context.Context` as input also check for canceled, or only those that perform some async behavior, e.g. sleep or send request.

skmcgrail (approved):

PR #1579, opened 2022-02-02
Update code generation to depend on latest version of Smithy

jasdel:
Updates the SDK's code generation to depend on the latest version of
smithy, and smithy-go modules. This change fixes unit tests to be
compatible with the latest smithy protocol tests.

Depends on: https://github.com/aws/smithy-go/pull/351


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
```diff
@@ -300,4 +281,36 @@ private void writeJsonTypeAssertStub(GoWriter writer, Shape shape) {
         writer.openBlock("if !ok {", "}", () -> writer.write("return fmt.Errorf(\"unexpected JSON type %v\", value)"));
         writer.write("");
     }
+
+    public static final class Builder implements SmithyBuilder<JsonShapeDeserVisitor> {
```
skmcgrail, line 285: Our typical pattern has been to make the Builder class private to the top-level class, and the top level class provide a static builder method that returns a new builder. That's consistent with how I believe most/all the Smithy builders are in their implementations.

.changelog/90358f5074924d8d8b736922ebb7e586.json
```diff
@@ -0,0 +1,330 @@
+{
+    "id": "90358f50-7492-4d8d-8b73-6922ebb7e586",
```
skmcgrail, line 2: You probably don't actually want to include this changelog note since this will be part of the release. Otherwise there will be multiple changelog entries with the same information.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
jasdel: will take a look at those and duplicate that pattern

skmcgrail (approved):

Bartascnk (approved):

PR #1624, opened 2022-03-15
Ignore Retryer field when marshaling Config, fixing SA1026: trying to marshal unsupported type func()

SuperFluffy:
Applying `golangci-lint` to our codebase, it raised this lint on `aws.Config`:
```
SA1026: trying to marshal unsupported type func() github.com/aws/aws-sdk-go-v2/aws.Retryer, via x[k].AWSConfig.Retryer (staticcheck)
```
This commit ensures the field is ignored and no lints are raised when calling `json.Marshal` on it.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -84,7 +84,7 @@ type Config struct {
 	//
 	// If not nil, RetryMaxAttempts, and RetryMode will be ignored by API
 	// clients.
-	Retryer func() Retryer
+	Retryer func() Retryer `json:"-"`
 
 	// ConfigSources are the sources that were used to construct the Config.
 	// Allows for additional configuration to be loaded by clients.


jasdel:
Thanks for taking the time to create this PR @SuperFluffy. The `aws.Config` type is not intended to be json serialized. It could have additional members and types added in the future that are not compatible with JSON serialization. In addition, many members such as `APIOptions`, `Credential`, `HTTPClient`, and others have no JSON serialization equivalent.

Since the SDK's types are not intended to be JSON serialized these members do not need additional struct tags.

One way to work with this type and serialize it for your needs is to define a type you own, and copy the fields that are important from the `aws.Config` to a type you'll have control over. This allows you to filter specific fields that are important to your use case.

PR #1626, opened 2022-03-16
internal/awstesting: Update SDK custom certificate testing utils

jasdel:
Updates the SDK's custom certificate testing utilities to be compatible
with Go 1.18's deprecating of SHA1. Switch from static openssl generated
certificates to generate certificates at package initialization.

Related to https://github.com/aws/aws-sdk-go/issues/4316

### DIFF
--- a/internal/awstesting/certificate_utils.go
+++ b/internal/awstesting/certificate_utils.go
@@ -20,16 +20,20 @@ import (
 )
 
 var (
-	// CA PEM
+	// TLSBundleCA is the CA PEM
 	TLSBundleCA []byte
 
-	// Server PEM
+	// TLSBundleCert is the Server PEM
 	TLSBundleCert []byte
-	TLSBundleKey  []byte
 
-	// Client PEM
+	// TLSBundleKey is the Server private key PEM
+	TLSBundleKey []byte
+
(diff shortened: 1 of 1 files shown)


internal/awstesting/certificate_utils.go
```diff
@@ -0,0 +1,291 @@
...
+	if ok := server.TLS.ClientCAs.AppendCertsFromPEM(certPem); !ok {
+		return nil, fmt.Errorf("failed to append client certs")
+	}
+
+	return server, nil
+}
+
+// CreateClientTLSCertFiles returns a set of temporary files for the client
+// certificate and key files.
+func CreateClientTLSCertFiles() (cert, key string, err error) {
```
skmcgrail, line 184: Do these actually need to be written out to disk?

jasdel: For the v1 SDK yes since it supports the [AWS_SDK_GO_CLIENT_TLS_KEY](https://pkg.go.dev/github.com/aws/aws-sdk-go/aws/session#hdr-Custom_Client_TLS_Certificate) environment variables. Though it doesn't look like the v2 SDK has this behavior for some reason. Will have to investigate why that is missing.

PR #1627, opened 2022-03-17
Support for custom time formats for encoding and decoding DynamoDB AttributeValues

### LINKED ISSUE #1628: Allow Encode/Decode custom time formats in DynamoDB
### Describe the feature

When saving `time.Time` fields into dynamo db it defaults to format `time.RFC3339Nano`. It would be helpful to make this configurable via EncodingOptions

### Is your Feature Request related to a problem?

In my use case I'd like to save the time fields formatted with milliseconds keeping the trailing zeroes to keep sorting correct, 

e.g.: I want to save`2022-03-17T09:46:11.100Z` rather than the default `2022-03-17T09:46:11.1Z`.

In Go I could format a time with the following format to achieve that: `"2006-01-02T15:04:05.000Z07:00"` but currently I cannot use `MarshalWithOptions` with my custom format.

### Proposed Solution

https://github.com/aws/aws-sdk-go-v2/pull/1627

### Describe alternatives you've considered

currently I'm doing something like this:
```
itemValues, err := attributevalue.MarshalMap(item)
if err != nil {
   return fmt.Errorf("marshalling item: %w", err)
}
itemValues["createdAt"] = &types.AttributeValueMemberS{Value: item.CreatedAt.Format("2006-01-02T15:04:05.000Z07:00")}
```

### Acknowledge

- [x] I may be able to implement this feature request

### AWS Go SDK version used

v1.43.20

### Go version used

go version go1.18 darwin/amd64

plopezlpz:
Adds a new `TimeFormat` decode and encode option
to override the default `time.RFC3339Nano`

resolves #1628

### DIFF
--- a/.changelog/d10ed36c72e84ccaa5befe7975f10af6.json
+++ b/.changelog/d10ed36c72e84ccaa5befe7975f10af6.json
@@ -1,0 +1,9 @@
+{
+    "id": "d10ed36c-72e8-4cca-a5be-fe7975f10af6",
+    "type": "feature",
+    "description": "Support has been added for specifying a custom time format when encoding and decoding DynamoDB AttributeValues. Use `EncoderOptions.EncodeTime` to specify a custom time encoding function, and use `DecoderOptions.DecodeTime` for specifying how to handle the corresponding AttributeValues using the format. Thank you [Pablo Lopez](https://github.com/plopezlpz) for this contribution.",
+    "modules": [
+        "feature/dynamodb/attributevalue",
+        "feature/dynamodbstreams/attributevalue"
+    ]
+}


jasdel (changes requested):
Thanks for taking the time to create this PR @plopezlpz. We should be able to take this feature to customize (de)serialization of `time.Time` members.

In order to take this change I recomend chaning the `TimeFormat` option from a string to a function. As a function the user can control much more granularity how the time is parsed, and potentially provide support for parsing multiple formats. As a string, only a very ridged format could be supported.

To do this I think `EncodeOptions.TimeFormat` would be a member of a function type with a signature like the following. (Where `ddbtypes` is alias for `github.com/aws/aws-sdk-go-v2/service/dynamodb/types`).

```go
TimeFormat func (time.Time) (ddbtypes.AttributeValue, error)
```

The `DecodeOptions.TimeFormat` would be similar member function type for deserialization.

```go
TimeFormat func (ddbtypes.AttributeValue) (time.Time, error)
```

With this pattern members of time.Time could also target other AttributeValue members such as Number.

jasdel:
We should also consider renaming `TimeFormat` members to be specific to their usecase. e.g. `FormatTime` for encoding, and `ParseTime` for decoding workflows.

jasdel:
The `feature/dynamodb/attributevalue` package could also include helper utilities that provide `MarshalTime`/`UnmarshalTime` implementation for static time format strings. Either as standalone functions or a type with method for encoding the time value.

```go
type StaticStringTimeFormat struct {
    Format string
}

func (s StaticStringTimeFormat) MarshalAttributeValueTime(t time.Time) (ddbtypes.AttributeValue, error) {
     v, err := t.Format(s.Format)
     if err != nil {
          return nil, fmt.Errorf("failed to marshal attribute value time, %w", err)
     }
     return &ddbtypes.AttributeValueMemberS{Value: v}, nil
}

func (s StaticStringTimeFormat) UnmarshalAttributeValueTime(av ddbtypes.AttributeValue) (time.Time, error) {
     avS, ok := av.(*ddbtypes.AttributeValueMemberS)
     if !ok {
          reutrn time.Time{}, fmt.Errorf("failed to unmarshal attribute value time, expect %T type, got %T, avS, av)
     }

     t, err := time.Parse(s.Format)
     if err != nil {
          return nil, fmt.Errorf("failed to unmarshal attribute value time, %w", err)
     }
     return t, nil
}
```

plopezlpz:
@jasdel The encoding function is easy to do with that signature but the decoding is challenging with this signature:
`ParseTime func (ddbtypes.AttributeValue) (time.Time, error)`

Parsing currently is done inside functions that accept `string`s not `ddbtypes.AttributeValue`s (i.e.: `decodeString` and `decodeNumber`)

Number Sets and String Sets will eventually call those two functions as well. In the case of sets it would not make sense to return a single `time.Time` but a slice instead:
I mean, it does not make sense to create `PaseTime` function like this:
```
func PaseTime(av ddbtypes.AttributeValue) (time.Time, error) {
     avSS, ok := av.(*ddbtypes.AttributeValueMemberSS)
     // ... eventually return a time.Time does not make sense for sets
}
```

This signature would make more sense for the decoder, the encoder is fine:
`ParseTime func (string) (time.Time, error)`
So that it can be called from inside `decodeString` and `decodeNumber`

plopezlpz:
@jasdel
The best I could come up with is a decoding function like this:
```
		DecodeTime: func(av types.AttributeValue) (time.Time, error) {
			switch val := av.(type) {
			case *types.AttributeValueMemberS:
				t, err := time.Parse(time.RFC3339, val.Value)
				if err != nil {
					return time.Time{}, &UnmarshalError{Err: err, Value: val.Value, Type: timeType}
				}
				return t, nil
			case *types.AttributeValueMemberN:
				v, err := strconv.ParseInt(val.Value, 10, 64)
				if err != nil {
					return time.Time{}, &UnmarshalError{Err: err, Value: val.Value, Type: timeType}
				}
				return time.Unix(v, 0), nil
			default:
				return time.Time{}, fmt.Errorf("failed to unmarshal attribute value time, expect %T or %T type, got %T", &types.AttributeValueMemberS{}, &types.AttributeValueMemberN{}, av)
			}
		},
```

That will be called from `decodeString` and `decodeNumber` problem is that it is a pain to write a custom function because it will need to handle both string and number ddbtypes. If you just want to decode number times you will have to copy/paste the default string decoding code inside your function.

Another option would be to provide multiple decode functions for multiple ddb types:
`DecodeTimeS func (ddbtypes.AttributeValue) (time.Time, error)`
`DecodeTimeN func (ddbtypes.AttributeValue) (time.Time, error)`
The counterpart encoding function could be called:
`EncodeTime func (time.Time) (ddbtypes.AttributeValue, error)`

plopezlpz:
@jasdel I've changed the implementation trying to follow your suggestions as close as I could. Feel free to provide further feedback or change the code yourself (sometimes it is faster/more practical to change the code oneself than giving PR feedback).
Thanks a lot for your input.

jasdel:
Taking the time to update the PR. We'll review the update and get back with you.

I see the reasoning for using separate options for Number vs String decoding, but this does make it asymmetric compared to Encoding which has a single option. This may be the best pattern though.

feature/dynamodb/attributevalue/decode_test.go
```diff
@@ -797,6 +797,148 @@ func (t *testUnmarshalMapKeyComplex) UnmarshalDynamoDBAttributeValue(av types.At
...
+			if err := UnmarshalWithOptions(inputMap, &actualValue, func(options *DecoderOptions) {
+				if c.decodeTimeS != nil {
+					options.DecodeTimeS = c.decodeTimeS
+				}
+			}); err != nil {
+				t.Errorf("expect no error, got %v", err)
+			}
+			if !reflect.DeepEqual(expectedValue, actualValue) {
+				t.Errorf("expect %+v, got %+v", expectedValue, actualValue)
+			}
```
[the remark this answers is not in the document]
plopezlpz: done

jasdel (approved):
Thanks for taking the time to update this PR with feedback. The change looks good from my perspective. Getting an additional set of eyes from @skmcgrail to get his thoughts.

Only part that sticks out is the `DecodeTimeS` and `DecodeTimeN`, but I'm not seeing an obvious alternative to this given how Decode is structured.

feature/dynamodb/attributevalue/decode.go
```diff
@@ -212,6 +212,16 @@ type DecoderOptions struct {
...
+
+	// Will decode S attribute values and SS attribute value elements into time.Time
+	//
+	// Default string parsing format is time.RFC3339
+	DecodeTimeS func(string) (time.Time, error)
+
+	// Will decode N attribute values and NS attribute value elements into time.Time
+	//
+	// Default number parsing format is seconds since January 1, 1970 UTC
+	DecodeTimeN func(string) (time.Time, error)
```
skmcgrail, line 224: What are your thoughts about putting these on a separate structure to group them for discoverability?

```go
type DecodeTimeAttributes struct {
    S func(string) (time.Time, error)
    N func(string) (time.Time, error)
}
```

This would allow for easy extension for adding `SS` and `NS` attributes down the road if we want to, but without having to have an explosion of top-level fields on the `DecoderOptions` structure. As on the top level structure we would have `DecodeTime DecodeTimeAttributes`.

plopezlpz: done

skmcgrail:
Thanks for the updates @plopezlpz , we will work on getting this merged for release this week.

jasdel (approved):

skmcgrail (approved):

skmcgrail:
This went out as part of today's release. Thanks again for the contribution @plopezlpz 

https://github.com/aws/aws-sdk-go-v2/releases/tag/release-2022-04-15

PR #1659, opened 2022-04-04
Fix sqs receive message input

shamaton:
Issue: https://github.com/aws/aws-sdk/issues/275


jasdel:
Thanks for reaching out and creating this PR @shamaton. This is a change we cannot accept. At least not via PR. This seems to be an issue with the API model's definition, which is not owned by the SDK team, but consumed from an upstream source.

Lets use https://github.com/aws/aws-sdk-go-v2/issues/1660 to tracking and upstreaming the issue to the service team.

PR #1664, opened 2022-04-09
local-mod-replace.sh: fix shebang

mlloreda:
- Additionally double-quoted variable to prevent globbing and word splitting


skmcgrail:
Looks good, thanks for the fixes @mlloreda

skmcgrail (approved):

PR #1681, opened 2022-04-23
set default value for ExpiredObjectDeleteMarker

prembhaskal:
When ExpiredObjectDeleteMarker is empty, the parsing fails.
Few S3 storage eg. ODF/Noobaa are returning empty tag "`<ExpiredObjectDeleteMarker></ExpiredObjectDeleteMarker>`" causing parsing to fail.

### DIFF
--- a/service/s3/deserializers.go
+++ b/service/s3/deserializers.go
@@ -16887,7 +16887,7 @@ func awsRestxml_deserializeDocumentLifecycleExpiration(v **types.LifecycleExpira
 			if err != nil {
 				return err
 			}
-			if val == nil {
+			if val == nil  || len(val) == 0 {
 				break
 			}
 			{


jasdel (changes requested):
Thanks for reaching out with this issue. We cannot take this change as it is because it modifies the SDK's generated code. In addition, could you provide details about the error that you're seeing parsing the `ExpiredObjectDeleteMarker` member? It probably will be better to close this PR and create a [bug issue](https://github.com/aws/aws-sdk-go-v2/issues/new?assignees=&labels=bug%2Cneeds-triage&template=bug-report.yml&title=%28short+issue+description%29) instead to investigate the issue you're seeing.

prembhaskal:
Oh k. I didn't realise this is generated code. Yes i will create a issue with the results.

PR #1682, opened 2022-04-25
Fully populates `aws.Config` before calling `resolveCredentials`

### LINKED ISSUE #1398: Disabling EC2 IMDS Client in `config.LoadDefaultConfig()` does not disable EC2 IMDS Client
  
**Describe the bug**
Setting the `EC2IMDSClientEnableState` to `imds.ClientDisabled` does not disable the EC2 IMDS Client when resolving configuration using `config.LoadDefaultConfig()`.

Setting `AWS_EC2_METADATA_DISABLED = true` does correctly disable the EC2 IMDS Client.

**Version of AWS SDK for Go v2?**

* github.com/aws/aws-sdk-go-v2 v1.9.0
* github.com/aws/aws-sdk-go-v2/config v1.7.0
* github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.5.0

go1.16 darwin/amd64

The following code

```go
imdsEnableState := imds.ClientDefaultEnableState
if c.SkipMetadataApiCheck {
	imdsEnableState = imds.ClientDisabled
}

...

cfg, err := config.LoadDefaultConfig(ctx,
	...
	config.WithEC2IMDSClientEnableState(imdsEnableState),
)

_, err = cfg.Credentials.Retrieve(ctx)
if err != nil {
	return cfg, fmt.Errorf("retrieving credentials: %w", err)
}
```

returns the following error

> retrieving credentials: no EC2 IMDS role found, operation error ec2imds: GetMetadata, canceled, context deadline exceeded

If the environment variable `AWS_EC2_METADATA_DISABLED` is set as follows

```go
imdsEnableState := imds.ClientDefaultEnableState
if c.SkipMetadataApiCheck {
	imdsEnableState = imds.ClientDisabled
	os.Setenv("AWS_EC2_METADATA_DISABLED", "true")
}
```

the following is returned

> retrieving credentials: no EC2 IMDS role found, operation error ec2imds: GetMetadata, access disabled to EC2 IMDS via client option, or "AWS_EC2_METADATA_DISABLED" environment variable

**Expected behavior**

The EC2 IMDS Client should not be called when `config.WithEC2IMDSClientEnableState(imds.ClientDisabled)` is set.

gdavison:
Setting the `EC2IMDSClientEnableState` to `imds.ClientDisabled` does not disable the EC2 IMDS Client when resolving configuration using `config.LoadDefaultConfig()`.

Setting `AWS_EC2_METADATA_DISABLED = true` does correctly disable the EC2 IMDS Client.

The `awsConfigResolver` functions take the parameters `(ctx context.Context, cfg *aws.Config, configs configs)` where `cfg` has no `ConfigSources` set, and the config providers are passed in `configs`.

This works until the resolvers need to create a service client, such as `imds.Client`. The clients are created using their `NewFromConfig` function. The `aws.Config` with zero `ConfigSources` is then passed in, ignoring the config providers in `configs`.

This PR *works*, but doesn't make any deep changes. For instance, either of the following could be done:

1. Since `resolveCredentials ` is no longer called in a loop with other `awsConfigResolver` functions, it doesn't need to match the interface, and could replace uses of `config` with the `ConfigSources` on the `aws.Config`
2. Change how the `aws.Config` is populated in the loop of `awsConfigResolver` functions. This could also allow the use of  the `ConfigSources` on the `aws.Config` instead of the `config` parameter.

Closes #1398

### DIFF
--- a/config/resolve_credentials_test.go
+++ b/config/resolve_credentials_test.go
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
 	"github.com/aws/aws-sdk-go-v2/service/sso"
@@ -478,11 +477,11 @@ func TestResolveCredentialsCacheOptions(t *testing.T) {
 
 func TestResolveCredentialsIMDSClient(t *testing.T) {
 	expectEnabled := func(t *testing.T, err error) {
-		var re *retry.MaxAttemptsError
-		if !(err == nil || // If running in EC2, there will be no error
-			errors.As(err, &re) || // When not running in EC2, the IMDS call can either reach max attempts
-			errors.Is(err, context.DeadlineExceeded)) { // or reach the context deadline
-			t.Fatalf("unexpected error: %v", err)
+		if err == nil {
+			t.Fatalf("expect error got none")
+		}
+		if e, a := "expected HTTP client error", err.Error(); !strings.Contains(a, e) {
+			t.Fatalf("expected %v error in %v", e, a)
 		}
 	}
 
@@ -508,7 +507,7 @@ func TestResolveCredentialsIMDSClient(t *testing.T) {
 		expectedState imds.ClientEnableState
 		expectedError func(*testing.T, error)
 	}{
-		"nothing": {
+		"default no options": {
 			expectedState: imds.ClientDefaultEnableState,
 			expectedError: expectEnabled,
 		},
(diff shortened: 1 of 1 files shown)


gdavison:
I'm not able to reproduce the test errors

jasdel (approved):
Thanks for taking the time to create this PR @gdavison. While reviewing the change I noticed the `ConfigSources` were being after before the config resolvers were being processed. This was incorrect. By moving assigning ConfigSources in `resolveDefaultAWSConfig`, it ensures the config sources are available for all downstream config resolvers, such as `resolveCredentials`. This also allows `resolvedCredentials` to stay in the set of resolvers, without being special cased.

jasdel:
Not sure why unit tests are failing. I'm not able to reproduce that locally. They seem to only fail in GitHub actions.

jasdel:
I think the reason the tests were failing in GitHub actions but not locally, must have something to do with environment. I updated the tests to use stub HTTP clients instead of real clients. This should prevent the issue.

gdavison:
One twist to this approach is that the subsequent `awsConfigResolver` functions no longer need the `configs` parameter, since the configs are now contained in the `aws.Config`. That was what I meant by my option 2 above

skmcgrail (approved):

jasdel:
Thanks for the feedback @gdavison. There is a subtle usage difference between `ConfigSources` in `aws.Config` and `configs` used by the resolvers. `configs` where the original set of configuration providers the config resolvers will search through when initializing the SDK's generic config. Whereas `ConfigSources` was added later in order to provide a way for API client's to lazy load configuration data without exposing the API client specific fields in `aws.Config` directly. 

These two values do overlap currently, but the intention is that `ConfigSources` could be appended with additional configuration sources by resolvers to be used by API clients. Whereas the set `configs` are only intended to be the initial set of configuration providers such as shared config, environment variables, and LoadOptions.

PR #1683, opened 2022-04-26
service/s3: Update hostCompatibleBucketName doc to reflect #848 changes

lggomez:
Updated the comment for the method as S3ForcePathStyle was deprecated and at a first glance it is not obvious why the default behavior changes when migrating from SDK v1 to V2 and how to enforce the endpoint to not append the bucket name. This includes #848 

PS: I saw the PR template about opening an issue first but it wasn't clear which files are autogenerated and which ones are not (if everything under feature/* falls into this category). If the former is needed please tell me


jasdel (approved):
Thanks for updating the documentation for this behavior. I've made a minor tweak to the change, and adding link to the member driving this behavior.

PR #1700, opened 2022-05-11
Fix typo github.coma instead of github.com

bpaquet:


jasdel:
Looks like this Pr is duplicated by https://github.com/aws/aws-sdk-go-v2/pull/1709, we can pull the fix in with that change.

PR #1702, opened 2022-05-13
fix feature/dynamodb/expression/condition.go Or documentation

yeehenchen:
Fix the documentation to `Or` function.


jasdel (approved):
Thanks for updating the documentation for the expressions package. The change looks good, and we'll include it in our next release.

PR #1711, opened 2022-05-16
Fix iotsecuretunneling and mobile API clients to use correct signing name

jasdel:
Fixes iotsecuretunneling and mobile API clients to use the correct name for signing requests. The models for these APIs contained the wrong name. This PR fixes that, and regenerates the API clients. 

Fixes [#1686](https://github.com/aws/aws-sdk-go-v2/issues/1686).

### DIFF
--- a/.changelog/e4cddb5c443e4c7ab265dd02f294fbde.json
+++ b/.changelog/e4cddb5c443e4c7ab265dd02f294fbde.json
@@ -1,0 +1,9 @@
+{
+    "id": "e4cddb5c-443e-4c7a-b265-dd02f294fbde",
+    "type": "bugfix",
+    "description": "Fixes iotsecuretunneling and mobile API clients to use the correct name for signing requests, Fixes [#1686](https://github.com/aws/aws-sdk-go-v2/issues/1686).",
+    "modules": [
+        "service/iotsecuretunneling",
+        "service/mobile"
+    ]
+}


codegen/sdk-codegen/aws-models/lex-runtime-v2.json
```diff
@@ -42,7 +42,10 @@
                 "aws.auth#sigv4": {
                     "name": "lex"
                 },
-                "aws.protocols#restJson1": {},
+                "aws.protocols#restJson1": {
+                    "http": ["http/1.1", "h2"],
+                    "eventStreamHttp": ["h2"]
```
skmcgrail, line 47: These are new to the trait, wonder if there is any guidance on how we should use them.

skmcgrail (approved):

codegen/sdk-codegen/aws-models/lex-runtime-v2.json
jasdel: Looks like they are [documented on the protocols](https://awslabs.github.io/smithy/1.0/spec/aws/aws-restjson1-protocol.html?highlight=http#aws-protocols-restjson1-trait). Basically the priority order of the HTTP protocol version used. `eventStreamHttp` basically stating only h2 should be used. The v1 SDK models had similar ideas, but I think it was in the `metadata` section. The AWS SDK for Go never really used these traits since h2 is handled so transparently by the HTTP client.

> Each entry in http and eventStreamHttp SHOULD be a valid [Application-Layer Protocol Negotiation (ALPN) Protocol ID](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids) (for example, http/1.1, h2, etc). Clients SHOULD pick the first protocol in the list they understand when connecting to a service. A client SHOULD assume that a service supports http/1.1 when no http or eventStreamHttp values are provided.

Issue #1732, opened 2022-06-16
emrcontainers.DescribeJobRun deserialization failed


lgiuliani-mwb:
### Describe the bug

Deserialization for the `emrcontainers.DescribeJobRun` call fails with `operation error EMR containers: DescribeJobRun, https response error StatusCode: 200, RequestID: <REDACTED>, deserialization failed, failed to decode response body with invalid JSON, expected Date to be a JSON Number, got string instead`

### Expected Behavior

The type of the response to be what it's advertised in the documentation.

### Current Behavior

I can see `"createdAt" : "2022-06-16T13:38:25Z",` and `"finishedAt" : "2022-06-16T14:39:29Z",` returned as strings while the deserializer expects a `JSON Number`.

### Reproduction Steps

Sample program:

```go
package main

import (
	"context"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/emrcontainers"
	"github.com/op/go-logging"
)

var log = logging.MustGetLogger("emr-tool")

func main() {
	ctx := context.Background()
	awsConfig, err := config.LoadDefaultConfig(ctx)

	client := emrcontainers.NewFromConfig(awsConfig)

	jobInfo, err := client.DescribeJobRun(ctx, &emrcontainers.DescribeJobRunInput{
		Id:               aws.String("REDACTED"),
		VirtualClusterId: aws.String("REDACTED"),
	}, func(o *emrcontainers.Options) {
		o.ClientLogMode = aws.LogRequestWithBody | aws.LogResponseWithBody
	})
	if err != nil {
		log.Errorf("emrcontainers.DescribeJobRun %s", err.Error())
	}

	log.Infof("result: %s", jobInfo)
}

```

### Possible Solution

Fix the AWS documentation and implement the right data types in the deserializator.

### AWS Go SDK V2 Module Versions Used

go mod graph:

```github.com/lgiuliani-mwb/scripts/emr_tools github.com/aws/aws-sdk-go-v2@v1.16.5
github.com/lgiuliani-mwb/scripts/emr_tools github.com/aws/aws-sdk-go-v2/config@v1.15.10
github.com/lgiuliani-mwb/scripts/emr_tools github.com/aws/aws-sdk-go-v2/service/emrcontainers@v1.13.6
[...]

PR #1737, opened 2022-06-24
service/emrcontainers: Fixes bug with incorrect modeled timestamp format

jasdel:
Updates the `emr-containers` model to fixing the incorrectly modeled
timestamp's format to be a date-time instead of the default Unix epoch.

Fixes: https://github.com/aws/aws-sdk-go-v2/issues/1732


jasdel:
When PR https://github.com/aws/aws-sdk-go-v2/pull/1738 is merged, Go 1.18 and 1.17 will start passing.

PR #1739, opened 2022-06-24
Remove retracted service module source

jasdel:
Removes the retracted service module source.


skmcgrail (approved):

jasdel:
Ignoring license scan failure as its a false negative on module being removed.

PR #1748, opened 2022-06-30
service/sqs: Add support for validating message checksums by default

jasdel:
Adds support for the Amazon SQS client to automatically validate message checksums for `SendMessage`, `SendMessageBatch`, and `ReceiveMessage`. This brings the v2 SDK up to speed with the v1 SDK's behavior.

A `DisableMessageChecksumValidation` parameter has been added to the Options struct for the `sqs` package. Setting this to true will disable the checksum validation. This can be set when creating a client, or per operation call.

Based on the [v1 SDK's SQS customization](https://github.com/aws/aws-sdk-go/blob/main/service/sqs/customizations.go)

### DIFF
--- a/service/sqs/cust_checksum_validation.go
+++ b/service/sqs/cust_checksum_validation.go
@@ -12,13 +12,16 @@ import (
 	"github.com/aws/smithy-go/middleware"
 )
 
-//********************************
-// SendMessage checksum validation
-//********************************
+// addValidateSendMessageChecksum adds the ValidateMessageChecksum middleware
+// to the stack configured for the SendMessage Operation.
 func addValidateSendMessageChecksum(stack *middleware.Stack, o Options) error {
 	return addValidateMessageChecksum(stack, o, validateSendMessageChecksum)
 }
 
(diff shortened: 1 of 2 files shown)


service/sqs/cust_checksum_validation.go
```diff
@@ -0,0 +1,214 @@
...
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	sqstypes "github.com/aws/aws-sdk-go-v2/service/sqs/types"
+	"github.com/aws/smithy-go/middleware"
+)
+
+//********************************
+// SendMessage checksum validation
+//********************************
```
skmcgrail, line 17: minor nit on the styling of the comments. Even if its an internal function should we stick to the typical expected formatting as if it was a public doc string?

skmcgrail (approved):

service/sqs/cust_checksum_validation.go
jasdel: Will update this. Used these as section separators, but regular doc string probably are better

skmcgrail (approved):

PR #1751, opened 2022-07-01
service/appsync: Fix broken API client operation URL paths

jasdel:
Regenerates the appsync API client with the latest version of the model. This update fixes the incorrectly modeled URL path. The incorrect URL path prevented operation calls from being made.


kstich (approved):

jasdel:
Changes already synced, PR no longer needed.

PR #1766, opened 2022-07-14
Fix codegen bug for header-only events

skmcgrail:
* Fixes a bug where the event message serializers and deserializers for event streams would be missing the function return statement if the event only contained headers.

Do not impact existing services, as they don't currently have models that exercise this capability.

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEventStreamUtils.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEventStreamUtils.java
@@ -1296,7 +1296,7 @@ public final class AwsEventStreamUtils {
                     if (!headerBindings.isEmpty() || payloadBinding.isPresent()) {
                         for (var headerBinding : headerBindings) {
                             new HeaderShapeSerVisitor(writer, model, headerBinding, "msg",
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEventStreamUtils.java
```diff
@@ -1878,7 +1882,7 @@ protected Void getDefault(Shape shape) {
 
         private void writeSetter(String valueTypeSymbolName) {
             var ds = CodegenUtils.getAsValueIfDereferencable(pointableIndex, memberShape, dataSource);
-            writer.write("$L.Set($S, $T($L))", target, headerName,
+            writer.write("$L.Headers.Set($S, $T($L))", target, headerName,
```
bkaws, line 1885: Is this still valid if the event didn't only contain headers?

skmcgrail: This is valid, this snippet of code actually comes from a helper class within the `AwsEventStreamUtils` class that is responsible for setting headers.

Otherwise there is separate code after using this utility class that will handle the message payload.

RanVaknin (approved):
Looks good to me 😀

Purnaak:
Taking a look

PR #1779, opened 2022-07-29
Fix go1.19 vet failures

skmcgrail:
Fixes Go 1.19 vet failures due to the following:
> The vet checker “errorsas” now reports when [errors.As](https://tip.golang.org/pkg/errors/#As) is called with a second argument of type *error, a common mistake.


Purnaak (approved):
Looks good for me

PR #1791, opened 2022-08-05
feature/s3/manager: Suppress WriterAt which can bypass Windows optimizations.

skmcgrail:
Fixes issue with Go 1.19 discovered in the S3 transfer manager's windows upload and downloader buffer pools. It looks like the update to Go 1.19 changes the code path that is used by the pool because `io.NopCloser` was updated to forward `WriterTo` if the underlying `Writer` implemented it. Looks likes we'd made some assumptions about this behavior not being present.  This PR suppresses the `WriterTo` of the buffer pool to prevent the unexpected usage.

### DIFF
--- a/feature/s3/manager/download.go
+++ b/feature/s3/manager/download.go
@@ -100,7 +100,8 @@ func WithDownloaderClientOptions(opts ...func(*s3.Options)) func(*Downloader) {
 // interface.
 //
 // Example:
-// 	// Load AWS Config
+//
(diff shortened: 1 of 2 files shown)


jasdel (approved):

cgang:
It's a little confusion that title & code suggest that this commit is suppressing **WriterAt**, while above comment suggest suppressing **WriterTo**, could you please clarify which one is correct? @skmcgrail

PR #1793, opened 2022-08-08
feature/dynamodb/attribute: Add String method for UnixTime

jasdel:
Adds a `String` method to `UnixTime`, so that when structs with this field get logged it prints a human readable time.

v2 port of https://github.com/aws/aws-sdk-go/pull/4504


skmcgrail (approved):

jasdel:
🥳

PR #1799, opened 2022-08-11
ci: correct deprecated Go version in deprecated-windows-tests

Juneezee:
A small typo in `go-version` of `deprecated-windows-tests`, it should be `1.17`, not `1.19`.


jasdel (approved):
Thanks for catching this and correcting the ci test cases.

Issue #1809, opened 2022-08-19
Invalid field type for ReceiveMessageInput.AttributeNames in api_op_ReceiveMessage


anan44:
### Describe the bug

AttributeNames type set as []types.QueueAttributeName which is not in align with type requirements of actual AWS API.

### Expected Behavior

AttributeNames type should be []types.MessageSystemAttributeName which contains all the necessary input values described in the documentation & doc strings

### Current Behavior

AttributeNames type is []types.QueueAttributeName which contains wrong values. It is still currently possible to use the functionality, but it requires not using pre-provided enums, since they are incorrect. 

### Reproduction Steps

There is no real way to Reproduce this. The input type is just incorrect and does not align with the required field values.

As you can see here:

https://github.com/aws/aws-sdk-go-v2/blob/8c989d2921563236cbc7f672b39464917911eeab/service/sqs/api_op_ReceiveMessage.go#L128

it is listed that following values are accepted:

```Golang
	// A list of attributes that need to be returned along with each message. These
	// attributes include:
	//
	// * All – Returns all values.
	//
	// *
	// ApproximateFirstReceiveTimestamp – Returns the time the message was first
	// received from the queue (epoch time (http://en.wikipedia.org/wiki/Unix_time) in
	// milliseconds).
	//
	// * ApproximateReceiveCount – Returns the number of times a
	// message has been received across all queues but not deleted.
	//
	// * AWSTraceHeader –
	// Returns the X-Ray trace header string.
	//
	// * SenderId
	//
	// * For an IAM user, returns
	// the IAM user ID, for example ABCDEFGHI1JKLMNOPQ23R.
	//
	// * For an IAM role, returns
	// the IAM role ID, for example ABCDE1F2GH3I4JK5LMNOP:i-a123b456.
	//
	// * SentTimestamp
	// – Returns the time the message was sent to the queue (epoch time
	// (http://en.wikipedia.org/wiki/Unix_time) in milliseconds).
	//
	// *
	// SqsManagedSseEnabled – Enables server-side queue encryption using SQS owned
	// encryption keys. Only one server-side encryption option is supported per queue
	// (e.g. SSE-KMS
	// (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-configure-sse-existing-queue.html)
	// or SSE-SQS
[...]

PR #1810, opened 2022-08-19
fixed but bug with AttributeNames field type

anan44:
# Issue

https://github.com/aws/aws-sdk/issues/337

# Solution

Fixe the invalid field type to correct type: []types.MessageSystemAttributeName


jasdel:
Thanks for taking the time to create this PR @anan44. We cannot take this change directly as it modifies code generated by the SDK. In addition, the source of these parameters are defined in a model owned by the Amazon SQS team. This issue will need to be raised with them to determine the correct fix. We can use #1809 for further discussion.

Issue #1825, opened 2022-08-31
DynamoDB returns operation error `use of closed network`


satorunooshie:
### Describe the bug

SDK often returns the error below after updated SDK version.

acutal error: 
```
operation error DynamoDB: Query, $tableName: https response error StatusCode: 200, RequestID: 123456, deserialization failed, failed to decode response body, read tcp $src->$dst: use of closed network connection
```

### Expected Behavior

No errors

### Current Behavior

```
operation error DynamoDB: Query, $tableName: https response error StatusCode: 200, RequestID: 123456, deserialization failed, failed to decode response body, read tcp $src->$dst: use of closed network connection
```

### Reproduction Steps

```go
	ddb := dynamodb.NewFromConfig(defaultCfg)
	result, err := ddb.Query(
		ctx,
		&dynamodb.QueryInput{
			TableName:                 aws.String(tableName),
			ExpressionAttributeNames:  expr.Names(),
			ExpressionAttributeValues: expr.Values(),
			KeyConditionExpression:    expr.KeyCondition(),
		},
	)
```

### AWS Go SDK V2 Module Versions Used

```
	github.com/aws/aws-sdk-go-v2 v1.16.8
	github.com/aws/aws-sdk-go-v2/config v1.15.15
	github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue v1.9.8
	github.com/aws/aws-sdk-go-v2/feature/dynamodb/expression v1.4.14
	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.15.10
	github.com/aws/smithy-go v1.12.0
```

### Compiler and Version used

1.19

### Operating System and version

Linux

RanVaknin:
Hi @satorunooshie ,

I saw that you opened 2 issues but Im going to answer them both here:

The reason you are seeing a `status 200` followed by a serialization error is that the DynamoDB server is able to send back a valid response, and before the client is finished reading the body of the successful response, the connection errors out. This can happen for many many reasons and is difficult to pinpoint especially when the pattern is inconsistent like in your case. 

Regarding your other ticket and PR:
We cannot accept your PR to override the Retryer behavior because not every API operation may be idempotent. The service API models don't have information to determine if an operation is idempotent, thus the SDKs cannot assume if it's safe to retry.

**Proposed solution:**
You can implement your own Retryer on the client level, or on operation level and configure it to retry when the connection fails.

[...]

PR #1826, opened 2022-08-31
aws/retry: Allows `use of closed network connection` to be retryable

satorunooshie:
handled the `use of closed network connection` error as same as https://github.com/aws/aws-sdk-go.
ref: https://github.com/aws/aws-sdk-go/pull/3476

https://github.com/aws/aws-sdk-go-v2/issues/1825

### DIFF
--- a/aws/retry/retryable_error.go
+++ b/aws/retry/retryable_error.go
@@ -100,7 +100,8 @@ func (r RetryableConnectionError) IsErrorRetryable(err error) aws.Ternary {
 	case errors.As(err, &conErr) && conErr.ConnectionError():
 		retryable = true
 
-	case strings.Contains(err.Error(), "connection reset"):
+	case strings.Contains(err.Error(), "connection reset"),
+		strings.Contains(err.Error(), "use of closed network connection"):
 		retryable = true
 
 	case errors.As(err, &urlErr):


RanVaknin:
We cannot accept this PR to override the Retryer behavior because not every API operation may be idempotent. The service API models don't have information to determine if an operation is idempotent, thus the SDKs cannot assume if it's safe to retry.

**Proposed solution:**
Just take your solution and implement in on the client level instead of the SDK level by overriding the default implementation of `RetryableConnectionError`

```go
// your implementaion:
type RetryableConnectionError2 struct{}

func (r RetryableConnectionError2) IsErrorRetryable(err error) aws.Ternary {
	if err == nil {
		return aws.UnknownTernary
	}
	var retryable bool
	var conErr interface{ ConnectionError() bool }
	var tempErr interface{ Temporary() bool }
	var timeoutErr interface{ Timeout() bool }
	var urlErr *url.Error
	var netOpErr *net.OpError
	switch {
	case errors.As(err, &conErr) && conErr.ConnectionError():
		retryable = true

	case strings.Contains(err.Error(), "connection reset"),
		strings.Contains(err.Error(), "use of closed network connection"):
		retryable = true

	case errors.As(err, &urlErr):

		// Refused connections should be retried as the service may not yet be
		// running on the port. Go TCP dial considers refused connections as
		// not temporary.
		if strings.Contains(urlErr.Error(), "connection refused") {
			retryable = true
		} else {
			return r.IsErrorRetryable(errors.Unwrap(urlErr))
		}
	case errors.As(err, &netOpErr):
		// Network dial, or temporary network errors are always retryable.
		if strings.EqualFold(netOpErr.Op, "dial") || netOpErr.Temporary() {
			retryable = true
		} else {
			return r.IsErrorRetryable(errors.Unwrap(netOpErr))
		}
	case errors.As(err, &tempErr) && tempErr.Temporary():
		// Fallback to the generic temporary check, with temporary errors
		// retryable.
		retryable = true
	case errors.As(err, &timeoutErr) && timeoutErr.Timeout():
		// Fallback to the generic timeout check, with timeout errors
		// retryable.
		retryable = true
	default:
		return aws.UnknownTernary
	}
	return aws.BoolTernary(retryable)
}

func main(){
	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion("us-east-1"), config.WithClientLogMode(aws.LogRequestWithBody|aws.LogResponseWithBody))
	if err != nil {
		log.Fatalf("unable to load SDK config, %v", err)
	}

	svc := dynamodb.NewFromConfig(cfg, func(options *dynamodb.Options) {
			options.Retryer = retry.NewStandard(func(opts *retry.StandardOptions) {
				opts.Retryables = append([]retry.IsErrorRetryable{RetryableConnectionError2{}}, opts.Retryables...)
			})
	})
}
```

PR #1832, opened 2022-09-05
Updating code examples to be correct

MovieStoreGuy:
When upgrading a service to use the v2 of the SDK, I notice the code examples weren't correct if someone was to copy and paste (thinking if someone is new to the language and don't currently have their head around errors).

This should correct the code examples so that others don't stumble by mistake.


MovieStoreGuy:
Sorry, I am wrong.

I didn't expect a stdlib to require a double pointer.

PR #1837, opened 2022-09-09
Update BuildAuthToken to validate endpoint contains a port

RanVaknin:
- Validated that the right side of the colon has to be an string representation of an integer
- Added test cases to validate type of error and if an error should be returned.

### DIFF
--- a/.changelog/aaba2642a8f64293a3ad7dd5bc0e9ef7.json
+++ b/.changelog/aaba2642a8f64293a3ad7dd5bc0e9ef7.json
@@ -1,0 +1,8 @@
+{
+    "id": "aaba2642-a8f6-4293-a3ad-7dd5bc0e9ef7",
+    "type": "feature",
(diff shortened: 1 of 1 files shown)


skmcgrail (approved):
Looks good, lets get a release note added to this before merging.

skmcgrail (approved):

feature/rds/auth/connect.go
```diff
@@ -94,3 +100,26 @@ func BuildAuthToken(ctx context.Context, endpoint, region, dbUser string, creds
...
+	if colon != -1 {
+		host, port = hostPort[:colon], hostPort[colon+1:]
+	}
+	if !validatePort(port) {
+		port = ""
+		return
+	}
+	if strings.HasPrefix(host, "[") && strings.HasSuffix(host, "]") {
+		host = host[1 : len(host)-1]
+	}
```
skmcgrail, line 115: Hmmm, this assumes that the position of `[` and `]` is the first and last character.

PR #1866, opened 2022-10-02
Directly Run Credentials Process

ijrsvt:
* Directly call credential process instead of `/bin/sh -c "credential process + args`
* This more directly follows the botocore variant. https://github.com/boto/botocore/blob/409b4043602eacfd0951ea1b28aa52c5f6dbb6bd/botocore/credentials.py#L1052
* Makes using `credential_process` usable in distroless containers (which lack `/bin/sh`)

### DIFF
--- a/credentials/processcreds/provider.go
+++ b/credentials/processcreds/provider.go
@@ -86,21 +86,13 @@ type DefaultNewCommandBuilder struct {
 // Args. The command is also initialized current process environment variables,
 // stderr, and stdin pipes.
 func (b DefaultNewCommandBuilder) NewCommand(ctx context.Context) (*exec.Cmd, error) {
-	var cmdArgs []string
-	if runtime.GOOS == "windows" {
-		cmdArgs = []string{"cmd.exe", "/C"}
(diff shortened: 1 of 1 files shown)


skmcgrail:
Sorry we can't accept this pull request due to several issues:
* This would be a backwards incompatible change in behavior users who expect that the shell is being loaded for the sourcing of environment configuration. This would need minimally be gated behind some sort of configuration value to toggle the desired behavior. Likely something that can be toggled programmatically when calling `config.LoadDefaultConfig` 
* The given implementation lacks parsing/lexing of the command necessary to support commands that take arguments. Currently the way this is invoking the process it would pass the arguments as a single argument to the process, which would be incorrect.
* Windows platforms would also need their own level of special handling of command argument parsing that are specific to how commands are pared there.

skmcgrail:
See this issue that was originally cut towards the Go repository https://github.com/aws/aws-sdk/issues/152 which covers tracking this as a cross-sdk issue, as this is not just limited to the Go SDK.

PR #1882, opened 2022-10-12
remove old broken fuzzer

AdamKorcz:
Signed-off-by: AdamKorcz <adam@adalogics.com>

The fuzzer in `internal/ini/v1.3.11` is broken.

I am trying to remove in from the `internal/ini/v1.3.11` tag as it is breaking builds in projects that use this dependency even transitively when using the `gofuzz` tag.

It seems I cannot merge into `internal/ini/v1.3.11` directly and would appreciate a bit of help here.


skmcgrail:
This was already removed in `v1.3.11` of the ini module. There isn't a method for us to radioactively modify an existing module version. You need to roll forward to a newer version.

Issue #1846, opened 2022-09-15
Should not backfill session-name for SSO Token Provider


skmcgrail:
https://github.com/aws/aws-sdk-go-v2/blob/4b5d0ddcae9bb5134c0756befbe12512d9f4d6b7/config/resolve_bearer_token.go#L59-L67

isaiahvita:
specifically, i believe the conditional linked above should be deleted, and an error should be returned if the SSOSession is not set. this is because the SSOTokenProvider is required to rely on a certain format (the existence of the SSOSession). Falling back to a legacy format used by by a different provider (SSOCredentialProvider) is not supported.

Issue #1845, opened 2022-09-15
Need to check for sso-session is present or not.


skmcgrail:
https://github.com/aws/aws-sdk-go-v2/blob/4b5d0ddcae9bb5134c0756befbe12512d9f4d6b7/config/shared_config.go#L1120-L1133

PR #1903, opened 2022-11-02
Add token refresh support to SSOCredentialProvider

isaiahvita:
This PR adds support for token refresh to the SSOCredentialProvider (by using the SSOTokenProvider) while maintaining support for legacy SSO configurations (i.e. SSO configurations that do not use an sso-session but specify sso fields directly in a profile section of a shared config file)

Testing procedure:
Followed testing instructions by SDKs feature owner @alextwoods which included:
- running the sso login flow with the AWS CLI
- using the access token, calling S3 List Buckets on the configured AWS account
- when the access token expires, re-calling S3 List buckets and see it succeed with an updated expiration in the SSO cache

Additonally: 
Fixes https://github.com/aws/aws-sdk-go-v2/issues/1845
Fixes https://github.com/aws/aws-sdk-go-v2/issues/1846

### DIFF
--- a/.changelog/2f60f375046b466789bff94b7406ba45.json
+++ b/.changelog/2f60f375046b466789bff94b7406ba45.json
@@ -1,0 +1,9 @@
+{
+    "id": "2f60f375-046b-4667-89bf-f94b7406ba45",
+    "type": "feature",
+    "description": "Adds token refresh support (via SSOTokenProvider) when using the SSOCredentialProvider",
+    "modules": [
+        "config",
+        "credentials"
+    ]
+}
--- a/.changelog/e5c705b5a4f34a99a85f12cee232ae51.json
+++ b/.changelog/e5c705b5a4f34a99a85f12cee232ae51.json
@@ -1,0 +1,9 @@
+{
+    "id": "e5c705b5-a4f3-4a99-a85f-12cee232ae51",
+    "type": "announcement",
+    "description": "When using the SSOTokenProvider, a previous implementation incorrectly compensated for invalid SSOTokenProvider configurations in the shared profile. This has been fixed via PR #1903 and tracked in issue #1846",
+    "modules": [
+        "config",
+        "credentials"
+    ]
+}


sbiscigl (approved):

config/resolve_bearer_token.go
```diff
@@ -54,20 +54,9 @@ func resolveBearerAuthTokenProviderChain(ctx context.Context, cfg *aws.Config, c
...
-		ssoSession := sharedConfig.SSOSession
-		if ssoSession == nil {
-			// Fallback to legacy SSO session config parameters, if the
-			// sso-session section wasn't used.
-			ssoSession = &SSOSession{
-				Name:        sharedConfig.SSOStartURL,
-				SSORegion:   sharedConfig.SSORegion,
-				SSOStartURL: sharedConfig.SSOStartURL,
-			}
-		}
```
syall, line 67: Due to incorrect implementation before, removing the transformation of the legacy SSO session configuration is considered a break/fix and will need to be noted in the changelog / release notes.

config/shared_config.go
```diff
@@ -900,31 +892,20 @@ func (c *SharedConfig) setFromIniSections(profiles map[string]struct{}, profile
...
+		var ssoSession SSOSession
+		ssoSession.setFromIniSection(section)
+		ssoSession.Name = c.SSOSessionName
+		c.SSOSession = &ssoSession
 	}
 
 	return nil
 }
 
-func getSSOSession(name string, sections ini.Sections, logger logging.Logger) (*SSOSession, error) {
```
syall, line 913: Make sure this function that was removed is not referenced anywhere else.

config/shared_config.go
```diff
@@ -1088,17 +1069,66 @@ func (c *SharedConfig) validateCredentialType() error {
...
 		len(c.WebIdentityTokenFile) != 0,
 	) {
-		return fmt.Errorf("only one credential type may be specified per profile: source profile, credential source, credential process, web identity token, or sso")
+		return fmt.Errorf("only one credential type may be specified per profile: source profile, credential source, credential process, web identity token")
 	}
 
 	return nil
 }
 
 func (c *SharedConfig) validateSSOConfiguration() error {
```
syall, line 1078: note: validation in general has become stricter for both the legacy and current format.

syall (changes requested):
Minor notes about the PR, add:

- More context in the description
- How the changes was tested (commands, etc.)
- How changes will impact customers (only due to the possible break/fix)

config/resolve_credentials.go
```diff
@@ -171,7 +172,29 @@ func resolveSSOCredentials(ctx context.Context, cfg *aws.Config, sharedConfig *S
...
+
+	if sharedConfig.SSOSession != nil {
+		ssoTokenProviderOptionsFn, found, err := getSSOTokenProviderOptions(ctx, configs)
+		if err != nil {
+			return fmt.Errorf("failed to get SSOTokenProviderOptions from config sources, %w", err)
+		}
+		var optFns []func(*ssocreds.SSOTokenProviderOptions)
+		if found {
+			optFns = append(optFns, ssoTokenProviderOptionsFn)
+		}
```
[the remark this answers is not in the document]
jasdel: looks like this should of bee used as input for `ssoidc.NewFromConfig`?

config/shared_config.go
```diff
@@ -1088,17 +1069,66 @@ func (c *SharedConfig) validateCredentialType() error {
...
+
+	if c.SSOSession == nil {
+		missing = append(missing, ssoSectionPrefix)
+	} else {
+		if len(c.SSOSession.SSORegion) == 0 {
+			missing = append(missing, fmt.Sprintf("%s", ssoRegionKey))
+		}
+
+		if len(c.SSOSession.SSOStartURL) == 0 {
+			missing = append(missing, fmt.Sprintf("%s", ssoStartURLKey))
```
jasdel, line 1111: are the `fmt.Sprintf` needed here?

config/shared_config.go
```diff
@@ -1088,17 +1069,66 @@ func (c *SharedConfig) validateCredentialType() error {
...
 		return nil
 	}
 
+	if c.hasLegacySSOConfiguration() {
+		err := c.validateLegacySSOConfiguration()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
alextwoods: Yes - both could exist in one profile (its unlikely, but someone who is migrating from legacy to new could have such a setup).  If so - the new format should be used, and the sso_start_url and sso_region MUST match (ie, if there is a mismatch between the profile's sso_region and the sso-session's, then we should raise an error).

config/shared_config.go
isaiahvita: ah right, its not. will remove

config/shared_config.go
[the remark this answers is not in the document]
isaiahvita: @skmcgrail yes both can exist in the same profile. i intentionally check to see if the new configuration is present and valid. if it is valid, then it should be used and if it is not valid, then it should error (even if the legacy configuration is present and valid). is there another scenario in mind you might be thinking about thats not covered here?

also, regarding @alextwoods comment about matching `sso_start_url` and `sso_region`, i do that additional check here https://github.com/aws/aws-sdk-go-v2/pull/1903/files#diff-ad732dd8295a0ce04a495056007062e13a140b7f97f173004d0d6523545805d2R1120

config/resolve_credentials.go
```diff
@@ -171,7 +172,29 @@ func resolveSSOCredentials(ctx context.Context, cfg *aws.Config, sharedConfig *S
...
+		cfgCopy.Region = sharedConfig.SSOSession.SSORegion
+		cachedPath, err := ssocreds.StandardCachedTokenFilepath(sharedConfig.SSOSession.Name)
+		if err != nil {
+			return err
+		}
+		oidcClient := ssooidc.NewFromConfig(cfgCopy)
```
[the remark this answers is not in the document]
isaiahvita: hmm im a bit confused at this requested change. 

this validation:
```
		cachedPath, err := ssocreds.StandardCachedTokenFilepath(sharedConfig.SSOSession.Name)
		if err != nil {
			return err
		}
```
isnt operating on any input related to `oidcClient := ssooidc.NewFromConfig(cfgCopy)`

additionally, the cached path conditional checks to see if the existing token is valid before creating an oidc client. if the existing token is not valid, then there is not reason to create an oidc client, so it should happen after

config/resolve_bearer_token.go
isaiahvita: @jasdel im looking at the `changelog` announcement option:
```
-t <change-type> The change annotation type (release, feature, bugfix, dependency, announcement)
```
it looks like the announcement option would determine the change type. since my overall change is actually a new feature (token refresh support for sso), is there a way to tell the changelog tool that this change is a feature and to make such an announcement?

jasdel: The best way to do that is to create two different change log entries. a `feature` and a `announcement`. Its kind of tedious, but the tool doesn't allow manual specify version change strategy outside of the `change-type`.

credentials/ssocreds/sso_credentials_provider.go
```diff
@@ -45,6 +45,10 @@ type Options struct {
 	// If custom cached token filepath is used, the Provider's startUrl
 	// parameter will be ignored.
 	CachedTokenFilepath string
+
+	// Used by the SSOCredentialProvider if a token configuration
+	// profile is used in the shared config
+	TokenProvider *SSOTokenProvider
```
jasdel, line 51: Probably should prefix this member with `SSO` in case there are other kinds of `TokenProvider` in the future. `CachedTokenFilepath` probably should of been prefixed with `SSO` as well, but that ship has passed.

isaiahvita: gotcha, just made this change with a new commit

Issue #1910, opened 2022-11-05
chimesdkmeetings and chimesdkmediapipelines do not return NotFoundException as expected


JamieEdge:
### Describe the bug

When using functions such as [`chimesdkmeetings.*Client.GetMeeting`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/chimesdkmeetings#Client.GetMeeting) and [`chimesdkmediapipelines.*Client.DeleteMediaCapturePipeline`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/chimesdkmediapipelines#Client.DeleteMediaCapturePipeline), an error matching `types.NotFoundException` is *not* returned when a resource is not found.

### Expected Behavior

An error matching `types.NotFoundException` is expected to be returned.

### Current Behavior

The returned error does not match `types.NotFoundException`.

### Reproduction Steps

The below program reproduces the issue when using the `chimesdkmeetings` package.

```go
package main

import (
	"context"
	"errors"
	"log"

	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/chimesdkmeetings"
	"github.com/aws/aws-sdk-go-v2/service/chimesdkmeetings/types"
)

func main() {
	ctx := context.Background()

	cfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		log.Fatalln(err)
	}

	c := chimesdkmeetings.NewFromConfig(cfg)
	id := "00000000-0000-0000-0000-000000000000"

	_, err = c.GetMeeting(ctx, &chimesdkmeetings.GetMeetingInput{
		MeetingId: &id,
	})
	if err == nil {
		log.Fatalln("err not nil")
	}

	var e *types.NotFoundException
	log.Println(errors.As(err, &e))
}
```

The below program reproduces the issue when using the `chimesdkmediapipelines` package.

```go
package main

import (
	"context"
	"errors"
	"log"

	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/chimesdkmediapipelines"
	"github.com/aws/aws-sdk-go-v2/service/chimesdkmediapipelines/types"
)

func main() {
	ctx := context.Background()

	cfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		log.Fatalln(err)
	}

	c := chimesdkmediapipelines.NewFromConfig(cfg)
[...]

PR #1918, opened 2022-11-10
Fix codegen to accommodate errorcode in response body

RanVaknin:
Service was returning both the ErrorCode in the body *and* in the header. Our logic was not covering that edge case. 

related issue: #1910


lucix-aws:
Closing in favor of https://github.com/aws/aws-sdk-go-v2/pull/2012

PR #1943, opened 2022-11-30
add smithy-ci gradle task

AndrewFossAWS:
### Description

> In the past, Smithy changes have been released without integration testing with Smithy-based SDK code generators. In addition, SDK code generators often fall behind in Smithy versions. As a result, we don’t have visibility on code generators’ failures until they are upgraded to the latest smithy version. 

> Upgrading versions and building all the code generators manually is also a labor-intensive task. Hence, the goal is to design a continuous integration system for testing the latest Smithy implementation with the SDK code generators.

This PR adds  `smithy-ci` gradle task to be executed by Smithy integration tests.

### DIFF
--- a/.changelog/0ef5418490454827aadfa882dcbcd781.json
+++ b/.changelog/0ef5418490454827aadfa882dcbcd781.json
@@ -1,8 +1,0 @@
-{
-    "id": "0ef54184-9045-4827-aadf-a882dcbcd781",
-    "type": "feature",
-    "description": "This release adds the LinkedAccountName field to the GetAnomalies API response under RootCause",
-    "modules": [
-        "service/costexplorer"
-    ]
-}
--- a/.changelog/224aedba11854b5a8c58faa9b845ab38.json
+++ b/.changelog/224aedba11854b5a8c58faa9b845ab38.json
@@ -1,8 +1,0 @@
-{
-    "id": "224aedba-1185-4b5a-8c58-faa9b845ab38",
-    "type": "feature",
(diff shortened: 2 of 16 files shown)


syall (approved):
Approving, equivalent to `make ci-test`.

codegen/build.gradle.kts
```diff
@@ -19,3 +19,17 @@ allprojects {
         mavenCentral()
     }
 }
+
+ext {
+    this.set("smithyVersion", project.file("SMITHY_VERSION").readText().replace(System.lineSeparator(), ""))
```
isaiahvita, line 24: why this change? is this some documented smithy convention to have an extra `SMITHY_VERSION` file that then gets read into `gradle.properties`?

AndrewFossAWS: During the integration testing, it updates the SDK's smithy version to the latest. I could have used `sed -i "25s/.*/smithyVersion=$smithyVersion/" gradle.properties` but using line number is too fragile as dependencies get updated in `gradle.properties` Hence why the extra file seems appropriate.

isaiahvita: hm im not sure im following, are you saying that with your proposed change, there will be integration testing code that updates the extra file `SMITHY_VERSION`. and if you didnt have this extra file then you would have to update `gradle.properties` via line number using `sed` because in whatever context youre in you cannot access `gradle.properties` via its normal getter/setter gradle api? 

if that understanding is correct, i dont see any code in this PR that writes to a `SMITHY_VERSION` file.

Issue #1097, opened 2021-02-01
route53: ListResourceRecordSets paginator


flood4life:
**Is your feature request related to a problem? Please describe.**
SDK v1 provides `ListResourceRecordSetsPagesWithContext` which is a convenient API to process all the records in a hosted zone. I'm currently upgrading to SDK v2 and would love to keep not having to worry about handling pagination.

**Describe the solution you'd like**
SDK provides a `NewListListResourceRecordSetsPaginator` (akin to other paginators, e.g. `NewListHostedZonesPaginator`)

**Describe alternatives you've considered**
Implementing that specific paginator in my application code.

**Additional context**
I feel like providing a new paginator should be quite straightforward if the existing paginators are used as reference. If you think that the proposed solution is the way to go, I could take a stab at contributing and implementing it.

skyzyx:
Hah. I just went to file this exact same ticket. 👍🏻

meshuga:
Lack of this paginator led to incorrect implementation and a bug https://github.com/GoogleCloudPlatform/terraformer/issues/874

PR #1984, opened 2023-01-16
R53 paginators a

RanVaknin:
Added handwritten paginator to r53 `listResourceRecordSets`
Per customer request https://github.com/aws/aws-sdk-go-v2/issues/1097

With Smithy’s `paginated` trait, the model is expected to have an `inputToken`  and an `outputToken`  but in this case, for the pagination to work we have to keep track of 3 inputs and outputs. From the 53 docs:
>**Displaying the next page of results**
_If a ListResourceRecordSets command returns more than one page of results, the value of IsTruncated is true. To display the next page of results, get the values of `NextRecordName`, `NextRecordType`, and `NextRecordIdentifier` (if any) from the response. Then submit another ListResourceRecordSets request, and specify those values for `StartRecordName`, `StartRecordType`, and `StartRecordIdentifier`._

### DIFF
--- a/service/route53/handwritten_paginators.go
+++ b/service/route53/handwritten_paginators.go
@@ -34,6 +34,7 @@ type ListResourceRecordSetsPaginator struct {
 	startRecordName       *string
 	startRecordType       types.RRType
 	startRecordIdentifier *string
+	isTruncated           bool
 }
 
 // NewListResourceRecordSetsPaginator returns a new ListResourceRecordSetsPaginator
@@ -64,7 +65,7 @@ func NewListResourceRecordSetsPaginator(client ListResourceRecordSetsAPIClient, 
 
 // HasMorePages returns a boolean indicating whether more pages are available
 func (p *ListResourceRecordSetsPaginator) HasMorePages() bool {
-	return p.firstPage || (p.startRecordName != nil && len(*p.startRecordName) != 0)
+	return p.firstPage || p.isTruncated
 }
 
 // NextPage retrieves the next ListResourceRecordSets page.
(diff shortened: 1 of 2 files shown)


service/route53/handwritten_paginators.go
```diff
@@ -0,0 +1,108 @@
...
+func (p *ListResourceRecordSetsPaginator) NextPage(ctx context.Context, optFns ...func(*Options)) (*ListResourceRecordSetsOutput, error) {
+	if !p.HasMorePages() {
+		return nil, fmt.Errorf("no more pages available")
+	}
+
+	params := *p.params
+	params.StartRecordName = p.startRecordName
+
+	params.StartRecordIdentifier = p.startRecordIdentifier
+	params.StartRecordType = p.startRecordType
```
RanVaknin, line 80: You need to keep track of all 3 parameters (`startRecordName`, `StartRecordIdentifier` and `StartRecordType`)
for the pagination to work.

service/route53/handwritten_paginators.go
```diff
@@ -0,0 +1,108 @@
...
+		firstPage:             true,
+		startRecordName:       params.StartRecordName,
+		startRecordType:       params.StartRecordType,
+		startRecordIdentifier: params.StartRecordIdentifier,
+	}
+}
+
+// HasMorePages returns a boolean indicating whether more pages are available
+func (p *ListResourceRecordSetsPaginator) HasMorePages() bool {
+	return p.firstPage || (p.startRecordName != nil && len(*p.startRecordName) != 0)
```
aajtodd, line 67: **correctness**: This is only looking at `startRecordName`, should pagination continue if either of the other two parameters are set?

service/route53/handwritten_paginators.go
```diff
@@ -0,0 +1,108 @@
...
+
+	prevToken := p.startRecordName
+	p.startRecordName = result.NextRecordName
+
+	p.startRecordIdentifier = result.NextRecordIdentifier
+	p.startRecordType = result.NextRecordType
+
+	if p.options.StopOnDuplicateToken &&
+		prevToken != nil &&
+		p.startRecordName != nil &&
```
aajtodd, line 102: **correctness/question**: Should this be looking at all 3 parameters?

service/route53/handwritten_paginators_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+type testListRRSMiddleware struct {
+	id int
+}
+
+func (m *testListRRSMiddleware) ID() string {
+	return fmt.Sprintf("mock middleware %d", m.id)
+}
+
+func (m *testListRRSMiddleware) HandleInitialize(ctx context.Context, input middleware.InitializeInput, next middleware.InitializeHandler) (
```
aajtodd, line 25: **fix**: Same comment as s3 paginators, this can be done with a mock client more cleanly

aajtodd (changes requested):

service/route53/handwritten_paginators.go
wty-Bryant: I didn't find similar duplication judgement from [java sdk v2](https://code.amazon.com/packages/AwsJavaSdk-Route53/blobs/aws-sdk-java-v2/release/--/src/software/amazon/awssdk/services/route53/paginators/ListResourceRecordSetsIterable.java). Maybe need to check other sdk as well

aajtodd (approved):

PR #1997, opened 2023-01-29
feat(s3): add missing DeleteObject interface

zekth:
Hello,

I spotted the missing interface for `DeleteObject`, this PR just adds it. I'd rather call the interface `DeleteAPIClient`. That's up to maintainer choice.

Happy to update it.

### DIFF
--- a/feature/s3/manager/api.go
+++ b/feature/s3/manager/api.go
@@ -6,8 +6,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 )
 
-// DeleteObjectsAPIClient is an S3 API client that can invoke the DeleteObjects operation.
+// DeleteObjectsAPIClient is an S3 API client that can invoke DeleteObject, DeleteObjects operations.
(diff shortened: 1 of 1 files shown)


lucix-aws:
Is there a particular reason you need this addition?

I think we'd prefer to see this added as its own interface i.e.

```go
// DeleteObjectsAPIClient is an S3 API client that can invoke the DeleteObject operation.
type DeleteObjectAPIClient interface {
	DeleteObject(context.Context, *s3.DeleteObjectInput, ...func(*s3.Options)) (*s3.DeleteObjectOutput, error)
}
```

The _reason_ these exist, however, is to prevent the transfer manager from having to depend on the actual S3 client and to facilitate plugging in custom implementations, mocking, etc. As of this writing, nothing in the actual transfer manager implementation calls `DeleteObject`, so it's not really required.

I'm going to close this PR since it doesn't add anything on its own - if you believe there's a use case for this. please open an issue to discuss.

PR #2009, opened 2023-02-02
updated ci to use go version 1.20

RanVaknin:
Go compiler only supports the latest 2 releases, so that is now 1.19 and 1.20.
Updated all CI workflows to use Go 1.20


aajtodd:
This has most of what we need but there are still some failures.

Closing in favor of: https://github.com/aws/aws-sdk-go-v2/pull/2023 where I'll track down the any remaining test/vet failures.

Thanks

PR #2017, opened 2023-02-10
Fix v4a tests broken in Go 1.20

djedward:
Go 1.20 introduced a change to crypto/ecdsa which won't generate keys from an input stream that is only zeros. I've opened an issue upstream to get clarity on that change. This change simply updates the test to use ones instead, so we can start building against 1.20.


aajtodd:
Adding upstream issue for reference: https://github.com/golang/go/issues/58454

We're taking a look at support for Go 1.20 holistically right now (e.g. we need CI support, etc). This may just be one part of that.

aajtodd (approved):

Issue #2015, opened 2023-02-08
Configuring Assume Role With Web Identity doesn't allow directly setting parameters when setting `AWS_WEB_IDENTITY_TOKEN_FILE`


gdavison:
### Describe the bug

When assuming a web identity role, we're not able to point to the token file using `AWS_WEB_IDENTITY_TOKEN_FILE` and set the role ARN and other parameters by using `config.WithWebIdentityRoleCredentialOptions`. For example, what we'd like to do is:

```go
loadOptions = append(loadOptions, config.WithWebIdentityRoleCredentialOptionsFunc(opts *stscreds.WebIdentityRoleOptions) {
  opts.RoleARN = "<...>"
 })
// ...
cfg, err := config.LoadDefaultConfig(ctx, loadOptions...)
```

If configuration is set using only a combination of environment variables and/or named profiles, it works as intended.

### Expected Behavior

All config sources, i.e. `config.LoadOptions`, `config.EnvConfig`, and `config.SharedConfig` should be used when resolving the web identity settings.

### Current Behavior

Because `AWS_WEB_IDENTITY_TOKEN_FILE` is set, `config.resolveCredentialChain` assumes that the token file location and role ARN will be sourced from environment variables. This causes the checks in `config.assumeWebIdentity` to fail, since there is no value set for the file path.

If a dummy value is set in `AWS_WEB_IDENTITY_TOKEN_FILE`, the check in `config.assumeWebIdentity` will pass, and then the function passed to `config.WithWebIdentityRoleCredentialOptionsFunc` will be called and override the token retriever.

### Reproduction Steps

Run the following with `AWS_WEB_IDENTITY_TOKEN_FILE` and `opts.RoleARN` set

```go
package main

import (
	"context"
	"log"

	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
)

func main() {
	ctx := context.Background()

	var loadOptions []func(*config.LoadOptions) error

	loadOptions = append(loadOptions,
		config.WithWebIdentityRoleCredentialOptions(func(opts *stscreds.WebIdentityRoleOptions) {
			opts.RoleARN = "..."
		}),
	)

	_, err := config.LoadDefaultConfig(ctx, loadOptions...)
	if err != nil {
		log.Fatalf("failed: %s", err)
	}
}
```

### Possible Solution

[...]

PR #2030, opened 2023-02-17
delay WebIdentityRoleOptions validation

aajtodd:
When loading the default config and `AWS_WEB_IDENTITY_TOKEN_FILE` is set it triggers the assume role with web identity provider. This provider requires it's configuration either entirely in environment variables or shared config. The Go SDK offers an API to configure this provider further via code which was being ignored or causing early validation issues. This change allows `AWS_WEB_IDENTITY_TOKEN_FILE` to be set and all other options to be supplied by Go code relaxing the constraint that everything has to come from the environment.

fixes aws/aws-sdk-go-v2#2015

To help speed up the process and reduce the time to merge please ensure that `Allow edits and access to secrets by maintainers` is checked before submitting your PR. This will allow the project maintainers to make minor adjustments or improvements to the submitted PR, allow us to reduce the roundtrip time for merging your request.

### DIFF
--- a/config/resolve_credentials.go
+++ b/config/resolve_credentials.go
@@ -384,7 +384,11 @@ func assumeWebIdentity(ctx context.Context, cfg *aws.Config, filepath string, ro
 		return fmt.Errorf("token file path is not set")
 	}
 
-	var optFns []func(*stscreds.WebIdentityRoleOptions)
+	optFns := []func(*stscreds.WebIdentityRoleOptions){
+		func(options *stscreds.WebIdentityRoleOptions) {
+			options.RoleSessionName = sessionName
+		},
+	}
 
 	optFn, found, err := getWebIdentityCredentialProviderOptions(ctx, configs)
 	if err != nil {
@@ -396,8 +400,7 @@ func assumeWebIdentity(ctx context.Context, cfg *aws.Config, filepath string, ro
 	}
 
 	opts := stscreds.WebIdentityRoleOptions{
-		RoleARN:         roleARN,
-		RoleSessionName: sessionName,
+		RoleARN: roleARN,
 	}
 
 	for _, fn := range optFns {
--- a/config/resolve_web_identity_test.go
+++ b/config/resolve_web_identity_test.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"context"
-	"fmt"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
 	"os"
@@ -23,7 +22,7 @@ func TestResolveWebIdentityWithOptions(t *testing.T) {
 		os.Setenv("AWS_WEB_IDENTITY_TOKEN_FILE", tokenFile)
 		os.Setenv("AWS_REGION", "us-east-1")
 
-		config, err := LoadDefaultConfig(context.Background(),
+		_, err := LoadDefaultConfig(context.Background(),
 			WithWebIdentityRoleCredentialOptions(func(options *stscreds.WebIdentityRoleOptions) {
 				options.RoleARN = "test-arn"
 			}),
@@ -32,9 +31,6 @@ func TestResolveWebIdentityWithOptions(t *testing.T) {
 		if err != nil {
 			t.Fatalf("expect no error, got %v", err)
 		}
-
-		fmt.Println(config.Credentials)
-
 	})
 
 	t.Run("token from profile", func(t *testing.T) {


config/resolve_credentials.go
```diff
@@ -384,25 +384,36 @@ func assumeWebIdentity(ctx context.Context, cfg *aws.Config, filepath string, ro
...
+
 	if found {
 		optFns = append(optFns, optFn)
 	}
 
-	provider := stscreds.NewWebIdentityRoleProvider(sts.NewFromConfig(*cfg), roleARN, stscreds.IdentityTokenFile(filepath), optFns...)
+	opts := stscreds.WebIdentityRoleOptions{
+		RoleARN:         roleARN,
+		RoleSessionName: sessionName,
+	}
```
gdavison, line 401: The `opts` value isn't passed to `stscreds.NewWebIdentityRoleProvider`, so the `RoleSessionName` set here is lost.

The original declaration of `optFns` should be restored, i.e.

```go
optFns := []func(*stscreds.WebIdentityRoleOptions){
	func(options *stscreds.WebIdentityRoleOptions) {
		options.RoleSessionName = sessionName
	},
}
```

aajtodd: Good catch, I've restored the way `RoleSessionName` was being set.

gdavison:
HI @aajtodd, I've been doing some more testing. To clarify, when AWS_WEB_IDENTITY_TOKEN_FILE` _isn't set_, should calling `config.WithWebIdentityRoleCredentialOptions` with `RoleARN` be sufficient to trigger assuming a web identity role, or does the `stscreds.NewWebIdentityRoleProvider` still have to be configured manually in that case?

gdavison:
In our current setup, we have to do (among other setup), the following:

```go
func getCredentialsProvider(ctx context.Context, c *Config) (aws.CredentialsProvider, string, error) {
  ...
  
	envConfig, err := config.NewEnvConfig()
  
  ...
  
  // Set up the WebIdentityRoleCredentialOptions here, in case the envvar `AWS_WEB_IDENTITY_TOKEN_FILE` is set
  // Otherwise, the AWS SDK for Go expects that there will be configured credentials and will return an error.
  if web := c.AssumeRoleWithWebIdentity; web != nil {
  	loadOptions = append(loadOptions, config.WithWebIdentityRoleCredentialOptions(func(opts *stscreds.WebIdentityRoleOptions) {
  		opts.RoleARN = web.RoleARN
  		opts.RoleSessionName = web.SessionName
  		opts.Duration = web.Duration
  		if web.Policy != "" {
  			opts.Policy = aws.String(web.Policy)
  		}
  		if len(web.PolicyARNs) > 0 {
  			opts.PolicyARNs = getPolicyDescriptorTypes(web.PolicyARNs)
  		}
  		if web.HasValidTokenSource() {
  			opts.TokenRetriever = web
  		}
  	}))
  }

  cfg, err := config.LoadDefaultConfig(ctx, loadOptions...)
  if err != nil {
  	return nil, "", fmt.Errorf("loading configuration: %w", err)
  }

  // If the envvar `AWS_WEB_IDENTITY_TOKEN_FILE` has not been set, and we want to assume a web role, we now have
  // to handle it manually. Creating the STS client for assuming the role needs a populated `aws.Config`, so we have
  // to do this after calling `config.LoadDefaultConfig`
  if c.AssumeRoleWithWebIdentity != nil && envConfig.WebIdentityTokenFilePath == "" {
  	provider, err := webIdentityCredentialsProvider(ctx, cfg, c)
  	if err != nil {
  		return nil, "", err
  	}
  	cfg.Credentials = provider
  }

  ...
}

func webIdentityCredentialsProvider(ctx context.Context, awsConfig aws.Config, c *Config) (aws.CredentialsProvider, error) {
	web := c.AssumeRoleWithWebIdentity
	client := stsClient(ctx, awsConfig, c)

	appCreds := stscreds.NewWebIdentityRoleProvider(client, web.RoleARN, web, func(opts *stscreds.WebIdentityRoleOptions) {
		opts.RoleSessionName = web.SessionName
		opts.Duration = web.Duration

		if web.Policy != "" {
			opts.Policy = aws.String(web.Policy)
		}

		if len(web.PolicyARNs) > 0 {
			opts.PolicyARNs = getPolicyDescriptorTypes(web.PolicyARNs)
		}
	})

	_, err := appCreds.Retrieve(ctx)
	if err != nil {
		return nil, c.NewCannotAssumeRoleWithWebIdentityError(err)
	}
	return aws.NewCredentialsCache(appCreds), nil
}

func stsClient(ctx context.Context, awsConfig aws.Config, c *Config) *sts.Client {
	return sts.NewFromConfig(awsConfig, func(opts *sts.Options) {
		if c.StsRegion != "" {
			opts.Region = c.StsRegion
		}
		if c.StsEndpoint != "" {
			opts.EndpointResolver = sts.EndpointResolverFromURL(c.StsEndpoint)
		}
	})
}
```

We need to resolve all of the configuration first so that we can get an `aws.Config` that we can pass to `sts.NewFromConfig` so that we can do a "manual" assume role with web credentials. It would be great if `config.LoadDefaultConfig` would see the web identity options and assume the role.

aajtodd:
@gdavison 

From the code provided:
> If the envvar `AWS_WEB_IDENTITY_TOKEN_FILE` has not been set, and we want to assume a web role, we now have
  // to handle it manually.

This is correct and expected.

---

>  It would be great if config.LoadDefaultConfig would see the web identity options and assume the role.

This is a misunderstanding of both how `config.LoadDefaultConfig` works and it's purpose. As it's name suggests it's attempting to load/resolve the SDK configuration by looking in the most common (default) places. Each of the credential providers it's looking for are used in specific/comomon deployment scenarios (e.g. local development, EC2, lambda, code build, EKS, etc). Once a valid credential source is found the search stops. All the options provided do is influence the web identity provider IFF it is the provider that is selected. Admittedly this is a confusing API (`config.WithWebIdentityRoleCredentialOptions`) and probably shouldn't exist. 


Attempting to hook into this default behavior is understandable but also probably not optimal for your use case. It uses heuristics and a prioritized list to attempt to resolve configuration stopping at the first valid source. When you know the credential provider to choose already you'd be better off handling your own configuration with custom code as you are doing for the web identity provider in the code shared).

---

All that being said this PR enables more of the options from `config.WithWebIdentityRoleCredentialOptions` to be valid/merged with environment config including role ARN and explicitly set `sts` client (previously only the session name was actually used). It would not remove the need to explicitly configure a web identity provider if the environment processed by `config.LoadDefaultConfig` isn't configured for one.

gdavison:
Thanks, @aajtodd. Because of the purpose of `config.LoadDefaultConfig`, this meets our needs because it at least does not crash, and allows us to do our own handling.

lucix-aws (approved):

PR #2040, opened 2023-02-24
add doc support for S3 object integrity feature

### LINKED ISSUE #1689: Developer Guide: Document Support for Amazon S3 Object Integrity Features
A little bit more nuanced, but I think this highlights that it is worth creating a developer guide documentation page to help clarify the behaviors when using the S3 client and checksums.

* If a user specifies an explicit checksum value in either the `ChecksumCRC32`, `ChecksumCRC32C`, `ChecksumSHA1`, or `ChecksumSHA256` fields then the SDK will send that value specified using a standard HTTP request header using the appropriate `x-amz-checksum-*` header for each. This is not affected by whether the `HTTP` or `HTTPS` scheme is used. The request will be signed with SigV4 or SigV4a regardless of the scheme, the only distinction is whether the payload body hash is computed or not for the `x-amz-content-sha256` header. For HTTP schemes this will be the computed sha256 hash for the PutObject body, otherwise for HTTPS this will be `UNSIGNED-PAYLOAD`. If the payload body hash needs to be computed for sending to a non-HTTPS endpoint, then the body must be seekable so the SDK can compute the hash and reseek the body.
* If `ChecksumAlgorithm` field, and an algorithm `Checksum*` member is not set, then the SDK will compute the corresponding checksum for the request and will send it in the appropriate `x-amz-checksum-*`. The only distinction is if the checksum is sent as a header or trailer based on the endpoint scheme
  * If the scheme is HTTPS the computed checksum will be sent in a trailer header block using aws-chunked encoding. It will also do so use an unsigned payload variant called `STREAMING-UNSIGNED-PAYLOAD-TRAILER` which does not require signing each aws-chunk. The Go SDK only supports this variant of aws-chunked, which is why computed checksums sent via a aws-chunked trailer requires an HTTPS endpoint in this SDK. For trailer streaming the content-length of the PutObject body must be known. So either the PutObject request specifies `ContentLength` field, the body implements  `interface{ Len() int }`, or the body is seekable so SDK can figure out the length to be sent. The SDK may add support for the signed aws-chunked variants (e.g. `STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER`) in the future as to support sending of signed checksum trailers to non-HTTPS endpoints. That feature is supported in SDKs that had already supported signed aws-chunked payloads previously. 
  * If the scheme is HTTP then the SDK will fallback to to a behavior similar to the provided checksum case, but instead the SDK will compute the value (thus the body must be seekable), and attach it as a header similar to the provided checksum case.

_Originally posted by @skmcgrail in https://github.com/aws/aws-sdk-go-v2/issues/1667#issuecomment-1113754596_

wty-Bryant:
Add dev guide doc support for S3 object integrity feature to resolve #1689, including checksum/md5 header configuration under SDK and http/https PutObject request.

### DIFF
--- a/content/en/docs/sdk-utilities/s3.md
+++ b/content/en/docs/sdk-utilities/s3.md
@@ -101,6 +101,48 @@ before sending it to {{% alias service="S3" %}}. `Uploader` calculates the expec
 file to {{% alias service="S3" %}}. If the current value of `PartSize` requires more than 10,000 parts to upload the
 file, `Uploader` increases the part size value so that fewer parts are required.
 
+### S3 Object Integrity Features (checksum and md5)
+In old version `aws-sdk-go`, uploaded object's integrity is checked by computing and comparing `ContentMD5` and checksum related fields in `UploadInput` type, within which `ContentMD5` will be ignored if object is uploaded in multipart due to mismatch among full object md5, single part md5 and checksum of checksum of full object. In `aws-sdk-go-v2`, similarly, when object is larger than `PartSize` in `Uploader`, it will be multipart uploaded and `ContentMD5` in `PutObjectInput` will be invalid.
+
+Now the integrity of object is mainly checked by checksum. S3 supports 4 checksum algorithms: SHA-1, SHA-256, CRC-32, and CRC-32C, which could be set in `ChecksumAlgorithm` field of `PutObjectInput` and SDK will compute object checksum according to given algorithm. If a user specifies an explicit `Checksum*` value in either the `ChecksumCRC32`, `ChecksumCRC32C`, `ChecksumSHA1`, or `ChecksumSHA256` fields in `PutObjectInput`, SDK will compute corresponding checksum and compare with sent value, the upload will fail if they dismatch. 
+
+```go
+result, err := uploader.Upload(context.TODO(), &s3.PutObjectInput{
+	Bucket: aws.String("my-bucket"),
+	Key:    aws.String("my-object-key"),
+	Body:   uploadFile,
+	ChecksumAlgorithm: types.ChecksumAlgorithmSha256,
+})
+```
+
+If an explicit checksum value in either the `ChecksumCRC32`, `ChecksumCRC32C`, `ChecksumSHA1`, or `ChecksumSHA256` fields is specified, then the SDK will send that value specified using a standard HTTP request header using the appropriate `x-amz-checksum-*` header for each. This is not affected by whether the HTTP or HTTPS scheme is used. The request will be signed with SigV4 or SigV4a regardless of the scheme, the only distinction is whether the payload body hash is computed or not for the `x-amz-content-sha256` header. For HTTP schemes this will be the computed sha256 hash for the `PutObject` body, otherwise for HTTPS this will be `UNSIGNED-PAYLOAD`. If the payload body hash needs to be computed for sending to a non-HTTPS endpoint, then the body must be seekable so the SDK can compute the hash and reseek the body.
+
+If `ChecksumAlgorithm` field, and an algorithm `Checksum*` member is not set, then the SDK will compute the corresponding checksum for the request and will send it in the appropriate `x-amz-checksum-*` header in the REST request. The only distinction is if the checksum is sent as a header or trailer based on the endpoint scheme.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/sdk-utilities/s3.md
```diff
@@ -101,6 +101,48 @@ before sending it to {{% alias service="S3" %}}. `Uploader` calculates the expec
...
... is HTTPS the computed checksum will be sent in a trailer header block using aws-chunked encoding. It will also do so use an unsigned payload variant called `STREAMING-UNSIGNED-PAYLOAD-TRAILER` which does not require signing each aws-chunk. The Go SDK only supports this variant of aws-chunked, which is why computed checksums sent via a aws-chunked trailer requires an HTTPS endpoint in this SDK. For trailer streaming the content-length of the `PutObject` body must be known. So either the `PutObjectInput` specifies `ContentLength` field, the `Body` field implements `interface{ Len() int }`, or the body is seekable so SDK can figure out the length to be sent. The SDK may add support for the signed aws-chunked variants (e.g. `STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER`) in the future as to support sending of signed checksum trailers to non-HTTPS endpoints. That feature is supported in SDKs that had already supported signed aws-chunked payloads previously. For more information about checksum location and trailing checksum support for unsigned payload, please refer to [flexible-checksums](https://code.amazon.com/packages/AwsDrSeps/blobs/main/--/seps/accepted/shared/flexible-checksums.md)
```
aajtodd, line 122: this isn't true, where are you getting this information?

content/en/docs/sdk-utilities/s3.md
```diff
@@ -101,6 +101,48 @@ before sending it to {{% alias service="S3" %}}. `Uploader` calculates the expec
 file to {{% alias service="S3" %}}. If the current value of `PartSize` requires more than 10,000 parts to upload the
 file, `Uploader` increases the part size value so that fewer parts are required.
 
+### S3 Object Integrity Features (checksum and md5)
```
aajtodd, line 104: I think this is overall focused a bit too much on the feature itself, instead I think we want to just document the differences/behavior w.r.t `ContentMD5/Checksum*/ChecksumAlgorithm` settings. Perhaps we should work this out in a quip doc first rather than going back and forth in PR but here is a first stab at it:


```
S3 [Object Integrity Checks](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html) in {{% alias service="S3" %}} can be used with the transfer manager `Upload` function.

\```go
result, err := uploader.Upload(context.TODO(), &s3.PutObjectInput{
	Bucket: aws.String("my-bucket"),
	Key:    aws.String("my-object-key"),
	Body:   uploadFile,
	ChecksumAlgorithm: types.ChecksumAlgorithmSha256,
})
\```

There are some differences from `PutObject` and `UploadPart` to be aware of. It is not recommended to set explicit `ContentMD5` or any of the `Checksum*` values (e.g. `ChecksumCRC32`, `ChecksumCRC32C`, `ChecksumSHA1`, or `ChecksumSHA256`) directly when using the `Upload` function. This is because the `Uploader` will internally convert large upload requests to a multipart upload for performance. The provided checksum values would be invalid for individual parts and thus are ignored by the SDK. The recommended way to enable object integrity checks is to set a `ChecksumAlgorithm` which will be used for single and multi-part uploads. 

```

aajtodd (changes requested):

content/en/docs/sdk-utilities/s3.md
aajtodd: Ok we can talk about that, I'm not 100% sure that is all accurate but it's also talking about S3 integrity features for `PutObject`/`UploadPart` operations. We can document those in a different section and link to them from here. This section is on the S3 transfer manager which I think we should keep the discussion here related to the differences.

content/en/docs/sdk-utilities/s3.md
wty-Bryant: I share a quip containing the draft https://quip-amazon.com/1d1aAwaSz3ya/Dev-guide-draft-for-S3-object-integrity-check

Issue #1451, opened 2021-10-07
S3 Client does not send Expect: 100-continue header for HTTP PUT requests


skmcgrail:
Need to port the V1 SDK customization that set the `Expect: 100-continue` header for HTTP PUT operations that were larger 2 MB or streaming bodies where the content-length is unknown. Need to also provide a client configuration options that allows users to toggle this feature.

https://github.com/aws/aws-sdk-go/blob/abb93ecf8f7d631b6bc9d6701ff2938e56f1a055/service/s3/platform_handlers_go1.6.go#L18

PR #2051, opened 2023-03-10
Port v1 SDK customization for s3 HTTP PUT request 

wty-Bryant:
Add {Expect: 100-continue} http header customization middleware for s3 client HTTP PUT request larger than 2MB or with unknown-size body, solve #1451

### DIFF
--- a/.changelog/bbab7da0e2504bebb9d999dacb2de133.json
+++ b/.changelog/bbab7da0e2504bebb9d999dacb2de133.json
@@ -1,7 +1,7 @@
 {
     "id": "bbab7da0-e250-4beb-b9d9-99dacb2de133",
     "type": "feature",
-    "description": "port v1 sdk's 100-continue http header customization for s3 PutObject/UploadPart request and enable user config",
+    "description": "port v1 sdk 100-continue http header customization for s3 PutObject/UploadPart request and enable user config",
     "modules": [
         "service/internal/s3shared",
         "service/s3"


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,77 @@
...
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(ADD_100Continue_Header).build())
+                                .build())
+                        .build()
+        );
+    }
+
+    // returns true if service is either s3 or s3 control and needs s3 customization
+    private static boolean requiresS3Customization(Model model, ServiceShape service) {
+        return S3ModelUtils.isServiceS3(model, service) || S3ModelUtils.isServiceS3Control(model, service);
```
aajtodd, line 74: **fix**: Probably doesn't need to apply to s3control

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,77 @@
...
+            );
+        });
+        writer.insertTrailingNewline();
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                RuntimeClientPlugin.builder()
+                        .servicePredicate(S3100Continue::requiresS3Customization)
```
aajtodd, line 64: **fix**: IIRC this builder can also specify an operation predicate, we should use that to restrict this to just the operations that we want to use it on (`PutObject`, `UploadPart`, etc)

config/go.mod
```diff
@@ -3,13 +3,13 @@ module github.com/aws/aws-sdk-go-v2/config
...
 
 require (
-	github.com/aws/aws-sdk-go-v2 v1.17.5
-	github.com/aws/aws-sdk-go-v2/credentials v1.13.15
-	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.23
-	github.com/aws/aws-sdk-go-v2/internal/ini v1.3.30
-	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4
-	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.4
-	github.com/aws/aws-sdk-go-v2/service/sts v1.18.5
+	github.com/aws/aws-sdk-go-v2 v1.17.6
```
aajtodd, line 6: **fix**: There are way too many changes to `go.mod` and `go_module_metadata.go` un-related to the intent of this PR. We should figure out why these changed/revert them. It makes it very difficult to follow the PR otherwise

service/internal/s3shared/s3100continue.go
```diff
@@ -0,0 +1,41 @@
...
+	ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler,
+) (
+	out middleware.BuildOutput, metadata middleware.Metadata, err error,
+) {
+	req, ok := in.Request.(*smithyhttp.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown request type %T", req)
+	}
+
+	if req.Method == "PUT" && (req.ContentLength == -1 || (req.ContentLength == 0 && req.Body != nil) || req.ContentLength >= 1024*1024*2) {
```
aajtodd, line 36: **comment**: If we scope the customization correctly in codegen to only apply to specific operations we can probably remove the `req.Method` check here.

**fix**: Change limit (2MB) to a private constant in this file

service/s3/internal/customizations/s3100continue_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/aws/aws-sdk-go-v2/internal/awstesting"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/google/go-cmp/cmp"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// unit test for service/internal/s3shared/s3100continue.go
+func TestAdd100ContinueHttpHeader(t *testing.T) {
```
aajtodd, line 16: **fix**: Is there not a way to have the test live closer to where the code that defines it is (e.g. `s3shared`)? It would also seem like we don't really need to spinup a test server for this and can instead directly test the middleware ([example](https://github.com/aws/smithy-go/blob/main/transport/http/middleware_content_length_test.go) from smithy-go).

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,77 @@
...
+                        .registerMiddleware(MiddlewareRegistrar.builder()
+                                .resolvedFunction(SymbolUtils.createValueSymbolBuilder(ADD_100Continue_Header).build())
+                                .build())
+                        .build()
```
aajtodd, line 68: **fix**: This middleware needs someway to be configured by customers to turn off the behavior (opt out) and/or change the threshold. You can see an example in the equivalent kotlin [PR](https://github.com/awslabs/aws-sdk-kotlin/pull/845).

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,77 @@
...
+import software.amazon.smithy.go.codegen.integration.RuntimeClientPlugin;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.ServiceShape;
+import software.amazon.smithy.utils.ListUtils;
+
+/**
+ * Add middleware, which adds {Expect: 100-continue} header for s3 client HTTP PUT request larger than 2MB
+ * or with unknown size streaming bodies, during operation builder step
+ */
+public class S3100Continue implements GoIntegration {
```
aajtodd, line 20: **fix**: Not sure where to make this comment but we ran into an issue with this header being set and signing. See this [PR](https://github.com/awslabs/smithy-kotlin/pull/812) for more details. Can you verify that s3 put object requests with transfer acceleration enabled don't run into issues when this header is present? We may need to special case it in signing as we did in Kotlin.

aajtodd (changes requested):

service/s3/internal/customizations/s3100continue_test.go
wty-Bryant: If I put it under `s3shared` containing middleware, circular dependencies occur since `s3shared` and `s3` depends on each other in that case due to the test.

aajtodd: Right but if you don't make the test depend on the s3 client and instead just test the middleware in isolation then it should be fine (see the example tests linked in previous comment)

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
wty-Bryant: I'm confused about the test design of this validation. Since  `UseAccelerate` and `SupportsAccelerate` is configured in [UpdateEndpoint](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/internal/customizations/update_endpoint.go#:~:text=func-,UpdateEndpoint,-(stack%20*) at Serialize step in stack, is it possible to wire up inner middleware like `processARNResource`, `s3ObjectLambdaEndpoint` etc with current `s3100Continue` middleware in isolation in unit test to check issues?

wty-Bryant: I'm confused about the test design of this validation. Since UseAccelerate and SupportsAccelerate is configured in [UpdateEndpoints](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/internal/customizations/update_endpoint.go#:~:text=func-,UpdateEndpoint,-(stack%20*) at Serialize step in stack, is it possible to wire up inner middleware like processARNResource, s3ObjectLambdaEndpoint etc with current s3100Continue middleware in isolation in unit test to check issues?

wty-Bryant: I'm confused about the test design of this validation. Since UseAccelerate and SupportsAccelerate is configured in [UpdateEndpoints](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/internal/customizations/update_endpoint.go#:~:text=func-,UpdateEndpoint,-(stack%20* ) at Serialize step in stack, is it possible to wire up inner middleware like `processARNResource`, `s3ObjectLambdaEndpoint` etc with current s3100Continue middleware in isolation in unit test to check issues?

aajtodd: No need to check it in a unit test. Just test the actual real S3 service with your customization enabled and `UseAccelerate` enabled (might need to [configure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration-examples.html) it on the bucket).

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,103 @@
...
+            );
+        });
+        writer.insertTrailingNewline();
+    }
+
+    @Override
+    public List<RuntimeClientPlugin> getClientPlugins() {
+        return ListUtils.of(
+                RuntimeClientPlugin.builder()
+                        .operationPredicate((model, service, operation) -> isS3Service(model, service) && operation.
```
aajtodd, line 79: **fix**: We probably really only care about the `PutObject` and `UploadPart` operations. The other APIs don't have large binary payloads where an `Expect` header makes any difference (and may not be supported by the service for those operations IDK).

aajtodd: **Question**: Can service and operation predicates be combined such that you can move the `isS3Service` check to `servicePredicate(...)`?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,103 @@
...
+                        .build(),
+                RuntimeClientPlugin.builder()
+                        .servicePredicate(S3100Continue::isS3Service)
+                        .configFields(ListUtils.of(
+                            ConfigField.builder()
+                                    .name(Continue_Client_Option)
+                                    .type(SymbolUtils.createValueSymbolBuilder("int64")
+                                            .putProperty(SymbolUtils.GO_UNIVERSE_TYPE, true)
+                                            .build())
+                                    .documentation("The threshold ContentLength for HTTP PUT request to receive {Expect: 100-continue} header. " +
```
aajtodd, line 95: **nit**: 

> The threshold ContentLength in bytes ...

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,103 @@
...
+                                            "When set to -1, this header will be opt out of the operation request; when set to 0, the threshold" +
```
aajtodd, line 96: **fix**: 

> Setting to -1 will disable adding the `Expect` header to requests ...

service/s3/api_client.go
```diff
@@ -80,6 +80,11 @@ type Options struct {
 	// Configures the events that will be sent to the configured logger.
 	ClientLogMode aws.ClientLogMode
 
+	// The threshold ContentLength for HTTP PUT request to receive {Expect:
+	// 100-continue} header. When set to -1, this header will be opt out of the
+	// operation request; when set to 0, the thresholdwill be set to default 2MB
```
lucix-aws, line 85: the threshold will (space)

aws/signer/internal/v4/headers_test.go
```diff
@@ -33,3 +33,31 @@ func TestAllowedQueryHoisting(t *testing.T) {
 		})
 	}
 }
+
+func TestIgnoredHeaders(t *testing.T) {
+	cases := map[string]struct {
+		Header        string
+		ExpectIgnored bool
```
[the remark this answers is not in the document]
wty-Bryant: Agree, the initial `ExcludeList` rule within v4 is also a little confusing since it is valid when inner rule is invalid. I have changed the `ExpectIgnored` to be true if the header is ignored.

service/internal/s3shared/s3100continue.go
```diff
@@ -0,0 +1,54 @@
...
+		sizeLimit = m.continueHeaderThresholdBytes
+	default:
+	}
+
+	req, ok := in.Request.(*smithyhttp.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown request type %T", req)
+	}
+
+	if req.ContentLength == -1 || (req.ContentLength == 0 && req.Body != nil) || req.ContentLength >= sizeLimit {
```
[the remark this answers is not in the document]
wty-Bryant: I haven't met such case in real code, it is judged only because the initial issue put forward that "Need to port the V1 SDK customization that set the Expect: 100-continue header for HTTP PUT operations that were larger 2 MB or streaming bodies where the content-length is unknown." According to `http.Request` comment about `ContentLength`, "For client requests, a value of 0 with a non-nil Body is also treated as unknown."

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
wty-Bryant: Do you mean remove `isS3Service` by writing its logic directly within service predicate, I can do that. `isS3Service` is written just because it can be used in both `writeAdditionalFiles` and `getClientPlugins`.

wty-Bryant: Yep, but I'm confused that some middleware codegen like [S3GetBucketLocation](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3GetBucketLocation.java#:~:text=.operationPredicate((model%2C%20service%2C%20operation)%20%2D%3E%20%7B) actually adds `isS3Service` check to operation predicate. Not sure which format is better

.changelog/1d119df578fb4a57a2934d38ea28aaa5.json
```diff
@@ -0,0 +1,10 @@
+{
+    "id": "1d119df5-78fb-4a57-a293-4d38ea28aaa5",
+    "type": "feature",
+    "description": "enable s3100continue header config, exempt it from sigV4",
```
aajtodd, line 4: **fix**: There should only be one changelog entry for this entire PR/change.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/S3100Continue.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Add middleware, which adds {Expect: 100-continue} header for s3 client HTTP PUT request larger than 2MB
+ * or with unknown size streaming bodies, during operation builder step
+ */
+public class S3100Continue implements GoIntegration {
+    private static final String ADD_100Continue_Header = "add100Continue";
+    private static final String ADD_100Continue_Header_INTERNAL = "Add100Continue";
+    private static final String Continue_Client_Option = "ContinueHeaderThresholdBytes";
+    private static final Set<String> Put_Op_Set = new HashSet<>(Arrays.asList("PutObject", "UploadPart"));
```
aajtodd, line 30: **nit**: Use the full shape ID rather than just the name (e.g. `com.amazonaws.s3#PutObject` or whatever it actually is).

service/internal/s3shared/s3100continue.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/aws/smithy-go/middleware"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+)
+
+const s3100ContinueID = "S3100Continue"
+const defaultLimit int64 = 1024 * 1024 * 2
```
aajtodd, line 11: **nit**: `default100ContinueThresholdBytes` or similar.

aajtodd (approved):
Fix and ship.

PR #2057, opened 2023-03-17
Add aws standard lib functions

isaiahvita:
This PR adds the standard lib functions which are AWS specific. They were added to an internal package as required by the EP20 SEP.

### DIFF
--- a/internal/endpoints/awsrulesfn/arn.go
+++ b/internal/endpoints/awsrulesfn/arn.go
@@ -3,7 +3,7 @@ package awsrulesfn
 import (
 	"fmt"
 	"strings"
-	smithyerrep "github.com/aws/smithy-go/error/endpoints"
+	smithyrulesfn "github.com/aws/smithy-go/private/endpoints/rulesfn"
 )
 
 // ARN provides AWS ARN components broken out into a data structure.
@@ -32,9 +32,9 @@ const (
 // ParseARN returns an [ARN] value parsed from the input string provided. If
 // the ARN cannot be parsed nil will be returned, and error added to
 // [ErrorCollector].
-func ParseARN(input string, ec *smithyerrep.ErrorCollector) *ARN {
+func ParseARN(input string, ec *smithyrulesfn.ErrorCollector) *ARN {
 	if !strings.HasPrefix(input, arnPrefix) {
-		ec.AddError(smithyerrep.FnError{
+		ec.AddError(smithyrulesfn.FnError{
 			Name: "ParseARN",
 			Err:  fmt.Errorf("invalid ARN prefix, %q", input),
 		})
@@ -43,7 +43,7 @@ func ParseARN(input string, ec *smithyerrep.ErrorCollector) *ARN {
 
 	sections := strings.SplitN(input, arnDelimiters, arnSections)
 	if numSections := len(sections); numSections != arnSections {
-		ec.AddError(smithyerrep.FnError{
+		ec.AddError(smithyrulesfn.FnError{
 			Name: "ParseARN",
 			Err: fmt.Errorf("invalid ARN, not enough sections, %q, %d",
 				input, numSections),
@@ -52,7 +52,7 @@ func ParseARN(input string, ec *smithyerrep.ErrorCollector) *ARN {
 	}
 
 	if sections[sectionPartition] == "" {
-		ec.AddError(smithyerrep.FnError{
+		ec.AddError(smithyrulesfn.FnError{
 			Name: "ParseARN",
 			Err: fmt.Errorf("invalid ARN, partition section cannot be empty, %v",
 				input),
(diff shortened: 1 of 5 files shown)


internal/endpoints/awsrulesfn/generate.go
```diff
@@ -0,0 +1,7 @@
+//go:build codegen
+// +build codegen
+
+package awsrulesfn
+
+//go:generate go run -tags codegen ./internal/partition/codegen.go -model partitions.json -output partitions.go
```
aajtodd, line 6: **Question/Correctness**: Is this actually used? I'm assuming this is related to implementing the `partition` function but I'm not seeing how this is wired together actually and what it generates.

internal/endpoints/awsrulesfn/arn.go
```diff
@@ -0,0 +1,120 @@
...
+		offset += idx + 1
+	}
+
+	return parts
+}
+
+// OptionalStringSlice provides a helper to safely get the index of a string
+// slice that may be out of bounds. Returns pointer to string if index is
+// valid. Otherwise returns nil.
+type OptionalStringSlice []string
```
aajtodd, line 109: **Question**: I see `ResourceId` is set to this type but how is it used in practice? Why do we need the type to be `OptionalStringSlice`?

aajtodd (commented):
Couple of questions, overall looks good.

internal/endpoints/awsrulesfn/arn.go
lucix-aws: I'm assuming the rules generator leverages the receiver method `Get` to simplify the codegen.

lucix-aws: Confirmed--looking at the SEP / S3 rules, the expectation is that you can make a `getAttr` call with an array index, and the return will be null if OOB. This comes into play in practice when cracking open resource IDs in arns.

(we got this for free in kotlin via list `getOrNull`)

internal/endpoints/awsrulesfn/host.go
```diff
@@ -0,0 +1,69 @@
...
+				ec.AddError(smithyerrep.FnError{
+					Name: "IsVirtualHostableS3Bucket",
+					Err:  fmt.Errorf("host label %d cannot have capital letters, %q", i, label),
+				})
+				return false
+			}
+		}
+
+		// Validate not formatted as an IP address
+		// TODO implement
```
lucix-aws, line 56: ? This is done above.

internal/endpoints/awsrulesfn/host.go
```diff
@@ -0,0 +1,69 @@
...
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+
+)
+
+// IsVirtualHostableS3Bucket returns if the input is a DNS compatible bucket
+// name and can be used with Amazon S3 virtual hosted style addressing. Similar
+// to [rulesfn.IsValidHostLabel] with the added restriction that the length of label
+// must be [3:63] characters long, all lowercase, and not formatted as an IP
+// address.
+func IsVirtualHostableS3Bucket(input string, allowSubDomains bool, ec *smithyerrep.ErrorCollector) bool {
```
lucix-aws, line 17: We should reexamine this implementation in general, it should be greatly simplified once `IsValidHostLabel` in smithy-go becomes accessible here.

isaiahvita:
> As in smithy-go, we can't be in `internal` since generated code in `service/*` will use this.

@lucix-aws as discussed offline, this is actually ok since the `internal` directory is at the root of the SDK repository

internal/endpoints/awsrulesfn/generate.go
isaiahvita: gotcha, yah i agree. will remove.

internal/endpoints/awsrulesfn/arn.go
isaiahvita: Yah, I believe Luc is right here. As stated in the SEP, the `getAttr` function is supposed to return a nullable field on indexes into the `resourceId` returned by `parseArn`. So instead of having the implementation of the `getAttr` function handle the Go error for index out of bounds, we contain that error in the `OptionalStringSlice` because as far as the SEP is concerned, it either returns the item or `nil`.

internal/endpoints/awsrulesfn/host.go
isaiahvita: ah yah, thats correct. will remove.

internal/endpoints/awsrulesfn/host.go
isaiahvita: hmm yah i agree that `IsVirtualHostableS3Bucket` shares some logic with `IsValidHostLabel` but it looks like the only thing that could potentially be removed is lines 55-63 (since that case is already checked in `IsValidHostLabel`). But the added benefit of having it also here also could be better error reporting, since it identifies that the invalid host occurred during the check of `IsVirtualHostableS3Bucket`.
The other three things that are checked for in `IsVirtualHostableS3Bucket`, which are not in `IsValidHostLabel` are
1. not IP address
2. not less 3 characters and not longer than 63 characters
3. no capital letters

So not sure how those could be simplified by utilizing `IsValidHostLabel`. Or is there something else that you are seeing that could simplify this?

lucix-aws (approved):

Issue #2042, opened 2023-02-27
dynamodb/attributevalue: allow registration of custom type marshal/unmarshal functions


ztison:
### Describe the feature

As a DynamoDB user, I want to be able to register custom marshaler/unmarshaler functions. 

### Use Case

- I want to have common API types not coupled with the repository layer. So I don't want to implement  `MarshalDynamoDBAttributeValue` and `UnmarshalDynamoDBAttributeValue` functions with dependency on `AttributeValue` in my high-level code.

- I want to implement custom marshaler/unmarshaler for types that  I don't own. It can be required because of data masking or to omit empty values.

### Proposed Solution

```
func (e *Encoder) RegisterMarshaller(t reflect.Type, f func(interface{}) (types.AttributeValue, error)) error
func (d *Decoder) RegisterUnmarshaller(t reflect.Type, f func(types.AttributeValue) (interface{}, error)) error
```

```
type Foo string

func StringMarshaller(value interface{}) (types.AttributeValue, error) {
	if foo, ok := value.(Foo); ok {
		return &types.AttributeValueMemberS{Value: string(foo)}, nil
	}
	return nil, fmt.Errorf("type error")
}

var foo = "foo"
enc.RegisterMarshaller(reflect.TypeOf(foo), StringMarshaller)
```

### Other Information

Based on [customer marshaler/unmarshaler for json](https://github.com/golang/go/issues/5901)

### Rejected Alternatives

API without exposing reflect:
```
func (e *Encoder) RegisterFunc(f interface{})
func (d *Decoder) RegisterFunc(f interface{})
```

API based on generics:
```
func (e *Encoder)  Register[T any](f func(T) (types.AttributeValue, error)) { ... }
func (d *Decoder)  Register[T any](f func(types.AttributeValue) (T, error)) { ... }

```

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue@v1.10.14
github.com/aws/aws-sdk-go-v2/feature/dynamodb/expression@v1.4.39
github.com/aws/aws-sdk-go-v2/service/dynamodb@v1.18.4
github.com/aws/aws-sdk-go-v2/service/dynamodbstreams@v1.14.4

### Go version used

go version go1.19.5 darwin/arm64

raulvc:
[...]

PR #2060, opened 2023-03-21
Custom type marshal/unmarshal functions

ztison:
[dynamodb/attributevalue: allow registration of custom type marshal/unmarshal functions
](https://github.com/aws/aws-sdk-go-v2/issues/2042)

### DIFF
--- a/feature/dynamodb/attributevalue/decode.go
+++ b/feature/dynamodb/attributevalue/decode.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"sync"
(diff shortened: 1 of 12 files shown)


aajtodd:
Hi sorry for the delay and thanks for the PR. Unfortunately we aren't ready to accept contributions for this right now. In particular, while the proposal on which this feature request is based was accepted, it has not made it into a release and there seem to be open questions with the proposed interfaces. We would like to wait until the API stabilizes and makes it into an official release so that IF we end up adding similar capability the API is idiomatic and inline with the `stdlib`.

PR #2076, opened 2023-03-30
add partition generator, and run gofmt

isaiahvita:
Add partition generator via Go-based code generation. Also ran gofmt, so formatting changes are included on unrelated files.

### DIFF
--- a/internal/endpoints/awsrulesfn/partition.go
+++ b/internal/endpoints/awsrulesfn/partition.go
@@ -23,10 +23,14 @@ const defaultPartition = "aws"
 
 // GetPartition returns an AWS [Partition] for the region provided. If the
 // partition cannot be determined nil will be returned.
-func GetPartition(partitions []Partition, region string) *PartitionConfig {
+func GetPartition(region string) *PartitionConfig {
+	return getPartition(partitions, region)
+}
+
+func getPartition(partitions []Partition, region string) *PartitionConfig {
 	for _, partition := range partitions {
 		if v, ok := partition.Regions[region]; ok {
-			v.mergeWith(partition.DefaultConfig)
+			v = mergePartition(v, partition.DefaultConfig)
 			return &v
 		}
 	}
@@ -49,20 +53,21 @@ func GetPartition(partitions []Partition, region string) *PartitionConfig {
 	return nil
 }
 
-func (p *PartitionConfig) mergeWith(other PartitionConfig) {
-	if p.Name == nil {
-		p.Name = other.Name
+func mergePartition(into PartitionConfig, from PartitionConfig) PartitionConfig {
+	if into.Name == nil {
+		into.Name = from.Name
 	}
-	if p.DNSSuffix == nil {
-		p.DNSSuffix = other.DNSSuffix
+	if into.DNSSuffix == nil {
+		into.DNSSuffix = from.DNSSuffix
 	}
-	if p.DualStackDNSSuffix == nil {
-		p.DualStackDNSSuffix = other.DualStackDNSSuffix
+	if into.DualStackDNSSuffix == nil {
+		into.DualStackDNSSuffix = from.DualStackDNSSuffix
 	}
-	if p.SupportsFIPS == nil {
-		p.SupportsFIPS = other.SupportsFIPS
+	if into.SupportsFIPS == nil {
+		into.SupportsFIPS = from.SupportsFIPS
 	}
-	if p.SupportsDualStack == nil {
-		p.SupportsDualStack = other.SupportsDualStack
+	if into.SupportsDualStack == nil {
+		into.SupportsDualStack = from.SupportsDualStack
 	}
+	return into
 }
--- a/internal/endpoints/awsrulesfn/partition_test.go
+++ b/internal/endpoints/awsrulesfn/partition_test.go
@@ -2,7 +2,72 @@ package awsrulesfn
 
 import (
 	"testing"
+	"github.com/aws/smithy-go/ptr"
 )
+
+var mockPartitions = []Partition{
+	{
+		ID:          "aws",
+		RegionRegex: "^(us|eu|ap|sa|ca|me|af)\\-\\w+\\-\\d+$",
+		DefaultConfig: PartitionConfig{
+			Name:               ptr.String("aws"),
+			DNSSuffix:          ptr.String("amazonaws.com"),
+			DualStackDNSSuffix: ptr.String("api.aws"),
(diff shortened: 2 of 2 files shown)


internal/endpoints/awsrulesfn/partitions.go
```diff
@@ -0,0 +1,170 @@
+// Code generated by endpoint/awsrulesfn/internal/partition. DO NOT EDIT.
```
aajtodd, line 1: I'm going to make my comments on this generated file rather than on the codegen which I think will be clearer (just letting you know I realize this is generated).

internal/endpoints/awsrulesfn/partitions.go
```diff
@@ -0,0 +1,170 @@
...
+		p.SupportsFIPS = *vv
+	}
+	if vv := other.SupportsDualStack; vv != nil {
+		p.SupportsDualStack = *vv
+	}
+}
+
+const defaultPartition = "aws"
+
+var partitions = map[string]Partition{
```
aajtodd, line 77: **concern/discuss**: It seems like we take the partition data and split it into 3 different views (this partition map of parition id to base config, region partition map, and partition regexes). Why not just generate a list of partitions and have the `GetPartition` function take in a list of partitions?

This would move most of the runtime code for selecting a partition out of codegen and it would be easier to test (e.g. you can feed it a fake list of partitions rather than be tied to what is generated and may change over time). IIRC this is what we do in Kotlin.

internal/endpoints/awsrulesfn/partitions.json
```diff
@@ -0,0 +1,43 @@
+{
```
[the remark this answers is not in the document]
isaiahvita: came from Jasons original PR. swapped out with updated version

internal/endpoints/awsrulesfn/partitions.go
isaiahvita: I agree that the implementation you described in Kotlin seems like a cleaner implementation, I have modified the codegen to only include generation of the partitions json file into a Go data structure

internal/endpoints/awsrulesfn/partitions.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// GetPartition returns an AWS [Partition] for the region provided. If the
+// partition cannot be determined nil will be returned.
+func GetPartition(region string) *Partition {
+	if v, ok := regionPartition[region]; ok {
+		p, ok := partitions[v.PartitionID]
+		if !ok {
+			panic(fmt.Sprintf(
+				"bad codegen, unable to get region looked up partition %v from partitions map",
```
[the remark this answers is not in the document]
isaiahvita: removed

internal/endpoints/awsrulesfn/partition.go
```diff
@@ -0,0 +1,68 @@
...
+	DualStackDNSSuffix *string `json:"dualStackDnsSuffix"`
+	SupportsFIPS       *bool   `json:"supportsFIPS"`
+	SupportsDualStack  *bool   `json:"supportsDualStack"`
+}
+
+const defaultPartition = "aws"
+
+// GetPartition returns an AWS [Partition] for the region provided. If the
+// partition cannot be determined nil will be returned.
+func GetPartition(partitions []Partition, region string) *PartitionConfig {
```
aajtodd, line 26: **correctness**: The list of partitions is internal (`partitions`). This function should probably be internal and the public facing one wrap it. 

e.g.

```go
func GetPartition(region string): *PartitionConfig {
    return getPartition(partitions, region)
}

func getPartition(partitions []Partition, region string) *PartitionConfig { ... }
```

internal/endpoints/awsrulesfn/partition.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, partition := range partitions {
+		if v, ok := partition.Regions[region]; ok {
+			v.mergeWith(partition.DefaultConfig)
```
aajtodd, line 29: **Question**: I think this looks correct but just to clarify this is creating a copy of the struct from the partitions array and merging into the copy and not the original one right? (i.e. we aren't accidentally mutating the partitions in the partition array)

internal/endpoints/awsrulesfn/partition_test.go
```diff
@@ -0,0 +1,33 @@
...
+			Region: "aws-cn-global", PartitionName: "aws-cn",
+		},
+		"test invalid region; default partition": {
+			Region: "foo", PartitionName: "aws",
+		},
+	}
+
+	for n, c := range cases {
+		t.Run(n, func(t *testing.T) {
+			p := GetPartition(partitions, c.Region)
```
aajtodd, line 25: **Suggestion**: This is probably ok because the partitions don't change often but this test would be more immune to external partition changes if the test was against a mock list of partitions rather than the generated one. A change to `partitions.json` could theoretically break this. Perhaps it's unlikely enough to not matter though.

isaiahvita: I agree, unit tests should be more isolated to external changes. I can make this change since it shouldnt be too heavy of a lift.

internal/endpoints/awsrulesfn/partition.go
isaiahvita: gotcha, yah its correct. i even tested it to confirm its a copy

isaiahvita: since the performance here seems negligible, i altered the function to make the copying more explicit (it was also bothering me that when i quickly glance at it, it makes me do a double take to see if its actually copying)

internal/endpoints/awsrulesfn/partition_test.go
```diff
@@ -0,0 +1,103 @@
...
+			Region: "foo", PartitionName: "aws",
+		},
+	}
+
+	for n, c := range cases {
+		t.Run(n, func(t *testing.T) {
+
+			// monkey patch the partitions data structure
+			// thats used by the GetPartition func
+			partitions = mockPartitions
```
aajtodd, line 93: alternatively you could just call `getPartitions` with the mock partitions since you're in same package

aajtodd (approved):

internal/endpoints/awsrulesfn/partition_test.go
isaiahvita: yah i thought that would have been cleaner too,  but i wanted to test the exported function rather than the unexported one.

lucix-aws (approved):

PR #2077, opened 2023-03-30
add longrunning feature branch to workflows

isaiahvita:
Since this is a feature branch that will last several months, we should be continually running the main github CI to validate correctness

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, ep20 ]
 
 jobs:
   codegen-test:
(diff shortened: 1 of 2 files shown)


.github/workflows/codegen.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, ep20 ]
```
aajtodd, line 7: **fix**: This branch should just be renamed and use a wildcard e.g. [feat-*](https://github.com/awslabs/aws-sdk-kotlin/blob/main/.github/workflows/continuous-integration.yml#L9). Same thing in smithy-go which was already merged.

aajtodd (changes requested):

aajtodd:
OBE by https://github.com/aws/aws-sdk-go-v2/pull/2141

Issue #2025, opened 2023-02-15
Add support for recursion detection


aajtodd:
### Describe the feature

Implement support for the recursion detection SEP

### Use Case

see `recursion-detection.md` SEP

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

N/A

### Go version used

N/A

lucix-aws:
Completed by #2105

PR #2105, opened 2023-04-21
Add Recursion Detection middleware to all SDK requests

wty-Bryant:
Add Recursion Detection middleware to all SDK requests, which detects Lambda environment and sets its X-Ray trace ID to request header if absent to avoid recursion invocation in Lambda environment. Solves #2025

### DIFF
--- a/aws/middleware/recursion_detection.go
+++ b/aws/middleware/recursion_detection.go
@@ -9,8 +9,8 @@ import (
 )
 
 const envAwsLambdaFunctionName = "AWS_LAMBDA_FUNCTION_NAME"
-const envAmznTraceId = "_X_AMZN_TRACE_ID"
-const amznTraceIdHeader = "X-Amzn-Trace-Id"
+const envAmznTraceID = "_X_AMZN_TRACE_ID"
+const amznTraceIDHeader = "X-Amzn-Trace-Id"
 
 // AddRecursionDetection adds recursionDetection to the middleware stack
 func AddRecursionDetection(stack *middleware.Stack) error {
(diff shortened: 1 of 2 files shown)


aws/middleware/middleware_test.go
```diff
@@ -187,3 +188,98 @@ func TestAttemptClockSkewHandler(t *testing.T) {
...
+			TraceID:        "traceID4-=;:+&[]{}\"'",
+			HeaderAfter:    "traceID4-=;:+&[]{}\"'",
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			// has to pre-check if current os has lambda function and trace ID environment variable
+			// if exists, need to restore them after each test case
+			initialLambdaFunc, hasInitialLambdaFunc := os.LookupEnv(lambdaFunc)
```
aajtodd, line 234: It's unfortunate that the testing utils that already [exist](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/awstesting/util.go#L97) for this depend on `aws` package. 

I think we can simplify this by just clearing the environment and restoring it after the test such that each test gets a clean environment, [example](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/middleware/user_agent_test.go#L120)

Then you won't need the additional `setEnvVar`/`recordEnvVar` functions and we guarantee tests are isolated

aws/middleware/middleware_test.go
```diff
@@ -187,3 +188,98 @@ func TestAttemptClockSkewHandler(t *testing.T) {
 		})
 	}
 }
+
+func TestRecursionDetection(t *testing.T) {
+	const lambdaFunc = "AWS_LAMBDA_FUNCTION_NAME"
```
aajtodd, line 193: **suggestion**: You could define these as internal constants in the middleware and the test can just re-use them

e.g.

```go
// file: recursion_detection.go

const envAwsLambdaFunctionName = "AWS_LAMBDA_FUNCTION_NAME"
const envAmznTraceId = "_X_AMZN_TRACE_ID"
const amznTraceIdHeader = "X-Amzn-Trace-Id"
```

aajtodd (changes requested):
Looks pretty good overall!

lucix-aws (approved):

aajtodd (approved):

PR #2110, opened 2023-04-27
Add aws library functions for endpoint rules engine and fix partition generation

isaiahvita:
- endpoints2.0 needed aws library functions and they dont belong in the smithy-go repository.
- the partition function needed two different types for "partition config" this is because the default partition config should be guaranteed to always have values so those members needed to be Go value types. but the region overrides needed to be nullable.

### DIFF
--- a/service/s3/endpoints.go
+++ b/service/s3/endpoints.go
@@ -11,9 +11,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/internal/endpoints/awsrulesfn"
 	internalendpoints "github.com/aws/aws-sdk-go-v2/service/s3/internal/endpoints"
 	smithy "github.com/aws/smithy-go"
+	smithyendpoints "github.com/aws/smithy-go/endpoints"
 	"github.com/aws/smithy-go/endpoints/private/rulesfn"
 	"github.com/aws/smithy-go/middleware"
-	smithytransport "github.com/aws/smithy-go/transport"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
 	"net/http"
 	"net/url"
@@ -212,8 +212,8 @@ func finalizeClientEndpointResolverOptions(options *Options) {
 
 }
 
-// EndpointParameters provides the option parameters for influence how endpoints
-// are resolved.
+// EndpointParameters provides the parameters that influence how endpoints are
+// resolved.
 type EndpointParameters struct {
 	// The S3 bucket used to send the request. This is an optional parameter that will
 	// be set automatically for operations that are scoped to an S3 bucket.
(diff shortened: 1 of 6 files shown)


internal/endpoints/awsrulesfn/partitions.go
```diff
@@ -2,208 +2,210 @@
...
+			SupportsFIPS:       true,
+			SupportsDualStack:  true,
 		},
-		Regions: map[string]PartitionConfig{
+		Regions: map[string]RegionOverrides{
 			"af-south-1": {
 				Name:               nil,
-				DNSSuffix:          nil,
-				DualStackDNSSuffix: nil,
+				DnsSuffix:          nil,
```
aajtodd, line 25: **nit**: `nil` is the default value, we could generate only the overrides and not have to explicitly set them. This would shorten the region maps and make the generated partitions slightly more readable as the overrides would be more clear.

aajtodd (approved):

service/s3/endpoints.go
```diff
@@ -206,3 +211,6590 @@ func finalizeClientEndpointResolverOptions(options *Options) {
 	}
 
 }
+
+// EndpointParameters provides the option parameters for influence how endpoints
```
aajtodd, line 215: **fix**: Update codegen to fix grammar `provides the parameters that influence ...`

service/s3/endpoints.go
```diff
@@ -206,3 +211,6590 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+	// returning the endpoint if found. Otherwise an error is returned.
+	ResolveEndpoint(ctx context.Context, options EndpointParameters) (
+		smithytransport.Endpoint, error,
+	)
+}
+
+// resolverV2 provides the implementation for resolving endpoints.
+type resolverV2 struct{}
+
+func NewDefaultEndpointResolverV2() *resolverV2 {
```
aajtodd, line 351: **fix**: Type should be `EndpointResolverV2` interface not the concrete type right?

service/s3/endpoints.go
```diff
@@ -206,3 +211,6590 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+	}
+	_UseFIPS := *params.UseFIPS
+	_UseDualStack := *params.UseDualStack
+	_Accelerate := *params.Accelerate
+	_UseGlobalEndpoint := *params.UseGlobalEndpoint
+	_DisableMultiRegionAccessPoints := *params.DisableMultiRegionAccessPoints
+
+	if exprVal := params.Region; exprVal != nil {
+		_Region := *exprVal
+		_ = _Region
```
aajtodd, line 373: **question**: Why is this assignment necessary in each of these conditions again?

service/s3/endpoints.go
```diff
@@ -206,3 +211,6590 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+											}()
+
+											uri, err := url.Parse(uriString)
+											if err != nil {
+												return endpoint, fmt.Errorf("Failed to parse uri: %s", uriString)
+											}
+
+											return smithytransport.Endpoint{
+												URI:     *uri,
+												Headers: &http.Header{},
```
aajtodd, line 457: **suggestion**: No need to allocate here if there are no headers

aajtodd (approved):
Left a few comments on the generated code rather than tracking it down in codegen. Overall looks good.

Fix and ship

service/s3/endpoints.go
isaiahvita: yah so heres the original thread discussing this
https://github.com/aws/smithy-go/pull/424#discussion_r1182802251

service/s3/endpoints.go
isaiahvita: so letting this default to nil makes the generation logic slightly more complicated because omitting this line entirely (rather than creating the empty struct) requires some changes up an abstraction layer and its breaking things. since this is a suggestion, im going to add a TODO for now to reconsider

Issue #2115, opened 2023-05-05
(short issue description)


andrewmeissner:
### Describe the issue

Documentation says GetCredentialReportOutput.Content is base64 encoded, but it isn't.

### Links

Doc for the field: https://github.com/aws/aws-sdk-go-v2/blob/service/iam/v1.19.12/service/iam/api_op_GetCredentialReport.go#L40

This is where the Content field is being set, already decoded: https://github.com/aws/aws-sdk-go-v2/blob/service/iam/v1.19.12/service/iam/deserializers.go#L27216

### AWS Go SDK V2 Module Versions Used

$ go mod graph | grep -i github.com/aws 

creds github.com/aws/aws-sdk-go-v2@v1.18.0
creds github.com/aws/aws-sdk-go-v2/config@v1.18.23
creds github.com/aws/aws-sdk-go-v2/credentials@v1.13.22
creds github.com/aws/aws-sdk-go-v2/feature/ec2/imds@v1.13.3
creds github.com/aws/aws-sdk-go-v2/internal/configsources@v1.1.33
creds github.com/aws/aws-sdk-go-v2/internal/endpoints/v2@v2.4.27
creds github.com/aws/aws-sdk-go-v2/internal/ini@v1.3.34
creds github.com/aws/aws-sdk-go-v2/service/iam@v1.19.12
creds github.com/aws/aws-sdk-go-v2/service/internal/presigned-url@v1.9.27
creds github.com/aws/aws-sdk-go-v2/service/sso@v1.12.10
creds github.com/aws/aws-sdk-go-v2/service/ssooidc@v1.14.10
creds github.com/aws/aws-sdk-go-v2/service/sts@v1.18.11
creds github.com/aws/smithy-go@v1.13.5
github.com/aws/aws-sdk-go-v2@v1.18.0 github.com/aws/smithy-go@v1.13.5
github.com/aws/aws-sdk-go-v2@v1.18.0 github.com/google/go-cmp@v0.5.8
github.com/aws/aws-sdk-go-v2@v1.18.0 github.com/jmespath/go-jmespath@v0.4.0
github.com/aws/aws-sdk-go-v2/config@v1.18.23 github.com/aws/aws-sdk-go-v2@v1.18.0
github.com/aws/aws-sdk-go-v2/config@v1.18.23 github.com/aws/aws-sdk-go-v2/credentials@v1.13.22
[...]

PR #2136, opened 2023-05-24
Removing misleading GoDoc comment

Ahmadkashif:
Solves #2115


aajtodd:
This file is auto generated from a model, the documentation needs to be fixed upstream with the service model. Closing.

PR #2141, opened 2023-05-26
run workflows on feature branches

isaiahvita:
Amend github workflow config to run workflows on feature branches along with main. Pulled syntax from the github documentation page
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,9 +2,13 @@ name: Go Tests
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
+      - 'feat-**'
   pull_request:
-    branches: [ main ]
+    branches:
+      - main
+      - 'feat-**'
 
 env:
   EACHMODULE_CONCURRENCY: 2


.github/workflows/go.yml
```diff
@@ -2,9 +2,13 @@ name: Go Tests
 
 on:
   push:
-    branches: [ main ]
+    branches:
+      - main
+      - 'feat-**'
```
aajtodd, line 7: Should this just be `feat-*` (one *)?

aajtodd (approved):

.github/workflows/go.yml
isaiahvita: in github syntax, `**` to matches any character. `*` matches any character except `/`.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

 im fine with either, i kind of just copied one of the examples in the github documentation page which used `**`

wty-Bryant (approved):

PR #2146, opened 2023-06-05
Add client config adapters

isaiahvita:
To handle backwards compatibility constraints with old resolver interfaces, the new `EndpointResolverV2` needs to be wrapped in certain ways (adapters) before being initialized onto the client config. This PR implements these adapters.

### DIFF
--- a/service/s3/api_client.go
+++ b/service/s3/api_client.go
@@ -100,9 +100,8 @@ type Options struct {
 
 	// The service endpoint resolver.
 	//
-	// Deprecated: Deprecated: EndpointResolver is deprecated along with its
-	// corresponding With__ helper. See EndpointResolverV2 and corresponding With__
-	// helper
+	// Deprecated: EndpointResolver and WithEndpointResolver are deprecated. See
+	// EndpointResolverV2 and WithEndpointResolverV2
 	EndpointResolver EndpointResolver
 
 	// Resolves the endpoint used for a particular service. This should be used over
@@ -193,8 +192,8 @@ func WithAPIOptions(optFns ...func(*middleware.Stack) error) func(*Options) {
 	}
 }
 
-// Deprecated: EndpointResolver is deprecated along with its corresponding With__
-// helper. See EndpointResolverV2 and corresponding With__ helper
+// EndpointResolver and WithEndpointResolver are deprecated. See
+// EndpointResolverV2 and WithEndpointResolverV2
 func WithEndpointResolver(v EndpointResolver) func(*Options) {
 	return func(o *Options) {
 		o.EndpointResolver = v
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEndpointGenerator.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEndpointGenerator.java
@@ -72,10 +72,11 @@ public final class AwsEndpointGenerator implements GoIntegration {
                                         .documentation(String.format("The service endpoint resolver."))
                                         .deprecated(String.format(
                                             """
-                                            Deprecated: %s is deprecated along with its corresponding With__ helper.
-                                            See %s and corresponding With__ helper
+                                            %s and With%s are deprecated. See %s and With%s
                                             """,
                                             EndpointGenerator.RESOLVER_INTERFACE_NAME,
+                                            EndpointGenerator.RESOLVER_INTERFACE_NAME,
+                                            EndpointResolutionGenerator.RESOLVER_INTERFACE_NAME,
                                             EndpointResolutionGenerator.RESOLVER_INTERFACE_NAME
                                         ))
                                         .withHelper(true)
--- a/service/s3/go.mod
+++ b/service/s3/go.mod
@@ -3,15 +3,15 @@ module github.com/aws/aws-sdk-go-v2/service/s3
 go 1.15
 
 require (
-	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v0.0.0-00010101000000-000000000000
-	github.com/aws/aws-sdk-go-v2/internal/configsources v0.0.0-00010101000000-000000000000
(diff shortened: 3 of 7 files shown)


service/s3/endpoints.go
```diff
@@ -213,6 +215,127 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+		return smithyendpoints.Endpoint{
+			URI: *uri,
+		}, nil
+	}
+
+	if resolveEndpoint.Source == aws.EndpointSourceServiceMetadata {
+		return l.resolver.ResolveEndpoint(ctx, params)
+	}
+
+	params.UseDualStack = nil
```
lucix-aws, line 259: A few questions:
1. Shouldn't these booleans be set to their defaults instead?
2. How are we going to handle service-specific (`clientContextParam`) and operation-specific (`contextParam` / `staticContextParam`) bindings in this adapter?

service/s3/endpoints.go
```diff
@@ -213,6 +215,127 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+			pr, fipsInfix, "-"), fipsPrefix, ""), fipsSuffix, "")
+		fips = aws.FIPSEndpointStateEnabled
+	} else {
+		region = pr
+	}
+
+	return region, fips
+}
+
+func finalizeEndpointResolverV2(options *Options) {
```
lucix-aws, line 321: Is this needed for something down the road? I'm not seeing it used anywhere, only generated.

service/s3/api_client.go
```diff
@@ -193,6 +196,14 @@ func WithEndpointResolver(v EndpointResolver) func(*Options) {
 	}
 }
 
+// WithEndpointResolverV2 returns a functional option for setting the Client's
+// EndpointResolverV2 option.
+func WithEndpointResolverV2(v EndpointResolverV2) func(*Options) {
```
aajtodd, line 201: The existing `WithEndpointResolver` should be deprecated as well right?

service/s3/endpoints.go
```diff
@@ -213,6 +215,127 @@ func finalizeClientEndpointResolverOptions(options *Options) {
...
+	endpoint = aws.Endpoint{
+		URL:               resolved.URI.String(),
+		HostnameImmutable: false,
+		Source:            aws.EndpointSourceServiceMetadata,
+	}
+
+	return endpoint, nil
+}
+
+func mapPseudoRegion(pr string) (region string, fips aws.FIPSEndpointState) {
```
aajtodd, line 303: **suggestion** This could use some documentation on why it exists (I'm _assuming_ because the legacy resolver allowed pseudo regions but the new resolver doesn't need them so we have to map them to the appropriate region + param).

service/s3/endpoints.go
aajtodd: For (2) aren't the params already bound?

service/s3/endpoints.go
```diff
@@ -49,7 +49,9 @@ func resolveDefaultEndpointConfiguration(o *Options) {
 	if o.EndpointResolver != nil {
 		return
 	}
-	o.EndpointResolver = NewDefaultEndpointResolver()
+	o.EndpointResolver = &compatibleEndpointResolver{
```
aajtodd, line 52: Question: Is this necessary? If this is setting the default config couldn't we just set `EndpointResolverV2` directly and set `EndpointResolver` to `nil`? Maybe I'm missing something 🤔 .

service/s3/endpoints.go
isaiahvita: 1. yah i could see it going either way.  i think setting them to `nil` signals that this input was already used when resolving a URL. but i can make this change for now
2. its handled a layer above in the middleware. i can explain this in more depth offline

service/s3/endpoints.go
isaiahvita: yah this will be needed later down the road. ill add it to this PR

service/s3/endpoints.go
isaiahvita: good point, i think the reason to actually have it set (rather than nil) is if there is customer code that is relying on that value to actually be set and not `nil`. 
i can see a benefit in removing it, since i dont think its going to be directly used by our code. but this might be outweighed by the reason above. do you see another benefit to setting it to nil?

service/s3/endpoints.go
lucix-aws: Yeah you're right on (2) - I missed that it was being passed into this function.

service/s3/endpoints.go
aajtodd: No it just struck me as odd to not be setting resolver v2 in "set defaults" function since that should now be our default. I can see the value in ensuring there is something set though if it's relied upon elsewhere that we aren't aware of.

service/s3/endpoints.go
isaiahvita: gotcha, will add documentation. and yah your assumption seems to be correct
https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/endpoints.go#L183

service/s3/endpoints.go
isaiahvita: ah right, so i believe the question is why are we setting `Options.EndpointResolverV2` in `finalizeEndpointResolverV2` rather than `resolveDefaultEndpointConfiguration`. and i believe its because theres a concept of "resolving" a setting vs "finalizing" a setting where the former happens before customer mutation via api options, and the latter happens after customer mutation.

https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/integration/ConfigFieldResolver.java#L114
https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_client.go#L212

and we need `EndpointResolverV2` to be set after all customer mutations to `Options.EndpointResolver` has been finished, so we can detect whether its a customer provided resolver or the one that we provide

lucix-aws (approved):

aajtodd (approved):

PR #2147, opened 2023-06-07
Make S3 SelectObjectContentInput ScanRange members pointers

cad:
### Problem

**TLDR: Int64 members `Start` and `End` of `ScanRange` for the  `SelectObjectContentInput` are omitted when serialized if they are 0.**

When calling SelectObjectContent, you can specify a ScanRange to only request
records between start and end positions.

For example we want to process only the records starting from the 0th byte to the 100th byte we can have a Go code like this:

```go
cfg, err := config.LoadDefaultConfig(context.TODO(),
    config.WithClientLogMode(aws.LogRequestWithBody|aws.LogResponse), // Enable debug logging API calls
)
if err != nil {
    panic(err)
}

// Create S3 client
client := s3.NewFromConfig(cfg)


// Send a SelectObjectContent request
if _, err := s3Client.SelectObjectContent(ctx, &s3.SelectObjectContentInput{
    Bucket: aws.String("bucket"),
    Key:    aws.String("key"),
    Expression: aws.String("SELECT * FROM S3Object"),
    ExpressionType: s3.ExpressionTypeSql,
    InputSerialization: &s3.InputSerialization{
        JSON: &s3.JSONInput{
            Type: s3.JSONTypeDocument,
        },
    },
    OutputSerialization: &s3.OutputSerialization{
        JSON: &s3.JSONOutput{},
    },
    ScanRange: &s3.ScanRange{ // Expected to get records starting from the 0th byte to the 100th byte
        Start: 0, 
        End:   100,
    },
}); err != nil {
    panic(err)
}
```

Expected API call should have a ScanRange section that looks like this:

```xml
...
<ScanRange>
    <End>100</End>
    <Start>0</Start>
</ScanRange>
...
```

But when checked the log output, it looks like this:

```xml
...
<ScanRange>
    <End>100</End>
</ScanRange>
...
```

Which meant, *"process only the records within the last 100 bytes of the file"*
according to the
[docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#AmazonS3-SelectObjectContent-request-ScanRange).

**So there is no way to specify the start position as 0 and end position as an arbitrary number, of the ScanRange at the current state of the SDK.**

#### Hypothesis

The client just omits the `Start` or `End` member of the `ScanRange` shape when it is
equal to 0. Since it is an unboxed type and its empty value is 0 it is not serialized to the request.

### Solution

Added a customization that adds `clientOptional` trait to the `Start` and `End`
members of the S3 `ScanRange` shape. And regenerate the client code to make
these members optional.

Since this will break any existing application that specifies `ScanRange` with `Start` and `End` fields set, an announcement is added to the changelog regarding the breaking change.

### DIFF
--- a/.changelog/0f5bdd8a631d4196bba7718e0f50e426.json
+++ b/.changelog/0f5bdd8a631d4196bba7718e0f50e426.json
@@ -1,0 +1,8 @@
+{
+    "id": "0f5bdd8a-631d-4196-bba7-718e0f50e426",
+    "type": "bugfix",
+    "description": "Fix the issue in the Amazon S3 SelectObjectContent method's parameter serialization where the Start and End parameters of the ScanRange were not being serialized correctly. The Amazon S3 API client has been updated with a breaking change to make ScanRange members pointer types.",
+    "modules": [
+        "service/s3"
+    ]
+}
--- a/.changelog/1a2c46b4f6044155b3dd5979b26fd38e.json
+++ b/.changelog/1a2c46b4f6044155b3dd5979b26fd38e.json
@@ -1,0 +1,8 @@
+{
+    "id": "1a2c46b4-f604-4155-b3dd-5979b26fd38e",
+    "type": "announcement",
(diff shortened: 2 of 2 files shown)


lucix-aws:
This is actually part of a larger-scale issue with default value serialization. See https://github.com/aws/aws-sdk-go-v2/issues/2162.

We can't take this PR in isolation due to the breaking change, but we're investigating the issue at large to try to determine a more comprehensive way forward.

PR #2154, opened 2023-06-23
Modify user agent header syntax and add support for app ID

wty-Bryant:
Modify http request User-Agent header syntax and add support for optional app ID section in the UA header.
Customer code could offer an App ID for identifying their applications, which can be obtained from following sources in descending priority:
1. Client configuration
2. Environment variables with name of AWS_SDK_UA_APP_ID
3. Shared config profile field called sdk_ua_app_id

Once a service operation is called with app ID set, it will be concatenated to UA header string in format `{app/appID}`. One example UA string calling s3 service is shown here:
`User-Agent: aws-sdk-go-v2/1.18.1 os/macos lang/go#1.19.5 md/GOOS#darwin md/GOARCH#amd64 api/s3#1.36.0 app/123`

### DIFF
--- a/service/appstream/api_op_AssociateAppBlockBuilderAppBlock.go
+++ b/service/appstream/api_op_AssociateAppBlockBuilderAppBlock.go
@@ -89,7 +89,7 @@ func (c *Client) addOperationAssociateAppBlockBuilderAppBlockMiddlewares(stack *
 	if err = awsmiddleware.AddRecordResponseTiming(stack); err != nil {
 		return err
 	}
-	if err = addClientUserAgent(stack); err != nil {
+	if err = addClientUserAgent(stack, options); err != nil {
 		return err
 	}
 	if err = smithyhttp.AddErrorCloseResponseBodyMiddleware(stack); err != nil {
--- a/service/appstream/api_op_CreateAppBlockBuilder.go
+++ b/service/appstream/api_op_CreateAppBlockBuilder.go
@@ -140,7 +140,7 @@ func (c *Client) addOperationCreateAppBlockBuilderMiddlewares(stack *middleware.
 	if err = awsmiddleware.AddRecordResponseTiming(stack); err != nil {
 		return err
 	}
-	if err = addClientUserAgent(stack); err != nil {
+	if err = addClientUserAgent(stack, options); err != nil {
 		return err
 	}
 	if err = smithyhttp.AddErrorCloseResponseBodyMiddleware(stack); err != nil {
--- a/service/appstream/api_op_CreateAppBlockBuilderStreamingURL.go
+++ b/service/appstream/api_op_CreateAppBlockBuilderStreamingURL.go
@@ -91,7 +91,7 @@ func (c *Client) addOperationCreateAppBlockBuilderStreamingURLMiddlewares(stack 
 	if err = awsmiddleware.AddRecordResponseTiming(stack); err != nil {
 		return err
 	}
-	if err = addClientUserAgent(stack); err != nil {
+	if err = addClientUserAgent(stack, options); err != nil {
 		return err
 	}
 	if err = smithyhttp.AddErrorCloseResponseBodyMiddleware(stack); err != nil {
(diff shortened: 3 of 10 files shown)


aws/middleware/user_agent.go
```diff
@@ -74,7 +79,7 @@ type requestUserAgent struct {
 // X-Amz-User-Agent example:
 //
 //	aws-sdk-go-v2/1.2.3 md/GOOS/linux md/GOARCH/amd64 lang/go/1.15
-func newRequestUserAgent() *requestUserAgent {
+func newRequestUserAgent(appID ...string) *requestUserAgent {
```
aajtodd, line 82: **fix**: `appId` is singular, also this is forcing us to know/care about appId when creating the user agent structure.

There is already a `SDKAgentKeyType` of type `ApplicationIdentifier` for this purpose. We should revert this back and just make an explicit call in setting up the middleware to add two keys. I'll expand elsewhere.

service/accessanalyzer/api_client.go
```diff
@@ -354,8 +358,8 @@ func resolveAWSEndpointResolver(cfg aws.Config, o *Options) {
 	o.EndpointResolver = withEndpointResolver(cfg.EndpointResolver, cfg.EndpointResolverWithOptions, NewDefaultEndpointResolver())
 }
 
-func addClientUserAgent(stack *middleware.Stack) error {
-	return awsmiddleware.AddSDKAgentKeyValue(awsmiddleware.APIMetadata, "accessanalyzer", goModuleVersion)(stack)
+func addClientUserAgent(stack *middleware.Stack, options Options) error {
+	return awsmiddleware.AddSDKAgentKeyValue(awsmiddleware.APIMetadata, "accessanalyzer", goModuleVersion, options.AppID)(stack)
```
aajtodd, line 362: Rather than force AppId through this API and abuse variadic arguments we can just call it twice I believe.

e.g.

```go
if err := awsmiddleware.AddSDKAgentKeyValue(awsmiddleware.APIMetadata, "accessanalyzer", goModuleVersion)(stack); err != nil { 
    return err
}

if options.AppID != nil {
    return awsmiddleware.AddSDKAgentKey(awsmiddleware.ApplicationIdentifier, options.AppID)(stack)
}

return nil
```

aajtodd (changes requested):

aws/config.go
```diff
@@ -132,6 +132,13 @@ type Config struct {
 	// `config.LoadDefaultConfig`. You should not populate this structure
 	// programmatically, or rely on the values here within your applications.
 	RuntimeEnvironment RuntimeEnvironment
+
+	// The AppID could be retrieved from environment variable AWS_SDK_UA_APP_ID
```
aajtodd, line 136: > AppId is an optional application specific identifier that can be set. When set it will be appended to the User-Agent header of every request in the form of `App/{AppId}`. This variable is sourced from environment variable `AWS_SDK_UA_APP_ID` or the shared config profile attribute `sdk_ua_app_id`. See https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html for more information on environment variables and shared config settings.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -199,6 +201,13 @@ public class AddAwsConfigFields implements GoIntegration {
                     .name(LOG_MODE_CONFIG_NAME)
                     .type(getAwsCoreSymbol("ClientLogMode"))
                     .documentation("Configures the events that will be sent to the configured logger.")
+                    .build(),
+            AwsConfigField.builder()
+                    .name(SDK_APP_ID)
+                    .type(getUniversalSymbol("string"))
+                    .documentation("The aws sdk app ID that could be retrieved from env var or shared config file" +
```
aajtodd, line 208: This doesn't match the current generated setting:
> The sdk user agent app id to be added to http request's User-Agent header
	AppID string


Maybe need to regenerate?

**suggestion**: 
> The (optional) application specific identifier appended to the  User-Agent header.

aajtodd (approved):
Approved, but wait for second review and green light to merge.

lucix-aws (approved):

PR #2158, opened 2023-06-25
Add endpoint-based auth scheme resolution

isaiahvita:
This PR adds auth scheme resolution at endpoint resolution time to satisfy endpoints2.0 requirements.


service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +209,52 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
 		)
 	}
 
+	authSchemes, err := internalauth.GetAuthenticationSchemes(&resolvedEndpoint.Properties)
+	if err != nil || len(authSchemes) == 0 {
```
aajtodd, line 213: **fix**: The rules are not required to contain auth schemes, the logic here should just update the context as needed _if_ there is an auth scheme we support

service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +209,52 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
...
+		return out, metadata, fmt.Errorf("Failed to resolve authentication scheme")
+	}
+
+	supportedAuthSchemeFound := false
+	for _, authScheme := range authSchemes {
+		name := authScheme.GetName()
+		_, supportedAuthSchemeFound = internalauth.SupportedSchemes[name]
+		if supportedAuthSchemeFound {
```
aajtodd, line 221: **nit/style**:

We don't really need the name (or a set of supported auth schemes for that matter) if we are going to do a type assertion over the same set.

e.g.


```go
switch v := authScheme.(type) {
    case *internalauth.AuthenticationSchemeV4:
        if v.SigningName != nil {
            ctx = awsmiddleware.SetSigningName(ctx, v.SigningName)
        }
        if v.SigningRegion != nil {
            ctx = awsmiddleware.SetSigningRegion(ctx, v.SigningRegion)
        }
    case *internalauth.AuthenticationSchemeV4A:
}
```


Alternatively you could just move some of this into the runtime and let each auth scheme handle updating the context:

```go
interface AuthenticationScheme {
    fun SetAuthContext(context.Context): context.Context
}

func (s *AuthenticationSchemeV4) SetAuthContext(ctx context.Context): context.Context {
        if s.SigningName != nil {
            ctx = awsmiddleware.SetSigningName(ctx, s.SigningName)
        }

        if s.SigningRegion != nil {
            ctx = awsmiddleware.SetSigningRegion(ctx, s.SigningRegion)
        }

        return ctx
}
```

service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +209,52 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
...
+				ctx = awsmiddleware.SetSigningName(ctx, signingName)
+				ctx = awsmiddleware.SetSigningRegion(ctx, v4aScheme.SigningRegionSet[0])
+				break
+			}
+			if name == internalauth.None {
+				break
+			}
+		}
+	}
+	if !supportedAuthSchemeFound {
```
aajtodd, line 250: **fix**: Again the endpoints aren't required to have an auth scheme, if one is found it basically can help influence signing. Otherwise it should just do what it normally would do.

aajtodd (changes requested):

service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +210,54 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
 		)
 	}
 
+	authSchemes, err := internalauth.GetAuthenticationSchemes(&resolvedEndpoint.Properties)
+	if err != nil {
+		var nfe *internalauth.NoAuthenticationSchemesFoundError
+		if errors.As(err, &nfe) {
```
aajtodd, line 216: **correctness**: What if the error _isn't_ `NoAuthenticationSchemesFoundError`? It looks like we're just dropping the error.

service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +210,54 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
...
+		}
+		authSchemes = []internalauth.AuthenticationScheme{}
+	}
+
+	for _, authScheme := range authSchemes {
+		switch v := authScheme.(type) {
+		case *internalauth.AuthenticationSchemeV4:
+			v4Scheme, _ := authScheme.(*internalauth.AuthenticationSchemeV4)
+			var signingName, signingRegion string
+			if v4Scheme.SigningName == nil {
```
aajtodd, line 232: **comment**: If these are always the default (and I believe they are) we don't really need a no auth schemes found error/branch. We can just always set the defaults. 

e.g.

```go

ctx = awsmiddleware.SetSigningName(ctx, "lambda")
ctx = awsmiddleware.SetSigningRegion(ctx,  signingRegion)

authSchemes, err := internalauth.GetAuthenticationSchemes(...)
if err != nil { 
     return err
}

for _, authScheme := rangeAuthSchemes {
     switch v := authScheme.(type) {
            ...
     }
}

```

service/lambda/api_op_AddLayerVersionPermission.go
```diff
@@ -208,6 +210,54 @@ func (m *opAddLayerVersionPermissionResolveEndpointMiddleware) HandleSerialize(c
...
+			var signingName string
+			if v4aScheme.SigningName == nil {
+				signingName = "lambda"
+			}
+			ctx = awsmiddleware.SetSigningName(ctx, signingName)
+			ctx = awsmiddleware.SetSigningRegion(ctx, v4aScheme.SigningRegionSet[0])
+			break
+		case *internalauth.AuthenticationSchemeNone:
+			break
+		default:
```
aajtodd, line 252: **correctness**: This will error out possibly before exhausting the set of auth schemes.

e.g. if the auth schemes returned looked like: `[notSupported, sigv4]` we would fail too soon. The default case should just be a `continue` and you'll have to keep track of whether a valid authscheme was found via another variable.

e.g.

```go
foundSupportedAuthScheme := false

for _, authScheme in range authSchemes {
      case supported1:
          foundSupportedAuthScheme = true
          ...
      case supported2:
          foundSupportedAuthScheme = true
          ...
}

if len(authSchemes) >= && !foundSupportedAuthScheme {
    return out, metadata, fmt.Errorf("....")
}
```

aajtodd (changes requested):

service/lambda/api_op_AddLayerVersionPermission.go
isaiahvita: yah these are always the defaults. i believe the no auth schemes found error is part of the stated requirements.

service/lambda/api_op_AddLayerVersionPermission.go
isaiahvita: doh yah thats right ... brain was a bit fried last night, will fix

service/lambda/api_op_AddLayerVersionPermission.go
isaiahvita: i do like this. and attempted this. but there were a couple of hiccups:
1. there is a s3 specific context setter `s3cust.SetSignerVersion` that i would need to import into this package. and doing so, would result in an import cycle: s3cust package imports a v4a package which imports this package (in order to use the context getters i define in this `internal/auth` package)
2. the signing name and signing region defaults are service specific, so a shared library function like `SetAuthContext` wouldnt be as clean as taking in a single `ctx` param. and will potentially get messy with the fact that sigv4 and sigv4a use different data structures to model signing region (`string` vs `[]string`).

given, that this endpoint-based auth scheme resolution code is likely to be replace with the auth scheme resolution refactor work in the future. im not sure if much value is gained in style optimizing it.

aajtodd: > given, that this endpoint-based auth scheme resolution code is likely to be replaced with the auth scheme resolution refactor work in the future. im not sure if much value is gained in style optimizing it.

Endpoints will still be allowed to influence signing (they just won't eventually be setting e.g. signer version). Let's ignore that for now though to ship EP2.0.

> there is a s3 specific context setter s3cust.SetSignerVersion that i would need to import into this package. and doing so, would result in an import cycle: s3cust package imports a v4a package which imports this package (in order to use the context getters i define in this internal/auth package)

It looks like we are duplicating this logic in [event bridge](https://github.com/isaiahvita/aws-sdk-go-v2/blob/feat-ep20-authscheme/service/eventbridge/internal/customizations/signer_wrapper.go#L28)

> the signing name and signing region defaults are service specific, so a shared library function like SetAuthContext wouldnt be as clean as taking in a single ctx param. and will potentially get messy with the fact that sigv4 and sigv4a use different data structures to model signing region (string vs []string)

I'm not sure I understand this, the function would be a method on the authentication scheme allowing each to set context properties appropriate for the scheme.

internal/v4a/middleware.go
```diff
@@ -71,10 +71,23 @@ func (s *SignHTTPRequestMiddleware) HandleFinalize(
...
+	signerOptions := []func(o *SignerOptions){
+		func(o *SignerOptions) {
+			o.Logger = middleware.GetLogger(ctx)
+			o.LogSigning = s.logSigning
+		},
+	}
+
+	// existing DisableURIPathEscaping is equivalent in purpose
+	// to authentication scheme property DisableDoubleEncoding
+	disableDoubleEncoding, overridden := internalauth.GetDisableDoubleEncoding(ctx)
```
aajtodd, line 83: **correctness**: This change to allow overriding this setting should apply to v4 as well should it not?

internal/auth/scheme.go
```diff
@@ -0,0 +1,175 @@
...
+		default:
+			unsupportedAuthSchemeSpecified = true
+			unsupportedName = authScheme["name"].(string)
+			continue
+		}
+	}
+
+	if unsupportedAuthSchemeSpecified && len(result) == 0 {
+		return nil, &UnSupportedAuthenticationSchemeSpecifiedError{
+			UnsupportedName: unsupportedName,
```
aajtodd, line 125: **correctness**: This should probably include all of the auth schemes. You don't really need a flag for unsupported, if `result` is 0 length then we know only unsupported auth schemes were specified in the properties. I would also update the error message returned to include all of the schemes instead of just one.

internal/v4a/middleware.go
isaiahvita: agh ... oversight. fixing.

service/lambda/api_op_AddLayerVersionPermission.go
isaiahvita: > It looks like we are duplicating this logic in [event bridge](https://github.com/isaiahvita/aws-sdk-go-v2/blob/feat-ep20-authscheme/service/eventbridge/internal/customizations/signer_wrapper.go#L28)
yah thats correct. if you `git diff` that file `signer_wrapper` in `s3` and `eventbridge`, theyre identical.

internal/auth/scheme.go
isaiahvita: > I would also update the error message returned to include all of the schemes instead of just one.

which error message are you referring to? in the calling code (the operation file), i return an error message which states all of the supported auth schemes:
```go
			return out, metadata, fmt.Errorf(
				"This operation requests signer version %s but the client only supports %v",
				ue.UnsupportedName,
				internalauth.SupportedSchemes,
			)
```

isaiahvita: resolved offline, adding all unsupported auth schemes that are specified

service/lambda/api_op_AddLayerVersionPermission.go
isaiahvita: in any case, i got rid of the initial nit style ask which was to remove `GetName`. so resolving this thread.

aajtodd (approved):

lucix-aws (approved):

Issue #1993, opened 2023-01-23
Unit tests not isolated from local aws config file


aajtodd:
### Describe the bug

Running `make unit` on clean project checkout fails unit tests when `~/.aws/credentials` has stale credentials

### Expected Behavior

Tests should run without failure

### Current Behavior

Failure from `make unit`
```
2023/01/23 16:10:22 config: go test -tags example,codegen,integration,ec2env,perftest -run NONE ./... =>                                                                           
--- FAIL: TestResolveCredentialsIMDSClient (0.01s)                                                                                                                                 
    --- FAIL: TestResolveCredentialsIMDSClient/env_var_DISABLED_true (0.00s)                                                                                                       
        resolve_credentials_test.go:505: unexpected error: <nil>                                                                                                                   
    --- FAIL: TestResolveCredentialsIMDSClient/env_var_DISABLED_false (0.00s)                                                                                                      
        resolve_credentials_test.go:495: expect error got none                                                                                                                     
    --- FAIL: TestResolveCredentialsIMDSClient/option_state_enabled_overrides_env_var_DISABLED_true (0.00s)                                                                        
        resolve_credentials_test.go:495: expect error got none                                                                                                                     
    --- FAIL: TestResolveCredentialsIMDSClient/option_state_disabled_overrides_env_var_DISABLED_false (0.00s)                                                                      
        resolve_credentials_test.go:505: unexpected error: <nil>                                                                                                                   
    --- FAIL: TestResolveCredentialsIMDSClient/default_no_options (0.00s)                                                                                                          
        resolve_credentials_test.go:495: expect error got none                                                                                                                     
    --- FAIL: TestResolveCredentialsIMDSClient/state_enabled (0.00s)                                                                                                               
        resolve_credentials_test.go:495: expect error got none                                                                                                                     
    --- FAIL: TestResolveCredentialsIMDSClient/state_disabled (0.00s)                                                                                                              
        resolve_credentials_test.go:505: unexpected error: <nil>                                                                                                                   
2023/01/23 16:10:31 http: TLS handshake error from 127.0.0.1:60862: remote error: tls: bad certificate                                                                             
FAIL                                                                                                                                                                               
FAIL    github.com/aws/aws-sdk-go-v2/config     11.296s                                                                                                                            
FAIL                                                                                                                                                                               
2023/01/23 16:10:36 config: go test -timeout=1m  ./... => error: failed to run command, exit status 1                                                                              
[...]

PR #2160, opened 2023-06-27
isolate unit tests from local aws config file

Sasaki1994:
fixes https://github.com/aws/aws-sdk-go-v2/issues/1993

It is not needed  `WithSharedConfigFiles([]string{"creds-file"})` to just resolve issue #1993 (only `WithSharedConfigProfile("profile-name")` needed), but for completely separating from the default config, I added.

### DIFF
--- a/config/resolve_credentials_test.go
+++ b/config/resolve_credentials_test.go
@@ -577,8 +577,9 @@ func TestResolveCredentialsIMDSClient(t *testing.T) {
 			opts := []func(*LoadOptions) error{
 				WithRetryer(func() aws.Retryer { return aws.NopRetryer{} }),
 				WithHTTPClient(httpClient),
-				WithSharedConfigProfile("profile-name"),
-				WithSharedConfigFiles([]string{"creds-file"}),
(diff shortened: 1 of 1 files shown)


config/resolve_credentials_test.go
```diff
@@ -577,6 +577,8 @@ func TestResolveCredentialsIMDSClient(t *testing.T) {
 			opts := []func(*LoadOptions) error{
 				WithRetryer(func() aws.Retryer { return aws.NopRetryer{} }),
 				WithHTTPClient(httpClient),
+				WithSharedConfigProfile("profile-name"),
+				WithSharedConfigFiles([]string{"creds-file"}),
```
lucix-aws, line 581: Let's be more explicit about shutting this off - specifically, by giving it a profile name that categorically cannot exist, and no config files.

```suggestion
                                // separate from the local config, should it exist - the default loader will ignore nonexistent profiles
				WithSharedConfigProfile(" "),
				WithSharedConfigFiles([]string{}),
```

lucix-aws (approved):

PR #2185, opened 2023-07-11
Add back unintentionally deleted test cases

isaiahvita:
some test cases got accidentally deleted in my previous PR to this branch. this PR adds them back


isaiahvita:
as discussed with @lucix-aws , merging this into the primary feature branch without approval since its just adding back test cases that werent supposed to get deleted

PR #2199, opened 2023-07-20
Use `config.LoadOptionsFunc` type alias across SDK

### LINKED ISSUE #2193: Possible improvement on config.LoadDefaultConfig to accept existing config.LoadOptionsFunc type instead of dynamic function signature
### Describe the bug

The function `config.LoadDefaultConfig` second parameter is typed `optFns ...func(*LoadOptions) error` instead of using the previouisly defined `config.LoadOptionsFunc`.

This forces users to redefine the whole type if they want to dinamically use the variadic options.

### Expected Behavior

If the function `config.LoadDefaultConfig` used `config.LoadOptionsFunc` for its second parameter's type, this would make it a more elegant and concise code.

### Current Behavior

If a user tries to use `config.LoadOptionsFunc` they will get the following message:

```
cannot use awsConfigOptions (variable of type []"github.com/aws/aws-sdk-go-v2/config".LoadOptionsFunc) as []func(*"github.com/aws/aws-sdk-go-v2/config".LoadOptions) error value in argument to awsConfig.LoadDefaultConfig
```

### Reproduction Steps

```golang
// awsConfigOptions := []awsConfig.LoadOptionsFunc{} // this won't work properly as although the types are "the same" golang won't allow a runtime lambda-like function to match agaist a defined type signarture (afaik)
awsConfigOptions := []func(*awsConfig.LoadOptions) error{}

if config.Debug { // let's say config.Debug comes from the user's config
	awsConfigOptions = append(
		awsConfigOptions,
		awsConfig.WithClientLogMode(aws.LogRequestWithBody|aws.LogRequestEventMessage),
	)
}

cfg, err := awsConfig.LoadDefaultConfig(
	context.TODO(),
	awsConfigOptions...,
)
```

### Possible Solution

Make it so the function `LoadDefaultConfig` accepts a variadic type of `[]awsConfig.LoadOptionsFunc{}`.

_No response_

```
github.com/aws/aws-sdk-go-v2 v1.18.1
```

### Compiler and Version used

go1.20.1 darwin/arm64

MacOS

davidjb:
This fixes #2193 by replacing the relevant optFn types with the `LoadOptionsFunc` type alias. 

In particular, this allows the passing of previously-defined `LoadOptionsFunc` functions, as the reproduction example in https://github.com/aws/aws-sdk-go-v2/issues/2193#issue-1806909973 illustrates.  The other type replacements in this PR are otherwise just cleaning up test files.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -179,7 +179,7 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
 // The default configuration sources are:
 // * Environment Variables
 // * Shared Configuration and Shared Credentials files.
-func LoadDefaultConfig(ctx context.Context, optFns ...func(*LoadOptions) error) (cfg aws.Config, err error) {
+func LoadDefaultConfig(ctx context.Context, optFns ...LoadOptionsFunc) (cfg aws.Config, err error) {
 	var options LoadOptions
 	for _, optFn := range optFns {
 		if err := optFn(&options); err != nil {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -12,7 +12,7 @@ import (
 
 func TestConfigs_SharedConfigOptions(t *testing.T) {
 	var options LoadOptions
-	optFns := []func(*LoadOptions) error{
+	optFns := []LoadOptionsFunc{
 		WithSharedConfigProfile("profile-name"),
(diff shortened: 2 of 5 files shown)


lucix-aws:
See #2193.

PR #2200, opened 2023-07-20
Fix host prefix bug

isaiahvita:
1. adding the disable host prefix context checker to endpoint v2 resolution invocation only on certain cases, when previously it was added in all cases
2. unexported some middlewares that didnt need to be exported

Depends on:
https://github.com/aws/smithy-go/pull/439

### DIFF
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/DisableEndpointHostPrefix.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/DisableEndpointHostPrefix.java
@@ -1,0 +1,45 @@
+package software.amazon.smithy.aws.go.codegen.customization;
+
+import java.util.Optional;
+
+
+import software.amazon.smithy.go.codegen.GoSettings;
+import software.amazon.smithy.go.codegen.GoWriter;
+import software.amazon.smithy.go.codegen.SmithyGoDependency;
+import software.amazon.smithy.go.codegen.SymbolUtils;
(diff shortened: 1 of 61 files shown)


aajtodd (approved):

lucix-aws (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/DisableEndpointHostPrefix.java
```diff
@@ -0,0 +1,45 @@
...
+                if (segment.isLabel() && segment.getContent().equals("AccountId") && !written) {
+                    writer.write(
+                        """
+                        ctx = $T(ctx, true)
+                        """,
+                        SymbolUtils.createPointableSymbolBuilder(
+                            "DisableEndpointHostPrefix", SmithyGoDependency.SMITHY_HTTP_TRANSPORT).build()
+                    );
+                    // we only want the content above written ONCE per operation.
+                    written = true;
```
aajtodd, line 39: **nit**: Then just write it and break out of the loop

aajtodd (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/DisableEndpointHostPrefix.java
isaiahvita: right, will do, in next PR -- artifact left over from my previous logic. a break is more straightforward

PR #2202, opened 2023-07-21
feat: add WithHeaderComment API

### LINKED ISSUE #2198: HTTP User-Agent string-building no longer supports comments
### Describe the bug

In the HTTP spec, [User-Agent strings](https://www.rfc-editor.org/rfc/rfc9110#section-10.1.5) support a combination of `product`s and `comment`s. [A `comment`](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.5) is text contained within parentheses.

Prior to `aws-sdk-go-v2` v1.19.0, comments could be added using

```go
middleware.AddSDKAgentKey("(" + comment + ")")
```

### Expected Behavior

I expected

```go
middleware.AddSDKAgentKey("(" + comment + ")")
```

to continue working.

For example, a comment with the text `this is a comment` should be included in the User-Agent string as

```
(this is a comment)
```

### Current Behavior

Several characters in the comment are replaced by `-`. For example, a comment with the text `this is a comment` is included in the User-Agent string as

```
-this-is-a-comment-
```

instead of

```
(this is a comment)
```

### Reproduction Steps

Generate an HTTP User-Agent string after setting

```go
comment := "this is a comment"
middleware.AddSDKAgentKey("(" + comment + ")")
```

### Possible Solution

One of two solutions:

1. Directly support HTTP User-Agent comment with a function like `middleware.AddUserAgentComment()`

1. Update the escaping rules to not escape keys that match the `comment` pattern defined in [the HTTP spec](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.5)

_No response_

aws-sdk-go-v2 v1.19.0

### Compiler and Version used

go version go1.20.3 darwin/arm64

macOS 12.6.7

lucix-aws:
Adds new `middleware.WithHeaderComment` API:

```go
// WithHeaderComment instruments a middleware stack to append an HTTP field
// comment to the given header as specified in RFC 9110
// (https://www.rfc-editor.org/rfc/rfc9110#name-comments).
//
// The header is case-insensitive. If the provided header exists when the
// middleware runs, the content will be inserted as-is enclosed in parentheses.
//
// Note that per the HTTP specification, comments are only allowed in fields
// containing "comment" as part of their field value definition, but this API
// will NOT verify whether the provided header is one of them.
//
// WithHeaderComment MAY be applied more than once to a middleware stack and/or
// more than once per header.
func WithHeaderComment(header, content string) func(*middleware.Stack) error
```

This behavior was previously unintentionally possible via `middleware.AddUserAgentKey` but broken in changes made to `User-Agent` in #2154.

Closes #2198.

### DIFF
--- a/aws/middleware/header.go
+++ b/aws/middleware/header.go
@@ -1,0 +1,82 @@
+package middleware
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/aws/smithy-go/middleware"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+)
+
+// WithHeaderComment instruments a middleware stack to append an HTTP field
+// comment to the given header as specified in RFC 9110
+// (https://www.rfc-editor.org/rfc/rfc9110#name-comments).
+//
+// The header is case-insensitive. If the provided header exists when the
+// middleware runs, the content will be inserted as-is enclosed in parentheses.
+//
+// Note that per the HTTP specification, comments are only allowed in fields
+// containing "comment" as part of their field value definition, but this API
+// will NOT verify whether the provided header is one of them.
+//
+// WithHeaderComment MAY be applied more than once to a middleware stack and/or
+// more than once per header.
+func WithHeaderComment(header, content string) func(*middleware.Stack) error {
+	return func(s *middleware.Stack) error {
+		m, err := getOrAddHeaderComment(s)
+		if err != nil {
+			return err
+		}
+
+		m.values.Add(header, content)
+		return nil
+	}
+}
+
+type headerCommentMiddleware struct {
+	values http.Header // hijack case-insensitive access APIs
+}
+
+func (*headerCommentMiddleware) ID() string {
+	return "headerComment"
+}
+
+func (m *headerCommentMiddleware) HandleBuild(ctx context.Context, in middleware.BuildInput, next middleware.BuildHandler) (
+	out middleware.BuildOutput, metadata middleware.Metadata, err error,
+) {
+	r, ok := in.Request.(*smithyhttp.Request)
+	if !ok {
+		return out, metadata, fmt.Errorf("unknown transport type %T", in.Request)
+	}
+
+	for h, contents := range m.values {
+		for _, c := range contents {
+			if existing := r.Header.Get(h); existing != "" {
+				r.Header.Set(h, fmt.Sprintf("%s (%s)", existing, c))
+			}
+		}
+	}
+
+	return next.HandleBuild(ctx, in)
+}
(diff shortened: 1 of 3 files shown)


aws/middleware/header.go
```diff
@@ -0,0 +1,82 @@
...
+// The header is case-insensitive. If the provided header exists when the
+// middleware runs, the content will be inserted as-is enclosed in parentheses.
+//
+// Note that per the HTTP specification, comments are only allowed in fields
+// containing "comment" as part of their field value definition, but this API
+// will NOT verify whether the provided header is one of them.
+//
+// WithHeaderComment MAY be applied more than once to a middleware stack and/or
+// more than once per header.
+func WithHeaderComment(header, content string) func(*middleware.Stack) error {
```
aajtodd, line 25: **nit**: Seems like it could have existed in `smithy-go` since it's not really tied to AWS

aws/middleware/header.go
```diff
@@ -0,0 +1,82 @@
...
+	return func(s *middleware.Stack) error {
+		m, err := getOrAddHeaderComment(s)
```
aajtodd, line 27: **Question**: Does this get wired up by default in a way that runs after the UserAgent middleware?

lucix-aws: It's not there by default at all, so if you add it, it's guaranteed to come after.

aajtodd: Sorry not what I meant, I meant that if you add this option/middleware is it (by default) after the UA middleware? If it's not then this would have no effect on the UA. I think the answer is yes but I didn't actually go verify.

aws/middleware/header.go
isaiahvita: ^agree with the above. i think this is prob more than a `nit` given that its an exported function, and we cant ever move it later (this is my only issue, and fine with approving after a response or fixed).

aws/middleware/header.go
lucix-aws: Correct - confirmed offline

aws/middleware/header.go
lucix-aws: I think my main objection to doing so is that `smithy-go/middleware` doesn't actually _implement_ any middleware, it's just the middleware definition itself. Shoehorning this extra API, generic as it may be, doesn't feel quite right organizationally.

isaiahvita: @lucix-aws would it fit anywhere else in smithy-go? if not, and you still think it fits organizationally better where it is -- then im fine with that and feel free to merge

isaiahvita (approved):

PR #2204, opened 2023-07-24
don't serde the bucket

lucix-aws:
Swaps `httpLabel` with `NoSerialize` on relevant buckets so we don't inadvertently add them into the request document (which we were doing after removing `httpLabel`).


isaiahvita (approved):

Makefile
```diff
@@ -81,6 +81,10 @@ generate: smithy-generate update-requires gen-repo-mod-replace update-module-met
 gen-config-asserts gen-internal-codegen copy-attributevalue-feature gen-mod-dropreplace-smithy-. min-go-version-. \
 tidy-modules-. add-module-license-files gen-aws-ptrs format
 
+generate-tmpreplace-smithy: smithy-generate update-requires gen-repo-mod-replace update-module-metadata smithy-annotate-stable \
```
lucix-aws, line 84: I didn't intend to add this but I've been using it heavily during development.

Essentially it temporarily replaces the smithy-go dependency to point to your branch version. The `format` step will fail on `make generate` if you're consuming smithy-go APIs that aren't yet released otherwise.

aajtodd (approved):

PR #2206, opened 2023-07-24
Added fallback behavior to custom v1 resolvers

isaiahvita:
When a custom legacy endpoint resolver (pre-EP20) was set via one of the config setter helper functions, it was wrapped with a fallback endpoint resolver
https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/

This alters that previous fallback behavior to fallback to the endpoint resolver v2

### DIFF
--- a/service/s3/internal/customizations/update_endpoint_test.go
+++ b/service/s3/internal/customizations/update_endpoint_test.go
@@ -965,7 +965,7 @@ func TestVPC_CustomEndpoint(t *testing.T) {
 				}),
 				Region: "us-west-2",
 			},
-			expectedReqURL:        "https://bucketname.s3.us-west-2.amazonaws.com/g",
+			expectedReqURL:        "https://bucketname.s3.us-west-2.amazonaws.com/",
 			expectedSigningName:   "s3",
 			expectedSigningRegion: "us-west-2",
 		},


lucix-aws (approved):

service/s3/endpoints.go
```diff
@@ -132,27 +132,20 @@ func removeResolveEndpointMiddleware(stack *middleware.Stack) error {
...
-	}
 	endpoint, err = w.awsResolver.ResolveEndpoint(ServiceID, region, options)
 	if err == nil {
 		return endpoint, nil
 	}
 
-	if nf := (&aws.EndpointNotFoundError{}); !errors.As(err, &nf) {
-		return endpoint, err
+	if nf := (&aws.EndpointNotFoundError{}); errors.As(err, &nf) {
+		return endpoint, nil
```
lucix-aws, line 144: question: Does downstream code handle this properly? i.e. if the v1 middleware gets a zero-value endpoint, it does nothing and lets v2 pick up the rest?

isaiahvita: it didnt. pushed new commits with a fix. and tested via debugger with sample program and unit tests. heres my sample program:
```go
package main

import (
	"context"
	"log"
)

func main() {
	customResolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
		return aws.Endpoint{}, &aws.EndpointNotFoundError{}
	})
	
	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithEndpointResolverWithOptions(customResolver), config.WithRegion("us-west-2"))

	// Create an Amazon S3 service client
	client := s3.NewFromConfig(cfg)

	// Get the first page of results for ListObjectsV2 for a bucket
	_, err = client.GetObject(context.TODO(), &s3.GetObjectInput{
		Bucket: aws.String("test-bucket"),
		Key: aws.String("test-key"),
	})
	if err != nil {
		log.Fatal(err)
	}

}
```

aajtodd (approved):
Would be nice to see a unit test added to cover this behavior if possible.

lucix-aws (approved):

PR #2214, opened 2023-07-27
Add a new smithy-modeled rules based endpoint resolution.

isaiahvita:
This PR adds a significant feature to support smithy-modeled rules-based endpoint resolution which will supercede and deprecate existing endpoint resolution.


aajtodd (approved):
Fix and ship

PR #2220, opened 2023-08-01
add cross-branch smithy-go runtime detection in CI

lucix-aws:
See new script description for an explanation of the checkout workflow.

Note that I also pushed a `feat-ci.xbranch` on the smithy-go side - you can inspect the workflow log to see that it detects and uses it.

### DIFF
--- a/ci-find-smithy-go.sh
+++ b/ci-find-smithy-go.sh
@@ -13,6 +13,7 @@ if [ -z "$SMITHY_GO_REPOSITORY" ]; then
     echo env SMITHY_GO_REPOSITORY is required
     exit 1
 fi
+
 if [ -z "$RUNNER_TMPDIR" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -12,6 +12,8 @@ on:
 
 env:
   EACHMODULE_CONCURRENCY: 2
+  SMITHY_GO_REPOSITORY: ${{ vars.SMITHY_GO_REPOSITORY }}
```
isaiahvita, line 15: how is `vars.SMITHY_GO_REPOSITORY` being set?

lucix-aws: GitHub has support for env vars at the repo level (org level as well, I believe)

isaiahvita (approved):

aajtodd (approved):

Issue #2276, opened 2023-09-13
INI Parsing Failures silently ignored


daemonl:
### Describe the bug

Config INI files with certain strings which look like numbers silently cause the config file to be entirely skipped.

E.g.
```
color = 970b00
```

### Expected Behavior

Either an error should be thrown for corrupted ini files, or in this case the string should be parsed correctly.

### Current Behavior

- The config file is skipped entirely, with no warning or error
- Meaning the profile specified in the $AWS_PROFILE variable isn't found
- Which also silently fails
- The SDK falls back to attempting to load EC2 IMDS credentials

### Reproduction Steps

Add the color line to ~/.aws/config under any profile
 ```
[profile foobar]
color = 970b00
... existing creds ...
```
`export AWS_PROFILE=foobar`

and attempt to use that profile through `config.LoadDefaultConfig`

```
cfg, err := config.LoadDefaultConfig(ctx)
if err != nil {
return fmt.Errorf("failed to load configuration: %w", err)
}

cfClient := cloudformation.NewFromConfig(cfg)
res, err := d.CFClient.DescribeStacks(ctx, &cloudformation.DescribeStacksInput{
StackName: aws.String(stackName),
})
```

### Possible Solution

Error is thrown [here](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/ini/value_util.go#L117)
Ignored [here](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/ini/ini_lexer.go#L121) 

```
func countTokens(runes []rune) int {
...
		if err != nil {
			return 0
		}
```

Leading to the tokenize function simply not iterating over any tokens.

Returning an error from countTokens passes it up the chain:
```
failed to load configuration: failed to load shared config file, [redacted], incorrect base format found b at 3 index
```

### AWS Go SDK V2 Module Versions Used

ithub.com/google/go-cmp@v0.5.8
github.com/aws/aws-sdk-go-v2/service/sso@v1.13.6 github.com/aws/aws-sdk-go-v2@v1.21.0
github.com/aws/aws-sdk-go-v2/service/sso@v1.13.6 github.com/aws/aws-sdk-go-v2/internal/configsources@v1.1.41
github.com/aws/aws-sdk-go-v2/service/sso@v1.13.6 github.com/aws/aws-sdk-go-v2/internal/endpoints/v2@v2.4.35
[...]

PR #2277, opened 2023-09-13
Throw errors from countTokens in ini tokenizer

daemonl:
Fixes, or at least improves, https://github.com/aws/aws-sdk-go-v2/issues/2276
The parser likely should be able to handle the hex string as-is, but the format isn't a standard part of the AWS config so likely not a real issue.

### DIFF
--- a/internal/ini/ini_lexer.go
+++ b/internal/ini/ini_lexer.go
@@ -59,9 +59,11 @@ func (l *iniLexer) Tokenize(r io.Reader) ([]Token, error) {
 
 func (l *iniLexer) tokenize(b []byte) ([]Token, error) {
 	runes := bytes.Runes(b)
-	var err error
 	n := 0
-	tokenAmount := countTokens(runes)
+	tokenAmount, err := countTokens(runes)
+	if err != nil {
+		return nil, err
+	}
 	tokens := make([]Token, tokenAmount)
 	count := 0
 
@@ -95,7 +97,7 @@ func (l *iniLexer) tokenize(b []byte) ([]Token, error) {
 	return tokens[:count], nil
 }
 
-func countTokens(runes []rune) int {
+func countTokens(runes []rune) (int, error) {
 	count, n := 0, 0
 	var err error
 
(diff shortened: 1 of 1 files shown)


lucix-aws:
We can't change the behavior in this way - I don't want to suddenly break consumers that have "invalid" values in their configs (according to the current implementation, at least). We should be correcting the actual failure to parse instead.

daemonl:
@lucix-aws would you accept a PR which changes the behavior to throw global errors in all error cases (parsing issue, profile not found) based on an environment variable?

lucix-aws:
@daemonl Sorry for the delayed response. As an FYI in case you hadn't seen, we did fix the underlying parsing issue in #2286.

I would definitely be open to supporting a "strict" parsing mode but it would have to be through an in-code setting, probably on [`LoadOptions`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config#LoadOptions):

```go
type LoadOptions struct {
    // ...

    // Whether "strict" parsing of shared config is enabled. By default, config loading or parsing
    // failures are silently ignored. Enabling this setting will instead propagate these errors to config
    // load calls.
    SharedConfigParseStrict bool
}
```

We tend to prefer not to add arbitrary environment variables to drive SDK settings as it creates confusion in a cross-SDK context.

lucix-aws:
If you're interested in having that kind of behavior independent of the fix, I'd ask that you create a feature request such that we can discuss further.

daemonl:
The thing is that when using a tool like terraform, I won't be able to set that config. I guess I have to hope they will set it.

The default behaviour to silently ignore some but not all errors seems very surprising, especially with no way for me as a consumer of various tools to even log the error.

Implementing that flag would be even more strange as the library does throw some errors, so this would really only apply to errors when counting the number of tokens, that or the function would be changed to throw all errors, and it's caller would choose based on that config if the error should be dropped - which is also a 'breaking change'.

I really believe this is a simple bug fix, not a breaking change. I would be interested to compare the way an actually corrupt file behaves in the various SDKs and the CLI, I suspect it will be inconsistent but I would be very surprised if they silently ignore the file all together.

daemonl:
How about something a bit different - but haven't groked enough of the code to know if it'll be easy: When the configs all fall back and finally do error out, that final error could include that it failed to parse the config, or couldn't find the profile in the config etc, so it's only changing the error message on a final failure situation.
That and/or re-instating the verbose logging env var and starting to implement it in places like this.
Really, I'm just looking for some way to know what is wrong with my setup rather than getting an error about EC2 metadata when I'm trying to deploy terraform through an SSO profile using the 'new stuff'

PR #2290, opened 2023-09-25
Upgrade to Smithy 1.39

isaiahvita:
Primarily adds support for endpoints2.0 refactor. Smithy side endpoints changes were all java changes with no intended changes to Go code.

Tested via:
```
make generate && make unit
```

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -7,6 +7,7 @@ on:
     branches: [ main ]
 
 env:
+  # get owner of the repository. used by forks.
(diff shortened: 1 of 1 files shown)


syall (approved):
LGTM, only nit would be using the iterator directly.

Want to also confirm that there are no changes to the generated clients?

isaiahvita:
> Want to also confirm that there are no changes to the generated clients?

@syall yes, thats correct

lucix-aws (approved):

PR #2293, opened 2023-09-28
refactor/regen for upstream endpoint middleware/binding changes

lucix-aws:
Track changes in https://github.com/aws/smithy-go/pull/455 and regenerate s3, s3control, and eventbridge services.

The behavior to disable modeled `AccountId` prefix for v2 endpoint resolution in s3control was ported to a standalone middleware, this was previously a per-op "post-resolution" hook in codegen, which doesn't exist anymore.


service/eventbridge/endpoints.go
```diff
@@ -624,3 +562,357 @@ func (r *resolver) ResolveEndpoint(
...
+	params.UseDualStack = aws.Bool(options.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled)
+	params.UseFIPS = aws.Bool(options.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled)
+	params.Endpoint = options.BaseEndpoint
+
+	input.bindEndpointParams(params)
+
+	return params
+}
+
+func (in *ActivateEventSourceInput) bindEndpointParams(p *EndpointParameters) {
```
lucix-aws, line 583: Now that I see all of these empty bindings I'm thinking we probably want to just not generate in that case. The outer `bind` call could then delegate accordingly.

service/s3/endpoints.go
```diff
@@ -4325,3 +4239,602 @@ func (r *resolver) ResolveEndpoint(
...
+
+}
+
+func (in *WriteGetObjectResponseInput) bindEndpointParams(p *EndpointParameters) {
+
+	p.UseObjectLambdaEndpoint = ptr.Bool(true)
+}
+
+type resolveEndpointV2Middleware struct {
+	options  Options
```
isaiahvita, line 4730: is this the client `Options` in `api_client.go`?
reason why im asking is because it seems that there was a previous effort to not pass the client `Options` around directly in middleware, but rather to create intermediary objects and only copy out what you need from the client `Options`
https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_client.go#L468
https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_op_GetObject.go#L517

would you be able to comment on whether you think that previous pattern should or should not be followed?

also, why give `resolver` its own field if it already exists in the same `options` field
```go
func addResolveEndpointV2Middleware(stack *middleware.Stack, options Options) error {
	return stack.Serialize.Insert(&resolveEndpointV2Middleware{
		options:  options,
		resolver: options.EndpointResolverV2,
	}, "ResolveEndpoint", middleware.After)
}
```

service/s3/endpoints.go
```diff
@@ -4325,3 +4239,602 @@ func (r *resolver) ResolveEndpoint(
...
+	params.Accelerate = aws.Bool(options.UseAccelerate)
+	params.DisableMultiRegionAccessPoints = aws.Bool(options.DisableMultiRegionAccessPoints)
+	params.UseArnRegion = aws.Bool(options.UseARNRegion)
+
+	input.bindEndpointParams(params)
+
+	return params
+}
+
+func (in *AbortMultipartUploadInput) bindEndpointParams(p *EndpointParameters) {
```
isaiahvita, line 4264: just curious, was there a strong reason to put all the operation input receivers for `bindEndpointParams` here rather than in the operation specfic files where the operation input struct is defined? or was it more of a toss up between "it being endpoints related functionality" and "it being an operation method"?

service/s3/endpoints.go
```diff
@@ -4325,3 +4239,602 @@ func (r *resolver) ResolveEndpoint(
 	}
 	return endpoint, fmt.Errorf("endpoint rule error, %s", "A region must be set when sending requests to S3.")
 }
+
+type endpointParamsBinder interface {
+	bindEndpointParams(*EndpointParameters)
+}
+
+func bindEndpointParams(input endpointParamsBinder, options Options) *EndpointParameters {
```
isaiahvita, line 4247: this does look much simpler. i think when i first implemented it i was trying to follow a pattern common in this SDK of only copying out of `Options` what you need (see my other comment). but seeing how much simpler this is makes me seem to prefer this more.

isaiahvita (approved):

service/s3/endpoints.go
lucix-aws: Discussed offline - this is more important for exported middlewares or situations where code can't take a dependency on `Options` directly. For codegenned middleware internal to services, passing options at large is generally fine.

service/s3/endpoints.go
lucix-aws: Moved after discussing offline.

service/eventbridge/endpoints.go
lucix-aws: Did the above after discussing offline.

PR #2308, opened 2023-10-10
Adds subproperty as new literal type

isaiahvita:
To support subproperties in the aws config file, it was decided to extend the ini lexer to treat subproperties as literals. Then parse these literals in the config parser.

It was specifically avoided to add this support in the INI parser. This is because the current state of the INI parser:
- does not follow traditional language parser construction from a CFG (or BNF) into a parse table.
- parser logic from a parse table is obscure and is non-standard
- all institutional knowledge of this implementation is lost

### DIFF
--- a/internal/ini/literal_tokens.go
+++ b/internal/ini/literal_tokens.go
@@ -130,7 +130,7 @@ func newLitToken(b []rune) (Token, int, error) {
 		}
 		token = newToken(TokenLit, b[:n], QuotedStringType)
 	} else if isSubProperty(b) {
-		offset := 0 // include leading newline and whitespace
+		offset := 0
 		end, err := getSubProperty(b, offset)
 		if err != nil {
 			return token, n, err


lucix-aws:
Is there going to be followup to bubble this further up to the actual parsed `Section`s? We have test data right now that still expects nested fields to be skipped:

```
[foo]                            
aws_access_key_id =              
    aws_secret_access_key = valid
aws_session_token = valid        
[bar]                            
aws_access_key_id = valid        
aws_secret_access_key = valid    
aws_session_token = valid        
[baz]                            
aws_access_key_id = valid        
aws_secret_access_key = valid    
aws_session_token = valid        
```
expects
```
{
    "foo": {
        "aws_session_token": "valid"
        // doesn't expect {"aws_access_key_id": { "aws_secret_access_key": "valid"} }
    },
    "bar": {
        "aws_access_key_id": "valid",
        "aws_secret_access_key": "valid",
        "aws_session_token": "valid"
    },
    "baz": {
        "aws_access_key_id": "valid",
        "aws_secret_access_key": "valid",
        "aws_session_token": "valid"
    }
}
```

I would think that test case should fail as written and be updated with this change.

isaiahvita:
@lucix-aws 
> Is there going to be followup to bubble this further up to the actual parsed Sections? We have test data right now that still expects nested fields to be skipped:

with the last commit, the data is now bubbled up in `Sections.container["section_name"].values["item_name"].mp`
additionally, the referenced test data has been amended

internal/ini/literal_tokens.go
```diff
@@ -60,7 +60,6 @@ const (
 	NoneType = ValueType(iota)
 	StringType
 	QuotedStringType
-	// FUTURE(2226) MapType
```
lucix-aws, line 63: Why no `MapType`? I don't fully remember what all uses this, though.

internal/ini/literal_tokens.go
```diff
@@ -124,6 +145,49 @@ func newLitToken(b []rune) (Token, int, error) {
...
+func getSubProperty(runes []rune, offset int) (int, error) {
+	for idx, val := range runes[offset:] {
+		if val == '\n' && !isSubProperty(runes[offset+idx:]) {
+			return offset + idx, nil
+		}
+	}
+	return 0, fmt.Errorf("no sub property")
+}
+
+
```
lucix-aws, line 189: Do you have `gofmt` running? It should clean up extra newlines like this.

lucix-aws (approved):
So given that this new token is "greedy" (like, it's made up of smaller tokens that could be recognized on their own) we probably want to extend the test data (can do `nested_fields`) to cover a few more things

* multiple nested fields
* nested fields at the end of a profile where the next line is a `[profile]` declaration
* nested fields at the end of a file
* nested fields at the end of a file + newlines

LGTM otherwise, fixandship.

PR #2309, opened 2023-10-10
fix: fail to load config if configured profile doesn't exist

### LINKED ISSUE #2058: Default authentication flow should fail if invalid profile name is passed
### Describe the bug

When the default authentication flow is used with `config.LoadDefaultConfig()`, providing an invalid profile name, either with `config.WithSharedConfigProfile()` or by setting `AWS_PROFILE`, does not cause an error, but falls back to trying IMDS.

Previously reported as #1591, but the proposed fix did not resolve the issue

### Expected Behavior

In the AWS CLI, I get

```console
$ AWS_PROFILE=no-such-profile aws sts get-caller-identity

The config profile (no-such-profile) could not be found
```

The SDK should fail with a `config.SharedConfigProfileNotExistError` error, or at least have the option to do so.

### Current Behavior

The AWS SDK falls back to calling IMDS

### Reproduction Steps

```go
```

and set `AWS_PROFILE` to an invalid profile name

or

```go
cfg, err := config.LoadDefaultConfig(ctx, config.WithSharedConfigProfile("no-such-profile"))
```

### Possible Solution

_No response_

_No response_

github.com/aws/aws-sdk-go-v2 v1.13.0, github.com/aws/aws-sdk-go-v2/config v1.13.1

and

github.com/aws/aws-sdk-go-v2 v1.17.6, github.com/aws/aws-sdk-go-v2/config v1.18.18

### Compiler and Version used

go version go1.20.2 darwin/arm64

N/A

lucix-aws:
Closes #2058.

This isn't really possible to unit test in a meaningful way since we don't have well-defined os/fs separations in our config loading unit tests. I've verified it manually, however.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,14 +7,6 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 )
 
-// defaultLoaders are a slice of functions that will read external configuration
-// sources for configuration values. These values are read by the AWSConfigResolvers
-// using interfaces to extract specific information from the external configuration.
-var defaultLoaders = []loader{
-	loadEnvConfig,
-	loadSharedConfigIgnoreNotExist,
-}
-
 // defaultAWSConfigResolvers are a slice of functions that will resolve external
 // configuration values into AWS configuration values.
 //


wty-Bryant (approved):

config/config.go
```diff
@@ -202,3 +195,17 @@ func LoadDefaultConfig(ctx context.Context, optFns ...func(*LoadOptions) error)
...
+
+func resolveConfigLoaders(options *LoadOptions) []loader {
+	loaders := make([]loader, 2)
+	loaders[0] = loadEnvConfig
+
+	// specification of a profile should cause a load failure if it doesn't exist
+	if os.Getenv(awsProfileEnvVar) != "" || options.SharedConfigProfile != "" {
+		loaders[1] = loadSharedConfig
+	} else {
+		loaders[1] = loadSharedConfigIgnoreNotExist
```
isaiahvita, line 207: am i missing something here? the behavior looks the same: its still using the `loadSharedConfigIgnoreNotExist` which swallows the error (i.e. doesnt return an error message).

lucix-aws: Unless the check on 204 is `true` though (which means a profile was specified). That's the change here.

isaiahvita (approved):

PR #2313, opened 2023-10-12
feat: add loaders for pkcs8 keys in cloudfront/sign

### LINKED ISSUE #2270: Support PKCS#8PrivateKey Parsing in CloudFront Signed URL Signer
### Describe the feature

Since OpenSSL version >3.0.0 was deprecated `genrsa` and by default generates Private Key using `PKCS#8` standard and encodes in PEM format,  I would suggest that this SDK would include the `PKCS#8` in reading Private Key file.

Following the instructions in [CloudFront documents]((https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-trusted-signers.html)), there would be generated PKCS#8 format. If users would have PKCS#1 format, users should add [`-traditional`](https://www.openssl.org/docs/man3.0/man1/openssl-rsa.html) flag on their command.

### Use Case

Creating the Signed URL

### Proposed Solution

Either one at the below:
1. add new function `LoadEncryptedPEMPKCS8PrivKey`
2. amend the implementation of `LoadEncryptedPEMPrivKey` with using `x509.ParsePKCS8PrivateKey` for parsing the key file

### Other Information

This OpenSSL feature updates may effect the AWS CLI with `aws cloudfront sign`

### Acknowledgements

github.com/aws/aws-sdk-go-v2/feature/cloudfront/sign@v1.3.49 github.com/aws/aws-sdk-go-v2@v1.21.0
github.com/aws/aws-sdk-go-v2@v1.21.0 github.com/aws/smithy-go@v1.14.2
github.com/aws/aws-sdk-go-v2@v1.21.0 github.com/google/go-cmp@v0.5.8
github.com/aws/aws-sdk-go-v2@v1.21.0 github.com/jmespath/go-jmespath@v0.4.0
github.com/aws/smithy-go@v1.14.2 github.com/google/go-cmp@v0.5.8
github.com/aws/smithy-go@v1.14.2 github.com/jmespath/go-jmespath@v0.4.0

### Go version used

1.21.0

lucix-aws:
Closes #2270.

Add the following APIs:

```go
// LoadPEMPrivKeyPKCS8 reads a PEM-encoded RSA private key in PKCS8 format from
// the given reader.
//
// [x509.ParsePKCS8PrivateKey] can return multiple key types and this API does
// not discern between them. Callers in need of the underlying value must
// obtain it via type assertion:
//
//  key, err := sign.LoadPEMPrivKeyPKCS8(r)
//  if err != nil { /* ... */ }
//
//  switch key.(type) {
//  case *rsa.PrivateKey:
//      // ...
//  case *ecdsa.PrivateKey:
//      // ...
//  case ed25519.PrivateKey:
//      // ...
//  default:
//      panic("unrecognized private key type")
//  }
//
// See aforementioned API docs for a full list of possible key types.
//
// If calling code can opaquely handle the returned key as a [crypto.Signer],
// use [LoadPEMPrivKeyPKCS8AsSigner] instead.
func LoadPEMPrivKeyPKCS8(reader io.Reader) (interface{}, error)

// LoadPEMPrivKeyPKCS8AsSigner wraps [LoadPEMPrivKeyPKCS8] to expect a
// [crypto.Signer].
func LoadPEMPrivKeyPKCS8AsSigner(reader io.Reader) (crypto.Signer, error)
```

Additionally--
* improve package documentation
* deprecate `LoadEncryptedPEMPrivKey`

### DIFF
--- a/.changelog/71a17e4b0d334a06a7b2398a0dc1ec87.json
+++ b/.changelog/71a17e4b0d334a06a7b2398a0dc1ec87.json
@@ -1,0 +1,8 @@
+{
+    "id": "71a17e4b-0d33-4a06-a7b2-398a0dc1ec87",
+    "type": "feature",
+    "description": "Add support for loading PKCS8-formatted private keys.",
+    "modules": [
+        "feature/cloudfront/sign"
+    ]
+}
--- a/feature/cloudfront/sign/privkey.go
+++ b/feature/cloudfront/sign/privkey.go
@@ -1,6 +1,9 @@
 package sign
 
 import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
@@ -44,6 +47,10 @@ func LoadPEMPrivKey(reader io.Reader) (*rsa.PrivateKey, error) {
 // LoadEncryptedPEMPrivKey decrypts the PEM encoded private key using the
 // password provided returning a RSA private key. If the PEM data is invalid,
 // or unable to decrypt an error will be returned.
+//
+// Deprecated: RFC 1423 PEM encryption is insecure. Callers using encrypted
+// keys should instead decrypt that payload externally and pass it to
+// [LoadPEMPrivKey].
 func LoadEncryptedPEMPrivKey(reader io.Reader, password []byte) (*rsa.PrivateKey, error) {
 	block, err := loadPem(reader)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


isaiahvita:
@lucix-aws LGTM, just one question though: why create a new API `LoadPEMPrivKeyPKCS8` rather than amending `LoadEncryptedPEMPrivKey`? is it to avoid an unexpected behavioral change? that way its better to explicitly deprecate and add rather than replace?

wty-Bryant (approved):

isaiahvita (approved):

PR #2314, opened 2023-10-13
Optimize logic of retrieving user agent appID from env config

wty-Bryant:
In a previous [PR](https://github.com/aws/aws-sdk-go-v2/pull/2154),  a new `appID` field is introduced into request's user agent header. In that old PR, `appID` set via env var is retrieved separately after iterating  resolver chain's all configs, this pr now optimize this step by setting `appID` as a field of env config so that it be retrieved directly from env config struct

### DIFF
--- a/config/env_config.go
+++ b/config/env_config.go
@@ -69,6 +69,7 @@ const (
 
 	awsRetryMaxAttempts = "AWS_MAX_ATTEMPTS"
 	awsRetryMode        = "AWS_RETRY_MODE"
+	awsSdkAppID         = "AWS_SDK_UA_APP_ID"
 )
 
 var (
@@ -248,6 +249,9 @@ type EnvConfig struct {
 	//
 	// aws_retry_mode=standard
 	RetryMode aws.RetryMode
+
+	// aws sdk app ID that can be added to user agent header string
+	AppID string
 }
 
 // loadEnvConfig reads configuration values from the OS's environment variables.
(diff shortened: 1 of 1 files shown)


config/env_config.go
```diff
@@ -335,6 +335,11 @@ func (c EnvConfig) getDefaultsMode(ctx context.Context) (aws.DefaultsMode, bool,
 	return c.DefaultsMode, true, nil
 }
 
+func (c EnvConfig) getAppID(context.Context) (string, bool, error) {
+	appID := os.Getenv(`AWS_SDK_UA_APP_ID`)
```
isaiahvita, line 339: while this works, i dont think this is the typical setup. see the other fields in `env_config.go`

wty-Bryant: Just want to lazy set this field in case it is modified, but you are right, I will change that

isaiahvita:
@wty-Bryant also, while i already know the context of this PR. it will be good to clearly explain the motivation for it in the description

config/env_config.go
isaiahvita: i dont think there is actually any reasonable time interval between when `NewEnvConfig` is called and when `getAppId` is called. so i dont think lazily evaluating it here makes a difference

lucix-aws: Yeah - `loadEnvConfig` first reads everything out of the environment into the env config struct, and then the "resolvers" e.g. `getAppID` just read out of the struct. `getAppID` shouldn't be querying the environment directly. See other fields in `env_config` as @isaiahvita said.

lucix-aws: > Just want to lazy set this field in case it is modified,

> i dont think there is actually any reasonable time interval

There's no change of changing or interval to speak of - a process' environment is static.

isaiahvita (approved):
LGTM, wait until the relevant CI passes though before merging

lucix-aws (approved):

PR #2317, opened 2023-10-17
Fix timeout_read_closer corruption bug

iodeal:
The length returned(n) maybe large than the real length of bytes read when an error occurs.
In this case, it will generate a slice overflow due to bufio read operation (golang/go#49795).
To prevent crashes, in the event of an error, set n to 0.

### DIFF
--- a/aws/transport/http/timeout_read_closer.go
+++ b/aws/transport/http/timeout_read_closer.go
@@ -46,6 +46,12 @@ func (r *timeoutReadCloser) Read(b []byte) (int, error) {
 
 	go func() {
 		n, err := r.reader.Read(b)
+		// The length returned("n") maybe large than the real length of bytes read when an error occurs.
+		// In this case, it will generate a slice overflow due to bufio read operation (golang/go#49795).
+		// To prevent crashes, in the event of an error, set "n" to 0.
+		if err != nil {
+			n = 0
+		}
 		timer.Stop()
 		c <- readResult{n: n, err: err}
 	}()


iodeal:
### Here is the error log：
panic: runtime error: slice bounds out of range [:5042] with capacity 4096

goroutine 76399 [running]:
bufio.(*Reader).Read(0xc000516ba0, {0xc00042a080?, 0xc000675e30?, 0x100a936?})
        /usr/local/go/src/bufio/bufio.go:250 +0x2f8
io.ReadAtLeast({0x166c240, 0xc000516ba0}, {0xc00042a080, 0x8, 0x200}, 0x8)
        /usr/local/go/src/io/io.go:332 +0x9a
io.ReadFull(...)
        /usr/local/go/src/io/io.go:351
compress/gzip.(*Reader).Read(0xc00042a000, {0xc000612000, 0x2000, 0x2000})
        /usr/local/go/src/compress/gzip/gunzip.go:261 +0x1d2
net/http.(*gzipReader).Read(0xc00059a300, {0xc000612000, 0x2000, 0x2000})
        /usr/local/go/src/net/http/transport.go:2856 +0x18a
github.com/aws/aws-sdk-go-v2/aws/transport/http.(*timeoutReadCloser).Read.func1()
        /Users/xiaoleishi/go/pkg/mod/github.com/aws/aws-sdk-go-v2@v1.21.0/aws/transport/http/timeout_read_closer.go:48 +0x47
created by github.com/aws/aws-sdk-go-v2/aws/transport/http.(*timeoutReadCloser).Read
        /Users/xiaoleishi/go/pkg/mod/github.com/aws/aws-sdk-go-v2@v1.21.0/aws/transport/http/timeout_read_closer.go:47 +0x11e

lucix-aws:
This makes sense to me but could we instead fix it by not using the invalid index?

It _seems_ to me like the consumer of that channel should _not_ attempt to do anything with the value of `n` if `err` is set, but I don't have full context.

lucix-aws:
@iodeal ping for visibility

iodeal:
> This makes sense to me but could we instead fix it by not using the invalid index?
> 
> It _seems_ to me like the consumer of that channel should _not_ attempt to do anything with the value of `n` if `err` is set, but I don't have full context.

Firstly, the "bufio/bufio" is a official standart package, and it also mention with "Note: if the slice panics here, it is probably because the underlying reader returned a bad count. See issue 49795". For more: https://github.com/golang/go/issues/49795
Secondly, I am not able to "not using the invalid index", because it is not exists in my code. It seems that something unexcepted in the kinesis client's http-response underlying reader.
After of all, since there is a problem with the reader and 'n' does not match the actual value, it is better to set it to 0 to prevent the program from crashing.

iodeal:
I found that every time my program crashes, it aways throws "slice bounds out of range [:LARGE_THAN_4095] with capacity 4096" error, e.g. "slice bounds out of range [:5042] with capacity 4096"。
Is there any possible that kinesis server truncated response body and this action will cause the kinesis client's bad underlying reader?

lucix-aws:
Took another look at this, I see what you're saying now.

This isn't something we should be handling, though:

* I agree with the sentiment in golang/go#49795 that any reader (beyond that, any consumer of a reader) could be misbehaving, it's not reasonable to try and catch all misbehaviors in calling code.
* This `timeout_read_closer` implementation _is_ correct as written in the sense that it performs no processing or inference of the results of the underlying `Read` call, it simply passes them on, the logic add is entirely to include the timeout.
* This change violates the semantics of `Read` which _should_ return a nonzero value of `n` if any bytes were read, even if an `error` is returned (again, see point 2, this implementation is and should be just a proxy for the actual read data).

Closing, this most likely needs to be filed against the standard lib, since that is providing both the `Reader` implementation and its consumer here.

PR #2318, opened 2023-10-17
feat(i&a): track smithy-go auth middleware

lucix-aws:
* Add internal adapters for SDK -> smithy-go I&A types (credentials, v4 signer, bearer token, bearer token signer)
* Relocate `HttpBearerAuth` integration here. This is effectively now the legacy implementation of bearer.
* Add internal implementation of `http.AuthScheme` to handle per-op config updates to signers on client config (the actual config finalizer to do this will be in a future PR).
* Remove old signing middleware.

Regenerated services `codecatalyst` and `dynamodb` as examples of each auth scheme. The removal of generated `operationName` on input structs was applied in the downstream branch to avoid cluttering the diff.


service/dynamodb/auth.go
```diff
@@ -59,3 +62,153 @@ func serviceAuthOptions(params *AuthResolverParameters) []*smithyauth.Option {
...
+	}
+
+	rscheme := getResolvedAuthScheme(ctx)
+	if rscheme == nil {
+		return out, metadata, fmt.Errorf("no resolved auth scheme")
+	}
+
+	identity := getIdentity(ctx)
+	if identity == nil {
+		return out, metadata, fmt.Errorf("no identity")
```
isaiahvita, line 201: probably want to make these error messages more informative before this gets merged to `main`

service/dynamodb/endpoints.go
```diff
@@ -513,68 +519,14 @@ func (m *resolveEndpointV2Middleware) HandleSerialize(ctx context.Context, in mi
...
-			ctx = awsmiddleware.SetSigningName(ctx, *v4aScheme.SigningName)
-			ctx = awsmiddleware.SetSigningRegion(ctx, v4aScheme.SigningRegionSet[0])
-			break
-		case *internalauth.AuthenticationSchemeNone:
-			break
-		}
+	opts, _ := smithyauth.GetAuthOptions(&resolvedEndpoint.Properties)
+	for _, o := range opts {
+		rscheme.SignerProperties.SetAll(&o.SignerProperties)
 	}
```
syall, line 530: I'm pretty sure it does this, but wanted to double check this section means the resolved endpoint properties are merged into the the auth scheme identity / signer properties?

lucix-aws: Correct - `SetAll` was added on the `smithy-go` runtime side, which just blindly accepts all given `Properties`.

syall (approved):

Issue #2322, opened 2023-10-18
DynamoDB expression.NameBuilder.Contains() forces input to string, but interface{} should be supported


andrewcretin:
### Describe the bug

The NameBuilder.Contains() does not allow for filtering sets of types other than strings. For example, a Set of Number []int{1,2,3} can not be filtered using Contains as the value must be cast to a string. 

### Expected Behavior

I should be able to provide a value of any type, represented by an interface and contains should work so long as the provided value is the same value of the attribute (either primitive or list)

### Current Behavior

The contains value is being forced to string (implementation likely done for substring or larger string) but this should also work for searching for a value contained within a list (of any supported type)

### Reproduction Steps

See additional info and test case included in possible solution

### Possible Solution

In `condition.go`

```
// Contains returns a ConditionBuilder representing the result of the
// contains function in DynamoDB Condition Expressions. The resulting
// ConditionBuilder can be used as a part of other Condition Expressions or as
// an argument to the WithCondition() method for the Builder struct.
//
// Example:
//
//	// condition represents the boolean condition of whether the item
//	// attribute "InviteList" has the value "Ben"
//	condition := expression.Name("InviteList").Contains("Ben")
//
//	// condition represents the boolean condition of whether the item
//	// attribute "LocationIds" has the value 2
//	condition := expression.Name("LocationIds").Contains(2)
//
//	// Used in another Condition Expression
//	anotherCondition := expression.Not(condition)
//	// Used to make an Builder
//	builder := expression.NewBuilder().WithCondition(condition)
//
// Expression Equivalent:
//
//	expression.Name("InviteList").Contains("Ben")
//	// Let :ben be an ExpressionAttributeValue representing the value "Ben"
//	"contains (InviteList, :ben)"
func (nb NameBuilder) Contains(val interface{}) ConditionBuilder {
	return Contains(nb, val)
}
```

and 

```
// Contains returns a ConditionBuilder representing the result of the
// contains function in DynamoDB Condition Expressions. The resulting
// ConditionBuilder can be used as a part of other Condition Expressions or as
// an argument to the WithCondition() method for the Builder struct.
//
// Example:
//
//	// condition represents the boolean condition of whether the item
//	// attribute "InviteList" has the value "Ben"
//	condition := expression.Contains(expression.Name("InviteList"), "Ben")
//
//	// condition represents the boolean condition of whether the item
//	// attribute "LocationIds" has the value 2
[...]

PR #2324, opened 2023-10-19
fix: NameBuilder.Contains() supports all types for usage on lists

andrewcretin:
PR to address issue laid out in https://github.com/aws/aws-sdk-go-v2/issues/2322

TestContainsCondition passing

image

### DIFF
--- a/feature/dynamodb/expression/condition.go
+++ b/feature/dynamodb/expression/condition.go
@@ -1346,6 +1346,10 @@ func (nb NameBuilder) BeginsWith(prefix string) ConditionBuilder {
 //	// attribute "InviteList" has the value "Ben"
 //	condition := expression.Contains(expression.Name("InviteList"), "Ben")
 //
+//	// condition represents the boolean condition of whether the item
+//	// attribute "LocationIds" has the value 2
+//	condition := expression.Contains(expression.Name("LocationIds"), 2)
+//
 //	// Used in another Condition Expression
 //	anotherCondition := expression.Not(condition)
 //	// Used to make an Builder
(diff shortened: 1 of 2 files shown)


feature/dynamodb/expression/condition.go
```diff
@@ -1387,8 +1395,8 @@ func Contains(nameBuilder NameBuilder, substr string) ConditionBuilder {
 //	expression.Name("InviteList").Contains("Ben")
 //	// Let :ben be an ExpressionAttributeValue representing the value "Ben"
 //	"contains (InviteList, :ben)"
-func (nb NameBuilder) Contains(substr string) ConditionBuilder {
-	return Contains(nb, substr)
+func (nb NameBuilder) Contains(val interface{}) ConditionBuilder {
```
lucix-aws, line 1398: Do we want to open up the type completely like this? versus
```
func ContainsInt(...)
func ContainsFloat(...)
func ContainsBool(...)
```

I don't fully understand the capabilities of the `contains` dynamodb expression (does it work with maps, sublists, etc.)

andrewcretin: image

The only snag with current implementation would be if the user provides a type that is different than the type of the list. (ie: list of strings and providing number) but I would imagine this would just fail the dynamo filter expression case. 

Only reason I opted for opening it up is that the value is cast back to interface in `ValueBuilder`. I'm happy to implement the above methods if that is desired implementation. This would leave some ambiguity around the default `Contains()` handling both substring in string and string is list of strings. (user would have no way to differentiate, even though would yield proper results)

lucix-aws (approved):

lucix-aws:
@andrewcretin Thanks again for the contribution. We're going to accept this as-is. Mering now.

andrewcretin:
@lucix-aws Great! Happy to help. I'll watch for the next release and update on our end 🤝

PR #2327, opened 2023-10-20
build(docs): improve `Makefile` for docs

agilgur5:
## Summary

Used the docs `Makefile` in #2326 and made some improvements while working on it

## Modifications

- the `docker build` can be extracted into a separate `make` command and re-used

- volumes need to be mounted in order to make changes to local dir and reflect in container (and vice-versa)

## Notes to Reviewers

I actually still had trouble getting changes to reflect in the local dir properly. Moved to local usage of `hugo` and didn't debug further, but thought this was still a worthwhile improvement.

## Future Work

- May want to update versions in the `Dockerfile` (e.g. to Go 1.21+), newer version of Hugo, etc
- `$(COMMON_DOCKER_BUILD_ARGS)` is no longer necessary as it's in a single place. But args for `-v ...` and `-t ...` could be consolidated if one wanted to do so (nbd either way imo)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,16 @@ init-node-modules:
 generate:
 	hugo --environment $(SITE_ENV) -d docs --gc
 
-container-generate:
+container-build:
 	docker build \
 		$(COMMON_DOCKER_BUILD_ARGS) \
 		-t aws-sdk-go-v2-docs \
 		-f ./Dockerfile \
 		.
+
+container-generate: container-build
 	docker run \
+		-v .:/aws-sdk-go-v2 \
 		-t aws-sdk-go-v2-docs \
 		make setup generate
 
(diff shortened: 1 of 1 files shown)


lucix-aws:
We have internal CI that consumes this makefile, we're not generally in a position to accept changes to it without risk of breakage.

agilgur5:
> without risk of breakage.

This isn't a breaking change though?

> We have internal CI that consumes this makefile

The container variant also doesn't work locally, so I am guessing that your internal CI isn't using it either

> we're not generally in a position to accept changes to it

That's totally your prerogative, but I would ask that you make that explicit to all potential contributors. The `CONTRIBUTING.md` does not mention that nor does the `Makefile` itself, say, in a comment at the top.

There is also just no contributing guidelines for the docs what-so-ever, which made contributing a docs fix a good bit more difficult as a result as well

PR #2328, opened 2023-10-20
Add support for configured endpoints.

isaiahvita:
While SDK consumers can already configure endpoint overrides in code, consumers have expressed the desire for the same configuration options in environment variables and in the aws config file. This PR fulfills this feature.

This PR regenerated code for all services

In order to support this feature, support need to be added for subproperties in the aws config file. It was decided to extend the INI lexer to treat subproperties as literals. Then parse these literals in the config parser.

It was specifically avoided to add this support in the INI parser. This is because the current state of the INI parser:
- does not follow traditional language parser construction from a CFG (or BNF) into a parse table.
- LL(1) parser logic from a parse table is obscure and is non-standard
- all institutional knowledge of this implementation is lost

Feature context: https://github.com/aws/aws-sdk/issues/229
Depends on: https://github.com/aws/smithy-go/pull/463

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go-version: ["1.20", "1.21"]
+        go-version: [1.19, "1.20"]
     steps:
     - uses: actions/checkout@v2
 
@@ -48,7 +48,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go-version: [1.15, 1.16, 1.17, 1.18, 1.19]
+        go-version: [1.15, 1.16, 1.17, 1.18]
         exclude:
           - os: macos-latest
             go-version: 1.15
@@ -76,7 +76,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest]
-        go-version: ["1.20", "1.21"]
+        go-version: [1.19, "1.20"]
     env:
       EACHMODULE_SKIP: "internal\\repotools\\changes"
     steps:
@@ -103,7 +103,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest]
-        go-version: [1.15, 1.16, 1.17, 1.18, 1.19]
+        go-version: [1.15, 1.16, 1.17, 1.18]
     env:
       EACHMODULE_SKIP: "internal\\repotools\\changes"
     steps:


config/env_config.go
```diff
@@ -482,6 +494,34 @@ func (c EnvConfig) getCustomCABundle(context.Context) (io.Reader, bool, error) {
...
+	return *c.IgnoreConfiguredEndpoints, true, nil
+}
+
+func (c EnvConfig) getBaseEndpoint(context.Context) (string, bool, error) {
+	return c.BaseEndpoint, len(c.BaseEndpoint) > 0, nil
+}
+
+// GetServiceBaseEndpoint is used to retrieve a normalized SDK ID for use
+// with configured endpoints.
+func (c EnvConfig) GetServiceBaseEndpoint(ctx context.Context, sdkID string) (string, bool, error) {
```
lucix-aws, line 513: [nonblocking] In new "resolvers" like this I've just left off the `error` return if it can't actually return an error.

config/env_config.go
```diff
@@ -482,6 +494,34 @@ func (c EnvConfig) getCustomCABundle(context.Context) (io.Reader, bool, error) {
 	return bytes.NewReader(b), true, nil
 }
 
+// GetIgnoreConfiguredEndpoints is used in knowing when to disable configured
+// endpoints feature.
+func (c EnvConfig) GetIgnoreConfiguredEndpoints(context.Context) (bool, bool, error) {
```
lucix-aws, line 499: Why `context.Context`? Don't see it used anywhere, it's also not possible for the caller to propagate it with the structure of how these are used (I see you're setting it to `context.Background()` later on).

config/shared_config.go
```diff
@@ -548,6 +599,7 @@ func LoadSharedConfigProfile(ctx context.Context, profile string, optFns ...func
 
 	cfg := SharedConfig{}
 	profiles := map[string]struct{}{}
+	// isaiah this is where global endpoint_url is dropped.
```
lucix-aws: remove

aws/config.go
```diff
@@ -146,6 +146,10 @@ type Config struct {
 	// See https://docs.aws.amazon.com/sdkref/latest/guide/settings-reference.html for
 	// more information on environment variables and shared config settings.
 	AppID string
+
+	// BaseEndpoint is an intermediary transfer location to a service specific
+	// BaseEndpoint on a service's Options.
+	BaseEndpoint *string
```
lucix-aws, line 152: [nonblocking] Can we `string` this?

service/elasticbeanstalk/api_client.go
```diff
@@ -296,6 +296,7 @@ func NewFromConfig(cfg aws.Config, optFns ...func(*Options)) *Client {
 	resolveAWSEndpointResolver(cfg, &opts)
 	resolveUseDualStackEndpoint(cfg, &opts)
 	resolveUseFIPSEndpoint(cfg, &opts)
+	resolveConfigEndpointResolverV2(cfg, &opts)
```
lucix-aws, line 299: [nonblocking] `resolveBaseEndpoint`?

config/shared_config.go
```diff
@@ -272,6 +280,15 @@ type SharedConfig struct {
 
 	// aws sdk app ID that can be added to user agent header string
 	AppID string
+
+	IgnoreConfiguredEndpoints *bool
```
lucix-aws, line 284: [nonblocking] `bool`?

config/shared_config.go
```diff
@@ -272,6 +280,15 @@ type SharedConfig struct {
...
+
+	BaseEndpoint string
+
+	// Services section values
+	// {"serviceId": {"key": "value"}}
+	// e.g. {"s3": {"endpoint_url": "example.com"}}
+	Services map[string]map[string]string
```
lucix-aws, line 291: * The values in `[service whatever]` could be non-objects though (they're just regular sections with another name) so I think we have to generalize to `Services map[string]Section`
* [nonblocking] probably want to leave unexported for now.

config/env_config.go
```diff
@@ -251,7 +254,10 @@ type EnvConfig struct {
 	RetryMode aws.RetryMode
 
 	// aws sdk app ID that can be added to user agent header string
-	AppID string
+	AppID                     string
+	IgnoreConfiguredEndpoints *bool
```
lucix-aws, line 258: Doc these.

config/shared_config.go
lucix-aws: Also doc.

config/env_config.go
lucix-aws: Oh I see it's part of that static `resolver` signature. I would just not propagate it further down wherever it's not needed.

config/shared_config.go
lucix-aws: We actually _must_ not export this, `Section` is internal.

isaiahvita: re-reading the spec, its not actually clear whether the `services` section is a normal section. it seems to say that the properties in a `services` section are all service subproperties? if not though, its not a big change, just need to move the logic that knows how to parse it up, and should be a small lift.

isaiahvita: ok, instead of making the value a `Section`, i wrapped the map in a Services struct, and top-level non map values can be defined in this new struct. this follows a similar pattern with sso session sections. i dont think we should be propagating the `Section` to the layer above. all `Section` processing into `SharedConfig` fields seems like it should happen in `setFromIniSection`. will push this change to this branch soon

aws/config.go
isaiahvita: i agree, i should have done that. but reason why i didnt is because the corresponding field in a services `Options` struct is also a `*string`. and its much easier to do the transfer from aws config to service options when they are the same types

config/shared_config.go
isaiahvita: i initially had it as a pointer because without it theres not a good way to distinguish between the types zero value and a reasonable value set by the user. (although since false is the default i suppose it may not matter as much). 

additionally, every boolean field type in this file is a `*bool` rather than a `bool`.

im going to keep it for now, but let me know if you disagree

config/env_config.go
isaiahvita: sure, but in this impl in env config and shared config, i dont propagate it down to any further functions. feel free to elaborate if im misunderstanding

config/shared_config.go
lucix-aws: Agreed, new way LGTM.

config/shared_config.go
lucix-aws: That's fine.

.github/workflows/go.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go-version: [1.19, "1.20"]
+        go-version: ["1.20", "1.21"]
```
lucix-aws, line 24: Why are we doing this here?

lucix-aws (approved):
We should regenerate everything to see CI pass for every service here but LGTM.

.github/workflows/go.yml
isaiahvita: this was before i root caused my issue. and thought this might fix it. but i left it because 1.21 is the latest released version. all tests were passing with it, but i can change it back if you think we should

wty-Bryant (approved):

.github/workflows/go.yml
lucix-aws: Yeah revert this. We technically still need to check 1.19 and I'm going to handle this in a separate PR today anyway.

PR #2337, opened 2023-10-27
feat: sra i&a: add sigv4a and stabilize s3,s3control,eventbridge,sso,sts,protocoltest

lucix-aws:
**IMPORTANT: Ignore failing CI for now. It has passed on a previous run but the downstream changes to bump to go 1.19 have made the modules here irreconcilable without either merging smithy-go or committing replace statements.**

* Add sigv4a
  * For s3 and eventbridge the synthetic trait from upstream is backfilled onto those models to facilitate the service loading the scheme accordingly
* Add endpoint-based auth resolvers for s3 and eventbridge
  * Note that eventbridge has an additional "fallback" step in its custom resolver to just default to sigv4. This is because eventbridge's ruleset doesn't always specify a list of schemes in each endpoint, unlike s3
* Stabilize s3, s3control, and eventbridge tests
  * s3, eventbridge: remove signing customization middleware
  * s3control: this service performs some mutation of the input during the serialize step. This was previously fine, but endpoint resolution was broken by it after moving to finalize because the binding of endpoint parameters MUST use the original input. To facilitate this I've basically stashed the "original" input at the start of serialize in s3control, to then re-apply it to the operation input field on the context at the start of build.
  * update all of these tests to unwrap the actual signature from the finalize phase and assert on those values instead of checking the context, since that isn't always set now

misc:
* Do not inject `AnonymousCredentials` by default anymore. This is effectively no longer necessary since `AnonymousIdentity` handles this opaquely.
* Respect caller context overrides for signing name and signing region
  * I've left the issue of the API for signing region set for sigv4a unaddressed. The plan is to actually deprecate these and instead add new client `With` helpers to set them during instantiation - `WithSigningRegionSet`, etc.
* The interface to presigning remains untouched. We likely want to re-examine this holistically in the future. As it stands, SRA handles choosing the scheme, and a "polyfill" middleware hoists signing version, name, and region to the context so the presigning middlewares can work exactly as they did before.

### DIFF
--- a/internal/endpoints/config/go.mod
+++ b/internal/endpoints/config/go.mod
@@ -32,8 +32,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../internal/
 
 replace github.com/aws/aws-sdk-go-v2/internal/ini => ../../../internal/ini/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../../../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../../../service/sso/
--- a/config/go.mod
+++ b/config/go.mod
@@ -32,8 +32,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../internal/endpoi
 
 replace github.com/aws/aws-sdk-go-v2/internal/ini => ../internal/ini/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../service/sso/
--- a/credentials/go.mod
+++ b/credentials/go.mod
@@ -26,8 +26,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/configsources => ../internal/confi
 
 replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../internal/endpoints/v2/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../service/sso/
--- a/feature/ec2/imds/internal/configtesting/go.mod
+++ b/feature/ec2/imds/internal/configtesting/go.mod
@@ -34,8 +34,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../../../int
 
 replace github.com/aws/aws-sdk-go-v2/internal/ini => ../../../../../internal/ini/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../../../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../../../../../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../../../../../service/sso/
--- a/internal/configsources/configtesting/go.mod
+++ b/internal/configsources/configtesting/go.mod
@@ -34,8 +34,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../internal/
 
 replace github.com/aws/aws-sdk-go-v2/internal/ini => ../../../internal/ini/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../../../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../../../service/sso/
--- a/service/internal/eventstreamtesting/go.mod
+++ b/service/internal/eventstreamtesting/go.mod
@@ -26,8 +26,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/configsources => ../../../internal
 
 replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../internal/endpoints/v2/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../../../service/internal/presigned-url/
 
 replace github.com/aws/aws-sdk-go-v2/service/sso => ../../../service/sso/
--- a/service/kinesis/internal/testing/go.mod
+++ b/service/kinesis/internal/testing/go.mod
@@ -32,8 +32,6 @@ replace github.com/aws/aws-sdk-go-v2/internal/configsources => ../../../../inter
 
 replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../../internal/endpoints/v2/
 
-replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../../service/internal/accept-encoding/
-
 replace github.com/aws/aws-sdk-go-v2/service/internal/eventstreamtesting => ../../../../service/internal/eventstreamtesting/
 
 replace github.com/aws/aws-sdk-go-v2/service/internal/presigned-url => ../../../../service/internal/presigned-url/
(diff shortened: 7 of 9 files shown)


service/dynamodb/endpoints.go
```diff
@@ -192,9 +192,7 @@ func finalizeClientEndpointResolverOptions(options *Options) {
 }
 
 func resolveEndpointResolverV2(options *Options) {
-	if options.EndpointResolverV2 == nil {
-		options.EndpointResolverV2 = NewDefaultEndpointResolverV2()
-	}
+	options.EndpointResolverV2 = NewDefaultEndpointResolverV2()
```
isaiahvita, line 195: this looks like were going to overwrite anything EndpointResolverV2 that the user may have set? why?

service/dynamodb/auth.go
```diff
@@ -15,6 +16,42 @@ func bindAuthParamsRegion(params *AuthResolverParameters, _ interface{}, options
...
+	return "setLegacyContextSigningOptions"
+}
+
+func (m *setLegacyContextSigningOptionsMiddleware) HandleFinalize(ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler) (
+	out middleware.FinalizeOutput, metadata middleware.Metadata, err error,
+) {
+	rscheme := getResolvedAuthScheme(ctx)
+	schemeID := rscheme.Scheme.SchemeID()
+
+	if sn := awsmiddleware.GetSigningName(ctx); sn != "" {
```
isaiahvita, line 32: do we know of an actual occurrence where users were setting the signing name manually on the context? or is it just our code and the possibility of users being able to do this that we need to support it?

service/dynamodb/endpoints.go
```diff
@@ -508,33 +506,32 @@ func (m *resolveEndpointV2Middleware) HandleSerialize(ctx context.Context, in mi
...
+	params := bindEndpointParams(getOperationInput(ctx), m.options)
+	endpt, err := m.options.EndpointResolverV2.ResolveEndpoint(ctx, *params)
 	if err != nil {
 		return out, metadata, fmt.Errorf("failed to resolve service endpoint, %w", err)
 	}
 
-	req.URL = &resolvedEndpoint.URI
-	for k := range resolvedEndpoint.Headers {
-		req.Header.Set(k, resolvedEndpoint.Headers.Get(k))
+	if endpt.URI.RawPath == "" && req.URL.RawPath != "" {
```
isaiahvita, line 515: im guessing this logic was pulled from `OperationSerializer`? should we maybe put that into its own function and document it or just add a comment so it doesnt feel out of place?
and if thats true, was this logic removed from `OperationSerializer` so its not duplicated (since i dont see a change in `serializers.go`?

service/s3/internal/customizations/update_endpoint_test.go
```diff
@@ -134,9 +135,9 @@ func Test_UpdateEndpointBuild(t *testing.T) {
 				},
 				tests: []s3BucketTest{
 					{"abc", "key", "https://example.region.amazonaws.com/abc/key?x-id=GetObject", ""},
-					{"a$b$c", "key", "https://example.region.amazonaws.com/a%24b%24c/key?x-id=GetObject", ""},
-					{"a.b.c", "key", "https://example.region.amazonaws.com/a.b.c/key?x-id=GetObject", ""},
-					{"a..bc", "key", "https://example.region.amazonaws.com/a..bc/key?x-id=GetObject", ""},
+					{"a$b$c", "key", "", "cannot be used with"},
```
isaiahvita, line 138: this is breaking, no? why are these use cases no longer supported?

service/dynamodb/endpoints.go
lucix-aws: It's can't, this happens before user config fns run.

service/dynamodb/auth.go
lucix-aws: Someone from lightsail was definitely doing it after the endpoints2 refactor (v1 resolution let you bake in a signing region, v2 doesn't).

service/s3/internal/customizations/update_endpoint_test.go
lucix-aws: It's the "it never worked anyway and now it's just explicitly caught instead of producing garbage" case. I mentioned this in a standup a week or so back.

lucix-aws: I believe in this case it was accelerate+path style now being caught correctly (it's one of those endpoint combinations).

service/dynamodb/endpoints.go
lucix-aws: > im guessing this logic was pulled from OperationSerializer? should we maybe put that into its own function and document it or just add a comment so it doesnt feel out of place?

We _could_ but then it's part of the public API for something that is very much internal. Defaulting to "don't export unless necessary" here, duplication aside.

> was this logic removed from OperationSerializer so its not duplicated (since i dont see a change in serializers.go?

It wasn't, I will check to see if it's safe to remove this. If we just hard set the path in serialization before endpoints2 then I'd say it should be removed (as it exists right now after this change it basically does nothing because the path at that point is empty - **unless** caller middleware did something to it). This is a subtle thing behaviorally but if we've _always_ constructed the path in serde and joined it to anything existing then that's a behavioral quirk we should leave be. On the flip side, if it wasn't there before EP2 we should drop it.

service/dynamodb/endpoints.go
isaiahvita: but the user could have set `EndpointResolverV2` directly on the options struct and then call `New`
https://github.com/aws/aws-sdk-go-v2/blob/main/service/dynamodb/api_client.go#L46

service/dynamodb/endpoints.go
isaiahvita: > We could but then it's part of the public API for something that is very much internal. Defaulting to "don't export unless necessary" here, duplication aside.

oh i more meant an unexported function. pulling it out into its own function was more of a code organization comment. but nvm, its fine as is.

> It wasn't, I will check to see if it's safe to remove this
sounds good

> if we've always constructed the path in serde and joined it to anything existing then that's a behavioral quirk we should leave be.
is there a behavioral difference though between constructing the path in serialize instead of finalize? are there specific customer hooks between serialize and finalize where moving this logic from serialize to finalize makes a difference?

service/dynamodb/endpoints.go
lucix-aws: You're right I always forget about New being used in isolation. Will fix.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
```diff
@@ -204,6 +212,8 @@ public void writeAdditionalFiles(
             // generate client helpers such as copyAPIClient, GetAPIClientOptions()
             writePresignClientHelpers(writer, model, symbolProvider, serviceShape);
 
+            writer.write(new PresignContextPolyfillMiddleware(serviceShape));
+
```
syall, line 216: nit: I'm not too familiar with Go conventions, but is using the term polyfill idiomatic?

internal/endpoints/config/go.mod
```diff
@@ -32,6 +32,8 @@ replace github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 => ../../../internal/
 
 replace github.com/aws/aws-sdk-go-v2/internal/ini => ../../../internal/ini/
 
+replace github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding => ../../../service/internal/accept-encoding/
```
syall, line 35: clarifying question: is this `accept-encoding` change part of identity and auth?

lucix-aws: There's no change to speak of, this must have been some intermediate artifact of codegen during development. The module in question isn't used here nor in the other places that the replace statement has appeared. I'll remove it.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpPresignURLClientGenerator.java
lucix-aws: Guess it is now? The resulting middleware is also unexported fwiw.

service/dynamodb/endpoints.go
lucix-aws: Fixed in codegen but will actually save regeneration for merge.

service/dynamodb/endpoints.go
lucix-aws: It looks like we've always "appended" to any existing path at the serialize step even before endpoint resolution, so that's going to have to remain.

e.g. https://github.com/aws/aws-sdk-go-v2/blob/v1.17.7/service/dynamodb/serializers.go#L592

Essentially the change here is that instead of resolving endpoint and then serializing we're now doing that in reverse. So what was once a one-step `join(existing url that happens to have resolved endpoint, serde)` is now happening in two parts:
1. `join(some caller injected url, serde)`
2. `join(resolved endpoint, result of step 1)`

internal/endpoints/config/go.mod
isaiahvita: the mod replace command is recursive. so an indirect dependency might be using it

isaiahvita (approved):

PR #2357, opened 2023-11-08
sra i&a: odds/ends

lucix-aws:
**ignore the mod version changes, churn from being out of date w/ main**

Address a few loose ends (and regen s3):

* Fix a bug with the interaction between presigning and checksums (we were incorrectly trying to do something in the checksum workflow that we shouldn't have after the phase reorder. Added a simple check of "is presigning" to skip it)
* Add config helpers to set sigv4[a] signing name and region[s], and deprecate the old context versions
* S3 custom auth resolver
  * rebase identity / signer properties from modeled ones if they exist, certain properties like unsigned payload do not ever come from endpoint rules
  * include anonymous as an option for s3 operations (this is a possibility in various use cases but it's not modeled at all)
 * zero out `options.Credentials` if caller set `AnonymousCredentials` so we can correctly fall back to anonymous if supported

### DIFF
--- a/service/s3/options.go
+++ b/service/s3/options.go
@@ -3,7 +3,10 @@
 package s3
 
 import (
+	"context"
+	"fmt"
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
 	internalauthsmithy "github.com/aws/aws-sdk-go-v2/internal/auth/smithy"
 	"github.com/aws/aws-sdk-go-v2/internal/v4a"
 	smithyauth "github.com/aws/smithy-go/auth"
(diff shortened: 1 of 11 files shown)


service/s3/options.go
```diff
@@ -200,3 +203,82 @@ func WithEndpointResolverV2(v EndpointResolverV2) func(*Options) {
...
+			)
+		})
+	}
+}
+
+func ignoreAnonymousAuth(options *Options) {
+	if _, ok := options.Credentials.(aws.AnonymousCredentials); ok {
+		options.Credentials = nil
+	}
+}
```
syall, line 284: Does the S3 client allow anonymous credentials currently?

I feel like the request should always be signed unless S3 explicitly models that it doesn't need it.

lucix-aws: I've been told you can configure a bucket for anonymous access at a minimum, among a host of other things, of course none of this is modeled. Hence I've added it as an option. It's obviously not ideal but the worst case scenario is we send an anonymous request that gets rejected.

syall (approved):

isaiahvita (approved):

PR #2416, opened 2023-12-07
fix: translation of ini service sections into shared config

lucix-aws:
Two fixes:
* Regression from ini parse refactor that prevented [services] sections from being translated into `SharedConfig`
* Correct referenced loading of [services ] section. The [services ] section must be specified in the config profile to be loaded.

**NOTE**: due to the pending release of smithy-go request compression changes, CI will fail until today's release.

### DIFF
--- a/config/shared_config.go
+++ b/config/shared_config.go
@@ -31,7 +31,7 @@ const (
 
 	// Prefix for services section. It is referenced in profile via the services
 	// parameter to configure clients for service-specific parameters.
-	servicesPrefix = `services`
+	servicesPrefix = `services `
 
 	// string equivalent for boolean
 	endpointDiscoveryDisabled = `false`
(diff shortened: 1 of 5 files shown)


wty-Bryant (approved):
A little question, do we add an extra `serviceSectionKey` to just distinguish it from `servicesPrefix`

lucix-aws:
@wty-Bryant They're fundamentally different - one is the actual prefix for the section, and then the other (the new one) is the literal name of the key we expect for a related config.

config/shared_config.go
```diff
@@ -31,7 +31,7 @@ const (
 
 	// Prefix for services section. It is referenced in profile via the services
 	// parameter to configure clients for service-specific parameters.
-	servicesPrefix = `services`
+	servicesPrefix = `services `
```
isaiahvita, line 34: is this extra space important? if so, why?

lucix-aws: Yeah, it matches the way we define all the other section headers because the profile name is stored as declared (e.g. `Sections.Get("services foo")`).

PR #2433, opened 2023-12-20
simplify base64 encoding of Cloudfront signing

shogo82148:
By introducing a customized base64 encoder, we have eliminated the replacement step.
This enhancement simplifies the encoding process and increases efficiency.

### DIFF
--- a/feature/cloudfront/sign/policy.go
+++ b/feature/cloudfront/sign/policy.go
@@ -15,6 +15,8 @@ import (
 	"unicode"
 )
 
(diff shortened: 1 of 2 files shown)


lucix-aws:
I don't think this is right, the expected output for the encoding shouldn't be changing here (specifically the padding at the end).

Beyond that I get the motivation, but I'm not worried about the double iteration over the policy document, so we're going to leave it alone for now.

shogo82148:
I have modified the test for `encodePolicy`, but since it is a private function, there are no changes to its behavior when viewed from outside the package. Indeed, other tests are passing as well.

PR #2446, opened 2024-01-10
Allow unknown traits when generating SDK

milesziemer:
Updates Smithy Gradle Plugin config to allow unkown traits, stopping builds from failing if there are traits in models which haven't been added as a dependency in the code generator. Usually new traits are added to existing packages, but if/when a trait is added in a new package, we will have to update the dependencies of the code generator, even if the code generator isn't actually using the trait yet.


milesziemer:
Closing in favor of https://github.com/aws/aws-sdk-go-v2/pull/2447

PR #2453, opened 2024-01-14
Update getting-started.md

akashgreninja:
There is an error while using the normal syntax referred to by the doc  on creating seperate folders for each where in one we put region and other the creds it works like a charm

image

image

this loads the credentials and region else we get error like this in the normal syntax 

```
2024/01/14 16:52:57 operation error S3: ListObjectsV2, resolve auth scheme: resolve endpoint: endpoint rule error, Invalid region: region was not a valid DNS name.
exit status 1
```

### DIFF
--- a/content/en/docs/getting-started.md
+++ b/content/en/docs/getting-started.md
@@ -105,8 +105,9 @@ import (
 )
 
 func main() {
-	// Load the Shared AWS Configuration (~/.aws/config)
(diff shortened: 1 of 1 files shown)


lucix-aws:
The way the doc calls LoadDefaultConfig() isn't wrong (assuming your aws config file is well-formed) and is the most "basic" way to use the API which is what we want to show in a getting-started context.

PR #2458, opened 2024-01-16
Remove invalid defaults from a few services

milesziemer:
Updates the RemoveDefaults customization to include a few more services/ shapes, specifically ones where the shape has a range incompatible with its default value. Also refactors RemoveDefaults slightly.

Tested by `make generate` and confirming the expected diff.

This is a breakfix for evidently and emrserverless - the types for ResultsPeriod and WorkerCounts will become nillable.


lucix-aws:
PR should include re-generated code.

lucix-aws (approved):

PR #2465, opened 2024-01-23
codegen: normalize adding of sigv4a config based on trait presence

lucix-aws:
Normalize auth codegen for sigv4a:
* internal signer config and resolution is now driven by presence of `SigV4ATrait`
* sha256 payload middleware will now generate when either of the v4 traits are present

_All_ services were regenerated with this change and in terms of codegen this is largely a no-op. Explanation of diffs:
* Removed redundant code leftover from SRA refactor:
  * defaulting of `DisableURIPathEscaping`, this is now controlled by signer properties
  * `resolveCredentialProvider` in each sigv4a service, the `Options`' `GetIdentityResolver` API now handles this (the asymmetric wrapper)
  * `swapWithCustomHTTPSignerMiddleware` - obsolete by new SRA auth resolution
  * The check for a credentials provider in s3 presigning w/ v4a is redundant w/ SRA refactor (scheme would never be selected in the first place)
* Incidental rearrangement of some codegen due to switching integrations (no change in behavior)
  * various sigv4a internals in eventbridge
  * `resolveHTTPSignerV4a`


wty-Bryant (approved):
Need a changelog for the PR?

isaiahvita (approved):
Nice, thanks for this. this does make it more standard and cleaner rather than having the s3, eventbridge, and cfkvs all implement similar logic

PR #2470, opened 2024-01-26
Fix name of network interface

lukemassa:
Per https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInterface.html
> api_gateway_managed | aws_codestar_connections_managed | branch | ec2_instance_connect_endpoint | efa | efs | gateway_load_balancer | gateway_load_balancer_endpoint | global_accelerator_managed | interface | iot_rules_managed | lambda | load_balancer | nat_gateway | network_load_balancer | quicksight | transit_gateway | trunk | vpc_endpoint

I'm not sure why nat_gateway was spelled `natGateway`, but as you can tell none of the others are camelCase (either in the code or in the docs) so it's very likely a mistake.

### DIFF
--- a/service/ec2/types/enums.go
+++ b/service/ec2/types/enums.go
@@ -5831,7 +5831,7 @@ type NetworkInterfaceType string
 // Enum values for NetworkInterfaceType
 const (
 	NetworkInterfaceTypeInterface                     NetworkInterfaceType = "interface"
-	NetworkInterfaceTypeNatGateway                    NetworkInterfaceType = "natGateway"
+	NetworkInterfaceTypeNatGateway                    NetworkInterfaceType = "nat_gateway"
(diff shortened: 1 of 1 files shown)


lucix-aws:
We are unable to accept this PR as you have directly modified code-generated source.

All AWS SDKs are generated from API models which are defined by each service and pushed to this repository every time a service releases a feature.

For future reference, most code under `services/*` packages are auto-generated. You can check whether a file is generated by the presence of a header comment at the top:

`// Code generated by smithy-go-codegen DO NOT EDIT.`

lukemassa:
Ah my apologies! To be fair, it's much easier to read than most generated code I've seen :) I will try to find where this is generated *from* an open a new PR, thanks again!

PR #2475, opened 2024-01-31
Add support for AWS_CREDENTIAL_EXPIRATION environment variable

calebstewart:
This PR adds support for loading the credential expiration time from the `AWS_CREDENTIAL_EXPIRATION` environment variable. Similar support can be seen in the AWS Javascript SDK ([docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/) under the `fromEnv()`  section).

This essentially looks for `AWS_CREDENTIAL_EXPIRATION` after finding valid credentials in the environment, and if the value is a valid RFC3339 datetime string, sets `creds.CanExpire` to `true` and saves the time in `creds.Expires`.

Also, I read through the contributing guidelines, which references adding a `CHANGELOG_PENDING.md` file, but I was confused by  the expected format of that file, and when I looked at other PRs here, I didn't see it included. So, I've left it out, but I'm happy to add it if I can get some more clarification on what that is supposed to look like. :)

### DIFF
--- a/config/env_config.go
+++ b/config/env_config.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
@@ -28,6 +29,8 @@ const (
 	awsSecretKeyEnvVar       = "AWS_SECRET_KEY"
 
 	awsSessionTokenEnvVar = "AWS_SESSION_TOKEN"
+
+	awsCredentialExpirationEnvVar = "AWS_CREDENTIAL_EXPIRATION"
 
 	awsContainerCredentialsEndpointEnvVar     = "AWS_CONTAINER_CREDENTIALS_FULL_URI"
 	awsContainerCredentialsRelativePathEnvVar = "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"
@@ -310,6 +313,14 @@ func NewEnvConfig() (EnvConfig, error) {
 	setStringFromEnvVal(&creds.SecretAccessKey, credSecretEnvKeys)
 	if creds.HasKeys() {
 		creds.SessionToken = os.Getenv(awsSessionTokenEnvVar)
+
+		if expireTimeString, hasExpire := os.LookupEnv(awsCredentialExpirationEnvVar); hasExpire {
+			if expireTime, err := time.Parse(time.RFC3339, expireTimeString); err == nil {
+				creds.Expires = expireTime
+				creds.CanExpire = true
+			}
+		}
+
 		cfg.Credentials = creds
 	}
 
--- a/config/env_config_test.go
+++ b/config/env_config_test.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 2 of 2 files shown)


lucix-aws:
@calebstewart Can you outline the motivation behind this PR?

Despite its apparent presence in botocore and jsv3, we had previously decided not to standardize this env setting.

calebstewart:
My use-case is that `aws-vault` passes this information along with the credentials, and I would like to base some of the functionality of my application on the potential expiration of the credentials, since it is assumed they are short-lived.

Given, the number of different ways that the AWS SDK can load credentials, if I were to manually parse out the `AWS_CREDENTIAL_EXPIRATION` variable, I would still need to check `credentials.CanExpire`/`credentials.Expires` as well. Doing so is feasible, but feels messy when there is already a (apparently semi-)standard way of communicating this expiration information, and a place to store it in the credential object.

I do think there is something to be said for uniformity of the credential interface across SDKs/languages as well. I'm not sure how many of the other AWS SDKs in other languages support it, but `botocore` and the AWS Javascript SDK are two relatively heavy hitters. A user coming from a different language could easily be misguided by the lack of consistency.

On the flip-side, is there any specific negative consequence of parsing and including this information when loading credentials of which I am ignorant?

calebstewart:
I thought I'd also mention that this environment variable is also exported by the builtin `aws configure export-credentials --format env` command. It is odd that any given SDK would explicitly make the decision not to support something that the official CLI exports for external consumption.

calebstewart:
Understood. The internal context is helpful. Thank you for looking into it!

calebstewart:
@lucix-aws were you able to find out any more information from the internal side?

lucix-aws:
It's scheduled for review. I'll post here when there's an update.

lucix-aws:
We reviewed this today. **In summary, we're not ready to add this to any additional SDKs right now.**

We've confirmed that this is a legacy setting, added long before a time where SDKs had a standardization process for keeping things consistent. It was essentially added into botocore on a whim to support the CLI for some obscure internal tooling purpose. JS received it for similarly arbitrary reasons. These days we do not commit to SDK configuration features unless they're vetted and agreed upon universally beforehand, and that consensus is not in place here. The fact that aws-vault has picked up on it before then is unfortunate, but as it stands the AWS ecosystem does not formally recognize it.

We have some additional concerns about supporting this at large as well:
* The value really means nothing if our compute environments such as Lambda do not set it. I approached the design review of the mindset that we should just commit all the way with this across SDKs, but doing so would require a non-trivial amount of lift to upstream it to be supported within the business at large.
* Adoption of an intrinsic credentials value like this at large across first- and third-party tooling may be perilous. More specifically, were this env to snap into existence tomorrow, tools today that operate around refreshing these environment variables would not immediately recognize or refresh the new expiration field. This could then resulting in essentially invalid credentials being forwarded to whatever application is using them (because expiration hasn't been refreshed), and the SDKs would have no way of knowing otherwise.
* There are also usability issues across SDKs in general here - not all of them actually readily expose their sourced credentials, which essentially just further tips the scale on the balance of usefulness vs. lift to implement.

In short, we haven't decided _against_ it, but we've agreed that SDKs shouldn't pick it up for the time being until the above concerns are addressed and we have a better idea of the use cases and user segment at play. I'm not going to be able to provide any sort of timeline for that effort, unfortunately.

In the interest of keeping this on the record (and since our answer isn't definitively, eternally "no") I'm going to spawn this as a feature request in this repository and cross-link.

PR #2479, opened 2024-02-02
Add accountID endpoint builtin and accountID endpoint mode config

wty-Bryant:
Merge accountID into endpoint rules codegen as an endpoint param builtin retrieved from credentials, add accountIDEndpoint mode config to opt in/out accountID based endpoint routing

### DIFF
--- a/aws/accountid_endpoint_mode.go
+++ b/aws/accountid_endpoint_mode.go
@@ -7,7 +7,7 @@ const (
 	// AccountIDEndpointModeUnset indicates the AWS account ID will not be used for endpoint routing
 	AccountIDEndpointModeUnset AccountIDEndpointMode = ""
 
-	// AccountIDEndpointModePreferred indicates the AWS account ID will be used for endpoint routing if offered
+	// AccountIDEndpointModePreferred indicates the AWS account ID will be used for endpoint routing if present
 	AccountIDEndpointModePreferred = "preferred"
 
 	// AccountIDEndpointModeRequired indicates an error will be returned if the AWS account ID is not resolved from identity
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
@@ -39,13 +39,9 @@ public class AccountIDEndpointRouting implements GoIntegration {
             case $aidModeDisabled:T:
             case $aidModeRequired:T:
                 if ca, ok := identity.(*$credentialsAdapter:T); !ok {
-                    return $errorf:T("the accountID is configured to be required, but the " +
-                        "identity provider could not be converted to a valid credentials adapter " +
-                        "and provide an accountID, should switch to a valid credentials provider " +
-                        "or change to another account id endpoint mode")
+                    return $errorf:T("accountID is required but not set")
                 } else if ca.Credentials.AccountID == "" {
-                    return $errorf:T("the required accountID could not be empty, should switch to a valid " +
-                        "credentials provider or change to another account id endpoint mode")
+                    return $errorf:T("accountID is required but not set")
                 }
             // default check in case invalid mode is configured through request config
             default:
--- a/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsEndpointBuiltins.java
+++ b/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsEndpointBuiltins.java
@@ -54,7 +54,7 @@ public class AwsEndpointBuiltins implements GoIntegration {
         }
         goDelegator.useShapeWriter(settings.getService(model), goTemplate("""
         func resolveAccountID(identity $auth:T, mode $accountIDEndpointMode:T) *string {
-            if mode == $aidModeDisabled:T || mode == $aidModeUnset:T {
+            if mode == $aidModeDisabled:T {
                 return nil
             }
             


aws/accountid/mode/mode.go
```diff
@@ -0,0 +1,29 @@
+package mode
+
+import "fmt"
+
+// AIDMode switches on/off the account ID based endpoint routing
+type AIDMode string
+
+// enums of valid AIDMode
+const (
+	Preferred AIDMode = "preferred"
```
lucix-aws, line 10: We should have a none/unset sentinel i.e. `AccountIDEndpointModeUnset` (which is just the empty string).

aws/accountid/mode/mode.go
```diff
@@ -0,0 +1,29 @@
...
+	Required  AIDMode = "required"
+	Disabled  AIDMode = "disabled"
+)
+
+// SetFromString converts config string to corresponding AIDMode or reports error if the value is not enumerated
+func (mode *AIDMode) SetFromString(s string) error {
```
lucix-aws, line 16: There's no real reason to expose this, since sourcing the value from config is an implementation detail for us, let's keep it that way.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/LegacyEndpointContextSetter.java
```diff
@@ -63,7 +64,13 @@ public void renderPreEndpointResolutionHook(GoSettings settings, GoWriter writer
                 if $T(ctx) {
                     return next.HandleFinalize(ctx, in)
                 }
-                """, SdkGoTypes.Aws.Middleware.GetRequiresLegacyEndpoints);
+                if err := $T(getIdentity(ctx), m.options.AccountIDEndpointMode); err != nil {
```
lucix-aws, line 67: This doesn't really belong in this "legacy context" integration, it should have its own entity in codegen that generates it.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -235,6 +237,11 @@ public class AddAwsConfigFields implements GoIntegration {
                     .documentation("The inclusive min request body size to be compressed.")
                     .servicePredicate(RequestCompression::isRequestCompressionService)
                     .generatedOnClient(false)
+                    .build(),
+            AwsConfigField.builder()
+                    .name(SDK_AID_ENDPOINT_MODE)
+                    .type(getAwsAccountIDModeSymbol("AIDMode"))
```
lucix-aws, line 243: Add and use `SdkGoTypes.Aws.AccountIDEndpointMode` (once you've moved the enum there) so you can use that directly.

service/dynamodb/endpoints.go
```diff
@@ -489,7 +490,7 @@ type endpointParamsBinder interface {
 	bindEndpointParams(*EndpointParameters)
 }
 
-func bindEndpointParams(input interface{}, options Options) *EndpointParameters {
+func bindEndpointParams(input interface{}, options Options, ctx context.Context) *EndpointParameters {
```
lucix-aws, line 493: Shouldn't we be seeing the account ID binding getting generated here? Specifically somewhere from line 496 on?

lucix-aws (changes requested):

service/dynamodb/endpoints.go
wty-Bryant: I regenerated the client using model without accountID builtin in my last commit.  I will regenerate that using mocked model

aws/accountid_endpoint_mode.go
```diff
@@ -0,0 +1,12 @@
+package aws
+
+// AccountIDEndpointMode switches on/off the account ID based endpoint routing
```
lucix-aws, line 3:
```suggestion
// AccountIDEndpointMode controls how a resolved AWS account ID is handled for endpoint routing.
```

aws/accountid_endpoint_mode.go
```diff
@@ -0,0 +1,12 @@
...
+type AccountIDEndpointMode string
+
+// enums of valid AccountIDEndpointMode
+const (
+	AccountIDEndpointModeUnset     AccountIDEndpointMode = ""
+	AccountIDEndpointModePreferred AccountIDEndpointMode = "preferred"
```
lucix-aws, line 9: [nonblocking] You can leave the type declaration off of everything but the first line, the compiler figures it out

e.g.
```go
type Foo = string

const (
    Foo1 Foo = "1"
    Foo2     = "2" // is implicitly type Foo as is everything after it
)

File: aws/config.go [hunk 8]
```diff
@@ -162,6 +162,10 @@ type Config struct {
 	// This variable is sourced from environment variable AWS_REQUEST_MIN_COMPRESSION_SIZE_BYTES or
 	// the shared config profile attribute request_min_compression_size_bytes
 	RequestMinCompressSizeBytes int64
+
+	// AccountIDEndpointMode indicates how aws account ID is applied in endpoint2.0 routing.
+	// Will be set to preferred by default. Supported modes are: Preferred, Required and Disabled
```
lucix-aws, line 167:
```suggestion
	// Controls how a resolved AWS account ID is handled for endpoint routing.
```

aws/accountid_endpoint_mode.go
```diff
@@ -0,0 +1,12 @@
+package aws
+
+// AccountIDEndpointMode switches on/off the account ID based endpoint routing
+type AccountIDEndpointMode string
+
+// enums of valid AccountIDEndpointMode
```
lucix-aws, line 6: Rather than having the comment on the declaration block, give each one a comment explaining what it does.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AddAwsConfigFields.java
```diff
@@ -81,6 +81,8 @@ public class AddAwsConfigFields implements GoIntegration {
 
     private static final String REQUEST_MIN_COMPRESSION_SIZE_BYTES = "RequestMinCompressSizeBytes";
 
+    private static final String SDK_AID_ENDPOINT_MODE = "AccountIDEndpointMode";
```
lucix-aws, line 84: [nonblocking] We'd rather see this spelled out `SDK_ACCOUNTID_ENDPOINT_MODE` but it's not a huge deal.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
```diff
@@ -0,0 +1,77 @@
...
+import software.amazon.smithy.go.codegen.integration.GoIntegration;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.rulesengine.traits.EndpointRuleSetTrait;
+import software.amazon.smithy.utils.MapUtils;
+
+import static software.amazon.smithy.go.codegen.GoWriter.goTemplate;
+
+public class AccountIDEndpointRouting implements GoIntegration {
+    @Override
+    public void renderPreEndpointResolutionHook(GoSettings settings, GoWriter writer, Model model) {
```
lucix-aws, line 15: We should make sure this happens _after_ the early return that's triggered when we detect they're using v1 endpoint resolution. You may need to fiddle with the integration order to accomplish that.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
```diff
@@ -0,0 +1,77 @@
...
+            GoSettings settings,
+            Model model,
+            SymbolProvider symbolProvider,
+            GoDelegator goDelegator
+    ) {
+        if (!settings.getService(model).hasTrait(EndpointRuleSetTrait.class)) {
+            return;
+        }
+        goDelegator.useShapeWriter(settings.getService(model), goTemplate("""
+        func accountID(identity $auth:T, mode $accountIDEndpointMode:T) *string {
```
lucix-aws, line 35: A few things here
1. the name here is too broad, this is likely to collide with random identifiers we generate in the package. I'd call it `resolveAccountID`, `bindAccountID`, etc.
2. you're adding the builtin that calls this in a different location than where it's defined. We should move the `accountID` function you have here to be generated in the builtins integration instead. (checkAccountID _is_ used here and can stay in this integration)

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
```diff
@@ -0,0 +1,77 @@
...
+            if ca, ok := identity.(*$credentialsAdapter:T); ok && (mode == $aidModePreferred:T || mode == $aidModeRequired:T) {
```
lucix-aws, line 36: Instead of checking all the modes where it's on, I'd just check the one mode where it's off (disabled) and return early

aws/accountid_endpoint_mode.go
```diff
@@ -0,0 +1,18 @@
+package aws
+
+// AccountIDEndpointMode controls how a resolved AWS account ID is handled for endpoint routing.
+type AccountIDEndpointMode string
+
+const (
+	// AccountIDEndpointModeUnset indicates the AWS account ID will not be used for endpoint routing
+	AccountIDEndpointModeUnset AccountIDEndpointMode = ""
+
+	// AccountIDEndpointModePreferred indicates the AWS account ID will be used for endpoint routing if offered
```
lucix-aws, line 10:
```suggestion
	// AccountIDEndpointModePreferred indicates the AWS account ID will be used for endpoint routing if present
```

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
```diff
@@ -0,0 +1,70 @@
...
+            switch mode {
+            case $aidModeUnset:T:
+            case $aidModePreferred:T:
+            case $aidModeDisabled:T:
+            case $aidModeRequired:T:
+                if ca, ok := identity.(*$credentialsAdapter:T); !ok {
+                    return $errorf:T("the accountID is configured to be required, but the " +
+                        "identity provider could not be converted to a valid credentials adapter " +
+                        "and provide an accountID, should switch to a valid credentials provider " +
+                        "or change to another account id endpoint mode")
```
lucix-aws, line 45: We don't have to specific about the type assertion failing in this path - we can just say an accountID wasn't found.

```suggestion
                    return $errorf:T("accountID is required but not set")
```

(use this value in the next branch as well)

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AwsEndpointBuiltins.java
```diff
@@ -32,6 +39,42 @@ public class AwsEndpointBuiltins implements GoIntegration {
...
+            Model model,
+            SymbolProvider symbolProvider,
+            GoDelegator goDelegator
+    ) {
+        if (!settings.getService(model).hasTrait(EndpointRuleSetTrait.class)) {
+            return;
+        }
+        goDelegator.useShapeWriter(settings.getService(model), goTemplate("""
+        func resolveAccountID(identity $auth:T, mode $accountIDEndpointMode:T) *string {
+            if mode == $aidModeDisabled:T || mode == $aidModeUnset:T {
```
lucix-aws, line 57: An unset value is supposed to imply preferred, according to the spec:

> preferred, or no value - Default behavior

So this should only check for disabled

lucix-aws (approved):
fixandship!

Issue #2507, opened 2024-02-20
Refactor operation builders to only modify the middleware stack within their own module's context


lucix-aws:
### Describe the feature

This will category prevent compatibility fractures such as #2370 going forward.

### Use Case

n/a

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

n/a

### Go version used

n/a

prochac:
Did you also consider using `retract` go.mod directive?
I know that these things happen, but the retract directive could help module users identify the root cause much faster.
https://go.dev/doc/go1.17#go-command

PR #2516, opened 2024-02-22
move common middleware stack ops to service client modules

lucix-aws:
This is the majority of the work for #2507, though it doesn't completely address it, as some smithy-side middlewares and service-specific ones (e.g. flex checksums) remain.

This is a **net-zero** change behaviorally, as the new snapshot tests will indicate.

I've also removed golangci-lint from the workflows, we don't really pay attention to it and it seems to be overopinionated in a lot of different ways that don't necessarily mesh w/ codegen (e.g. unused unexported funcs).

### DIFF
--- a/aws/middleware/request_id_retriever.go
+++ b/aws/middleware/request_id_retriever.go
@@ -24,6 +24,8 @@ func (m *RequestIDRetriever) ID() string {
 	return "RequestIDRetriever"
 }
 
+// HandleDeserialize pulls the AWS request ID from the response, storing it in
+// operation metadata.
 func (m *RequestIDRetriever) HandleDeserialize(ctx context.Context, in middleware.DeserializeInput, next middleware.DeserializeHandler) (
 	out middleware.DeserializeOutput, metadata middleware.Metadata, err error,
 ) {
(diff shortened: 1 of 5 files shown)


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -187,4 +199,47 @@ public List<RuntimeClientPlugin> getClientPlugins() {
                         .build()
         );
     }
+
+    @Override
+    public void writeAdditionalFiles(GoSettings settings, Model model, SymbolProvider symbolProvider, GoDelegator goDelegator) {
+        goDelegator.useFileWriter("api_client.go", settings.getModuleName(), addMiddleware());
+    }
+
+    // TODO symbols
```
lucix-aws: The commit two entries down addresses this: https://github.com/aws/aws-sdk-go-v2/pull/2516/commits/0d1b977e61194563ccf40ed58af31499d8d6681b

wty-Bryant (approved):

Issue #2518, opened 2024-02-26
endpointdiscovery.Endpointcache will panic on 32-bit archs


dswarbrick:
### Describe the bug

`atomic.panicUnaligned` will occur due to unaligned struct member `size`. This occurs since the size of the `endpoints` sync.Map is variable, depending on the machine word size. On 64-bit archs, sync.Map is 32 bytes. On 32-bit archs, sync.Map is 20 bytes - resulting in the misalignment of the subsequent members in the `EndpointCache` struct:
```go
type EndpointCache struct {
    endpoints     sync.Map
    endpointLimit int64
    // size is used to count the number elements in the cache.
    // The atomic package is used to ensure this size is accurate when
    // using multiple goroutines.
    size int64
}
```
```
$ GOARCH=386 go run foo.go 
size of EndpointCache.endpoints: 20
offset of EndpointCache.endpoints: 0
offset of EndpointCache.endpointLimit: 20
offset of EndpointCache.size: 28
panic: unaligned 64-bit atomic operation

goroutine 1 [running]:
runtime/internal/atomic.panicUnaligned()
        /usr/lib/go-1.21/src/runtime/internal/atomic/unaligned.go:8 +0x2d
runtime/internal/atomic.Xadd64(0x9ca201c, 0x1)
        /usr/lib/go-1.21/src/runtime/internal/atomic/atomic_386.s:125 +0x11
main.main()
        /tmp/foo/foo.go:30 +0x1e6
exit status 2
```

### Expected Behavior

No panic to occur.

### Current Behavior

The following test fails on 32-bit archs:
```
=== RUN   TestEndpointCache_Get_prune
--- FAIL: TestEndpointCache_Get_prune (0.00s)
panic: unaligned 64-bit atomic operation [recovered]
        panic: unaligned 64-bit atomic operation

goroutine 6 [running]:
testing.tRunner.func1.2({0x81c33c0, 0x822a528})
        /usr/lib/go-1.21/src/testing/testing.go:1545 +0x2ab
testing.tRunner.func1()
        /usr/lib/go-1.21/src/testing/testing.go:1548 +0x43e
panic({0x81c33c0, 0x822a528})
        /usr/lib/go-1.21/src/runtime/panic.go:914 +0x1ec
runtime/internal/atomic.panicUnaligned()
        /usr/lib/go-1.21/src/runtime/internal/atomic/unaligned.go:8 +0x2d
[...]

PR #2519, opened 2024-02-27
Fix misaligned struct member used in atomic operation

dswarbrick:
This fixes a panic caused by attempting to atomically access a struct member which is not 64-bit aligned when running on 32-bit arch, due to the smaller sync.Map struct.

Fixes: #2518

Please consider adding testing with a 32-bit GOARCH (e.g. "386") to your CI testing pipeline to avoid such oversights in future.

### DIFF
--- a/.changelog/448aaf1509f448d385e5c243933a39d8.json
+++ b/.changelog/448aaf1509f448d385e5c243933a39d8.json
@@ -1,0 +1,8 @@
+{
+    "id": "448aaf15-09f4-48d3-85e5-c243933a39d8",
+    "type": "bugfix",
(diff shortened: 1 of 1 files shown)


lucix-aws:
I saw when this came in yesterday, solid find. I was actually planning on adding a 386 arch test in CI as you've called out there, I'll try and add that to this changeset when I get capacity.

lucix-aws (approved):

wty-Bryant (approved):

lucix-aws:
Added the x86 test externally and rebased this branch. Test running now, will merge on a pass.

PR #2595, opened 2024-04-03
docextractor: correct docs, type signatures, and links

lucix-aws:


wty-Bryant (approved):

internal/repotools/cmd/docextractor/extract.go
```diff
@@ -189,6 +168,125 @@ func extractTypes(packageName string, types map[string]*ast.TypeSpec, items map[
 	return nil
 }
 
+// We've already converted the model's HTML to Go docs, now for ref docs we
+// must convert back. We can't use the model's original docs directly because
+// that doesn't include extra content we may inject at codegen.
```
isaiahvita, line 173: do you have an example of where we would add extra content at codegen on top of what we get from the service?

isaiahvita (approved):

Issue #2247, opened 2023-08-18
Support for io.Reader Interface in S3 Transfer Manager's Downloader


yacchi:
### Describe the feature

Currently, the Download function implemented in the Downloader accepts `io.WriteAt`. Due to this, after writing to a file or buffer, there is a need to create an `io.Reader`.

Frequently when working with files in Go, the `io.Reader` interface is commonly required. I believe that if the Downloader could directly produce an `io.Reader`, it would significantly improve usability.

### Use Case

- Decompressing archive files (e.g., tar, zip) stored in S3
  - The standard Go modules for tar and zip utilize the `io.Reader` interface. With this feature, it would be possible to directly load and decompress files from S3.
  - Especially when handling large files, the efficiency is improved as writing to storage occurs only once.

- Stream processing large files in environments with limited memory or storage
  - There are scenarios, for instance, where you might want to compute digests (e.g., sha, md5) using AWS Lambda.
  - By leveraging the `io.Reader` to process streams directly, it becomes possible to handle large files without using storage and with minimal memory consumption.

### Proposed Solution

The behavior of the AWS CLI's cp command closely aligns with my expectations. For instance, it can be used as follows:

```shell
aws s3 cp s3://BUCKET/key.tar.gz | tar zxf -
```

Internally, it appears to use the heap to sequentially output chunks from the beginning.
I've created code in my repository that operates in a similar manner using the current Downloader.

https://github.com/yacchi/s3-fast-reader

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2 v1.20.1
github.com/aws/aws-sdk-go-v2/config v1.18.33
github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.11.77
github.com/aws/aws-sdk-go-v2/service/s3 v1.38.2

### Go version used

1.20.4

RanVaknin:
Hi @yacchi, 

This seems like a reasonable request. We will likely work on this when we re-implement Downloader. 

For now we cannot prioritize this, but I will add this to our backlog.

Thanks!
Ran~

yacchi:
@lucix-aws 

Thanks for the helpful code. It looks like I can certainly do sequentialization that way.
However, if you don't want concurrency, I think the standard `GetObject` method is easier because it provides the `io.Reader` interface.

[...]

PR #2622, opened 2024-04-26
feat: add object io reader/writer to manager

jobstoit:
Hello I wanna contribute an io.Reader for the DownloadManager and io.WriteCloser for the UploadManager. 
I recently created my own wrapper [s3io](https://github.com/jobstoit/s3io) and figured I could just as well contribute the ObjectReader and ObjectWriter to the s3 sdk. (resolving issue #2247)

Having an io.Reader and io.Write(Close)r can significantly optimise memory consumption in your applications and is therefor a nice to have.
I designed it to also work in chunks/parts concurrently similar to how the current downloader/uploader work

This is my first PR here so maybe you can help me with the conventions. 

Let me know what you think and if you'd like me to finish this

### DIFF
--- a/feature/s3/manager/reader_test.go
+++ b/feature/s3/manager/reader_test.go
@@ -11,7 +11,38 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 )
 
-func TestObjectReader(t *testing.T) {
+func TestObjectReaderSinglePart(t *testing.T) {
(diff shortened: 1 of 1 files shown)


jobstoit:
Thank you for the response @lucix-aws. Hope the PR could give some inspiration on how to stream s3 objects concurrently. Best of luck with the redesign. 

I'll update my s3io package likely too cause I like my implementation I did  here better 😅.

Issue #2634, opened 2024-05-08
Support generic SigV4 and SigV4A signers


lucix-aws:
A handful of AWS services facilitate provision of an external resource that the user then communicates with using non-AWS SDK tooling, e.g.
* EKS creates kubernetes clusters which you then control with `kubectl`
* OpenSearch creates opensearch clusters which you then talk to through a separately-maintained library, https://github.com/opensearch-project/opensearch-go

Among these services, a handful require SigV4 or SigV4A authentication (e.g. opensearch, APIGateway, VPC lattice). The state of generic signer support across SDKs is fragmented and the Go v2 SDK is no exception:
* The sigv4 signer implementation is part of the public API but its implementation is not truly generic. It depends on various elements of the runtime and makes some decisions about header canonicalization and signing variables based on quirks within the AWS service fabric
* The sigv4a signer is entirely an implementation detail and correctly remains in an internal package. However, this means that users who require SigV4A functionality outside of an AWS service context have little to no recourse - as the algorithm isn't yet publicly documented, an implementation must be derived from an existing AWS-borne implementation (CRT, Rust/Go v2 SDKs).

This issue tracks the development work to support _generic, fully reusable_ APIs for modern AWS signing algorithms.

prashil-g:
Hi @lucix-aws we are blocked on this. Is there any ETA for this?

prashil-g:
Hi @lucix-aws  any update on when this support can be rolled out?

prashil-g:
Hi @lucix-aws any update on when this support can be rolled out?

lucix-aws:
Released: https://pkg.go.dev/github.com/aws/smithy-go/aws-http-auth

lucix-aws:
https://github.com/aws/aws-sdk-go-v2/issues/2634

prashil-g:
thanks a lot @lucix-aws Really appreciate it

PR #2648, opened 2024-05-21
Allow host header from request object

dink10:
In our team we actively use v4 signer for neptune DB. 

Just one case we faced. Sometimes for different environments we need to keep the host from created request and pass this request host to signer. Otherwise it won't work. 

My suggestion to keep this header host if it's not empty, otherwise pick up it from request url. 

I have quick&easy changes for a review. 

What do you think about my proposal and changes?


lucix-aws:
This is too behaviorally sensitive to change. Despite being public, the existing v4 signer in the SDK is more of an implementation detail and we don't intend to make any more adjustments to it to facilitate generic use.

We're tracking the addition of fully-generic signers (both v4 and v4a) in #2634, which I am actively working on.

PR #2666, opened 2024-06-04
add codec option to use encoding.Text/Binary(Un)Marshaler when present

### LINKED ISSUE #2596: Use `encoding.TextMarshaler` and `encoding.TextUnmarshaler` in DynamoDB `attributevalue` marshalers/unmarshalers, if available
### Describe the feature

When DynamoDB `attributevalue` marshals or unmarshals a struct, look to see if that struct implements `encoding.TextMarshaler` and `encoding.TextUnmarshaler` interfaces, and use them if available.

### Use Case

When dynamodb's `attributevalue` marshals and unmarshals a custom type, it requires that you implement the `attributevalue.Marshaler` interface on that type. This is somewhat awkward and requires dynamodb-specific code on a type that may have nothing to do with dynamodb, or it requires building some sort of wrapper marshaller
In many cases, all that we want to do is unmarshal or marshal a custom type as a dynamodb string (ie, a subfield of a struct that `MarshalMap` is being called upon, representing a DynamoDB record), thus we should be able to implement [encoding/text](https://pkg.go.dev/encoding#TextMarshaler)'s interfaces on our custom type interface that looks to see if a type has implemented this interface and, if so, use it to marshal (or unmarshal) to/from a dynamodb type using that interface, much like the https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/encode.go;l=36 encoding/json package does

### Proposed Solution

Update attributevalue's [Marshaler](https://github.com/aws/aws-sdk-go-v2/blob/feature/dynamodb/attributevalue/v1.13.13/feature/dynamodb/attributevalue/encode.go#L834) and Unmarshaler to try, in addition to the `attributevalue.Marshal` interface, the `encoding.Marshal` interface, converting to and from an `AttributeValueMemberS`.

### Other Information

_No response_

### Acknowledgements

1.13.13

### Go version used

1.22.1

lucix-aws:
Adds opt-in use of `encoding.TextMarshaler` / `encoding.BinaryMarshaler` (and unmarshal versions):

```
+
+       // When enabled, the encoder will use implementations of
+       // encoding.TextMarshaler and encoding.BinaryMarshaler when present on
+       // marshaled values.
+       //
+       // Implementations are checked in the following order:
+       //   - [Marshaler]
+       //   - encoding.TextMarshaler
+       //   - encoding.BinaryMarshaler
+       //
+       // The results of a MarshalText call will convert to string (S), results
+       // from a MarshalBinary call will convert to binary (B).
+       UseEncodingMarshalers bool
```

```
+
+       // When enabled, the decoder will use implementations of
+       // encoding.TextUnmarshaler and encoding.BinaryUnmarshaler when present on
+       // unmarshaling targets.
+       //
+       // If a target implements [Unmarshaler], encoding unmarshaler
+       // implementations are ignored.
+       //
+       // If the attributevalue is a string, its underlying value will be used to
+       // call UnmarshalText on the target. If the attributevalue is a binary, its
+       // value will be used to call UnmarshalBinary.
+       UseEncodingUnmarshalers bool
```

Closes #2596

### DIFF
--- a/feature/dynamodb/attributevalue/decode.go
+++ b/feature/dynamodb/attributevalue/decode.go
@@ -231,6 +231,18 @@ type DecoderOptions struct {
 	// Default string parsing format is time.RFC3339
 	// Default number parsing format is seconds since January 1, 1970 UTC
 	DecodeTime DecodeTimeAttributes
+
+	// When enabled, the decoder will use implementations of
+	// encoding.TextUnmarshaler and encoding.BinaryUnmarshaler when present on
+	// unmarshaling targets.
+	//
+	// If a target implements [Unmarshaler], encoding unmarshaler
+	// implementations are ignored.
+	//
+	// If the attributevalue is a string, its underlying value will be used to
+	// call UnmarshalText on the target. If the attributevalue is a binary, its
+	// value will be used to call UnmarshalBinary.
+	UseEncodingUnmarshalers bool
 }
 
 // A Decoder provides unmarshaling AttributeValues to Go value types.
@@ -288,16 +300,29 @@ func (d *Decoder) decode(av types.AttributeValue, v reflect.Value, fieldTag tag)
 	var u Unmarshaler
 	_, isNull := av.(*types.AttributeValueMemberNULL)
 	if av == nil || isNull {
-		u, v = indirect(v, indirectOptions{decodeNull: true})
+		u, v = indirect[Unmarshaler](v, indirectOptions{decodeNull: true})
 		if u != nil {
 			return u.UnmarshalDynamoDBAttributeValue(av)
 		}
 		return d.decodeNull(v)
 	}
 
-	u, v = indirect(v, indirectOptions{})
+	v0 := v
+	u, v = indirect[Unmarshaler](v, indirectOptions{})
 	if u != nil {
 		return u.UnmarshalDynamoDBAttributeValue(av)
+	}
+	if d.options.UseEncodingUnmarshalers {
+		if s, ok := av.(*types.AttributeValueMemberS); ok {
+			if u, _ := indirect[encoding.TextUnmarshaler](v0, indirectOptions{}); u != nil {
+				return u.UnmarshalText([]byte(s.Value))
+			}
+		}
+		if b, ok := av.(*types.AttributeValueMemberB); ok {
+			if u, _ := indirect[encoding.BinaryUnmarshaler](v0, indirectOptions{}); u != nil {
+				return u.UnmarshalBinary(b.Value)
+			}
+		}
 	}
 
 	switch tv := av.(type) {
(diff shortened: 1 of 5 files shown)


feature/dynamodb/attributevalue/decode.go
```diff
@@ -288,17 +300,30 @@ func (d *Decoder) decode(av types.AttributeValue, v reflect.Value, fieldTag tag)
...
-		u, v = indirect(v, indirectOptions{decodeNull: true})
+		u, v = indirect[Unmarshaler](v, indirectOptions{decodeNull: true})
 		if u != nil {
 			return u.UnmarshalDynamoDBAttributeValue(av)
 		}
 		return d.decodeNull(v)
 	}
 
-	u, v = indirect(v, indirectOptions{})
+	v0 := v
```
Madrigal, line 310: Are you using the same technique described by Go source code? [link](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/encoding/json/decode.go;l=438)

```go
func indirect(v reflect.Value, decodingNull bool) (Unmarshaler, encoding.TextUnmarshaler, reflect.Value) {
	// Issue #24153 indicates that it is generally not a guaranteed property
	// that you may round-trip a reflect.Value by calling Value.Addr().Elem()
	// and expect the value to still be settable for values derived from
	// unexported embedded struct fields.
	//
	// The logic below effectively does this when it first addresses the value
	// (to satisfy possible pointer methods) and continues to dereference
	// subsequent pointers as necessary.
	//

	// After the first round-trip, we set v back to the original value to
	// preserve the original RW flags contained in reflect.Value.
	v0 := v
```

If so, aren't we missing the last part of the method?

```go
if haveAddr {
	v = v0 // restore original value after round-trip Value.Addr().Elem()
	haveAddr = false
} else {
	v = v.Elem()
}

```

feature/dynamodb/attributevalue/encode_test.go
```diff
@@ -94,6 +94,95 @@ func (m customBoolStringMarshaler) MarshalDynamoDBAttributeValue() (types.Attrib
...
+
+	for _, testCase := range cases {
+		actual, err := MarshalWithOptions(testCase.input, func(o *EncoderOptions) {
+			o.UseEncodingMarshalers = testCase.useMarshalers
+		})
+		if err != nil {
+			t.Errorf("got unexpected error %v for input %v", err, testCase.input)
+		}
+		if diff := cmpDiff(testCase.expected, actual); len(diff) != 0 {
+			t.Errorf("expected match but got: %s", diff)
```
Madrigal, line 181: [minor] should we print the expected case as well, like `"expected %s but got %s"`?

feature/dynamodb/attributevalue/decode.go
lucix-aws: No this is different - this is just preserving the original reflect value so we can re-traverse to look for the new marshalers.

Madrigal (approved):

PR #2682, opened 2024-06-14
add SDK-specific feature tracking

lucix-aws:
Tracks usage of the following features in the user-agent string:
* retry mode (adaptive vs. standard)
* s3 express
* s3 transfer manager
* waiters
* paginators
* gzip compression


wty-Bryant (approved):

aws/middleware/user_agent_test.go
```diff
@@ -231,6 +231,71 @@ func TestAddUserAgentKeyValue(t *testing.T) {
...
+		},
+		"one": {
+			Features: []UserAgentFeature{
+				UserAgentFeatureWaiter,
+			},
+			Expect: "m/B " + expectedAgent,
+		},
+		"two": {
+			Features: []UserAgentFeature{
+				UserAgentFeatureRetryModeAdaptive, // ensure stable order, and idempotent
```
Madrigal, line 254: [nit] not sure if I'd call it stable order since this maps to a set at the end, and we end up sorting the feature set

aws/middleware/user_agent.go
```diff
@@ -64,9 +68,29 @@ var validChars = map[rune]bool{
 	'-': true, '.': true, '^': true, '_': true, '`': true, '|': true, '~': true,
 }
 
+// UserAgentFeature enumerates tracked SDK features.
+type UserAgentFeature string
+
+// Enumerates UserAgentFeature.
+const (
```
Madrigal, line 75: I double checked that these are correct (hate when I push a PR and then whoops typo)

Madrigal (approved):

PR #2688, opened 2024-06-18
add implicit global region to internal partition metadata

lucix-aws:
(note that this isn't used anywhere in rules currently*, but was just a missing field in the codegen for the partitions table)


wty-Bryant (approved):

Madrigal (approved):
Discussed offline. This is part of the model that comes with `partitions.json` as can be seen [here](https://github.com/aws/aws-sdk-go-v2/blob/077df5deb4f94e0cacc5c64c4538e49b6c711563/internal/endpoints/awsrulesfn/partitions.json#L7)

PR #2690, opened 2024-06-19
validate put object content length

### LINKED ISSUE #2683: non-`io.Seeker` body in streaming-blob operation causes content-length to not be set, resulting in 5xx
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

I've seen that there have been previous reports of this issue, e.g. https://github.com/aws/aws-sdk-go-v2/issues/1108

However the resolution is missing what feels like the actual problem: the SDK clearly provides `io.Reader` as the interface and then returns an error when given a Body that implements that interface.

I cannot use an `http.Response.Body`  as the `PutObjectInput.Body`, which is a bug either in implementation of the S3 client `PutObject` method or with the Go typing.

### Expected Behavior

If the SDK says that the `PutObjectInput` supports `io.Reader`: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#PutObjectInput it should support them.

Otherwise the SDK should state what interface it actually supports.

### Current Behavior

SDK operation returns a 501, `api error NotImplemented: A header you provided implies functionality that is not implemented`

### Reproduction Steps

- Make an http request and store the response body (io.ReadCloser)
- Use that response body as the `PutObjectInput.Body`

### Possible Solution

_No response_

_No response_

### Compiler and Version used

go version go1.22.4 darwin/arm64

MacOS 14.5 (23F79)

lucix-aws:
Add specific validation for content-length presence in PutObject, see middleware docs

Closes #2683 or at least mitigates it somewhat

### DIFF
--- a/service/s3/put_object_content_length_test.go
+++ b/service/s3/put_object_content_length_test.go
@@ -3,8 +3,8 @@ package s3
 import (
 	"bytes"
 	"context"
+	"errors"
 	"io"
-	"strings"
 	"testing"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -62,10 +62,10 @@ func TestValidatePutObjectContentLength(t *testing.T) {
 			})
 
 			_, err := svc.PutObject(context.Background(), cs.Input)
-			if cs.ExpectErr && !strings.Contains(err.Error(), errNoContentLength.Error()) {
+			if cs.ExpectErr && !errors.Is(err, errNoContentLength) {
 				t.Errorf("expected errNoContentLength, got %v", err)
 			}
-			if !cs.ExpectErr && strings.Contains(err.Error(), errNoContentLength.Error()) {
+			if !cs.ExpectErr && errors.Is(err, errNoContentLength) {
 				t.Errorf("expected no errNoContentLength but got it")
 			}
 		})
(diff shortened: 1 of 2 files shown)


service/s3/put_object_content_length_test.go
```diff
@@ -0,0 +1,73 @@
...
+			ExpectErr: false,
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			svc := New(Options{
+				Region: "us-east-1",
+			})
+
+			_, err := svc.PutObject(context.Background(), cs.Input)
+			if cs.ExpectErr && !strings.Contains(err.Error(), errNoContentLength.Error()) {
```
Madrigal, line 65: [nit] can this be compared with `errors.Is` or `errors.As` instead of doing string comparisons?

wty-Bryant (approved):

service/s3/put_object_content_length.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/aws/smithy-go/middleware"
+)
+
+var errNoContentLength = errors.New(
```
Madrigal, line 13: nice error message

Madrigal (approved):

service/s3/put_object_content_length_test.go
lucix-aws: Yeah not sure what I was doing here. This is `errors.Is` with extra steps

PR #2696, opened 2024-06-24
nuke opsworks test

lucix-aws:


wty-Bryant (approved):

Madrigal (approved):
Discussed offline. OpsWorks is no longer accepting new customers and this test is broken because of that.

PR #2703, opened 2024-07-03
fix LoadDefaultConfig optFns

kentokento:
I want to use config.LoadOptionsFunc as below

```
	// var opts []func(*config.LoadOptions) error
	var opts []config.LoadOptionsFunc // now this will result in an error
	if c.AccessKey != "" && c.SecretKey != "" {
		opts = append(opts, config.WithCredentialsProvider(
			&credentials.StaticCredentialsProvider{
				Value: aws.Credentials{
					AccessKeyID:     c.AccessKey,
					SecretAccessKey: c.SecretKey,
				},
			},
		))
	}

	if c.Profile != "" {
		opts = append(opts, config.WithSharedConfigProfile(c.Profile))
	}
	if c.Filename != "" {
		opts = append(opts, config.WithSharedConfigFiles([]string{c.Filename}))
	}
	return config.LoadDefaultConfig(context.TODO(), opts...)
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -183,7 +183,7 @@ func (cs configs) ResolveConfig(f func(configs []interface{}) error) error {
 // The default configuration sources are:
 // * Environment Variables
 // * Shared Configuration and Shared Credentials files.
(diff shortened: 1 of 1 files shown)


lucix-aws:
We can't do this, it's an API break. See previous discussion in #2193.

PR #2709, opened 2024-07-10
save sso cache token expiresAt in UTC

clhuang:
If the expiresAt field is saved with time zone (e.g. 2024-06-10T15:00:06-08:00) it will fail to load in certain SDKs such as the rust AWS SDK. To avoid this, ensure that it is always saved as a UTC format.

### DIFF
--- a/.changelog/ebcf2e12f2ba4d0f9aa44b48b850c41f.json
+++ b/.changelog/ebcf2e12f2ba4d0f9aa44b48b850c41f.json
@@ -1,0 +1,8 @@
+{
+    "id": "ebcf2e12-f2ba-4d0f-9aa4-4b48b850c41f",
+    "type": "bugfix",
+    "description": "Save SSO cached token expiry in UTC to ensure cross-SDK compatibility.",
+    "modules": [
+        "credentials"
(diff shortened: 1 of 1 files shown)


lucix-aws:
Took another look at this, a few comments
* The new tests are fine but you didn't need to create a new suite -- we appear to have an existing one where this could have fit. You're welcome to roll them up or leave as-is
* I'd like to see some additional guarantees:
  * that we can still _load_ the token that we save
  * that we can still _load_ the token in both the 3339Z (changed behavior) and 3339+offset (original behavior) formats

The 2nd point may be covered by existing tests, I scanned them but may have missed one. Please either confirm their presence or add if missing

clhuang:
I opted to not change the existing tests because the existing tests assert that the saved datetime object is the exact same as the loaded one, which is actually not the case when the token expiresAt is non-UTC. I would have to significantly refactor the existing tests and the use of cmpDiff.

clhuang:
updated by adding a new helper cmpDiffToken that ignores time zone diffs in expiresAt

lucix-aws (approved):

PR #2716, opened 2024-07-17
provide a region hint of "us-west-2" in GetBucketRegion example

kaovilai:

Related work: https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/210

### DIFF
--- a/.changelog/45d613f9451e478c866ae676b72ab246.json
+++ b/.changelog/45d613f9451e478c866ae676b72ab246.json
@@ -1,0 +1,8 @@
+{
+    "id": "45d613f9-451e-478c-866a-e676b72ab246",
+    "type": "documentation",
+    "description": "Clarify region hint and credential usage in HeadBucketRegion.",
+    "modules": [
+        "feature/s3/manager"
+    ]
+}


lucix-aws (commented):
Why the addition of AnonymousCredentials in the example? That's going to cause any requests against non-public buckets to fail here.

kaovilai:
The anonymous credentials was added because if you don't have a specified ~/.aws/credentials, this call would fail with the current example prior to this PR.

Error only went away once a credential provider is defined when loading config. If you say it's needed, I can make a note in the PR. WDYT?

kaovilai:
`// The request will not be signed, and will not use your AWS credentials.`

> That's going to cause any requests against non-public buckets to fail here.

So I guess creds could be used on a needed basis then?

kaovilai:
So I have two downstream PRs
- https://github.com/vmware-tanzu/velero-plugin-for-aws/pull/210 could potentially rely on private buckets as we have many users there.
- https://github.com/openshift/oadp-operator/pull/1470 this we have a unit test run in a container without credentials, but also contain end to end tests while it do not test private buckets, could certainly do so in customer environments. So I think your point is valid here.

I need to dynamically do config depending on env I guess. If in credentialess unit tests, anonymouse.. otherwise do not put in anonymous credentials.

lucix-aws:
That makes sense to me.

So in terms of this PR:
* adding region hint in the example is good
* let's remove the example's use of Anonymous
* you're welcome to remove the blurb in this API doc where it states `// The request will not be signed, and will not use your AWS credentials.` because that's definitely wrong - but you're equally welcome to just leave it for now

feature/s3/manager/bucket_region.go
```diff
@@ -25,7 +25,7 @@ const bucketRegionHeader = "X-Amz-Bucket-Region"
 // For example to get the region of a bucket which exists in "eu-central-1"
 // you could provide a region hint of "us-west-2".
 //
-//	cfg, err := config.LoadDefaultConfig(context.TODO())
+//	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion("us-west-2"), config.WithCredentialsProvider(aws.AnonymousCredentials{}))
```
kaovilai, line 28: How about this?
```suggestion
//	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion("us-west-2"))
```

kaovilai: got it
https://github.com/aws/aws-sdk-go-v2/pull/2716#issuecomment-2233770483

Will remove the blurb about not using cred as well.

kaovilai:
Corrected notes on credentials using notes from
https://github.com/kaovilai/aws-sdk-go-v2/blob/f7a89262286d7ca7f2307a2df4214d0993d19f5d/aws/credentials.go#L54-L61

lucix-aws (approved):

Issue #2731, opened 2024-08-02
Custom Marshaler returning NULL should be omitted from .MarshalMap() with omitempty tag


babattles:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

When passing a struct with a custom type field, tagged with `omitempty`, that implements the Marshaler interface to .MarshalMap(), the field should be omitted from the resulting map if the custom field's `.MarshalDynamoDBAttributeValue()` implementation returns `&types.AttributeValueMemberNULL{Value: true}, nil`.

### Expected Behavior

omitempty should omit the custom field from the resulting map returned from `.MarshalMap()` if the value is of type NULL

### Current Behavior

`.MarshalMap()` returns a map with a NULL field like: `map[string]types.AttributeValue{"Val":(*types.AttributeValueMemberNULL)(0x1400021ea08)}`

### Reproduction Steps

```go
import (
	"testing"

	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

type OnlyNull struct{}

func (o OnlyNull) MarshalDynamoDBAttributeValue() (types.AttributeValue, error) {
	return &types.AttributeValueMemberNULL{Value: true}, nil
}

func TestMaybe(t *testing.T) {
	t.Parallel()

	type MyStruct struct {
		Val OnlyNull `dynamodbav:",omitempty"`
	}

	// Test Fails
	t.Run("omitted", func(t *testing.T) {
		t.Parallel()

		o := MyStruct{}
		val, err := attributevalue.MarshalMap(o)
		require.NoError(t, err)
		assert.Empty(t, val)
	})
}
```

### Possible Solution

Either allow NULLs to be omitted by omitempty, or provide another type that implements `types.AttributeValue` that is able to be omitted. 

### AWS Go SDK V2 Module Versions Used

```
github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue v1.14.10
github.com/aws/aws-sdk-go-v2/service/dynamodb v1.34.4
[...]

Issue #1429, opened 2021-09-20
Add support for omitting attribute values returned by custom Marshaler implementations


lmika:
**Is your feature request related to a problem? Please describe.**

We are working on a project that uses a custom type to represent UUIDs.  This type, which extends `github.com/google/uuid`, stores the UUID in memory as a `[16]byte` array, but we’ve added a custom `Marshal` and `Unmarshal` implementation that marshals/unmarshals it to a string UUID representation.

We were hoping to allow the service writing rows to the DynamoDB table to omit an `id` field when the UUID is nil (i.e. 16 zero bytes).  However, we’ve discovered that this is not possible to do when using a type that implements
`github.com/aws/aws-sdk-go-v2/service/dynamodb/dynamodbattribute.Marshaler`.  We’re hoping that this could be added in some way.

**Describe the solution you'd like**

There’s not a huge preference as to how this is implemented, either the following suggestion or one of the alternatives will be find.  Depending on the stability of the API, an ideal solution would be to return double `nil` from `Marshal.MarshalDynamoDBAttributeValue` to indicate that no attribute value should be marshalled out.  Example:

```go
func (u UUID) MarshalDynamoDBAttributeValue() (types.AttributeValue, error) {
	if u == NilUUID {
		// Indicate that no value should be stored for this attribute.
		return nil, nil
	}

	return &types.AttributeValueMemberS{
		Value: u.String(),
	}, nil
}

```

**Describe alternatives you've considered**

A few other suggestions for this approach:

- Add support for `omitempty` in the struct tag for custom types.  So using the example above, if `u` is `NilUUID`, then the function will return a `types.AttributeValueMemberS` with an empty string.  Since the field that has this type also has the `omitempty` option in the struct tag, the value won’t be marshalled out to the database.
- Either a dedicated attribute type or error type that acts as a sentinal would also work.

**Additional context**

We’re using “aws-sdk-go-v2” version 1.5.0 with Go 1.16.6

lucix-aws:
I think this is addressed in #2739?

PR #2732, opened 2024-08-03
omitempty for NULL attribute values from custom marshalers

babattles:
The goal here is to allow implementors of the Marshaler interface to be able to return a value from `MarshalDynamoDBAttributeValue()` that is able to be omitted via the `omitempty` tag.

Related Issues:
- https://github.com/aws/aws-sdk-go-v2/issues/1429
- https://github.com/aws/aws-sdk-go-v2/issues/2731


lucix-aws:
https://github.com/aws/aws-sdk-go-v2/issues/2731#issuecomment-2269530592

PR #2733, opened 2024-08-04
Feat s3 transfer manager v2 PutObject

wty-Bryant:
Implement v2 s3 transfer manager's putobject api bound to single union client which mimics normal service client's initialization and api call.

Test: passed unit test for putobject and pool, passed integration test of uploading object via transfer manager v2 and validating its content after downloading through s3 client

### DIFF
--- a/feature/s3/transfermanager/api_op_PutObject.go
+++ b/feature/s3/transfermanager/api_op_PutObject.go
@@ -55,8 +55,6 @@ type multipartUploadError struct {
 
 // Error returns the string representation of the error.
 //
-// # See apierr.BaseError ErrorWithExtra for output format
-//
 // Satisfies the error interface.
 func (m *multipartUploadError) Error() string {
 	var extra string
@@ -936,7 +934,6 @@ func (u *multiUploader) seterr(e error) {
 	u.err = e
 }
 
-// fail will abort the multipart unless LeavePartsOnError is set to true.
 func (u *multiUploader) fail(ctx context.Context, clientOptions ...func(*s3.Options)) {
 	params := u.in.mapAbortMultipartUploadInput(u.uploadID)
 	_, err := u.options.S3.AbortMultipartUpload(ctx, params, clientOptions...)
--- a/feature/s3/transfermanager/internal/testing/upload.go
+++ b/feature/s3/transfermanager/internal/testing/upload.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"slices"
 	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -68,7 +69,7 @@ func (f httpDoFunc) Do(r *http.Request) (*http.Response, error) {
 }
 
 func (u *UploadLoggingClient) traceOperation(name string, params interface{}) {
-	if contains(u.ignoredOperations, name) {
+	if slices.Contains(u.ignoredOperations, name) {
 		return
 	}
 	u.Invocations = append(u.Invocations, name)
@@ -189,13 +190,4 @@ func NewUploadLoggingClient(ignoredOps []string) (*UploadLoggingClient, *[]strin
 	}
 
 	return c, &c.Invocations, &c.Params
-}
-
-func contains(src []string, s string) bool {
-	for _, v := range src {
-		if v == s {
-			return true
-		}
-	}
-	return false
 }


lucix-aws:
Tests are failing.

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+		o.partPool = &returnCapacityPoolCloser{byteSlicePool: o.partPool}
+		o.partPool.ModifyCapacity(poolCap)
+	}
+
+	return nil
+}
+
+// initSize tries to detect the total stream size, setting u.totalSize. If
+// the size is not known, totalSize is set to -1.
+func (u *uploader) initSize() error {
```
lucix-aws, line 352: I don't really see how this is necessary. We should just keep pulling bytes off of the input (into slices from the pool) until we run out.

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+	awsutil.Copy(&params, u.in)
+	params.Body = r
+
+	var locationRecorder recordLocationClient
+	out, err := u.cfg.options.S3.PutObject(u.ctx, &params, append(u.cfg.options.PutClientOptions, locationRecorder.WrapClient())...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &PutObjectOutput{
```
lucix-aws, line 398: Use a helper function, this is going to get big.

feature/s3/transfermanager/pool.go
```diff
@@ -0,0 +1,252 @@
...
+
+type byteSlicePool interface {
+	Get(context.Context) (*[]byte, error)
+	Put(*[]byte)
+	ModifyCapacity(int)
+	SliceSize() int64
+	Close()
+}
+
+type maxSlicePool struct {
```
lucix-aws, line 18: See my comment elsewhere about non-necessity of shared pool. I think we can greatly simplify this by just having one fixed pool per upload call.

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+			addFeatureUserAgent, // yes, there are two of these
+			func(s *smithymiddleware.Stack) error {
+				return s.Finalize.Insert(&setS3ExpressDefaultChecksum{}, "ResolveEndpointV2", smithymiddleware.After)
+			},
+		)
+	})
+	clientOptions = append(clientOptions, c.options.PutClientOptions...)
+	i.cfg.options.PutClientOptions = clientOptions
+	for _, opt := range opts {
+		opt(&i.cfg.options)
```
lucix-aws, line 266: You're mutating options here - you have to create a `Copy()` of them local to each client API and then apply functional options therein so the original is preserved.

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+			// Add one to the part size to account for remainders
+			// during the size calculation. e.g odd number of bytes.
+			u.cfg.options.PartSizeBytes = (totalSize / int64(u.cfg.options.MaxUploadParts)) + 1
+		}
+	}
+
+	return nil
+}
+
+func getLen(in *PutObjectInput) (int64, error) {
```
lucix-aws, line 373: This doesn't seem to be used anywhere? Nor should it be - we should never have to size the whole input body.

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+
+		ChecksumAlgorithm: u.in.ChecksumAlgorithm,
+		// Invalid to set any of the individual ChecksumXXX members from
+		// PutObject as they are never valid for individual parts of a
+		// multipart upload.
+
+		PartNumber: c.partNum,
+		UploadId:   u.uploadID,
+	}
+	// TODO should do copy then clear?
```
lucix-aws: ?

lucix-aws (changes requested):

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+	}
+
+	mu := multiUploader{
+		uploader: u,
+	}
+	return mu.upload(r, cleanUp)
+}
+
+func (u *uploader) init() error {
+	if err := validateSupportedARNType(aws.ToString(u.in.Bucket)); err != nil {
```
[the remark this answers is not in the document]
Madrigal: Was a bit curious about this to spend some time looking into it. S3 Object Lambda is a feature that allows you to front operation on a bucket like `GetObject` with a Lambda that can do transformations on the object. When you create an Object Lambda, you get an `ARN` that you call instead of the ARN of the bucket, since it's a new access point. This access point only supports GET, HEAD and LIST operations, so a PutObject call will fail (I suppose, haven't tried it to see *how* exactly it fails)

IMO this seems a bit niche to add a validation transfer manager

feature/s3/transfermanager/putobject.go
```diff
@@ -0,0 +1,772 @@
...
+}
+
+func (u *uploader) singleUpload(r io.Reader, cleanUp func()) (*PutObjectOutput, error) {
+	defer cleanUp()
+
+	var params s3.PutObjectInput
+	awsutil.Copy(&params, u.in)
+	params.Body = r
+
+	var locationRecorder recordLocationClient
```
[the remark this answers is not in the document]
Madrigal: It seems ergonomic enough to me, kind of like returning a `201 Created` [response](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-semantics#section-15.3.2).

I'm more concerned about standardization with other SDKs, so I'd like to hear if this was something on the SEP or if we just thought it was neat

feature/s3/transfermanager/transfermanager.go
```diff
@@ -0,0 +1,78 @@
...
+// using PutObject().
+const DefaultTransferConcurrency = 5
+
+type Client struct {
+	options Options
+}
+
+// New returns an initialized Client from the client Options. Provide
+// more functional options to further configure the Client
+func New(opts Options, optFns ...func(*Options)) *Client {
```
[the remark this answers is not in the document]
lucix-aws: Yeah. I think as a rule of thumb, we should have high-level clients in the form of

```go
func New(LowLevelClient, Options, ...func(*Options)
```

(same for NewFromConfig)

feature/s3/transfermanager/pool.go
```diff
@@ -0,0 +1,252 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+
+	p.mtx.RLock()
+
+	for {
+		select {
+		case bs, ok := <-p.slices:
```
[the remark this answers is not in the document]
lucix-aws: (same)

feature/s3/transfermanager/api_client.go
```diff
@@ -0,0 +1,47 @@
...
+
+// DefaultMaxUploadParts is the maximum allowed number of parts in a multi-part upload
+// on Amazon S3.
+const DefaultMaxUploadParts int32 = 10000
+
+// DefaultPartSizeBytes is the default part size when transferring objects to/from S3
+const DefaultPartSizeBytes int64 = 1024 * 1024 * 8
+
+// DefaultMultipartUploadThreshold is the default size threshold in bytes indicating when to use multipart upload.
+const DefaultMultipartUploadThreshold int64 = 1024 * 1024 * 16
```
Madrigal, line 17: [nit] I don't think this has to be typed to `int64`

feature/s3/transfermanager/api_client.go
```diff
@@ -0,0 +1,47 @@
...
+
+// New returns an initialized Client from the client Options. Provide
+// more functional options to further configure the Client
+func New(s3Client S3Client, opts Options, optFns ...func(*Options)) *Client {
+	opts.S3 = s3Client
+	for _, fn := range optFns {
+		fn(&opts)
+	}
+
+	opts.init()
```
Madrigal, line 37: The style we follow on all clients is that `NewFromConfig` initializes the options object inline and `New` just uses the Options object passed to it, using `resolveXX` functions to set defaults if none is set. For the sake of consistency I'd like to keep that style, although I'm open to hear if you think otherwise

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+
+	// UploadID returns the upload id for the S3 multipart upload that failed.
+	UploadID() string
+}
+
+// A multiUploadError wraps the upload ID of a failed s3 multipart upload.
+// Composed of BaseError for code, message, and original error
+//
+// Should be used for an error that occurred failing a S3 multipart upload,
+// and a upload ID is available. If an uploadID is not available a more relevant
```
Madrigal, line 48: The last sentence is incomplete

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+	return m.err
+}
+
+// UploadID returns the id of the S3 upload which failed.
+func (m *multipartUploadError) UploadID() string {
+	return m.uploadID
+}
+
+// PutObjectInput represents a request to the Upload() call.
+type PutObjectInput struct {
```
Madrigal, line 80: Expand the comment and mention that this is an object that is the common field of `PutObject` and `CreateMultiPartUpload`.

I'm not sure about calling this `PutObjectInput`. On one hand is likely what you'd look for when using the transfer manager. However, it is confusing that it's almost exactly the same as `PutObjectInput` on S3 but not really.

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+
+func (u *uploader) upload(ctx context.Context) (*PutObjectOutput, error) {
+	if err := u.init(); err != nil {
+		return nil, fmt.Errorf("unable to initialize upload: %w", err)
+	}
+	defer u.partPool.Close()
+
+	clientOptions := []func(o *s3.Options){
+		func(o *s3.Options) {
+			o.APIOptions = append(o.APIOptions,
```
Madrigal, line 746: These options don't seem dynamic, can we add them when we create the client instead of when we call `upload` ?

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+
+	mu := multiUploader{
+		uploader: u,
+	}
+	return mu.upload(ctx, r, cleanUp, clientOptions...)
+}
+
+func (u *uploader) init() error {
+	partSize := u.options.PartSizeBytes
+	if partSize < DefaultPartSizeBytes {
```
Madrigal, line 769: why do we err on this? if customers want to upload 8 bytes at a time, should we block them? 

(also if this is a legitimate error we want to throw, we should change the name from `Default` to `Min`)

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+	return &output, nil
+}
+
+type httpClient interface {
+	Do(r *http.Request) (*http.Response, error)
+}
+
+type recordLocationClient struct {
+	httpClient
+	location string
```
Madrigal, line 799: Is recording the location of the object on the client thread safe?

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+
+	if nextChunkLen == 0 {
+		// No need to upload empty part, if file was empty to start
+		// with empty single part would of been created and never
+		// started multipart upload.
+		return false, nil
+	}
+
+	// This upload exceeded maximum number of supported parts, error now.
+	if part > DefaultMaxUploadParts {
```
Madrigal, line 948: This should be `MaxUploadParts` then instead of just `DefaultMax...`. But same question as previous, why do we want to limit the number of parts used?

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+		return false, fmt.Errorf(fmt.Sprintf("exceeded total allowed S3 limit MaxUploadParts (%d). Adjust PartSize to fit in this limit", DefaultMaxUploadParts))
+	}
+
+	return true, err
```
Madrigal, line 952: [nit] while we checked that err is not nil here, it would be clearer to return `true, nil`

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+
+	u.err = e
+}
+
+// fail will abort the multipart unless LeavePartsOnError is set to true.
+func (u *multiUploader) fail(ctx context.Context, clientOptions ...func(*s3.Options)) {
+	params := u.in.mapAbortMultipartUploadInput(u.uploadID)
+	_, err := u.options.S3.AbortMultipartUpload(ctx, params, clientOptions...)
+	if err != nil {
+		//logMessage(u.cfg.S3, aws.LogDebug, fmt.Sprintf("failed to abort multipart upload, %v", err))
```
Madrigal, line 1025: [nit] commented code

feature/s3/transfermanager/api_op_PutObject.go
lucix-aws: I'm still not convinced that we need to support this feature...

feature/s3/transfermanager/pool.go
lucix-aws: You removed the shared-ness of the pool but you didn't really simplify anything:
1. The pool sized is now fixed, since it's not shared, it never has to grow or shrink. The entire `ModifyCapacity` API and business logic to support it should not longer be necessary.
2. None of the read/write locking here should be necessary. [Buffered channels](https://go.dev/tour/concurrency/3) inherently give us synchronization across goroutines, this should be as simple as Get() and Put() calls that pull and push to such a channel.

Frankly I would recommend you start from scratch here. A buffered-channel-based fixed size pool should be trivial to implement in significantly less code vs. inheriting/bastardizing the old version of this.

feature/s3/transfermanager/types.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+)
+
+// ChecksumAlgorithm indicates the algorithm used to create the checksum for the object
+type ChecksumAlgorithm string
+
+// Enum values for ChecksumAlgorithm
+const (
+	ChecksumAlgorithmCrc32  ChecksumAlgorithm = "CRC32"
+	ChecksumAlgorithmCrc32c ChecksumAlgorithm = "CRC32C"
```
lucix-aws, line 13: You can omit the `ChecksumAlgorithm` type declaration on every line but the first, compiler figures it out

feature/s3/transfermanager/types.go
```diff
@@ -0,0 +1,205 @@
...
+	// The base64-encoded, 32-bit CRC32 checksum of the object. This will only be
+	// present if it was uploaded with the object. When you use an API operation on an
+	// object that was uploaded using multipart uploads, this value may not be a direct
+	// checksum value of the full object. Instead, it's a calculation based on the
+	// checksum values of each individual part. For more information about how
+	// checksums are calculated with multipart uploads, see [Checking object integrity]in the Amazon S3 User
+	// Guide.
+	//
+	// [Checking object integrity]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#large-object-checksums
+	ChecksumCRC32 *string
```
lucix-aws, line 124: Why pointers? zero value for unset seems fine

feature/s3/transfermanager/api_op_PutObject.go
lucix-aws: There's an S3-enforced 5MB per-part minimum, your comment on changing to `Min` is accurate.

feature/s3/transfermanager/api_op_PutObject.go
lucix-aws: It appears there is in fact a 10k part limit for MPUs, I didn't catch this at first.

We're going to need to discuss offline how we want to handle this. The error case we're hitting here is unacceptable (failing a 10,000 part MPU only after we realize there are 10k parts).

lucix-aws (changes requested):
We need to seriously re-examine the part pool behavior. It seems entirely overcomplicated for what we need out of it.

We need to discuss how we manage maximum part size of 10k for MPUs - failing on an in-progress download from exceeding the limit after the fact isn't acceptable behavior.

feature/s3/transfermanager/api_client.go
lucix-aws: Agree

feature/s3/transfermanager/api_client.go
lucix-aws: It doesn't, none of these constants do, compiler is supposed to figure it out in context of usage

feature/s3/transfermanager/api_op_PutObject.go
lucix-aws: If we're trying to match what our normal clients look like, then it _should_ be PutObjectInput, since the operation is PutObject.

I'm fine with that naming since this is basically "put an object", where multipart split is an optimization therein.

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,1077 @@
...
+	//
+	//   - This functionality is not supported for directory buckets.
+	//
+	//   - This functionality is not supported for Amazon S3 on Outposts.
+	//
+	// [Managing ACLs Using the REST API]: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-using-rest-api.html
+	// [Access Control List (ACL) Overview]: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html
+	// [Canned ACL]: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#CannedACL
+	// [Controlling ownership of objects and disabling ACLs]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html
+	ACL ObjectCannedACL
```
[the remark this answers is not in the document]
Madrigal: that sound like a great idea

service/internal/integrationtest/s3transfermanager/setup_test.go
```diff
@@ -0,0 +1,259 @@
...
+	return cleanup, nil
+}
+
+type putObjectTestData struct {
+	Body        io.Reader
+	ExpectBody  []byte
+	ExpectError string
+}
+
+func testPutObject(t *testing.T, bucket string, testData putObjectTestData, opts ...func(options *tm.Options)) {
```
lucix-aws, line 220: nice

feature/s3/transfermanager/api_op_PutObject.go
lucix-aws: > Expand the comment and mention that this is an object that is the common field of PutObject and CreateMultiPartUpload.

This remains unaddressed.

lucix-aws (approved):

feature/s3/transfermanager/internal/testing/upload.go
```diff
@@ -0,0 +1,201 @@
...
+// NewUploadLoggingClient returns a new UploadLoggingClient.
+func NewUploadLoggingClient(ignoredOps []string) (*UploadLoggingClient, *[]string, *[]interface{}) {
+	c := &UploadLoggingClient{
+		ignoredOperations: ignoredOps,
+	}
+
+	return c, &c.Invocations, &c.Params
+}
+
+func contains(src []string, s string) bool {
```
Madrigal, line 194: nit: we now have `slices.Contains` that can replace this

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,995 @@
...
+
+// seterr is a thread-safe setter for the error object
+func (u *multiUploader) seterr(e error) {
+	u.m.Lock()
+	defer u.m.Unlock()
+
+	u.err = e
+}
+
+// fail will abort the multipart unless LeavePartsOnError is set to true.
```
Madrigal, line 939: Which option is this?

I suspect we are trying to do this

> As a result, it might be necessary to abort a given multipart upload multiple times in order to completely free all storage consumed by all parts.
> To verify that all parts have been removed and prevent getting charged for the part storage, you should call the [ListParts](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html) API operation and ensure that the parts list is empty.

feature/s3/transfermanager/api_op_PutObject.go
```diff
@@ -0,0 +1,995 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/middleware"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/types"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
+	smithymiddleware "github.com/aws/smithy-go/middleware"
+)
+
+// A MultipartUploadError wraps a failed S3 multipart upload. An error returned
```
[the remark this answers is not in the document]
Madrigal: it does, checked on my end

Madrigal (approved):

PR #2739, opened 2024-08-09
omitempty for NULL attribute values from custom marshalers

babattles:
Adds an `EncoderOption` that allows the encoder to omit NULL attribute values when marshaling a field tagged with `omitempty`.

Related issues:
- #2731
- #1429


lucix-aws (approved):

lucix-aws:
Once CI passes I'm going to push a clarification to the doc string on this new field, since `omitempty` _does_ chop out NULLs for values in the normal marshaling workflow e.g.

```go
type Foo struct {
	A string
	B *string `dynamodbav:",omitempty"` // will get omitted today if B == nil
}
```

Basically it'll be something like

```
	// When enabled, the encoder will omit null (NULL) attribute values returned
	// from custom marshalers tagged with `omitempty`.
```

Issue #2746, opened 2024-08-14
(short issue description)


dmsolutionsmn:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

Value REPOSITORY_NAME is missing from `com.amazonaws.codebuild#WebhookFilterType`

### Expected Behavior

as per https://github.com/aws/aws-sdk-go-v2/blob/main/service/codebuild/types/types.go#L2638 there should be a REPOSITORY_NAME value

### Current Behavior

no REPOSITORY_NAME value

### Reproduction Steps

check https://github.com/aws/aws-sdk-go-v2/blob/main/service/codebuild/types/enums.go#L975-L1000 and notice it is not there

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2/service/codebuild v1.40.3

### Compiler and Version used

1.23.0

### Operating System and version

mac OS (darwin)

RanVaknin:
Hi @dmsolutionsmn ,

This is now fixed.

Thanks,
Ran~

PR #2747, opened 2024-08-14
add missing value for codebuild WebhookFilterType

dmsolutionsmn:
This addresses https://github.com/aws/aws-sdk-go-v2/issues/2746

Updated generated code with the following:

- updated json
- make smithy-generate
- make format
- git restore 'go.mod'


lucix-aws:
You can't modify the model directly in this manner. These API models are owned by each service team and are shipped to the SDK as part of a daily release process.

PR #2758, opened 2024-08-22
Add a `PresignPostObject` to the `PresignClient` for `s3.PutObject` operation

### LINKED ISSUE #1224: Add convenience method for S3 presigned posts
**Is your feature request related to a problem? Please describe.**
The presign client for S3 currently supports presigning `GetObject` and `PutObject` URLs: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#PresignClient

It would be nice if an additional method is implemented to support presigning POST object URLs documented here: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html

This feature seems to be present in SDKs for other languages.

**Describe the solution you'd like**
Add a convenience method for S3 presigned posts to the `PresignClient`

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

Madrigal:

This is used as follows

```go
// conditions that you want your request to have
// see https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html#sigv4-PolicyConditions
conditions := []interface{}{
	[]interface{}{
		"content-length-range",
		1,
		10485760,
	},
}
// Duration on which you want this request to
expiresIn = 20 * time.Minute
req, err := presignClient.PresignPostObject(examplePresigner.ctx, &s3.PutObjectInput{
	Bucket: aws.String(bucketName),
	Key:    aws.String(uploadKey),
}, func(opts *s3.PresignPostOptions) {
	opts.Expires = expiresIn
	opts.Conditions = conditions
})
if err != nil {
	log.Printf("Couldn't get a presigned request to put %v:%v. Here's why: %v\n",
		bucketName, uploadKey, err)
}
fmt.Println(req.URL) 
// https://test-bucket.s3.us-west-2.amazonaws.com
fmt.Println(req.Values)
// map[X-Amz-Algorithm:AWS4-HMAC-SHA256
// X-Amz-Credential:ASIAXXXX/20240822/us-west-2/s3/aws4_request
// X-Amz-Date:20240822T215442Z
// X-Amz-Signature:d7c4f16a9a0b8d60a7ed4f01234b0f9c2e58f12d0d3a847f91f1b3e7d6f8192a
// key:my_file.json
// policy:base64encodedValue]
```

fixes #1224

### DIFF
--- a/service/s3/presign_post.go
+++ b/service/s3/presign_post.go
@@ -82,7 +82,7 @@ type PresignedPostRequest struct {
 	URL string
 	// Values is a key-value map of values to be sent as FormData
 	// these values are not encoded
-	Values map[string]interface{}
+	Values map[string]string
 }
 
 // postSignAdapter adapter to implement the presignPost interface
@@ -102,7 +102,7 @@ func (s *postSignAdapter) PresignPost(
 	credentials aws.Credentials,
 	bucket string, key string,
 	region string, service string, signingTime time.Time, conditions []interface{}, expirationTime time.Time, optFns ...func(*v4.SignerOptions),
-) (fields map[string]interface{}, err error) {
+) (fields map[string]string, err error) {
 	credentialScope := buildCredentialScope(signingTime, region, service)
 	credentialStr := credentials.AccessKeyID + "/" + credentialScope
 
@@ -121,8 +121,8 @@ func (s *postSignAdapter) PresignPost(
 	return fields, nil
 }
 
-func getPostSignRequiredFields(t time.Time, credentialStr string, awsCredentials aws.Credentials) map[string]interface{} {
-	fields := map[string]interface{}{
+func getPostSignRequiredFields(t time.Time, credentialStr string, awsCredentials aws.Credentials) map[string]string {
+	fields := map[string]string{
 		algorithmHeader:  algorithm,
 		dateHeader:       t.UTC().Format("20060102T150405Z"),
 		credentialHeader: credentialStr,
@@ -143,7 +143,7 @@ type PresignPost interface {
 		bucket string, key string,
 		region string, service string, signingTime time.Time, conditions []interface{}, expirationTime time.Time,
 		optFns ...func(*v4.SignerOptions),
-	) (fields map[string]interface{}, err error)
+	) (fields map[string]string, err error)
 }
 
 // PresignPostOptions represent the options to be passed to a PresignPost sign request
--- a/service/s3/presign_post_test.go
+++ b/service/s3/presign_post_test.go
@@ -107,7 +107,7 @@ func TestPresignPutObject(t *testing.T) {
 			if !ok {
 				t.Error("expected non-empty policy on postObject")
 			}
-			decoded, err := base64.StdEncoding.DecodeString(policy.(string))
+			decoded, err := base64.StdEncoding.DecodeString(policy)
 			if err != nil {
 				t.Error("expected base64 encoded policy, got error", err, "policy", policy)
 			}
@@ -297,11 +297,7 @@ func TestBuildPresignPostRequest(t *testing.T) {
 				t.Errorf("PresignPostHTTP did not contain expected \"key\" %s. Has %s", aBucketKey, actualKey)
 			}
 			policy := fields["policy"]
-			strPolicy, ok := policy.(string)
-			if !ok {
-				t.Errorf("Policy field is not a string as expected, is %v", policy)
-			}
-			decoded, err := base64.StdEncoding.DecodeString(strPolicy)
+			decoded, err := base64.StdEncoding.DecodeString(policy)
 			if err != nil {
 				t.Errorf("Decoding policy document %s failed with error %v", policy, err)
 			}
@@ -310,7 +306,7 @@ func TestBuildPresignPostRequest(t *testing.T) {
 			if err != nil {
 				t.Errorf("Policy document %s failed to parse to JSON with error %v", policy, err)
 			}
-			_, ok = doc["conditions"]
+			_, ok := doc["conditions"]
 			if !ok {
 				t.Errorf("Conditions field not present in policy document %s", policy)
 			}
--- a/service/internal/integrationtest/s3/presign_post_test.go
+++ b/service/internal/integrationtest/s3/presign_post_test.go
@@ -135,14 +135,14 @@ func TestInteg_PresigPost(t *testing.T) {
 	}
 }
 
-func sendMultipartRequest(url string, fields map[string]interface{}, filePath string) (*http.Response, error) {
+func sendMultipartRequest(url string, fields map[string]string, filePath string) (*http.Response, error) {
 	// Create a buffer to hold the multipart data
 	var requestBody bytes.Buffer
 	writer := multipart.NewWriter(&requestBody)
 
 	// Add form fields
 	for key, val := range fields {
-		err := writer.WriteField(key, val.(string))
+		err := writer.WriteField(key, val)
 		if err != nil {
 			return nil, err
 		}


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/s3/PresignPostCustomization.java
```diff
@@ -0,0 +1,102 @@
...
+ *  CodeGen class for a presigner that uses POST
+ *  This is a special kind of presigner that instead of generating and signing an HTTP request,
+ *  generates a URL and field values to be used on an HTML form.
+ * <p>
+ *  This method of sending request is only implemented service side for putting objects into an S3 bucket.
+ *  However, it could be expanded to other operations if other services start supporting it
+ *
+ *  @see <a href="https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html> the public docs</a>
+ */
+public class PresignPostCustomization implements GoIntegration {
```
lucix-aws, line 39: Is there a reason we need to codegen this? It could just live in its own file in s3 unless I'm missing something.

lucix-aws: I know other presigned APIs get generated into their operation's files but I'm not concerned about matching that convention for handwritten stuff like this. Always in favor of explicitly written Go code where we can instead of generated.

service/s3/presign_post.go
```diff
@@ -0,0 +1,408 @@
...
+	return hash.Sum(nil)
+}
+
+func buildSignature(strToSign, secret, service, region string, t time.Time) string {
+	key := deriveKey(secret, service, region, t)
+	return hex.EncodeToString(hmacsha256(key, []byte(strToSign)))
+}
+
+func deriveKey(secret, service, region string, t time.Time) []byte {
+	hmacDate := hmacsha256([]byte("AWS4"+secret), []byte(t.UTC().Format("20060102")))
```
lucix-aws, line 395: can we `const` the date format?

service/s3/presign_post.go
```diff
@@ -0,0 +1,408 @@
...
+		return err
+	}
+	err = presignedurlcust.AddAsIsPresigningMiddleware(stack)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func removeStepIfPresent(stack *middleware.Stack, id string) {
```
lucix-aws, line 291: This isn't really necessary, you can just call `Remove()` and ignore the returns (won't panic or anything)

service/s3/presign_post_test.go
```diff
@@ -0,0 +1,497 @@
...
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/awstesting/unit"
+	"github.com/aws/aws-sdk-go-v2/internal/sdk"
+)
+
+func TestPresignPutObject(t *testing.T) {
+	fixedTime := time.Date(2022, time.February, 1, 0, 0, 0, 0, time.UTC)
```
lucix-aws, line 18: Small tip, for testing like this if you find you're doing this temp-patch thing more often, this you can do something like

```go
func mockTime(t time.Time) func() {
    sdk.NowTime = func() time.Time { return t }
    return func() { sdk.NowTime = time.Now }
}

// in test
defer mockTime(time.Whatever)()
```

service/s3/presign_post.go
```diff
@@ -0,0 +1,408 @@
...
+// the http request using the SigV4 presign authentication scheme.
+//
+// Since the signed request is not a valid HTTP request
+func (s *presignPostRequestMiddleware) HandleFinalize(
+	ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler,
+) (
+	out middleware.FinalizeOutput, metadata middleware.Metadata, err error,
+) {
+	if s.presigner == nil {
+		return out, metadata, fmt.Errorf("no presigner provided. Double-check your configuration")
```
lucix-aws, line 180: nit, can this actually happen? Since we're defaulting to one? I guess it could if they explicitly `nil` it out but not sure  if we care to guard for things like that. I think there are any number of things you could nil out on client config right now that would cause panics.

lucix-aws (commented):
General comment, this seems worthy of an integration test if we can swing it (so like actually generate a presigned URL, form POST the object, and then download to verify).

We have a bunch of existing presign tests in service/internal/integrationtest/s3/presign_test.go

service/s3/presign_post.go
```diff
@@ -0,0 +1,408 @@
...
+// 4. The signature is made based on the policy document, not the whole request
+// More information can be found at https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html
+//
+// [presigned request] https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html
+func (s *postSignAdapter) PresignPost(
+	credentials aws.Credentials,
+	bucket string, key string,
+	region string, service string, signingTime time.Time, conditions []interface{}, expirationTime time.Time, optFns ...func(*v4.SignerOptions),
+) (fields map[string]interface{}, err error) {
+	if conditions == nil {
```
lucix-aws, line 67: This isn't necessary, `nil` slices and 0-length slices are programmatically treated basically the same everywhere (including in an `append(x, y...)` call where you're using it)

service/s3/presign_post.go
```diff
@@ -0,0 +1,408 @@
...
+	defaultExpiresIn = 15 * time.Minute
+)
+
+// PresignedPostRequest represents a presigned request to be sent using HTTP verb POST and FormData
+type PresignedPostRequest struct {
+	// Represents the Base URL to make a request to
+	URL string
+	// Values is a key-value map of values to be sent as FormData
+	// these values are not encoded
+	Values map[string]interface{}
```
lucix-aws, line 46: Can this just be map[string]string? Since the final result is just strings for the POST form, right?

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/s3/PresignPostCustomization.java
Madrigal: Moved to written Go.

Originally I wasn't sure how much of this could be extended to other presigners, like whether we could this kind of signer for other operations. After some internal discussions, it doesn't seem like there are plans to extend this to other operations

service/s3/presign_post.go
Madrigal: Done

service/s3/presign_post.go
Madrigal: Done

service/s3/presign_post_test.go
Madrigal: Nice, changed

service/s3/presign_post.go
Madrigal: Removed. 

This can't really happen "naturally", but I added it since during development I had a specific scenario where this was being set to nil, and it took me a while to figure it out.

service/s3/presign_post.go
Madrigal: Removed

service/s3/presign_post.go
Madrigal: You are right. Was getting tripped with the other JSON values having to be map[string]interface{}, but this can actually be just as string to string. Good catch!

lucix-aws (approved):

PR #2782, opened 2024-09-09
chore: Modified bug issue template to add checkbox to report potentia…

bhavya2109sharma:
- Modified bug issue template to add checkbox to report potential regression.
- Added GitHub action to add/remove label `potential-regression` when issue is created/edited.

**NOTE:**
- The GitHub action code was ported from CDK repo.
- The label `potential-regression` would need to be created manually (we could use color `#FF6700` and description `Marking this issue as a potential regression to be checked by team member` to make it consistent across SDK repos)

Label `potential-regression` would make issue standout in the list and help engineers to handle high severity issues effectively.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -22,6 +22,17 @@ body:
       description: What is the problem? A clear and concise description of the bug.
     validations:
       required: true
+
+  - type: checkboxes
+    id: regression
+    attributes:
+      label: Regression Issue
+      description: What is a regression? If it worked in a previous version but doesn't in the latest version, it's considered a
+        regression. In this case, please provide specific version number in the report.
+      options:
+        - label: Select this option if this issue appears to be a regression.
+          required: false
+
   - type: textarea
     id: expected
     attributes:
--- a/.github/workflows/issue-regression-labeler.yml
+++ b/.github/workflows/issue-regression-labeler.yml
@@ -1,0 +1,32 @@
+# Apply potential regression label on issues
+name: issue-regression-label
+on:
+  issues:
+    types: [opened, edited]
+jobs:
+  add-regression-label:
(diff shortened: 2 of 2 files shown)


lucix-aws:
I don't see any merit in having this workflow.
* This should be determined at triage, I don't expect or necessarily even trust an issue reporter to try and make the distinction of whether or not the behavior they're describing constitutes a regression.
* The term "regression" is ill-defined here especially when you consider that we have a v1 SDK. I know we have a separate issue template for that, but those guidelines aren't exactly infallible.

At best it isn't really giving us any value, at worst it's creating an impetus to triage/investigate something on the potentially false grounds of it being a regression.

ashishdhingra:
> I don't see any merit in having this workflow.
> 
> * This should be determined at triage, I don't expect or necessarily even trust an issue reporter to try and make the distinction of whether or not the behavior they're describing constitutes a regression.
> * The term "regression" is ill-defined here especially when you consider that we have a v1 SDK. I know we have a separate issue template for that, but those guidelines aren't exactly infallible.
> 
> At best it isn't really giving us any value, at worst it's creating an impetus to triage/investigate something on the potentially false grounds of it being a regression.

@lucix-aws Thanks for sharing the concern. The proposed change was first implemented in CDK in response to a regression that was lingering around for almost a month and was escalated to management. It was due to misinterpretation of the issue by the team which resulted in potential security breach in customer's infrastructure. After the implementation, it has helped team to quickly identify at least 50% of the regression issues flagged by this change.

Hence, based on discussion within the team, we felt it is valuable to implement this change across SDK(s) as it would help identify such issues more effectively. There is an assumption assumption that some of these issues might not be regression, but it's always wise to treat such issue as `p0`, which could be downgraded later if that is not the case. Please free to reach out to @kellertk if there is any concern.

Thanks,
Ashish

lucix-aws (approved):

PR #2784, opened 2024-09-11
add blurb on disabling endpoint prefix

lucix-aws:
Adds instructions on how to disable endpoint prefixes being applied, inspired by an internal conversation on the topic.


docs/404.html
```diff
@@ -23,7 +23,7 @@
 <meta property="og:title" content="404 Page not found" />
 <meta property="og:description" content="The AWS SDK for Go V2 provides APIs and utilities that developers can use to build Go applications that use AWS services." />
 <meta property="og:type" content="website" />
-<meta property="og:url" content="http://localhost:1313/aws-sdk-go-v2/404.html" />
```
Madrigal, line 26: lol

docs/404.html
```diff
@@ -99,7 +99,7 @@
   aria-label="Search this site…"
   autocomplete="off"
   
-  data-offline-search-index-json-src="/aws-sdk-go-v2/offline-search-index.30b245754e63db1f94dcbaa3ba0f9915.json"
+  data-offline-search-index-json-src="/aws-sdk-go-v2/offline-search-index.1ed5201c9429d6d49ab1e6d6c5e5409e.json"
```
Madrigal, line 102: Were you able to confirm if this offline search works?

Madrigal (approved):

docs/404.html
lucix-aws: I did actually check this and it worked for me both in the preview _and_ on the current render of the public site. We'll see what happens when this merges and gets propagated. The search index appears to get a new filename every time, it's possible that it just went out of sync in a previous pass on re-generating.

lucix-aws: cc @stefansundin

PR #2787, opened 2024-09-13
Docs: non-default profiles must start with `profile`

skyzyx:
AWS config profiles must start with the word `profile` in the INI file.


AlbertVilaCalvo:
I believe having "profile" is only needed for the `config` file, not in the `credentials` file, but you are adding it at the `credentials` file.

lucix-aws:
☝️

skyzyx:
I didn't realize they were different. I've not used the `~/.aws/credentials` file in several years. I use `~/.aws/config` exclusively.

Issue #1744, opened 2022-06-29
Support SDK metrics for go v2 AWS SDK


DanielBauman88:
### Describe the feature

The java feature is documented [here](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics.html).
The functionality is described in [this](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics.html#what-information-is-collected) section.

The request is to support the same for the go sdk so that it is trivial to get metrics for latencies/errors/retries to aws dependencies made in a customer application.

### Use Case

I want to have operational metrics for latency,error,num-calls for all my dependencies so that I can monitor the performance of my service and dig into problems and investigate the impact of outages.

### Proposed Solution

To implement this functionality with a simple option on SDK creation in the go sdk v2.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

This is applicable to all SDKs

### Go version used

This should be applicable to all go versions

jeichenhofer:
I'm also looking to integrate some metrics with the aws-sdk-go-v2 libraries, but I don't want to re-invent the wheel. Hopefully this will be an officially supported feature. Specifically, I want to record a tuple of service name, operation name, aws region, latency, retry count, and response code on every request sent to AWS. I can envision doing this with the "middleware" API, but these are the only docs I can find, and they don't do a great job explaining what information about the request is available (https://aws.github.io/aws-sdk-go-v2/docs/middleware/) (e.g., would we need to record the "sent time" in Initialize step, then check it in the deserialize step, or is latency already a populated metadata value).

While we wait for a response from the development team about incorporating this as an SDK feature, is there any guidance on implementing something ourselves?

jeichenhofer:
Here's what I could come up with by stepping through the middleware stack code. It seems to work as intended, but I'd be curious to hear from people more familiar with the API.

Of course, this would need to be incorporated with some existing metrics system, replacing the `ReportMetrics` function with something that feeds into monitoring systems or log files. If there's a chance that the function might return an error, then I'd have to think a bit more about how to handle that.

[...]

PR #2789, opened 2024-09-13
extend metrics with retry-related instruments

lucix-aws:
Metrics component of #1744.

Extends upstream metrics additions from https://github.com/aws/smithy-go/pull/536 with the following instruments (and their dimensions):
* `client.call.errors` - number of errors encountered (service, operation, error type)
* `client.call.attempts` - number of attempts made (service, operation)
* `client.call.attempt_duration` - time span for an entire attempt (service, operation)

This is not the final merge before release, as some minor tweaks need to be made to some of the component names to align with our internal guidelines.

### DIFF
--- a/service/s3/api_client.go
+++ b/service/s3/api_client.go
@@ -43,7 +43,7 @@ const ServiceAPIVersion = "2006-03-01"
 type operationMetrics struct {
 	Duration                metrics.Float64Histogram
 	SerializeDuration       metrics.Float64Histogram
-	GetIdentityDuration     metrics.Float64Histogram
+	ResolveIdentityDuration metrics.Float64Histogram
 	ResolveEndpointDuration metrics.Float64Histogram
 	SignRequestDuration     metrics.Float64Histogram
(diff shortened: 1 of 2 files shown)


service/s3/api_client.go
```diff
@@ -39,6 +40,129 @@ import (
...
+) (T, error) {
+	instr := getOperationMetrics(ctx).histogramFor(metric)
+	opts = append([]metrics.RecordMetricOption{withOperationMetadata(ctx)}, opts...)
+
+	start := time.Now()
+	v, err := fn()
+	end := time.Now()
+
+	elapsed := end.Sub(start)
+	instr.Record(ctx, float64(elapsed)/1e9, opts...)
```
Madrigal, line 83: not a blocker, but I still dislike this. Are we instructing customers when using their own metrics to conform to this format as well when recording duration? If so, what happens if they don't?

Madrigal (approved):

wty-Bryant (approved):

PR #2790, opened 2024-09-13
Read `AWS_CONTAINER_CREDENTIALS_FULL_URI` env variable if set when reading a profile with `credential_source`.

Madrigal:
## Context

EKS (Elastic Kubernetes Service) is a managed service for running Kubernetes clusters. Pods are the smallest deployable units in Kubernetes, consisting of one or more containers, and are managed within a cluster.[EKS Pod identity](https://aws.amazon.com/blogs/containers/amazon-eks-pod-identity-a-new-way-for-applications-on-eks-to-obtain-iam-credentials/) is a new feature of AWS, which allows individual pods to be mapped to a specific IAM roles 

For the SDKs, this means that EKS Pod Identity sets `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`, which are used to retrieve credentials.

This IAM role must exist within your own AWS account.

## use case
Customers want to be able to easily be able to assume the role of *another* AWS account, since it's a common pattern to have one AWS account managing the EKS cluster and others that host things like DynamoDB tables or S3 buckets. Right now, customers can setup IAM role chaining via code, but they want to be able to do it via profiles, like

```
[profile other-account]
role_arn = arn:aws:iam::1234567890:role/other-account-role
credential_source = EcsContainer
```

## Problem
Go SDK currently didn't read from `AWS_CONTAINER_CREDENTIALS_FULL_URI` when resolving credentials from a shared AWS config file, it only read `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`. Since this is not what EKS Pod identity sets, the request failed.

Additionally, while we have a SEP that defines the priority between these variables, this wasn't implemented.

### DIFF
--- a/config/resolve_credentials_test.go
+++ b/config/resolve_credentials_test.go
@@ -692,7 +692,7 @@ func TestResolveCredentialsEcsContainer(t *testing.T) {
 			expectedSecretKey: "SECRET-Full-Path",
 			envVar: map[string]string{
 				"AWS_CONTAINER_CREDENTIALS_FULL_URI": "placeholder-replaced-at-runtime",
-				"AWS_PROFILE":                            "ecscontainer",
+				"AWS_PROFILE":                        "ecscontainer",
 			},
 			configFile: filepath.Join("testdata", "config_source_shared"),
 		},


config/resolve_credentials.go
```diff
@@ -355,10 +355,13 @@ func resolveCredsFromSource(ctx context.Context, cfg *aws.Config, envConfig *Env
...
 	case credSourceECSContainer:
-		if len(envConfig.ContainerCredentialsRelativePath) == 0 {
-			return fmt.Errorf("EcsContainer was specified as the credential_source, but 'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI' was not set")
+		if len(envConfig.ContainerCredentialsRelativePath) != 0 {
+			return resolveHTTPCredProvider(ctx, cfg, ecsContainerURI(envConfig.ContainerCredentialsRelativePath), envConfig.ContainerAuthorizationToken, configs)
+		}
+		if len(envConfig.ContainerCredentialsEndpoint) != 0 {
+			return resolveLocalHTTPCredProvider(ctx, cfg, envConfig.ContainerCredentialsEndpoint, envConfig.ContainerAuthorizationToken, configs)
 		}
-		return resolveHTTPCredProvider(ctx, cfg, ecsContainerURI(envConfig.ContainerCredentialsRelativePath), envConfig.ContainerAuthorizationToken, configs)
```
lucix-aws, line 361: Why remove the fallback?

lucix-aws (approved):

wty-Bryant (approved):

config/resolve_credentials.go
Madrigal: This wasn't really a fallback but rather a function call after checking the args are valid

```
if (!isValid(a)) {
    return error
}
func(a)
```

This is however building an ECS path, and you are correct that we could pass an empty path to `ecsContainerURI` to resolve to the base path directly. However, the SEP doesn't like this

> If none of the above sources (the two env variables) yield a value, the provider MUST fail to resolve indicating as such.

PR #2795, opened 2024-09-17
BREAKFIX: Only generate AccountIDEndpointMode config for services that use it

lucix-aws:
Only generate AccountIDEndpointMode config for services that use it. This is a compiler break, but removes no actual functionality, as no services currently use the account ID in endpoint resolution.

Relevant commit here is https://github.com/aws/aws-sdk-go-v2/pull/2795/commits/fd86578a3ee950c0b4c649319c2749a64ca3c381


wty-Bryant (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIDEndpointRouting.java
```diff
@@ -67,4 +74,19 @@ func checkAccountID(identity $auth:T, mode $accountIDEndpointMode:T) error {
...
         ));
     }
+
+    public static boolean hasAccountIdEndpoints(Model model, ServiceShape service) {
+        if (!service.hasTrait(EndpointRuleSetTrait.class)) {
+            return false;
+        }
+
+        var rules = service.expectTrait(EndpointRuleSetTrait.class).getEndpointRuleSet();
+        for (var param : rules.getParameters()) {
```
Madrigal, line 84: this looks perfect for a stream, but Parameters is not a collection so while doable it's very awkward to handle it as such, so I like this version better

lucix-aws: There's no way to actually get a stream of `Parameters` that I could see. You can either iterate them, get() by string, or there's a forEach.

Madrigal: Again it's super awkward to do via streams and I like your version better, but just for completion sake, you could wrap the `Parameters` spliterator using StreamSupport

```
return StreamSupport.stream(rules.getParameters().spliterator(), false)
                .map(x -> x.getBuiltIn().orElse(""))
                .anyMatch(x -> x.equals("AWS::Auth::AccountId"));
```

PR #2799, opened 2024-09-19
add jitter instead of scaling by jitter

nalbion:
It's possible for a series of random numbers to be all `0`, more feasible for them to be all close to zero.

Multiplying by `ri` would yield:

- 0 * 1 = 0
- 0 * 2 = 0
- 0 * 4 = 0
- 0 * 8 = 0

To avoid this, jitter should be _added_:

- 0 + 1 = 1
- 0 + 2 = 2
- 0 + 4 = 4
- 0 + 8 = 8


lucix-aws:
The calculation for delay here is standardized across SDKs in the "standard" retry strategy, it's not something we can change.

PR #2808, opened 2024-09-26
feat: flexible checksum updates

wty-Bryant:
Add client cfg to opt-in/out checksum calculation/validation, change default algorithm to calculate checksum


service/internal/checksum/middleware_compute_input_checksum.go
```diff
@@ -264,6 +270,21 @@ func (m *addInputChecksumTrailer) HandleFinalize(
...
+	var set bool
+	algorithm, set, err = getInputAlgorithm(ctx)
+	if err != nil {
+		return out, metadata, computeInputTrailingChecksumError{
+			Msg: "failed to get algorithm",
+			Err: err,
+		}
+	}
+
+	checkRequireChecksum(&m.RequireChecksum, m.RequestChecksumCalculation, set)
```
lucix-aws, line 283: I see _why_ you did it this way, and I'm not saying it wouldn't work, but I think we're overcomplicating the way we figure out whether or not we're actually doing the checksum. Basically we've kind of warped the logic here by overloading the meaning of "required" to mean either _literally_ required according to the model, or synethically required if checksum calculation is on. Combined with the addition of the uncomfortable ternary value for requiredness, the ergonomics of this are not where we need them to be.

Basically, all this middleware should be doing is computing the checksum if an algorithm was previously selected. We can determine that in the earlier middleware which actually sets the algorithm on the context - using modeled requiredness and user config to decide accordingly what the algorithm should be.

That's going to give us a much cleaner separation between "should we checksum" and "do checksum" - which as written is fairly tangled.

aws/checksum.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// ResponseChecksumValidationWhenRequired indicates response checksum should be validated if modeled
+	// and user enable that in vlidation mode cfg
+	ResponseChecksumValidationWhenRequired = "whenRequired"
+)
+
+// RequireChecksum indicates if a checksum needs calculated/validated for a request/response
+type RequireChecksum string
+
+const (
```
lucix-aws, line 30: A few comments though I think they'll be moot in the end:
* this should definitely be an internal type
* this is not an idiomatic way to represent a ternary boolean - ternary bool should almost _never_ be necessary in Go, but if it absolutely is, you'd just use `*bool`

aws/checksum.go
```diff
@@ -0,0 +1,39 @@
+package aws
+
+// RequestChecksumCalculation controls request checksum calculation workflow
+type RequestChecksumCalculation string
+
+const (
+	// RequestChecksumCalculationWhenSupported indicates request checksum should be calculated if modeled
+	RequestChecksumCalculationWhenSupported RequestChecksumCalculation = "whenSupported"
```
lucix-aws, line 8: Idiomatically these should be `type <x> int`, I know we've deviated from that somewhat in the past, but I'd like us to course-correct.

.changelog/9ebe24c4791541e0840da49eab6f9d97.json
```diff
@@ -0,0 +1,12 @@
+{
+    "id": "9ebe24c4-7915-41e0-840d-a49eab6f9d97",
+    "type": "feature",
+    "description": "add client cfg to opt-in/out checksum behavior and change its default algorithm",
```
lucix-aws, line 4: 1. This needs to be a complete and correct sentence.
2. Frankly I think we're going to have to give a lot more detail here. This is an implicit behavioral change. We can discuss externally.

aws/checksum.go
```diff
@@ -0,0 +1,39 @@
+package aws
+
+// RequestChecksumCalculation controls request checksum calculation workflow
+type RequestChecksumCalculation string
+
+const (
+	// RequestChecksumCalculationWhenSupported indicates request checksum should be calculated if modeled
```
lucix-aws, line 7: "if modeled" does not mean anything to the caller. We have to find a different way to express this.

.changelog/9ebe24c4791541e0840da49eab6f9d97.json
wty-Bryant: Make sense after seeing the accountid changelog pr

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -93,33 +93,50 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
...
+						r.ContentLength = 11
 						return r
 					}(),
 				},
-				expectContentLength: -1,
+				expectContentLength: 11,
 				expectHeader:        http.Header{},
+				expectPayload:       []byte("hello world"),
+			},
+			"https no algorithm set": {
```
Madrigal, line 127: Is there any different treatment between http and https? Unless I'm missing something the tests seem the same

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -254,94 +271,27 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
...
-				expectContentLength: 11,
-				expectPayload:       []byte("hello world"),
-				expectChecksumMetadata: map[string]string{
-					"MD5": "XrY7u+Ae7tCTyyK7j1rNww==",
-				},
-			},
 			"http seekable": {
 				initContext: func(ctx context.Context) context.Context {
-					return internalcontext.SetChecksumInputAlgorithm(ctx, string(AlgorithmCRC32))
+					return internalcontext.SetChecksumInputAlgorithm(ctx, string(AlgorithmCRC32C))
```
Madrigal, line 276: It's fine to use 32C, but curious to hear why you changed this on this revision. It seems like existing values could have worked just as well

service/internal/checksum/middleware_setup_context.go
```diff
@@ -47,9 +60,11 @@ type setupOutputContext struct {
 	// GetValidationMode is a function to get the checksum validation
 	// mode of the output payload from the input parameters.
 	//
-	// Given the input parameter value, the function must return the validation
-	// mode and true, or false if no mode is specified.
```
Madrigal, line 51: Aren't we still returning true or false on the interface? Why remove the comment? If the value is required to prevent changing the interface, we should definitely mention that on the comment

service/internal/checksum/middleware_setup_context.go
```diff
@@ -64,13 +79,15 @@ func (m *setupOutputContext) HandleInitialize(
...
 	if m.GetValidationMode != nil {
 		// check is input resource has a checksum algorithm
-		mode, ok := m.GetValidationMode(in.Parameters)
-		if ok && len(mode) != 0 {
-			ctx = setContextOutputValidationMode(ctx, mode)
-		}
+		mode, _ = m.GetValidationMode(in.Parameters)
+	}
+
+	if m.ResponseChecksumValidation == aws.ResponseChecksumValidationWhenSupported || mode == "ENABLED" {
```
Madrigal, line 89: Seems like `"ENABLED"` would benefit by being defined as a constant somewhere

service/internal/checksum/middleware_validate_output.go
```diff
@@ -55,7 +55,7 @@ func (m *validateOutputPayloadChecksum) ID() string {
 }
 
 // HandleDeserialize is a Deserialize middleware that wraps the HTTP response
-// body with an io.ReadCloser that will validate the its checksum.
+// body with an io.ReadCloser that will validate its checksum.
```
Madrigal, line 58: praise: kudos for fixing these!

service/internal/checksum/middleware_add.go
```diff
@@ -122,10 +124,15 @@ type OutputMiddlewareOptions struct {
 	// GetValidationMode is a function to get the checksum validation
 	// mode of the output payload from the input parameters.
 	//
-	// Given the input parameter value, the function must return the validation
-	// mode and true, or false if no mode is specified.
+	// Given the input parameter value, the function must return the validation mode
```
lucix-aws, line 127: Why did this change?

service/internal/checksum/middleware_add.go
```diff
@@ -122,10 +124,15 @@ type OutputMiddlewareOptions struct {
...
 	GetValidationMode func(interface{}) (string, bool)
 
+	// Whether operation model forces middleware to validate checksum
```
lucix-aws, line 130: [no-op comment] - I'm totally OK w/ us talking about the model in this package, since it's `internal/` (callers can't use it).

service/internal/checksum/middleware_setup_context.go
```diff
@@ -31,13 +40,17 @@ func (m *setupInputContext) HandleInitialize(
 ) (
 	out middleware.InitializeOutput, metadata middleware.Metadata, err error,
 ) {
-	// Check if validation algorithm is specified.
+	var algorithm string
+	var ok bool
 	if m.GetAlgorithm != nil {
```
lucix-aws, line 45: I know you're working around what was already there, but commenting since I noticed - is `m.GetAlgorithm == nil` actually possible? It seems like GetAlgorithm shouldn't ever be nil, if it were I think we'd have done something wrong in codegen.

I bring it up because the pre-declaration of <algorithm, ok> is a little awkward and wouldn't be necessary if this nil check wasn't there.

service/internal/checksum/middleware_setup_context.go
```diff
@@ -31,13 +40,17 @@ func (m *setupInputContext) HandleInitialize(
...
-		// check is input resource has a checksum algorithm
-		algorithm, ok := m.GetAlgorithm(in.Parameters)
-		if ok && len(algorithm) != 0 {
-			ctx = internalcontext.SetChecksumInputAlgorithm(ctx, algorithm)
+		algorithm, ok = m.GetAlgorithm(in.Parameters)
+	}
+
+	if m.RequireChecksum || m.RequestChecksumCalculation == aws.RequestChecksumCalculationWhenSupported || ok {
```
lucix-aws, line 49: This works but the double-if is a little uncomfortable. Suggested workflow to improve the ergonomics:

```go
if alg, ok := m.GetAlgorithm(in.Parameters); ok {
    ctx = ... // user set an algorithm, nothing else matters
    return next.HandleInitialize(ctx, in)
}

if m.RequireChecksum || m.RequestChecksumCalculation == aws.RequestChecksumCalculationWhenSupported {
    ctx = ... // default to crc32
}
return next.HandleInitialize(ctx, in)

File: service/internal/checksum/middleware_compute_input_checksum.go [hunk 15]
```diff
@@ -130,29 +129,15 @@ func (m *computeInputPayloadChecksum) HandleFinalize(
...
-
 	// If the checksum header is already set nothing to do.
-	checksumHeader := AlgorithmHTTPHeader(algorithm)
-	if checksum = req.Header.Get(checksumHeader); checksum != "" {
-		return next.HandleFinalize(ctx, in)
+	for _, supportedAlgorithm := range supportedAlgorithms {
+		header := AlgorithmHTTPHeader(supportedAlgorithm)
+		if cs := req.Header.Get(header); cs != "" {
+			// need to add header value to metadata
+			algorithm = supportedAlgorithm
```
lucix-aws, line 137: ~~Are these two values doing anything? You return right after setting them.~~ Nevermind, I see the defer() above doing stuff with them.

service/internal/checksum/middleware_add.go
lucix-aws, line 1: This code is existing but I think there's breakage potential with this change -- on line 182, there's a check that returns an error if the input request body isn't seekable. Since we're now _defaulting_ to checksums, I don't think we can do this anymore -- if a caller is using a non-seekable body right now, they would start getting operation errors when this rolls out. We probably just have to switch to no-opping in that case.

service/internal/checksum/middleware_compute_input_checksum_test.go
wty-Bryant: they are tested to ensure checksum is calculated and set in header/trailer correctly with different schema cfg

service/internal/checksum/middleware_compute_input_checksum_test.go
wty-Bryant: just want to test different algorithm

service/internal/checksum/middleware_setup_context.go
wty-Bryant: checksum generator would generate getter for each op supporting checksum, I think we can remove that nil check

aws/checksum.go
```diff
@@ -0,0 +1,26 @@
...
+
+// RequestChecksumCalculation controls request checksum calculation workflow
+type RequestChecksumCalculation int
+
+const (
+	// RequestChecksumCalculationWhenSupported indicates request checksum should be calculated if
+	// client operation model has request checksum trait
+	RequestChecksumCalculationWhenSupported RequestChecksumCalculation = 1
+
+	// RequestChecksumCalculationWhenRequired indicates request checksum should be calculated
```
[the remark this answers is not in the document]
wty-Bryant: When set to WhenRequired, a checksum will only be calculated for operations modeled with checksum trait where requestChecksumRequired is true OR the algorithm is set by user in request

aws/checksum.go
```diff
@@ -0,0 +1,26 @@
+package aws
+
+// RequestChecksumCalculation controls request checksum calculation workflow
+type RequestChecksumCalculation int
+
+const (
+	// RequestChecksumCalculationWhenSupported indicates request checksum should be calculated if
+	// client operation model has request checksum trait
+	RequestChecksumCalculationWhenSupported RequestChecksumCalculation = 1
```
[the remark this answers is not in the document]
lucix-aws: Not sure I follow? The default/unset value is `WhenSupported` - which is the default setting we want.

We have an adage in Go - "make the zero value useful" which I believe applies here.

service/internal/checksum/middleware_add.go
wty-Bryant: make sense, I will check that in input middleware

service/internal/checksum/middleware_add.go
wty-Bryant: restore that

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -487,24 +437,9 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
...
-						return r
-					}(),
-				},
-				expectErr:      "unseekable stream is not supported",
-				expectBuildErr: true,
-			},
 			"http unseekable stream": {
 				initContext: func(ctx context.Context) context.Context {
-					return internalcontext.SetChecksumInputAlgorithm(ctx, string(AlgorithmCRC32))
+					return internalcontext.SetChecksumInputAlgorithm(ctx, string(AlgorithmSHA1))
```
Madrigal, line 442: I approve of the changes to typos you make on the PR. I don't like the random changes to algorithms which have no functional impact. They increase the diff without adding value to the PR, since the functionality (get an algorithm and use it) has already been tested and is not being changed.

service/internal/checksum/middleware_compute_input_checksum.go
```diff
@@ -130,29 +129,15 @@ func (m *computeInputPayloadChecksum) HandleFinalize(
...
-		return next.HandleFinalize(ctx, in)
-	}
-
 	// If the checksum header is already set nothing to do.
-	checksumHeader := AlgorithmHTTPHeader(algorithm)
-	if checksum = req.Header.Get(checksumHeader); checksum != "" {
-		return next.HandleFinalize(ctx, in)
+	for _, supportedAlgorithm := range supportedAlgorithms {
+		header := AlgorithmHTTPHeader(supportedAlgorithm)
+		if cs := req.Header.Get(header); cs != "" {
```
Madrigal, line 135: a corner case, but if multiple headers were to be set, do we have ranking on `supportedAlgorithms`?

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -514,8 +449,9 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
 						return r
 					}(),
 				},
-				expectErr:      "unseekable stream is not supported",
-				expectBuildErr: true,
+				expectContentLength: -1,
```
Madrigal, line 452: Why does unseekable stream works now when it didn't in the past?

wty-Bryant: Old code would not calculate checksum for unseekable stream unless required by codegen. We should not return error when user send unseekable stream now

service/internal/checksum/middleware_compute_input_checksum.go
wty-Bryant: No, the purpose of this block is stop the calculation if any valid checksum header is already set

config/load_options.go
```diff
@@ -340,6 +355,26 @@ func WithAccountIDEndpointMode(m aws.AccountIDEndpointMode) LoadOptionsFunc {
...
+		}
+		return nil
+	}
+}
+
+// WithResponseChecksumValidation is a helper function to construct functional options
+// that sets ResponseChecksumValidation on config's LoadOptions
+func WithResponseChecksumValidation(v aws.ResponseChecksumValidation) LoadOptionsFunc {
+	return func(o *LoadOptions) error {
+		o.ResponseChecksumValidation = v
```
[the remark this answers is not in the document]
wty-Bryant: zero check is unnecessary here since `unset` is a valid value

.changelog/9ebe24c4791541e0840da49eab6f9d97.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "9ebe24c4-7915-41e0-840d-a49eab6f9d97",
+    "type": "feature",
+    "description": "This feature adds new config fields to checksum-supported operations so user can switch on/off checksum workflow with those fields. The field RequestChecksumCalculation controls the request's checksum calculation and the field ResponseChecksumValidation controls the response's checksum validation. This feature also replaces the default MD5 checksum algorithm of checksum calculation with CRC32.",
+    "modules": [
+        ".",
+        "config",
```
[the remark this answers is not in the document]
lucix-aws: It needs to annotate every module that changed (otherwise the module will release with "no release notes available" which has confused people in the past).

.changelog/9ebe24c4791541e0840da49eab6f9d97.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "9ebe24c4-7915-41e0-840d-a49eab6f9d97",
+    "type": "feature",
+    "description": "This feature adds new config fields to checksum-supported operations so user can switch on/off checksum workflow with those fields. The field RequestChecksumCalculation controls the request's checksum calculation and the field ResponseChecksumValidation controls the response's checksum validation. This feature also replaces the default MD5 checksum algorithm of checksum calculation with CRC32.",
```
lucix-aws, line 4:
```suggestion
    "description": "Enable HTTP checksums in supported services by default. New config fields, RequestChecksumCalculation and ResponseChecksumValidation, allow the caller to opt-out of this new default behavior. This feature also replaces the default MD5 checksum with CRC32.",
```

aws/config.go
```diff
@@ -165,6 +165,22 @@ type Config struct {
...
+
+	// RequestChecksumCalculation allows user to opt-in/out request checksum calculation.
+	// When set to RequestChecksumCalculationWhenSupported by default, checksum is always calculated
+	// if operation supports that, when set to RequestChecksumCalculationWhenRequired, checksum
+	// is calculated only if user sets an algorithm in request
+	// This variable is sourced from environment variable AWS_REQUEST_CHECKSUM_CALCULATION or
+	// the shared config profile attribute request_checksum_calculation
+	RequestChecksumCalculation RequestChecksumCalculation
+
+	// ResponseChecksumValidation allows user to opt-in/out response checksum validation.
```
lucix-aws, line 177: I'd like to see this change to something like Luis suggested above.

aws/checksum.go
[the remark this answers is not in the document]
lucix-aws: Yes, that statement is true, but it's not what you have here. All you're saying is "checksum will be calculated if user elects to set an algorithm".

Like you said, it should be "checksum will be calculated **if required by the operation or** if user elects to set an algorithm

service/internal/checksum/middleware_compute_input_checksum.go
wty-Bryant: TBH I also wonder why they don't stop on first header detected if `supportedAlgorithm` is in preference order, I can change the order and make it stop as we want

wty-Bryant: latest SEP states If the user set any checksum header field, no calculation is needed, so the order here is not so important

aws/checksum.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	// ResponseChecksumValidationUnset is the unset value for ResponseChecksumValidation
+	ResponseChecksumValidationUnset ResponseChecksumValidation = iota
+
+	// ResponseChecksumValidationWhenSupported indicates response checksum will be validated
+	// if the operation supports output checksums
+	ResponseChecksumValidationWhenSupported
+
+	// ResponseChecksumValidationWhenRequired indicates response checksum will be validated
```
lucix-aws, line 30: Kind of like the WhenRequired variant for requests... "WhenRequired" for responses just means we're only doing it when required. Whether the user says to do it or not is immaterial

aws/checksum.go
```diff
@@ -0,0 +1,33 @@
...
+	// if user sets the validation mode to ENABLED in request
```
lucix-aws, line 31:
```suggestion
	// ResponseChecksumValidationWhenRequired indicates response checksum will only
	// be validated if the operation requires output checksum validation
```

aws/checksum.go
lucix-aws: I've patched it

lucix-aws (approved):

Madrigal (approved):
Had discussed and approved offline, but might as well give an explicit approval

service/internal/checksum/middleware_compute_input_checksum.go
```diff
@@ -130,29 +129,15 @@ func (m *computeInputPayloadChecksum) HandleFinalize(
...
-			algorithm = Algorithm("MD5")
-		}
-		return next.HandleFinalize(ctx, in)
-	}
-
 	// If the checksum header is already set nothing to do.
-	checksumHeader := AlgorithmHTTPHeader(algorithm)
-	if checksum = req.Header.Get(checksumHeader); checksum != "" {
-		return next.HandleFinalize(ctx, in)
+	for _, supportedAlgorithm := range supportedAlgorithms {
```
lucix-aws, line 133: This needs to check generically for _any_ type of checksum header being set, including ones we potentially don't know about. So basically the check should be "if there's any header on the request that starts with x-amz-checksum- (case insensitively)"

feature/s3/manager/integ_upload_test.go
```diff
@@ -161,8 +161,9 @@ func TestInteg_UploadPresetChecksum(t *testing.T) {
 	}{
 		"auto single part": {
 			"no checksum": {
-				payload:    bytes.NewReader(singlePartBytes),
-				expectETag: singlePartETag,
+				payload:             bytes.NewReader(singlePartBytes),
+				expectChecksumCRC32: singlePartCRC32,
```
Madrigal, line 165: 1. The tests is called "no checksum" but it has an expected checksum. We should change the name of the test
1. Does it still makes sense to have a `no checksum` test

service/internal/checksum/algorithms.go
```diff
@@ -30,6 +30,9 @@ const (
 
 	// AlgorithmSHA256 represents SHA256 hash algorithm
 	AlgorithmSHA256 Algorithm = "SHA256"
+
+	// AlgorithmCRC64NVME represents CRC64NVME hash algorithm
+	AlgorithmCRC64NVME Algorithm = "CRC64NVME"
```
Madrigal, line 35: Interesting, so this is used only when someone computes their own CRC64NVME checksum and pass their own implementation?

We should add a test with some pre-computed values to make sure we don't panic trying to use an algorithm that we don't have an implementation for

service/internal/integrationtest/s3/checksum_test.go
```diff
@@ -394,7 +422,7 @@ func TestInteg_ObjectChecksums(t *testing.T) {
 					}
 					// assert computed input checksums metadata
 					computedChecksums, ok := s3.GetComputedInputChecksumsMetadata(putResult.ResultMetadata)
-					if e, a := ok, (c.expectComputedChecksums != nil); e != a {
```
Madrigal, line 397: nice, this aligns with the actual `expected, actual` convention

config/load_options.go
```diff
@@ -299,6 +303,19 @@ func (o LoadOptions) getResponseChecksumValidation(ctx context.Context) (aws.Res
...
+func (o LoadOptions) getBaseEndpoint(context.Context) (string, bool, error) {
+	return o.BaseEndpoint, o.BaseEndpoint != "", nil
+}
+
+// GetServiceBaseEndpoint satisfies (internal/configsources).ServiceBaseEndpointProvider.
+//
+// The sdkID value is unused because LoadOptions only supports setting a GLOBAL
+// endpoint override. In-code, per-service endpoint overrides are performed via
+// functional options in service client space.
+func (o LoadOptions) GetServiceBaseEndpoint(context.Context, string) (string, bool, error) {
```
Madrigal, line 315: Can you elaborate on why we're adding this?

feature/s3/manager/integ_upload_test.go
wty-Bryant: I prefer to change its name to "no checksum algorithm passed"

service/internal/checksum/algorithms.go
wty-Bryant: s3 team already manual test that, but I can add that

Madrigal: yeah, I don't think S3 will panic, I want to ensure we don't have any logic on the SDK to do something like "if CRC64 try to compute checksum" and then we panic

Madrigal (approved):

aws/middleware/user_agent.go
```diff
@@ -76,19 +76,28 @@ type UserAgentFeature string
...
+	UserAgentFeatureProtocolRPCV2CBOR                              = "M"
+	UserAgentFeatureRequestChecksumWhenSupported                   = "Z"
+	UserAgentFeatureRequestChecksumWhenRequired                    = "a"
+	UserAgentFeatureResponseChecksumWhenSupported                  = "b"
+	UserAgentFeatureResponseChecksumWhenRequired                   = "c"
+	UserAgentFeatureRequestChecksumCRC32                           = "U"
+	UserAgentFeatureRequestChecksumCRC32C                          = "V"
+	UserAgentFeatureRequestChecksumCRC64                           = "W"
+	UserAgentFeatureRequestChecksumSHA1                            = "X"
+	UserAgentFeatureRequestChecksumSHA256                          = "Y"
```
Madrigal, line 100: Manually checked this against the spec, they match. However, should we look into ordering them by their alphabetic value? e.g. moving `UserAgentFeatureRequestChecksumWhenRequired` either to the top or to the bottom?

service/internal/checksum/middleware_checksum_metrics_tracking.go
```diff
@@ -0,0 +1,90 @@
...
+	switch m.RequestChecksumCalculation {
+	case aws.RequestChecksumCalculationWhenSupported:
+		m.UserAgent.AddUserAgentFeature(awsmiddleware.UserAgentFeatureRequestChecksumWhenSupported)
+	case aws.RequestChecksumCalculationWhenRequired:
+		m.UserAgent.AddUserAgentFeature(awsmiddleware.UserAgentFeatureRequestChecksumWhenRequired)
+	}
+
+	for algo, feat := range supportedChecksumFeatures {
+		checksumHeader := AlgorithmHTTPHeader(algo)
+		if checksum := req.Header.Get(checksumHeader); checksum != "" {
```
Madrigal, line 52: I suspect you already checked this, but I'm surprised these headers are already set so early on the middleware stack

service/internal/checksum/middleware_checksum_metrics_tracking.go
```diff
@@ -0,0 +1,90 @@
...
+
+// RequestChecksumMetricsTracking is the middleware to track operation request's checksum usage
+type RequestChecksumMetricsTracking struct {
+	RequestChecksumCalculation aws.RequestChecksumCalculation
+	UserAgent                  *awsmiddleware.RequestUserAgent
+}
+
+// ID provides the middleware identifier
+func (m *RequestChecksumMetricsTracking) ID() string {
+	return "AWSChecksum:RequestMetricsTracking"
```
Madrigal, line 28: [observation] we don't follow this pattern of `x:y`on other middlewares, but I like how you used it here

service/internal/checksum/middleware_checksum_metrics_tracking_test.go
```diff
@@ -0,0 +1,169 @@
+//go:build go1.16
```
Madrigal, line 1: 1. Why do we need to duplicate the user agent test? If we just want to test that it includes the features we're interested in, I wouldn't duplicate the whole test but just check that the user agent includes the features we're interested in
2. Why do we put this under `go:build` tags?
3. If we need to do the above, can we change this to go 1.21?

feature/s3/manager/integ_upload_test.go
```diff
@@ -223,6 +222,28 @@ func TestInteg_UploadPresetChecksum(t *testing.T) {
 			},
 		},
 		"auto multipart part": {
+			"no checksum": {
```
Madrigal, line 225: is this actually no checksum? it seems like we set CRC32

service/internal/checksum/middleware_checksum_metrics_tracking.go
wty-Bryant: they are set at [serialize](https://github.com/aws/aws-sdk-go-v2/blob/885de40869f9bcee29ad11d60967aa0f1b571d46/service/s3/serializers.go#L7729) step on stack

wty-Bryant: the check here is only for checksums marked with feature ID

.changelog/9ebe24c4791541e0840da49eab6f9d97.json
```diff
@@ -0,0 +1,11 @@
+{
+    "id": "9ebe24c4-7915-41e0-840d-a49eab6f9d97",
+    "type": "feature",
+    "description": "S3 client behavior is updated to always calculate a checksum by default for operations that support it (such as PutObject or UploadPart), or require it (such as DeleteObjects). The checksum algorithm used by default now becomes CRC32. Checksum behavior can be configured using `when_supported` and `when_required` options - in code using RequestChecksumCalculation, in shared config using request_checksum_calculation, or as env variable using AWS_REQUEST_CHECKSUM_CALCULATION. The S3 client attempts to validate response checksums for all S3 API operations that support checksums. However, if the SDK has not implemented the specified checksum algorithm then this validation is skipped. Checksum validation behavior can be configured using `when_supported` and `when_required` options - in code using ResponseChecksumValidation, in shared config using response_checksum_validation, or as env variable using AWS_RESPONSE_CHECKSUM_VALIDATION.",
```
Madrigal, line 4:
```suggestion
    "description": "S3 client behavior is updated to always calculate a checksum by default for operations that support it (such as PutObject or UploadPart), or require it (such as DeleteObjects). The checksum algorithm used by default now becomes CRC32. Checksum behavior can be configured using `when_supported` and `when_required` options in 3 different places - in code using RequestChecksumCalculation, in shared config using request_checksum_calculation, or as env variable using AWS_REQUEST_CHECKSUM_CALCULATION. The S3 client attempts to validate response checksums for all S3 API operations that support checksums. However, if a different checksum algorithm is specified and the SDK has not yet implemented it, this validation is skipped. Checksum validation behavior can be configured using `when_supported` and `when_required` options in 3 places as well - in code using ResponseChecksumValidation, in shared config using response_checksum_validation, or as env variable using AWS_RESPONSE_CHECKSUM_VALIDATION.",
```

Madrigal (approved):

Issue #2819, opened 2024-10-04
Excessive service module version churn


ewbankkit:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

Over recent weeks there have been a number of releases where every service module's version has increased and the only change for those services without API model changes has been a re-ordering of calls inside the service's `New` function.
For example with yesterday's `2024-10-03` release for the `accessanalyzer` service, calls to `resolveTracerProvider` and `resolveMeterProvider` were swapped. This module had no functionality changes but a new version of the module was published.

This module version churn causes projects that use many service modules (e.g. `terraform-provider-aws`) to update a large number of dependencies and can cause a problem with `dependabot` timeouts.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Service module only publish new versions when there is new functionality (or their own dependencies change).

### Current Behavior

Excessive module version churn.

### Reproduction Steps

n/a

### AWS Go SDK V2 Module Versions Used

n/a

### Compiler and Version used

go 1.23.1

### Operating System and version

linux

lucix-aws:
Do you have any examples of this happening for something _other_ than `resolveTracerProvider` and `resolveMeterProvider`?

The order of codegen for these two is evidently unstable - just checked on my own system, and I can re-generate multiple times and get either result. The abstraction through which they're inserted uses a [HashMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/HashSet.html), which is very much not guaranteed to retain order. How we have only run into this now, I have no idea. 🤦 

Unfortunately the release is about to happen again and so there's a 50/50 chance it'll swap today. We'll stabilize the order to prevent it in the future, at least.

lucix-aws:
I have a patch ready. Waiting to see what the order is in today's release so I can match that (which would prevent an extra chrun on monday's release).

ewbankkit:
@lucix-aws In my (relatively rather superficial) investigations I've only seen `resolveTracerProvider` and `resolveTracerMeter` swap.

lucix-aws:
We've structured things to be ordered in codegen by design (e.g. struct members are sorted, generated serde is sorted) so this should be an exception rather than the rule.

[...]

PR #2822, opened 2024-10-07
Remove deprecated nimble service

wty-Bryant:


.changelog/4490529bdb8044068d99216cba424862.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "4490529b-db80-4406-8d99-216cba424862",
+    "type": "feature",
+    "description": "Remove deprecated `nimble` service.",
+    "modules": [
+        "."
```
Madrigal, line 6: nit: shouldn't this be scoped to `nimble` instead of `.`?

Madrigal (approved):
If available, we should link any public announcement on these service deprecations

.changelog/4490529bdb8044068d99216cba424862.json
lucix-aws: So technically the `nimble` module is being deleted -- I'm not sure if our tooling is written to handle that as an edge case. I suspect it will try to go update `service/nimble/CHANGELOG.md` to include this new entry, but that file wouldn't exist anymore since we deleted it, so the release would probably explode.

The other problem we have (recently noted in #2819, not directly although I think this is another case of that) is that a changelog like this bumps the root SDK module for basically no reason (no actual change) which by proxy bumps every service client module for equally no reason.

TL;DR not sure what to do here. We should honestly probably just not have a changelog for these types of deprecations for now.

lucix-aws (changes requested):
Let's not merge this until we make a decision on ☝️

.changelog/4490529bdb8044068d99216cba424862.json
wty-Bryant: involved modules are generated by changelog cmd according to which commit we use, here I passed the commit including both model and service api removal

Issue #2827, opened 2024-10-08
PresignedPostObject doesn't map bucket name in URL


rdmdb:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

Generating presigned URLs with PresignPostObject while using a custom BaseEndPoint and UsePathStyle set to true in the s3 config options, returns a URL that does not include the provided bucket name. This issue appears related to a similar bug in the JS counter part https://github.com/aws/aws-sdk-js-v3/commit/5054ba5fd8622960b699dfeb91a4b06dd77f5a70#diff-7253082da28c79a9309bc2a05179a34663053b1901d8724ed7f12751d5ae7ae6R90.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

The returned URL should look like `<schema>://<baseendpoint>/<bucket>`

### Current Behavior

The returned URL is currently formatted as `<schema>://<baseendpoint>`

### Reproduction Steps

```
package main

import (
	"context"
	"fmt"
	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/credentials"
	"github.com/aws/aws-sdk-go-v2/service/s3"
	"time"
)

func main() {
	staticProvider := credentials.NewStaticCredentialsProvider(
		"someAccessKey",
		"someSecret",
		"",
	)
	cfg, _ := config.LoadDefaultConfig(
		context.Background(),
		config.WithCredentialsProvider(staticProvider),
		config.WithRegion("auto"),
	)
	client := s3.NewFromConfig(cfg, func(options *s3.Options) {
		options.BaseEndpoint = aws.String("https://localhost:9000")
		options.UsePathStyle = true
	})
	presignClient := s3.NewPresignClient(client)
	request, _ := presignClient.PresignPostObject(
		context.Background(),
		&s3.PutObjectInput{
			Bucket: aws.String("mybucket"),
			Key:    aws.String("myobject"),
		}, func(opts *s3.PresignPostOptions) {
			opts.Expires = time.Duration(60) * time.Second
		},
	)
	fmt.Println(request.URL) // prints https://localhost:9000
}
```

[...]

PR #2835, opened 2024-10-15
Added support for path style in presigned Post

RanVaknin:
Following https://github.com/aws/aws-sdk-go-v2/issues/2827

1. The Go v2 SDK needed to add presigned-post support for path style URLs. The change to `toBaseURL` adds support for the following casing:
- Checks if the input URL has a leading slash by inspecting the path segments. If so, it removes the leading empty segment.

- Determines if the URL is "path-style" (bucket name in path) or "virtual-hosted" (bucket name in host) by checking the length of the path segments. If there are multiple segments, it's path-style; otherwise, it's virtual-hosted. The function then constructs the appropriate base URL accordingly.

This is a heuristic approach of parsing the URL and will not cover all of the edge cases and ways for users of the SDK to provide an custom endpoint that might not work accordingly. 

2. To solve this, I added a `PresignPostOptions` option called `BaseURLOverride` which will let users directly override the URL that the presigner uses.

3. Added test cases for both additions.

### DIFF
--- a/service/s3/presign_post.go
+++ b/service/s3/presign_post.go
@@ -300,9 +300,8 @@ func toBaseURL(fullURL string) string {
 
 	if len(pathSegments) > 1 {
 		return fmt.Sprintf("%s://%s/%s", parsedURL.Scheme, parsedURL.Host, pathSegments[0])
-	} else {
-		return fmt.Sprintf("%s://%s", parsedURL.Scheme, parsedURL.Host)
 	}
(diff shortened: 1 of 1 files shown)


RanVaknin:
Should we rename the test function to "TestPresign**Post**Object" instead of "TestPresignPutObject"?

lucix-aws:
https://github.com/aws/aws-sdk-go-v2/pull/2842

PR #2838, opened 2024-10-15
compiler break: change LaunchTemplateConfiguration.SetDefaultVersion from bool to *bool

### LINKED ISSUE #2734: MIGRATION ISSUE: `imagebuilder.LaunchTemplateConfiguration.SetDefaultVersion` not sending `false` value
### Pre-Migration Checklist

- [X] I've read the [Migration Guide](https://aws.github.io/aws-sdk-go-v2/docs/migrating/).
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for similar migration issues.

### Go Version Used

go1.22.5

### Describe the Migration Issue

If the bool `SetDefaultVersion` field of the `imagebuilder.LaunchTemplateConfiguration` structure is set to `false`, that value is not being serialized and sent to the AWS service. The means that it is impossible to update the value from `true` to `false` via v2 of the Go SDK.

#### v1

```go
	apiObject := &imagebuilder.LaunchTemplateConfiguration{}

	if v, ok := tfMap["launch_template_id"].(string); ok && v != "" {
		apiObject.LaunchTemplateId = aws.String(v)
	}

	if v, ok := tfMap["default"].(bool); ok {
		apiObject.SetDefaultVersion = aws.Bool(v)
	}
```

#### v2

```go
	apiObject := &awstypes.LaunchTemplateConfiguration{}

	if v, ok := tfMap["launch_template_id"].(string); ok && v != "" {
		apiObject.LaunchTemplateId = aws.String(v)
	}

	if v, ok := tfMap["default"].(bool); ok {
		apiObject.SetDefaultVersion = v
	}
```

#### v1

```json
{"distributionConfigurationArn":"arn:aws:imagebuilder:us-west-2:...:distribution-configuration/ewbankkit-test-001","distributions":[{"launchTemplateConfigurations":[{"accountId":"...","launchTemplateId":"lt-00c9b21fbf4907bdb","setDefaultVersion":false}],"region":"us-west-2"}]}
```

#### v2

```json
{"distributionConfigurationArn":"arn:aws:imagebuilder:us-west-2:...:distribution-configuration/ewbankkit-test-001","distributions":[{"launchTemplateConfigurations":[{"accountId":"...","launchTemplateId":"lt-00c9b21fbf4907bdb"}],"region":"us-west-2"}]}
```

### Additional Context

_No response_

lucix-aws:
We are implementing this compiler break to fix abjectly broken behavior due to a service modeling issue. See the changelog entry for more information.

Closes #2734


Madrigal:
I'm a bit concerned about the behavior mentioned at aws/aws-sdk#736

> when the value is not explicitly sent, the service assigns a true default value for it.

With this change, we will always set it to false. Not sure what the service implications are for it, maybe it means that when you create new distributions you no longer have a default value until you set one explicitly?

wty-Bryant (approved):

lucix-aws:
> I'm a bit concerned about the behavior mentioned at [aws/aws-sdk#736](https://github.com/aws/aws-sdk/issues/736)
> 
> > when the value is not explicitly sent, the service assigns a true default value for it.
> 
> With this change, we will always set it to false. Not sure what the service implications are for it, maybe it means that when you create new distributions you no longer have a default value until you set one explicitly?

I just had a conversation with the service team about this and they've confirmed that the tri-state capability is important. So I'm going to switch this to _removing_ @default instead of _backfilling_ @required. It's a compiler break rather than a runtime break, basically.

Going forward we should take care to try to pin down service teams to get an idea of what the input surface should be, even if we can't get them to prioritize a fix.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/RemoveDefaults.java
```diff
@@ -32,7 +32,11 @@ public class RemoveDefaults implements GoIntegration {
         serviceToShapeIds("com.amazonaws.paymentcryptographydata#PaymentCryptographyDataPlane",
                 "com.amazonaws.paymentcryptographydata#IntegerRangeBetween4And12"),
         serviceToShapeIds("com.amazonaws.emrserverless#AwsToledoWebService",
-                "com.amazonaws.emrserverless#WorkerCounts"));
+                "com.amazonaws.emrserverless#WorkerCounts"),
+        serviceToShapeIds("com.amazonaws.imagebuilder#imagebuilder",
+                    // https://github.com/aws/aws-sdk-go-v2/issues/2734
```
Madrigal, line 37: Thanks for actually putting the issue and ticket here, otherwise we'd be like 

"why do we need this customization?" 
"🤷"

Curious as to what happens when/if it gets fixed by the service. What happens with this customization?

lucix-aws: Nothing, it's a no-op for `ShapeBuilder.removeTrait`:

```
    public B removeTrait(ShapeId traitId) {
        if (this.traits.hasValue()) {
            ((Map)this.traits.get()).remove(traitId);
        }

        return this;
    }
```

lucix-aws: > "why do we need this customization?"

Currently asking myself this for every single other thing on this list, in fact!

Madrigal (approved):

PR #2842, opened 2024-10-16
Update presign post URL resolution to use the exact result from EndpointResolverV2

### LINKED ISSUE #2827: PresignedPostObject doesn't map bucket name in URL

Madrigal:
- **Update presign post URL resolution to use the exact result from EndpointResolverV2**

Fixes #2827

Discussed offline. Our original approach was to update `toBaseURL` to fix this edge case. However, this had the pitfall of being hard to capture all corner cases. Instead of coming up with a heuristic to figure out how to best parse the URL from the bucket, use the URI directly from endpointResolverV2. This change does this by:

1. Adds a new code generator that has a `renderPostEndpointResolutionHook`, which generates the line on `endpoints.go` to store the URI on the context
2. PresigPost uses this value instead of getting the URL from the request object

Tests are copied from a draft PR with the 1st approach

## Testing
1. Ran all unit tests for presign_post_test
2. Ran integration tests manually (although we don't need this now since they are run by default 🎉 )

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -10,7 +10,6 @@ import (
 type s3BackendKey struct{}
 type checksumInputAlgorithmKey struct{}
 type clockSkew struct{}
-type s3resolvedURI struct{}
 
 const (
 	// S3BackendS3Express identifies the S3Express backend
@@ -50,15 +49,4 @@ func SetAttemptSkewContext(ctx context.Context, v time.Duration) context.Context
 func GetAttemptSkewContext(ctx context.Context) time.Duration {
 	x, _ := middleware.GetStackValue(ctx, clockSkew{}).(time.Duration)
 	return x
-}
-
-// SetS3ResolvedURI sets the URI as resolved by the EndpointResolverV2
-func SetS3ResolvedURI(ctx context.Context, value string) context.Context {
-	return middleware.WithStackValue(ctx, s3resolvedURI{}, value)
-}
-
-// GetS3ResolvedURI gets the URI as resolved by EndpointResolverV2
-func GetS3ResolvedURI(ctx context.Context) string {
-	v, _ := middleware.GetStackValue(ctx, s3resolvedURI{}).(string)
-	return v
 }
--- a/service/s3/presign_post_test.go
+++ b/service/s3/presign_post_test.go
@@ -33,6 +33,12 @@ func TestPresignPutObject(t *testing.T) {
 				Key:    aws.String("key"),
 			},
 		},
+		"bucket and key have the same value": {
+			input: PutObjectInput{
+				Bucket: aws.String("bucket"),
+				Key:    aws.String("bucket"),
+			},
+		},
 		"expires override": {
 			input: PutObjectInput{
 				Bucket: aws.String("bucket"),
@@ -76,6 +82,14 @@ func TestPresignPutObject(t *testing.T) {
 			expectedURL:      "https://s3.us-west-2.amazonaws.com/bucket",
 			pathStyleEnabled: true,
 		},
+		"use path style bucket and key have the same value ": {
+			input: PutObjectInput{
+				Bucket: aws.String("value"),
+				Key:    aws.String("value"),
+			},
+			expectedURL:      "https://s3.us-west-2.amazonaws.com/value",
+			pathStyleEnabled: true,
+		},
 		"use path style bucket with custom baseEndpoint": {
 			input: PutObjectInput{
 				Bucket: aws.String("bucket"),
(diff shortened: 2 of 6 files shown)


wty-Bryant (approved):

internal/context/context.go
```diff
@@ -50,3 +51,14 @@ func GetAttemptSkewContext(ctx context.Context) time.Duration {
 	x, _ := middleware.GetStackValue(ctx, clockSkew{}).(time.Duration)
 	return x
 }
+
+// SetS3ResolvedURI sets the URI as resolved by the EndpointResolverV2
+func SetS3ResolvedURI(ctx context.Context, value string) context.Context {
```
lucix-aws, line 56: If we _only_ need this in S3 right now then let's just have it unexported in `service/s3`. We're still committing to a pseudo-API contract by doing this, even though it's under internal/, if we ever wanted to expand it to other services we wouldn't be able to remove it because it could break cross-module compat.

service/s3/presign_post_test.go
```diff
@@ -66,6 +68,23 @@ func TestPresignPutObject(t *testing.T) {
 			},
 			expectedURL: "https://mfzwi23gnjvgw.mrap.accesspoint.s3-global.amazonaws.com",
 		},
+		"use path style bucket hosting pattern": {
```
lucix-aws, line 71: Can we also add tests for:
* bucket and key are the same value
* custom BaseEndpoint has an arbitrary path segment (e.g. `locals3.com/v1`)

lucix-aws (approved):
fixandship

internal/context/context.go
Madrigal: Added a new file where I moved these operations and made them package private

service/s3/presign_post_test.go
Madrigal: Added tests, thanks for the suggestion

Issue #2859, opened 2024-10-30
Error message isn't displayed when some ECS operations fail due to AccessDeniedException


abicky:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

No error message is displayed when some ECS operations depending on `ecs.awsAwsjson11_deserializeErrorAccessDeniedException` fail due to AccessDeniedException.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

An error message is displayed:

```
panic: operation error ECS: CreateService, https response error StatusCode: 400, RequestID: 4b95afb3-25ab-48dc-a2ba-1a9e4c0ecbda, AccessDeniedException: User: arn:aws:sts::xxxxxxxxx:assumed-role/xxxxxxxxxxx is not authorized to perform: ecs:CreateService on resource: arn:aws:ecs:ap-northeast-1:xxxxxxxxxxxxx:service/default/xxxxxxxx because no identity-based policy allows the ecs:CreateService action

goroutine 1 [running]:
main.main()
        example.com/test/main.go:23 +0x184
```

### Current Behavior

No error message is displayed:

```
panic: operation error ECS: CreateService, https response error StatusCode: 400, RequestID: 2ed99e65-d0e9-463f-97c9-8b8cd34ffedf, AccessDeniedException:

goroutine 1 [running]:
main.main()
        example.com/test/main.go:23 +0x184
```

### Reproduction Steps

1. Prepare an IAM user or role that doesn't have the permission `ecs:CreateService`
2. Run the following file with the user or role:
    - Note that you have to change the ARN of task definition.

```go
package main

import (
	"context"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/ecs"
)

func main() {
	cfg, err := config.LoadDefaultConfig(context.Background())
	if err != nil {
		panic(err)
	}

	client := ecs.NewFromConfig(cfg)
	_, err = client.CreateService(context.Background(), &ecs.CreateServiceInput{
		ServiceName: aws.String("xxxxx"),
[...]

PR #2860, opened 2024-10-30
Fix ecs.awsAwsjson11_deserializeErrorAccessDeniedException

abicky:
This PR fixes https://github.com/aws/aws-sdk-go-v2/issues/2859.
I've changed service/ecs/deserializers.go by running `make all` after I edited codegen/sdk-codegen/aws-models/ecs.json.

Although I've read the following notice, I thought it would be better to create a PR:

> For changes to files under the `/codegen/aws-models` folder, and manual edits to autogenerated code (e.g. `/service/s3/api.go`) please create an Issue instead of a PR for those type of changes.

If there are any problems, please feel free to close this PR.

### DIFF
--- a/codegen/sdk-codegen/aws-models/ecs.json
+++ b/codegen/sdk-codegen/aws-models/ecs.json
@@ -32,7 +32,7 @@
         "com.amazonaws.ecs#AccessDeniedException": {
             "type": "structure",
             "members": {
-                "message": {
+                "Message": {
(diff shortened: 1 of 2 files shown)


RanVaknin:
Hi @abicky ,

Your root causing seems to be correct, however we can't accept changes to the model files found under the codegen folder. These are directly code generated from the upstream API model which is considered "The source of truth". We will have to upstream this request on your behalf.

Thanks,
Ran~

Issue #2861, opened 2024-10-30
Cloudfront sign cookie expire time not update


deepenpatel19:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

Usually, we decide when to set cookie expire. So, we are passing expire time in cookie policy statement. But, it's not updating in cookie when generated. 

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

It should be set expire what we pass in policy statement or cookie options.

### Current Behavior

At this moment, it's giving default time.

### Reproduction Steps

```go
awsEpochTime := sign.NewAWSEpochTime(expireTimeNew)
	customStatement := sign.Statement{
		Resource: "https://*.<domain>/media/*",
		Condition: sign.Condition{
			DateLessThan: awsEpochTime,
		},
	}
	customPolicy := sign.Policy{
		Statements: []sign.Statement{
			customStatement,
		},
	}

	signedCookies, err := cookieSigner.SignWithPolicy(&customPolicy, func(co *sign.CookieOptions) {
		co.Path = "/"
		co.Domain = ".<domain>"
		co.Secure = true
	})
```

Here, we are passing date less than for expiry. But, it's not set expiry time in cookie.

### Possible Solution

May be we can pass expire in cookie option or pick from policy statement while generating cookie.

### AWS Go SDK V2 Module Versions Used

```
github.com/aws/aws-sdk-go-v2 v1.30.1
github.com/aws/aws-sdk-go-v2/config v1.27.11
github.com/aws/aws-sdk-go-v2/credentials v1.17.11
github.com/aws/aws-sdk-go-v2/service/rekognition v1.43.0
github.com/aws/aws-sdk-go-v2/service/s3 v1.53.1
github.com/aws/aws-sdk-go-v2/service/sqs v1.32.6
```

### Compiler and Version used

1.23.1

### Operating System and version

Mac

deepenpatel19:
Possible solution for this
https://github.com/aws/aws-sdk-go-v2/pull/2862

PR #2862, opened 2024-10-30
Cloudfront - add expire time in signed cookie.

deepenpatel19:
Raised bug: https://github.com/aws/aws-sdk-go-v2/issues/2861

This is possible solution to add expire time while cookie generation. 
So, do not need to manually pass while sending back to response header.

### DIFF
--- a/feature/cloudfront/sign/sign_cookie.go
+++ b/feature/cloudfront/sign/sign_cookie.go
@@ -24,7 +24,7 @@ type CookieOptions struct {
 	Domain   string
 	Secure   bool
 	SameSite http.SameSite
-	Expire   time.Time
+	Expires  time.Time
 }
 
 // apply will integration the options provided into the base cookie options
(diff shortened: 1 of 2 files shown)


feature/cloudfront/sign/sign_cookie.go
```diff
@@ -24,6 +24,7 @@ type CookieOptions struct {
 	Domain   string
 	Secure   bool
 	SameSite http.SameSite
+	Expire   *AWSEpochTime
```
lucix-aws, line 27: Let's just use `time.Time` - the wrapper is for serialization which isn't necessary if we're just passing it to an `http.Cookie`.

This looks fine otherwise.

lucix-aws (changes requested):

feature/cloudfront/sign/sign_cookie.go
deepenpatel19: Updated with `time.Time`.
Thanks for your feedback.

lucix-aws (approved):

lucix-aws:
Will merge pending CI pass. Ignore integration test failure.

PR #2881, opened 2024-10-31
Update Smithy to 1.52.1

Madrigal:
Includes the following changes

- **Update Smithy to 1.52.1**
- **Update protocol tests after Smithy 1.52.1 update**
- **Add changelog**

I wasn't expecting these many protocol tests to be updated by bringing in the latest version. Did a quick check of one that we brought in with this release ([this one](https://github.com/smithy-lang/smithy/blob/6b2c80cbea9e733560852d33b464d52fbf2b708a/smithy-aws-protocol-tests/model/awsJson1_0/required.smithy#L10)) and it seems to have been introduced in 1.41, so unsure why it wasn't being generated.

Relies on aws/smithy-go#550

### DIFF
--- a/codegen/protocol-test-codegen/smithy-build.json
+++ b/codegen/protocol-test-codegen/smithy-build.json
@@ -8,6 +8,8 @@
           "args": {
             "services": ["aws.protocoltests.ec2#AwsEc2"]
           }
+        },{
+          "name": "removeUnusedShapes"
         }
       ],
       "plugins": {
@@ -24,6 +26,8 @@
           "args": {
             "services": ["aws.protocoltests.json#JsonProtocol"]
           }
+        },{
+          "name": "removeUnusedShapes"
         }
       ],
       "plugins": {
@@ -40,6 +44,8 @@
           "args": {
             "services": ["aws.protocoltests.json10#JsonRpc10"]
           }
+        },{
+          "name": "removeUnusedShapes"
         }
       ],
       "plugins": {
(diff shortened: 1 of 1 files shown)


lucix-aws (approved):

wty-Bryant (approved):

SMITHY_GO_CODEGEN_VERSION
```diff
@@ -1 +1 @@
-e6338ca9b5ee30c6317ea908173a369a9d3eb236
+48250f49d98ad6bed6c302a8af97f62cc7f3bd83
```
lucix-aws, line 1: Hold on, is this right?

[this](https://github.com/aws/smithy-go/commit/a4c9efcda6aa54c75d1a130d1320a2709eebf51d) seems to be the latest (w/ smithy upgrade)

Madrigal: Great catch, will update it

Madrigal:
Discussed with Smithy, this is related to smithy-lang/smithy#2432

From Swift who also faced this issue

> Smithy 1.52.1 adds a new validator https://github.com/smithy-lang/smithy/pull/2426 which flags shapes that aren't attached to a service. The [removeUnusedShapes](https://smithy.io/2.0/guides/smithy-build-json.html#removeunusedshapes) transform is added to the build command for Smithy protocol test generation so that these shapes are stripped & the validation is not triggered.

Madrigal:
@erka will send another PR to add other common stuff to `.gitignore`

PR #2885, opened 2024-11-01
fix makefile to not spam releases for feature/dynamodbstreams/attributevalue

lucix-aws:
We've been spamming `feature/dynamodbstreams/attributevalue` releases, this fixes that.

## why

feature/dynamodbstreams is copied from feature/dynamodb as part of make recipe `copy-attributevalue-feature` - so whenever you update one, you technically also update the other. This happens on release.

That _includes_ the `go_module_metadata.go` file from the source module. At some point in the past few months the committed versions for these two files got desynced. So every release, this has been happening:

* release runs `make generate` - nothing changes in the streams version _except_ for go_module_metadata.go
* release system goes "hey, contents of `feature/dynamodbstreams/attributevalue` changed - new release!"
* release annotation runs, updating tags and module versions - which upon seeing no changelog entry in ddb streams, just does the default patch version bump, inserting a dummy changelog that says "no notes"
  * that process involves setting the value in go_module_metadata.go to the new version. so the incidental copy, which triggered this whole thing in the first place, is overwritten. rinse and repeat

## testing

Tested by running `make generate` with:
* nothing changed in dynamodb/attributevalue. verified that the module metadata file stream-side didn't update this time.
* some dummy change in dynamodb/attributevalue. verified that it still gets copied

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,13 +215,13 @@ sync-api-models:
 
 copy-attributevalue-feature:
 	cd ./feature/dynamodbstreams/attributevalue && \
-	find . -name "*.go" | grep -v "doc.go" | xargs -I % rm % && \
-	find ../../dynamodb/attributevalue -name "*.go" | grep -v "doc.go" | xargs -I % cp % . && \
-	ls *.go | grep -v "convert.go" | grep -v "doc.go" | \
+	find . -name "*.go" | grep -v "doc.go" | grep -v "go_module_metadata.go" | xargs -I % rm % && \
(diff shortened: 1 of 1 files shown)


wty-Bryant (approved):

Madrigal:
Is it also a valid use case to update `feature/dynamodbstreams` independently and ensure the metadata gets updated?

lucix-aws:
No - it should be a direct clone of the main one.

PR #2888, opened 2024-11-05
fix potential for user-agent lang value mismatch in tests

lucix-aws:
`runtime.Version()` may have extra data after the semver string that tests weren't accounting for (they were not normalizing it in the "expect" input, this fixes that)


Madrigal:
LGTM, just 2 nits

1. Can we describe better what the actual issue with this was on the PR? offline there was a good explanation of it
2. The current error message is rather unhelpful

> t.Error("User-Agent did not match expected")

Can we change the error message to 

> t.Error("User-Agent %v did not match expected %v", ua[0], c.Expect)

Madrigal (approved):

wty-Bryant (approved):

PR #2889, opened 2024-11-06
fix: handle error message field in structure deserializer

RanVaknin:
fixes: https://github.com/aws/aws-sdk-go-v2/issues/2859

Error messages from AWS services can be returned with inconsistent casing (`message` or `Message`). 
This change updates the error deserialization logic to handle case-insensitive message field matching.

### Driver code setup:
This code uses a profile configuration that would assume a restricted role that does not have permissions to call ecs:CreateService. This will trigger an `AccessDeniedException` in the server's response that includes a `Message` field which breaks deserialization since ECS models `message` (lower case). The result is that the message is omitted from the error.

```go
package main

import (
	"context"
	"fmt"
	"github.com/aws/aws-sdk-go-v2/service/ecs"
	"log"
)

func main() {
	cfg, err := config.LoadDefaultConfig(
		context.TODO(),
		config.WithSharedConfigProfile("restricted_role_profile"),
	)
	if err != nil {
		log.Fatalf("unable to load SDK config, %v", err)
	}

	client := ecs.NewFromConfig(cfg)

	out, err := client.CreateService(context.Background(), &ecs.CreateServiceInput{
		ServiceName:    aws.String("foo"),
		TaskDefinition: aws.String("arn:aws:ecs:us-east-1:REDACTED:task-definition/my-test-task:1"),
	})

	if err != nil {
		panic(err)
	}

	fmt.Println(out.Service.ServiceName)
}
```

### Outputs:

```
panic: operation error ECS: CreateService, https response error StatusCode: 400, RequestID: REDACTED, AccessDeniedException: 
```

### Output after change:

```
panic: operation error ECS: CreateService, https response error StatusCode: 400, RequestID: REDACTED, AccessDeniedException: User: arn:aws:sts::REDACTED:assumed-role/restricted_role/aws-go-sdk-1730854295921846000 is not authorized to perform: ecs:CreateService on resource: arn:aws:ecs:us-east-1:REDACTED:service/default/foo because no identity-based policy allows the ecs:CreateService action
```


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
```diff
@@ -180,8 +181,21 @@ protected void deserializeStructure(GenerationContext context, StructureShape sh
...
 
+        boolean isErrorStructure = shape.hasTrait(ErrorTrait.class);
+        if (isErrorStructure) {
+            writer.write("var errorMessage string");
+        }
+
         // Iterate through the decoder. The member visitor will handle popping tokens.
         writer.openBlock("for key, value := range shape {", "}", () -> {
+            if (isErrorStructure) {
+                writer.write("keyLower := strings.ToLower(key)");
```
lucix-aws, line 192: The cost of lowercasing every key in every structure is non-trivial, it's not something we want to add here. We should just check for `Message` and `message`.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
```diff
@@ -180,8 +181,21 @@ protected void deserializeStructure(GenerationContext context, StructureShape sh
...
+                writer.write("if keyLower == \"message\" {");
```
lucix-aws, line 193: JFYI you can and should do this in a single statement.

```
writer.write("""
        if foo = bar {
            baz qux
        }"""
);
```

lucix-aws (changes requested):
I can understand _why_ you might have gone about it this way, but you've pushed complexity into the runtime that should really be handled in codegen.

The real (generated) code change that needs to happen here is less complicated than what you have currently. Basically, if we had

```
case "message":
```

before, we should now just be able to have

```
case "message", "Message":
```

That's going to be of significantly less runtime complexity than what you have currently.

service/acm/deserializers.go
```diff
@@ -2261,15 +2261,14 @@ func awsAwsjson11_deserializeDocumentAccessDeniedException(v **types.AccessDenie
...
-		case "Message":
+		case "message", "Message":
 			if value != nil {
 				jtv, ok := value.(string)
 				if !ok {
 					return fmt.Errorf("expected ServiceErrorMessage to be of type string, got %T instead", value)
 				}
 				sv.Message = ptr.String(jtv)
 			}
-
```
lucix-aws, line 2272: The changes to codegen here dropped this newline between cases across the board - it's making the diff harder to read, we should put it back.

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
```diff
@@ -190,10 +192,16 @@ protected void deserializeStructure(GenerationContext context, StructureShape sh
...
+
+                    boolean isErrorStructure = shape.hasTrait(ErrorTrait.class);
+
+                    if(isErrorStructure && memberName.equals("Message")){
+                        writer.write("case \"message\", \"Message\":");
+                    } else {
+                        writer.write("case $S:",serializedMemberName);
+                    }
+                    String dest = "sv." + memberName;
+                    context.getModel().expectShape(member.getTarget()).accept(getMemberDeserVisitor(member, dest));
```
lucix-aws, line 204: This should fix the missing newline issue (I didn't check it myself though)

```suggestion
                    context.getModel().expectShape(member.getTarget()).accept(getMemberDeserVisitor(member, dest));
                    writer.write("");
```

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/JsonShapeDeserVisitor.java
```diff
@@ -190,10 +192,16 @@ protected void deserializeStructure(GenerationContext context, StructureShape sh
...
                     String memberName = symbolProvider.toMemberName(member);
                     String serializedMemberName = getSerializedMemberName(member);
-                    writer.openBlock("case $S:", "", serializedMemberName, () -> {
-                        String dest = "sv." + memberName;
-                        context.getModel().expectShape(member.getTarget()).accept(getMemberDeserVisitor(member, dest));
-                    });
+
+                    boolean isErrorStructure = shape.hasTrait(ErrorTrait.class);
+
+                    if(isErrorStructure && memberName.equals("Message")){
```
lucix-aws, line 198:
```suggestion
                    if(isErrorStructure && memberName.equalsIgnoreCase("Message")){
```

lucix-aws: (this will handle both message and Message in the model, and of course anything in between)

lucix-aws, line 198: This definitely does warrant a source comment explaining why.

e.g.

```
// The 'message' field in JSON protocol errors is known to not always match its over-the-wire casing with what is modeled.
// See (link github issue)
```

codegen/smithy-aws-go-codegen/src/main/resources/META-INF/services/software.amazon.smithy.go.codegen.integration.GoIntegration
```diff
@@ -88,3 +88,4 @@ software.amazon.smithy.aws.go.codegen.customization.s3.ExpressUserAgent
 software.amazon.smithy.aws.go.codegen.customization.BackfillRequiredTrait
 software.amazon.smithy.aws.go.codegen.customization.DeprecateService
 software.amazon.smithy.aws.go.codegen.customization.BasicUserAgentFeatures
+
```
lucix-aws, line 91: ?

lucix-aws (approved):

Madrigal (approved):

wty-Bryant (approved):

PR #2902, opened 2024-11-15
Bump smithy-go dependency

wty-Bryant:
Update smithy-go run-time dependency version


lucix-aws:
Rebase from main, that should fix the integration tests.

PR #2910, opened 2024-11-22
Fix a typo in migration example code

james-lambert-techcyte:
I found a migration example that had a typo


Madrigal:
The docs have been migrated to https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/welcome.html, which are no longer handled by this repository but rather an internal repository. This makes it easier for a specialized docs team to handle these requests.

I have made this change on your behalf this time and it should be reflected soon, but if you see other typos or thing you'd like to see change in the future, feel free to request them directly on the new page using the "feedback" button

image

PR #2934, opened 2024-12-17
Fix improper use of Printf-style functions

djedward:
Go 1.24rc1 dropped and it improved the ability to catch more misuses of Printf-style functions. This PR addresses the ones that Go 1.24 will now catch and report as compile errors.


Madrigal (approved):

djedward:
One way might be to add a test platform that uses `tip` of Go, but I don't know how stable that would actually be.

lucix-aws (approved):

wty-Bryant (approved):

PR #2944, opened 2025-01-06
Add a workflow for external contributors to add a changelog entry

Madrigal:
Add instructions on how external contributors can add a `.changelog` entry. We've been taking over adding them in the past, but it'd be nice if contributors can do it themselves. 

The workflow now would be 

1. Contributors would run `make external-changelog`
2. They would have a file on `.changelog/xxx.json` that they would need to modify

#### Why `external-changelog`?
Because we internally use [aws-go-multi-module-repository-tools](https://github.com/awslabs/aws-go-multi-module-repository-tools/tree/main/cmd/changelog). External contributors could use it too, but it's a bit too cumbersome to set it up for using only single time

#### Why not creating a uuid on behalf of contributors?
There's no built-in uuid tool that would work cross-platform

### DIFF
--- a/changelog-template.json
+++ b/changelog-template.json
@@ -2,6 +2,7 @@
     "id": "00000000-0000-0000-0000-000000000000",
     "type": "feature|bugfix|dependency",
     "description": "Description of your changes",
+    "collapse": false,
     "modules": [
         "service/amplify",
         "service/connect"


lucix-aws (approved):

changelog-template.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "00000000-0000-0000-0000-000000000000",
+    "type": "feature|bugfix|dependency",
+    "description": "Description of your changes",
```
wty-Bryant, line 4: do we need collapse here

wty-Bryant (approved):

changelog-template.json
lucix-aws: The doc mentions it, don't think we need it here, but also a default of false would be fine.

PR #2945, opened 2025-01-06
deprecate service/iot1clickdevicesservice and service/iot1clickprojects

wty-Bryant:
iot1clickdevicesservice and iot1clickprojects services will be deprecated, this PR pushes a final module update to add deprecation markers before actually removing the source.

PR #2946, opened 2025-01-07
drop service/iot1clickdevicesservice and service/iot1clickprojects

wty-Bryant:
Follow up to #2945


Madrigal (approved):
Approved, but remember we shouldn't merge it until we have the release of the deprecation

PR #2948, opened 2025-01-07
redirect to new developer guide

lucix-aws:
* replaces all index.html in docs/ with a static redirect to the new landing page
* updates the readme to mention migration
* deletes everything else


Madrigal:
lol at the approval workflow trying its best to build :P

I see that we ended up doing a blank redirect to the welcome page. We briefly discussed whether to do a site-per-site redirect, like https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/ -> https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/handle-errors.html 

However, I see that the URL is not exactly the same (handling-errors vs handle-errors) and there are other cases where this happens. Do we want to update the references, or are we OK just redirecting to main and letting customer find the content if linked on their own?

lucix-aws:
Yeah, I looked into doing it dynamically but unfortunately the URLs just don't line up.

This should be a one-time experience anyway if say people have these bookmarked. They get the redirect once, find the right thing, update their links, and life goes on.

Madrigal (approved):

Issue #2958, opened 2025-01-15
Infinite Pagination in cloudwatchlogs.NewGetLogEventsPaginator()


osmontero:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

When paginating through CloudWatch Logs events via cloudwatchlogs.NewGetLogEventsPaginator(), the end of the log stream is not properly detected. Instead, the final page of the result set is continually requested in an infinite loop.

The problem is that the GetLogEvents API signals the end of the stream by setting nextForwardToken or nextBackwardToken to the same as the requested value (and not null, like most other paginated APIs). GetLogEventsPaginator evaluates if the nextToken is null instead of evaluating if it is equal, resulting in an infinity loop.

![Image](https://github.com/user-attachments/assets/614afbea-9a03-41c3-aa63-851e8b8de3f4)

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

HasMorePages() must return false if there are no more pages

### Current Behavior

HasMorePages() always returns true

### Reproduction Steps

```
paginator := cloudwatchlogs.NewGetLogEventsPaginator(cwl, &cloudwatchlogs.GetLogEventsInput{
  LogGroupName:  aws.String(logGroup),
  LogStreamName: aws.String(stream),
  StartTime:     aws.Int64(startTime.Unix() * 1000),
  EndTime:       aws.Int64(endTime.Unix() * 1000),
  StartFromHead: aws.Bool(true),
})

for paginator.HasMorePages() {
  page, err := paginator.NextPage(ctx)
  if err != nil {
    return nil, catcher.Error("cannot get logs", err, nil)
  }
  
  for _, event := range page.Events {
    transformedLogs = append(transformedLogs, *event.Message)
  }
}
```

More details can be found here: [main.go](https://github.com/utmstack/UTMStack/blob/feature/10.6.0/new_correlation/plugins/aws/main.go)

### Possible Solution

evaluate if the nextForwardToken is equal to the requested params.NextToken

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2 v1.32.8
github.com/aws/aws-sdk-go-v2/config v1.28.11
[...]

PR #2959, opened 2025-01-15
Fix infinite pagination in cloudwatchlogs.NewGetLogEventsPaginator()

osmontero:
Added a `lastToken` field in `GetLogEventsPaginator` to track the last processed token and prevent repeated iterations. This addresses issue https://github.com/aws/aws-sdk-go-v2/issues/2958


osmontero:
Found an option to stop on duplicate token

PR #2966, opened 2025-01-16
pull retry loop forward to cover everything from resolving auth scheme onward

### LINKED ISSUE #2965: Retry loop does not refresh credentials after recent SRA identity&auth refactor
_Before_ the I&A rewrite (#2364), credentials were retrieved in the `Signing` middleware. The rewrite moved that step out into a separate middleware that now occurs before the retry loop, meaning credentials are not refreshed within said loop as they should:

```
    Finalize stack step
        DisableAcceptEncodingGzip
        ResolveAuthScheme
>>>>>>> GetIdentity
        ResolveEndpointV2
        disableHTTPS
        AWSChecksum:ComputeInputPayloadChecksum
        ComputePayloadHash
        SigV4ContentSHA256Header
        spanRetryLoop
>>>>>>> Retry
        addInputChecksumTrailer
        RetryMetricsHeader
        setLegacyContextSigningOptions
        Signing
```

lucix-aws:
Closes #2965

* Moves the retry loop forward to cover credential retrieval, which is the intended behavior per Smithy client reference architecture and is how we behaved before our auth refactor in #2364.
* Also adjusts the middleware insert logic for checksums to ensure that the _header_ and _trailer_ checksums (which are separate middlewares) happen after one another. The latter requires the former to determine whether to operate through context signals. In the future we should just combine these, see #2968.
* Fixes tests in service/s3 and service/route53 to correctly inspect the final request. These were technically retrieving the request at the wrong point in time before, it just happened to work because the retry loop hadn't kicked in and cloned the request yet (which it now does, making the captured request pointers stale).

This change is verified by our integration tests.


Madrigal (approved):
LGMT as soon as integ tests pass

wty-Bryant (approved):

wty-Bryant (approved):

lucix-aws:
Integ tests failed again but it looked transient that time, retried

service/internal/checksum/middleware_add.go
```diff
@@ -53,29 +53,6 @@ type InputMiddlewareOptions struct {
 // AddInputMiddleware adds the middleware for performing checksum computing
 // of request payloads, and checksum validation of response payloads.
 func AddInputMiddleware(stack *middleware.Stack, options InputMiddlewareOptions) (err error) {
-	// TODO ensure this works correctly with presigned URLs
```
lucix-aws: (it does)

service/internal/checksum/middleware_add.go
```diff
@@ -53,29 +53,6 @@ type InputMiddlewareOptions struct {
...
-
-	// Middleware stack:
```
lucix-aws: This is so beyond out-of-date at this point I've just removed it

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsHttpChecksumGenerator.java
```diff
@@ -279,6 +278,48 @@ private void writeInputMiddlewareHelper(
...
+                    err = stack.Initialize.Add(&internalChecksum.SetupInputContext{
+                        GetAlgorithm:               options.GetAlgorithm,
+                        RequireChecksum:            options.RequireChecksum,
+                        RequestChecksumCalculation: options.RequestChecksumCalculation,
+                    }, middleware.Before)
+                    if err != nil {
+                        return err
+                    }
+
+                    stack.Build.Remove("ContentChecksum")
```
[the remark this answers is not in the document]
lucix-aws: Yeah, ContentChecksum is the old MD5, which the recent changes to S3 object integrity effectively removes.

PR #2964, opened 2025-01-16
Fix flex checksum nil dereference issue

wty-Bryant:
The latest flexible checksum feature will nil deference and panic operations(e.g. `s3.PutBucketOwnershipControls`) that require checksum but don't have input algorithm setting, this pr fixes that via nil checking before getting checksum algorithm

It has been tested locally that operation mentioned above will not panic after this change and will use crc32 checksum by default 

close #2963 by merging this pr

PR #2967, opened 2025-01-16
Add integration test case for flex checksum

wty-Bryant:
Follow #2964 to add integration test case for operations requiring checksum without input checksum algorithm setting

### DIFF
--- a/service/internal/integrationtest/s3/checksum_test.go
+++ b/service/internal/integrationtest/s3/checksum_test.go
@@ -560,7 +560,7 @@ func TestInteg_RequireChecksumWithoutRequestAlgorithmMember(t *testing.T) {
 	}
 	computedChecksums, ok := s3.GetComputedInputChecksumsMetadata(result.ResultMetadata)
 	if !ok {
-		t.Fatalf("expect computed checksums metadata present, got %q", computedChecksums)
+		t.Fatalf("expect computed checksums metadata present, got %q", result)
 	}
 
 	expectComputedChecksums := s3.ComputedInputChecksumsMetadata{


service/internal/integrationtest/s3/checksum_test.go
```diff
@@ -541,6 +541,38 @@ func TestInteg_RequireChecksum(t *testing.T) {
...
+		t.Fatal("expect computed checksums metadata present, got nil")
+	}
+
+	expectComputedChecksums := s3.ComputedInputChecksumsMetadata{
+		ComputedChecksums: map[string]string{
+			"CRC32": "cK9COg==",
+		},
+	}
+	if diff := cmpDiff(expectComputedChecksums, computedChecksums); diff != "" {
+		t.Errorf("expect computed checksum metadata match\n%s", diff)
```
Madrigal, line 572: `\n%s`  will print a new line and then the result. It should be `%s\n`

service/internal/integrationtest/s3/checksum_test.go
```diff
@@ -541,6 +541,38 @@ func TestInteg_RequireChecksum(t *testing.T) {
...
+	}
+
+	t.Logf("putting bucket ownership control: %q", *params.Bucket)
+	result, err := s3client.PutBucketOwnershipControls(context.Background(), params)
+	if err != nil {
+		t.Fatalf("expect no error, got %v", err)
+	}
+	computedChecksums, ok := s3.GetComputedInputChecksumsMetadata(result.ResultMetadata)
+	if !ok {
+		t.Fatal("expect computed checksums metadata present, got nil")
```
Madrigal, line 563: [nit] print the full result so it's easier to debug

service/internal/integrationtest/s3/checksum_test.go
wty-Bryant: I guess the old test case did that to make diff clear? But will change that

service/internal/integrationtest/s3/checksum_test.go
```diff
@@ -541,6 +541,38 @@ func TestInteg_RequireChecksum(t *testing.T) {
...
+	}
+
+	t.Logf("putting bucket ownership control: %q", *params.Bucket)
+	result, err := s3client.PutBucketOwnershipControls(context.Background(), params)
+	if err != nil {
+		t.Fatalf("expect no error, got %v", err)
+	}
+	computedChecksums, ok := s3.GetComputedInputChecksumsMetadata(result.ResultMetadata)
+	if !ok {
+		t.Fatalf("expect computed checksums metadata present, got %q", computedChecksums)
```
Madrigal, line 563:
```suggestion
		t.Fatalf("expect computed checksums metadata present, got %q", result)
```

Sorry if I wasn't clear with my suggestion. If we print `computedChecksums` we already know the result is nil, so we don't get anything new by printing the value. If we print the whole result, we can see what we got and hopefully have a better clue of what went wrong

Madrigal (approved):

lucix-aws (approved):

PR #2975, opened 2025-01-21
feat: s3/transfermanager (v2): round-robin DNS and multi-NIC

lucix-aws:
Add two functional options to configure http.Transport to do the following:
* round-robin DNS - S3 returns up to 8 DNS records per query, this injects a custom resolver that ensures new connections rotate between these IP addresses.
* multi-value NIC - replaces the underlying `DialContext` hook with one that cycles between supplied local addresses (NICs).

### DIFF
--- a/feature/s3/transfermanager/rrdns.go
+++ b/feature/s3/transfermanager/rrdns.go
@@ -95,7 +95,14 @@ func (r *rrDNS) getAddr(ctx context.Context, host string) (string, error) {
 
 func (r *rrDNS) lookupHost(ctx context.Context, host string) (string, error) {
 	ch := r.sf.DoChan(host, func() (interface{}, error) {
-		return r.resolver.LookupHost(ctx, host)
+		addrs, err := r.resolver.LookupHost(ctx, host)
+		if err != nil {
+			return nil, err
+		}
+
+		expires := timeNow().Add(r.expiry)
+		r.cache.PutAddrs(host, addrs, expires)
+		return nil, nil
 	})
 
 	select {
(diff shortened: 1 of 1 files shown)


feature/s3/transfermanager/rrdns.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return r.dialContext(ctx, network, net.JoinHostPort(ipaddr, port))
+}
+
+func (r *rrDNS) getAddr(ctx context.Context, host string) (string, error) {
+	addr, ok := r.cache.GetAddr(host)
+	if ok {
+		return addr, nil
+	}
+	return r.lookupHost(ctx, host)
```
Madrigal, line 93: So when we have a cache miss we'll refresh the whole cache? is this what we want?

lucix-aws: Not the whole cache? We just re-query for that one host, or am I missing something?

lucix-aws: cache is hostname -> []<ip address>

Madrigal: You're right, my thinking was that we were storing things like

```
hostname -> {
  ip1/expiry1
  ip2/expiry2
  ip3/expiry3
}
```

But we're storing all results of the lookup with the same expiry (since we don't get TTL), so the entries are

```
hostname -> {
  ip1/expiry1
  ip2/expiry1
  ip3/expiry1
}
```

Then it doesn't really matter if one expires since all entries from the lookup will expire at the same time. And anyway, anytime we do a lookup for the same hostname, we replace existing values so they effectively always have the same expiry

Madrigal (approved):

PR #2976, opened 2025-01-21
add client config to disable logging skipped output checksum validation

### LINKED ISSUE #2974: add config option to disable logging of skipped output checksum validation
See #2971 

We have an internal switch to disable logging of skipped output checksum validation, but when we generate checksum validation for supported operations, we just hardcode it to `true`: https://github.com/aws/aws-sdk-go-v2/blob/fa8a15e1e023fd02edb11264d3e926d65f18f87f/service/s3/api_op_GetObject.go#L823

With the advent of crc64 in S3, people are now going to see this more often as other SDKs like the CLI now PutObject w/ crc64 by default. We should forward this config switch into client Options e.g.

```
package s3

type Options struct {
  // By default, the client will log a warning when an output checksum cannot be validated.
  // This can be disabled with this setting.
  DisableLogSkippedChecksumValidation bool
}
```

lucix-aws:
Closes #2974 

Adds `DisableLogOutputChecksumValidationSkipped` to service clients that have output checksum ops (which, right now, is only S3's GetObject).

## Testing

Uploaded a file with the latest cliv2 release, then ran:

```go
package main

import (
	"context"
	"fmt"

)

func main() {
	if err != nil {
		panic(err)
	}

	svc := s3.NewFromConfig(cfg)
	fmt.Println("1:----------------")
	svc.GetObject(context.Background(), &s3.GetObjectInput{
		Bucket: aws.String("fugazi"),
		Key:    aws.String("cli-upload-crc64"),
	})
	fmt.Println("------------------")
	fmt.Println("2:----------------")
	svc.GetObject(context.Background(), &s3.GetObjectInput{
		Bucket: aws.String("fugazi"),
		Key:    aws.String("cli-upload-crc64"),
	}, func(o *s3.Options) {
		o.DisableLogOutputChecksumValidationSkipped = true
	})
	fmt.Println("------------------")
}
```

result:

```
1:----------------
SDK 2025/01/21 13:03:34 WARN Response has no supported checksum. Not validating response payload.
------------------
2:----------------
------------------
```

### DIFF
--- a/.changelog/3b6c86fadda84a1ead6caf65d4aca7ea.json
+++ b/.changelog/3b6c86fadda84a1ead6caf65d4aca7ea.json
@@ -1,0 +1,8 @@
+{
+    "id": "3b6c86fa-dda8-4a1e-ad6c-af65d4aca7ea",
+    "type": "feature",
+    "description": "Add a client config option to disable logging when output checksum validation is skipped due to an unsupported algorithm.",
+    "modules": [
+        "service/s3"
+    ]
+}


Madrigal:
LGTM. I'd love to expand on the error message to say something like `Object was uploaded with algorithm "X", which the Go SDK has no support for. Not validating response payload.`, but won't block on it

Madrigal (approved):

PR #2981, opened 2025-01-22
Fix flex checksum validation cfg

wty-Bryant:
Operations supporting output checksum validation (e.g. `s3.GetObject`) would only return checksum if user enable that in request. This PR fixes flex checksum's validation cfg to enable validation by default

It has been tested that getting a s3 object uploaded with checksum will validate checksum from response by default, so there should be no warning like `WARN Response has no supported checksum. Not validating response payload.`. Sample test code is shown below:
```
import (
	"bytes"
	"context"
	"io"

)

func main() {
	if err != nil {
		panic(err)
	}

	svc := s3.NewFromConfig(cfg)
	_, err = svc.PutObject(context.Background(), &s3.PutObjectInput{ // should default to crc32
		Bucket: aws.String("your-bucket"),
		Key:    aws.String("sdk-upload-crc32"),
		Body:   bytes.NewReader([]byte("hello")),
	})
	if err != nil {
		panic(err)
	}

	out, err := svc.GetObject(context.Background(), &s3.GetObjectInput{ // should default to crc32
		Bucket: aws.String("your-bucket"),
		Key:    aws.String("sdk-upload-crc32"),
	})
	if err != nil {
		panic(err)
	}
	defer out.Body.Close()
	if _, err := io.Copy(io.Discard, out.Body); err != nil {
		panic(err)
	}
}
```


lucix-aws (commented):
Please include sample code that tests this and demonstrates the desired outcome in the PR description.

.changelog/df93fff8f662441fa12dac87614a5064.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "df93fff8-f662-441f-a12d-ac87614a5064",
+    "type": "bugfix",
+    "description": "Add a setter to output checksum middleware cfg to enable request validation mode by default",
```
Madrigal, line 4: Is this setter something that customers can interact with? If not, I'd just leave it as "Enable request validation mode by default"

Madrigal:
to double check, if an object is uploaded with CRC64, enabling response validation should still succeed and we have a test for it, right?

lucix-aws:
> to double check, if an object is uploaded with CRC64, enabling response validation should still succeed and we have a test for it, right?

Seconded. In fact, we need integration tests for _everything_ here. Please demonstrate all of the following in integration tests:

* round-tripping an object succeeds and defaults to crc32
* downloading an object that has a crc64 checksum succeeds, and logs skipped-validation warning
  * since we don't support it natively, you can do a crc64 checksum by precomputing one and just attaching it to the PutObject call
* uploading an object with _no checksum_ and then downloading it through the Go SDK succeeds

lucix-aws:
We may have tests already that cover some of those cases, just point to them it so

lucix-aws (approved):

service/internal/integrationtest/s3/checksum_test.go
```diff
@@ -92,23 +98,34 @@ func TestInteg_ObjectChecksums(t *testing.T) {
 					AlgorithmsUsed: []string{"CRC32C"},
 				},
 			},
+			"preset crc64 checksum": {
+				params: &s3.PutObjectInput{
+					Body:              strings.NewReader("Hello, precomputed checksum!"),
+					ChecksumCRC64NVME: aws.String("uxBNEklueLQ="),
+				},
+				expectPayload: []byte("Hello, precomputed checksum!"),
+				expectComputedChecksums: &s3.ComputedInputChecksumsMetadata{
```
Madrigal, line 107: nit: this is a bit of a misnomer since we don't really calculate anything, but ¯\_(ツ)_/¯

Madrigal (approved):

service/internal/integrationtest/s3/checksum_test.go
wty-Bryant: I just got that checksum from previous old manual test :(

RJKeevil:
Hello @wty-Bryant, this worked perfectly for me for single part gets, but i still get the warnings for multipart gets.  With a debugger i can see that the CRC32 header is returned for small files but not for large/multipart files, triggering the warning.  I am just using a library that calls `_, err = downloader.Download(ctx, writer, input)` and cannot see any missing options etc that would enable checksums in multipart, is it possible to add a larger file to your tests to confirm?

anishsana:
This change is causing my application to constantly throw the following warning -
```
SDK 2025/01/27 20:27:25 WARN Response has no supported checksum. Not validating response payload.
SDK 2025/01/27 20:27:25 WARN Response has no supported checksum. Not validating response payload.
SDK 2025/01/27 20:27:25 WARN Response has no supported checksum. Not validating response payload.
SDK 2025/01/27 20:27:25 WARN Response has no supported checksum. Not validating response payload.
SDK 2025/01/27 20:27:25 WARN Response has no supported checksum. Not validating response payload.
```
Is there any way to disable this validation to prevent the application from throwing this error?

RJKeevil:
@anishsana you probably need to set the env var AWS_RESPONSE_CHECKSUM_VALIDATION="WHEN_REQUIRED" in the meantime

PR #2987, opened 2025-01-28
move transfer manager v2 integration tests to within module

lucix-aws:
As written, the `service/internal/integrationtest` package depends on the s3 transfer manager module for its newly-added integration tests (as it should) - but our release tooling won't let us disable releases for any module that other things depend on. Since we need to put `feature/s3/transfermanager` into main, with releases disabled (to cut beta tags) - we need to remove that dependency.

Integration tests still work:
```
[feat-transfer-manager-v2-moveintegtests:~/git/aws-sdk-go-v2/feature/s3/transfermanager]$ go test -tags integration ./... -count=1
?       github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/internal/testing        [no test files]
?       github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/types   [no test files]
ok      github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager 17.072s
```

### DIFF
--- a/feature/s3/transfermanager/setup_integ_test.go
+++ b/feature/s3/transfermanager/setup_integ_test.go
@@ -220,6 +220,8 @@ type putObjectTestData struct {
 
 // UniqueID returns a unique UUID-like identifier for use in generating
 // resources for integration tests.
+//
+// TODO: duped from service/internal/integrationtest, remove after beta.
 func UniqueID() string {
 	uuid := make([]byte, 16)
 	io.ReadFull(rand.Reader, uuid)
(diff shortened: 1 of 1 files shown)


wty-Bryant (approved):

feature/s3/transfermanager/setup_integ_test.go
```diff
@@ -257,3 +266,159 @@ func testPutObject(t *testing.T, bucket string, testData putObjectTestData, opts
 		t.Errorf("expect %s, got %s", e, a)
 	}
 }
+
+const expressAZID = "usw2-az3"
```
Madrigal, line 270: should we TODO to move these back to shared once we merge this back into main?

feature/s3/transfermanager/api_op_PutObject_integ_test.go
```diff
@@ -1,7 +1,7 @@
 //go:build integration
 // +build integration
 
-package s3transfermanager
+package transfermanager
```
Madrigal, line 4: is the rename required or just something we prefer?

Madrigal (approved):

feature/s3/transfermanager/api_op_PutObject_integ_test.go
lucix-aws: required, this package is `transfermanager`

PR #2988, opened 2025-01-28
beta: feature/s3/transfermanager (S3 transfer manager v2)

lucix-aws:
* Adds current v2 transfermanager into main so we can start cutting beta tags
* Configures the module to not release via `modman.toml`
* Skip all `make` + `eachmodule` actions for the module, primarily to not run integration tests

The integration tests fail right now because of the recent S3 object integrity changes. So, integration tests in _this_ PR should pass, since the module is being skipped.

Verified that `make release` does NOT tag the module.


Madrigal (commented):
These changes (minus the modman and skip module) have been reviewed elsewhere, so approving as soon as integ tests pass

wty-Bryant (approved):

PR #2996, opened 2025-01-31
Feat S3 Transfer Manager v2 GetObject/DownloadObject

wty-Bryant:
Implement v2 s3 transfer manager's GetObject/DownloadObject api bound to single union client which mimics normal service client's initialization and api call. User could now download object sequentially with GetObject or asynchronously with DownloadObject.

Test: passed unit test for GetObject and DownloadObject, passed integration test of uploading object via s3 client and validating its content after downloading through v2 transfer manager.

### DIFF
--- a/feature/s3/transfermanager/api_op_GetObject.go
+++ b/feature/s3/transfermanager/api_op_GetObject.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math"
 	"strconv"
 	"strings"
 	"sync"
@@ -645,12 +644,12 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 			return nil, err
 		}
 
-		partsCount := int32(math.Max(float64(aws.ToInt32(out.PartsCount)), 1))
-		partSize := int64(math.Max(float64(aws.ToInt64(out.ContentLength)), 1))
-		sectionParts := int32(math.Max(1, float64(g.options.GetBufferSize/partSize)))
+		partsCount := max(aws.ToInt32(out.PartsCount), 1)
+		partSize := max(aws.ToInt64(out.ContentLength), 1)
+		sectionParts := int32(max(1, g.options.GetBufferSize/partSize))
 		capacity := sectionParts
 		g.r.setPartSize(partSize)
-		g.r.setCapacity(int32(math.Min(float64(capacity), float64(partsCount))))
+		g.r.setCapacity(min(capacity, partsCount))
 		g.r.setPartsCount(partsCount)
 
 		ch := make(chan getChunk, g.options.Concurrency)
@@ -666,7 +665,7 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 			}
 
 			if g.r.getRead() == capacity {
-				capacity = int32(math.Min(float64(capacity+sectionParts), float64(partsCount)))
+				capacity = min(capacity+sectionParts, partsCount)
 				g.r.setCapacity(capacity)
 			}
 
@@ -710,10 +709,10 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 		total := g.totalBytes - g.offset
 
 		partsCount := int32((total-1)/g.options.PartSizeBytes + 1)
-		sectionParts := int32(math.Max(1, float64(g.options.GetBufferSize/g.options.PartSizeBytes)))
+		sectionParts := int32(max(1, g.options.GetBufferSize/g.options.PartSizeBytes))
 		capacity := sectionParts
 		g.r.setPartSize(g.options.PartSizeBytes)
-		g.r.setCapacity(int32(math.Min(float64(capacity), float64(partsCount))))
+		g.r.setCapacity(min(capacity, partsCount))
 		g.r.setPartsCount(partsCount)
 
 		ch := make(chan getChunk, g.options.Concurrency)
@@ -729,7 +728,7 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 			}
 
 			if g.r.getRead() == capacity {
-				capacity = int32(math.Min(float64(capacity+sectionParts), float64(partsCount)))
+				capacity = min(capacity+sectionParts, partsCount)
 				g.r.setCapacity(capacity)
 			}
 
@@ -786,7 +785,7 @@ func (g *getter) singleDownload(ctx context.Context, clientOptions ...func(*s3.O
 		return nil, err
 	}
 
-	g.r.setPartSize(int64(math.Max(1, float64(aws.ToInt64(out.ContentLength)))))
+	g.r.setPartSize(max(1, aws.ToInt64(out.ContentLength)))
 	g.r.setCapacity(1)
 	g.r.setPartsCount(1)
 	g.r.ch <- outChunk{body: bytes.NewReader(buf), length: aws.ToInt64(out.ContentLength)}
@@ -881,7 +880,7 @@ func (g *getter) getDownloadRange() (int64, int64) {
 // byteRange returns an HTTP Byte-Range header value that should be used by the
 // client to request a chunk range.
 func (g *getter) byteRange() string {
-	return fmt.Sprintf("bytes=%d-%d", g.pos, int64(math.Min(float64(g.totalBytes-1), float64(g.pos+g.options.PartSizeBytes-1))))
+	return fmt.Sprintf("bytes=%d-%d", g.pos, min(g.totalBytes-1, g.pos+g.options.PartSizeBytes-1))
 }
 
 func (g *getter) getErr() error {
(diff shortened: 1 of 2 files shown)


feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,940 @@
...
+
+	d.setOutput(output)
+	d.pos += output.ContentLength
+	return output
+}
+
+// downloadChunk downloads the chunk from s3
+func (d *downloader) downloadChunk(ctx context.Context, chunk dlchunk, clientOptions ...func(*s3.Options)) (*GetObjectOutput, error) {
+	params := d.in.mapGetObjectInput(!d.options.DisableChecksumValidation)
+	if chunk.part != 0 {
```
Madrigal, line 750: is 0 also a valid part for "get the first part of this multi-part object"?

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,940 @@
...
+			return
+		}
+	} else {
+		parts := strings.Split(*resp.ContentRange, "/")
+
+		total := int64(-1)
+		var err error
+		totalStr := parts[len(parts)-1]
+
+		if totalStr != "*" {
```
Madrigal, line 862: Under which scenarios will this be `"*"`? Also, when this happens, `total` would be set to `-1`. Is that the desired behavior, or should we throw an error?

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,940 @@
...
+//
+// Additional functional options can be provided to configure the individual
+// download. These options are copies of the original Options instance, the client of which GetObject is called from.
+// Modifying the options will not impact the original Client and Options instance.
+func (c *Client) GetObject(ctx context.Context, input *GetObjectInput, opts ...func(*Options)) (*GetObjectOutput, error) {
+	i := downloader{in: input, options: c.options.Copy()}
+	for _, opt := range opts {
+		opt(&i.options)
+	}
+	i.options.Concurrency = 1
```
Madrigal, line 550: Why are we setting Concurrency to 1?

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,940 @@
...
+
+		if output.PartsCount > 1 {
+			partSize := output.ContentLength
+			ch := make(chan dlchunk, d.options.Concurrency)
+			for i := 0; i < d.options.Concurrency; i++ {
+				d.wg.Add(1)
+				go d.downloadPart(ctx, ch, clientOptions...)
+			}
+
+			for i := int32(2); i <= output.PartsCount; i++ {
```
Madrigal, line 625: why do we start at 2 here?

feature/s3/transfermanager/api_client.go
```diff
@@ -20,6 +20,8 @@ const defaultMultipartUploadThreshold = 1024 * 1024 * 16
 // using PutObject().
 const defaultTransferConcurrency = 5
 
+const defaultPartBodyMaxRetries = 3
```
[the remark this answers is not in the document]
wty-Bryant: yes, the latest SEP doesn't have such retry limitation in detail

feature/s3/transfermanager/api_op_GetObject.go
wty-Bryant: `GetObject` here mimics `s3.GetObject` to get an object's chunks and write to output body in sequence, thus there should be only one goroutine to write chunks to output in order

feature/s3/transfermanager/api_op_GetObject.go
wty-Bryant: part 1 should already be written to output when we call `getChunk()` to get total parts above

feature/s3/transfermanager/api_op_GetObject.go
wty-Bryant: Forgot to remove that check, s3 will always return content length

feature/s3/transfermanager/api_op_GetObject.go
lucix-aws: You'll need to benchmark it but I don't think we can do it this way. Doing multiple range GETs w/ concurrency 1 is going to be actively worse than just plain GetObject, in theory, because you're adding the overhead of multiple round-trips.

Ideally we dispatch concurrent range GETs and then we just arrange their output streams in sequence behind a single `io.ReadCloser` to the user.

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,927 @@
...
+
+// DownloadObject downloads an object from S3, intelligently splitting large
+// files into smaller parts/ranges according to config and getting them in parallel across
+// multiple goroutines. You can configure the download type, chunk size and concurrency
+// through the Options parameters.
+//
+// Additional functional options can be provided to configure the individual
+// download. These options are copies of the original Options instance, the client of which DownloadObject is called from.
+// Modifying the options will not impact the original Client and Options instance.
+func (c *Client) DownloadObject(ctx context.Context, w io.WriterAt, input *GetObjectInput, opts ...func(*Options)) (*GetObjectOutput, error) {
```
lucix-aws, line 559: Give this its own file.

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,927 @@
...
+	input.SSECustomerKey = nzstring(i.SSECustomerKey)
+	input.SSECustomerKeyMD5 = nzstring(i.SSECustomerKeyMD5)
+	input.VersionId = nzstring(i.VersionID)
+
+	return input
+}
+
+// GetObjectOutput represents a response from GetObject() or DownloadObject() call. It contains common fields
+// of s3 GetObject output
+type GetObjectOutput struct {
```
lucix-aws, line 275: We can't share these types between get and download. There's already divergence between what each one needs - as written, you have `Body` in the output for DownloadObject, which is functionally useless.

lucix-aws: (the same is true for the input)

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,927 @@
...
+	//
+	// [Checking object integrity]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
+	ChecksumSHA1 string
+
+	// The base64-encoded, 256-bit SHA-256 digest of the object. This will only be
+	// present if it was uploaded with the object. For more information, see [Checking object integrity]in the
+	// Amazon S3 User Guide.
+	//
+	// [Checking object integrity]: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html
+	ChecksumSHA256 string
```
lucix-aws, line 318: Are you actually verifying checksums in your GetObject implementation?

feature/s3/transfermanager/api_op_GetObject.go
wty-Bryant: Yea, we need `io.WriterAt` in `DownloadObjectInput` as dst, but not for `GetObject`

feature/s3/transfermanager/api_op_GetObject.go
wty-Bryant: Yes, the checksum validation is enabled by default unless user specify that in their input

wty-Bryant: The validation skip log also presented when multi-part download an large object uploaded without checksum

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,893 @@
...
+		}
+
+		// Assign work
+		for g.getErr() == nil {
+			if g.pos >= total {
+				break // We're finished queuing chunks
+			}
+
+			// Queue the next range of bytes to read.
+			ch <- getChunk{w: g.w, start: g.pos - g.offset, withRange: g.byteRange()}
```
lucix-aws, line 636: So as far as I can tell, you're now doing concurrent downloads but you're buffering the entire object in memory before returning it to the caller - which is all well and good until someone goes to try and pull a multi-gigabyte object.

We're basically going to have to return a reader that "composes" the response bodies from inner getPart requests in order. That will allow you to stream the response back to the caller without that extra memory pressure. You might have a "scratch" buffer for that internal composition but naturally it won't be the full object size.

feature/s3/transfermanager/options.go
```diff
@@ -16,7 +19,7 @@ type Options struct {
 	MultipartUploadThreshold int64
 
 	// Option to disable checksum validation for download
-	DisableChecksum bool
+	DisableChecksumValidation bool
```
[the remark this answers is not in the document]
wty-Bryant: This client field was listed in SEP as boolean, but we can change that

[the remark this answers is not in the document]
wty-Bryant: Okay I try to change that to `ResponseChecksumValidation` enum, but since s3 client in transfer manager is just an interface, we could not configure normal s3 checksum options like what we do for s3 service client, I think it's ok to just keep that bool which is only used to set request checksum mode

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,893 @@
...
+			)
+		}}
+
+	if g.in.PartNumber > 0 {
+		return g.singleDownload(ctx, clientOptions...)
+	}
+
+	var output *GetObjectOutput
+	if g.options.MultipartDownloadType == types.MultipartDownloadTypePart {
+		if g.in.Range != "" {
```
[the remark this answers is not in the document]
lucix-aws: This doesn't appear to have been addressed.

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+			Bucket:     aws.String(g.in.Bucket),
+			Key:        aws.String(g.in.Key),
+			PartNumber: aws.Int32(1),
+		}, clientOptions...)
+		if err != nil {
+			g.r.setErr(err)
+			return nil, err
+		}
+
+		partsCount := int32(math.Max(float64(aws.ToInt32(out.PartsCount)), 1))
```
lucix-aws, line 648: Don't use math.Max for this, it only supports float64 (hence why you had to do the conversion). It's more idiomatic to just define a local `max()` and use that, we do it in various places.

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+// GetObjectInput represents a request to the GetObject() or DownloadObject() call. It contains common fields
+// of s3 GetObject input
+type GetObjectInput struct {
+	// Bucket where the object is downloaded from
+	Bucket string
+
+	// Key of the object to get.
+	Key string
+
+	Reader *ConcurrentReader
```
lucix-aws, line 47: Why does the caller have to pass this? Shouldn't we just be creating one?

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+	//
+	// [PutBucketLifecycleConfiguration]: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLifecycleConfiguration.html
+	Expiration string
+
+	// The date and time at which the object is no longer cacheable.
+	//
+	// Deprecated: This field is handled inconsistently across AWS SDKs. Prefer using
+	// the ExpiresString field which contains the unparsed value from the service
+	// response.
+	Expires time.Time
```
lucix-aws, line 379: We don't need to address this now but I question whether we should have the `time.Time` version of this in the transfermanager. We should be using these APIs as an opportunity to get rid of old baggage like this.

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+
+	length int64
+	cur    int64
+}
+
+// ConcurrentReader receives object parts from working goroutines, composes those chunks in order and read
+// to user's buffer. ConcurrentReader limits the max number of chunks it could receive and read at the same
+// time so getter won't send following parts' request to s3 until user reads all current chunks, which avoids
+// too much memory consumption when caching large object parts
+type ConcurrentReader struct {
```
lucix-aws, line 923: Separate file?

lucix-aws: This should be an implementation detail and shouldn't need to be exported.

lucix-aws (approved):
We've discussed the nuances of exposing the GetObject reader offline, you're free to address that now or ship and address in a followup.

feature/s3/transfermanager/api_op_GetObject.go
Madrigal: since Go 1.21 there's also a built-in `max` function [ref](https://go.dev/ref/spec#Min_and_max)

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+			if g.getErr() != nil {
+				break
+			}
+
+			if g.r.getRead() == capacity {
+				capacity = int32(math.Min(float64(capacity+sectionParts), float64(partsCount)))
+				g.r.setCapacity(capacity)
+			}
+
+			if i == capacity {
```
Madrigal, line 673: if we hit this case, wouldn't we deadlock if `g.r.getRead() != capacity`?

wty-Bryant: `g.r.read` would be increased by reader in another goroutine, that's why we continue the loop to re-check `g.r.getRead()` at line 668

feature/s3/transfermanager/setup_integ_test.go
```diff
@@ -269,6 +278,100 @@ func testPutObject(t *testing.T, bucket string, testData putObjectTestData, opts
 	}
 }
 
+func testGetObject(t *testing.T, bucket string, testData getObjectTestData) {
+	key := UniqueID()
+
+	_, err := s3Client.PutObject(context.Background(),
```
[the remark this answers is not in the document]
wty-Bryant: those objects should be deleted when cleaning bucket

feature/s3/transfermanager/downloadobject_test.go
```diff
@@ -0,0 +1,458 @@
...
+		downloadRange     string
+		expectInvocations int
+		expectRanges      []string
+		partNumber        int32
+		partsCount        int32
+		expectParts       []int32
+		expectErr         string
+		dataValidationFn  func(*testing.T, *types.WriteAtBuffer)
+	}{
+		"range download in order": {
```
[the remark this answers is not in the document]
wty-Bryant: yea, for any range download where data length is not integral multiple of part size (e.g. `range download in order`)

feature/s3/transfermanager/concurrent_reader_test.go
```diff
@@ -0,0 +1,163 @@
...
+			concurrency:  3,
+			sectionParts: 6,
+		},
+		"multiple goroutines": {
+			partSize:     10,
+			partsCount:   1000,
+			concurrency:  5,
+			sectionParts: 6,
+		},
+		"multiple goroutines with only one section": {
```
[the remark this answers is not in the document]
wty-Bryant: yea, `sectionParts` means the number of parts in a section, since total parts count here is 6, there's only 1 section totally

feature/s3/transfermanager/internal/testing/upload.go
```diff
@@ -1,193 +0,0 @@
-package testing
```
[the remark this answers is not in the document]
wty-Bryant: I think so, we just leave it here and make any change if there's any checksum related update

feature/s3/transfermanager/getobject_test.go
```diff
@@ -0,0 +1,446 @@
...
+			downloadRange:     "bytes=0-100",
+			partsCount:        3,
+			expectInvocations: 1,
+			dataValidationFn: func(t *testing.T, bytes []byte) {
+				if e, a := "123", string(bytes); e != a {
+					t.Errorf("expect %q response, got %q", e, a)
+				}
+			},
+		},
+		"parts download with part number input": {
```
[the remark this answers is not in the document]
wty-Bryant: Nope, this test case verifies that we single download the whole object anytime user passes a part number

feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -0,0 +1,1106 @@
...
+	//
+	// This functionality is not supported for directory buckets.
+	WebsiteRedirectLocation string
+
+	// Metadata pertaining to the operation's result.
+	ResultMetadata smithymiddleware.Metadata
+}
+
+func (o *GetObjectOutput) mapFromGetObjectOutput(out *s3.GetObjectOutput, checksumMode s3types.ChecksumMode) {
+	o.AcceptRanges = aws.ToString(out.AcceptRanges)
```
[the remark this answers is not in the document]
wty-Bryant: added

PR #3000, opened 2025-02-04
Track accountID endpoint mode in user-agent

lucix-aws:
Adds the following user agent feature IDs:

```
UserAgentFeatureAccountIDModePreferred = "P"
UserAgentFeatureAccountIDModeDisabled  = "Q"
UserAgentFeatureAccountIDModeRequired  = "R"
```

Rather than writing a test script for this and demonstrating its output, I figured I'd just add a unit test. Why run something once to validate it when you could just have it validate forevermore?

### DIFF
--- a/.changelog/1c8f914e6b9a4aef9418a60e7fe62de2.json
+++ b/.changelog/1c8f914e6b9a4aef9418a60e7fe62de2.json
@@ -1,0 +1,8 @@
+{
+    "id": "1c8f914e-6b9a-4aef-9418-a60e7fe62de2",
+    "type": "feature",
+    "description": "Track AccountID endpoint mode in user-agent.",
+    "modules": [
+        "service/dynamodb"
+    ]
+}


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIdEndpointModeUserAgent.java
```diff
@@ -0,0 +1,69 @@
...
+                RuntimeClientPlugin.builder()
+                        .servicePredicate(AccountIDEndpointRouting::hasAccountIdEndpoints)
+                        .registerMiddleware(MIDDLEWARE)
+                        .build()
+        );
+    }
+
+    @Override
+    public void writeAdditionalFiles(GoCodegenContext ctx) {
+        if (!hasAccountIdEndpoints(ctx.model(), ctx.settings().getService(ctx.model()))) {
```
Madrigal, line 41: do we need this if we have the service predicate?

lucix-aws: We could remove it but then it would just generate in every service and force a bunch of module updates.

Madrigal: Ahh yeah, forgot that the predicate doesn't control whether we write something on the file or not. Yes, then this is perfect, let's not write superfluous code

Madrigal (approved):

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/AccountIdEndpointModeUserAgent.java
```diff
@@ -0,0 +1,69 @@
...
+            return;
+        }
+
+        ctx.writerDelegator().useFileWriter("api_client.go", ctx.settings().getModuleName(), goTemplate("""
+                $aws:D $awsMiddleware:D
```
wty-Bryant, line 46: love new format of import dependencies

wty-Bryant (approved):

Issue #2383, opened 2023-11-21
DynamoDB AttributeValue's MarshallMap() Does Not Support time.Time Correctly


kmacmcfarlane:
### Describe the bug

I am seeing that it's not possible to use `omitempty` on `time.Time` struct fields with `attributevalue.MarshalMap()` from `github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue`.

This behavior was observed on the latest version: `github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue` v1.12.3

### Expected Behavior

When `MarshalMap()` is supplied with a struct containing a `time.Time` field with the tag `omitempty` whos value is the zero value, the output map should not contain a key for that struct field.

### Current Behavior

Output attribute value map includes a key and value for the field that should be omitted.

### Reproduction Steps

repro_test.go
```
package service

import (
	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
	"time"
)

var _ = Describe("Serialize Struct", func() {

	Context("Struct with omitempty on a time.Time field", func() {
		It("Does not serialize the field", func() {

			e := struct {
				Created time.Time `dynamodbav:"created,omitempty"`
			}{
				Created: time.Time{},
			}
			av, err := attributevalue.MarshalMap(e)
			Ω(err).ShouldNot(HaveOccurred())

			Ω(av).ShouldNot(HaveKey("created"))
		})
	})
})
```
service_test_suite.go
```
package service_test

import (
	"testing"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

func TestService(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Service Suite")
}
```

### Possible Solution

As far as I've been able to reason by stepping through the repro above in a debugger, there should probably be a check for `fieldTag.OmitEmpty` in this code block in `encode.go` (just like the `fieldTag.AsUnixTime` tag is checked).

https://github.com/aws/aws-sdk-go-v2/blob/3bd97c063d962a34ca496720a3ce00ef4affe5fd/feature/dynamodb/attributevalue/encode.go#L478

### AWS Go SDK V2 Module Versions Used

```
        github.com/aws/aws-lambda-go v1.34.1
        github.com/aws/aws-sdk-go v1.44.128
[...]

PR #3004, opened 2025-02-06
Fix for #2383

badu:
DynamoDB AttributeValue's MarshallMap() Does Not Support time.Time Correctly

Closes https://github.com/aws/aws-sdk-go-v2/issues/2383

Added a EncoderOption called OmitEmptyTime, so we don't introduce breaking changes.

### DIFF
--- a/.changelog/adfe779a-b5b4-44bd-8e23-5cdb4db5aef8.json
+++ b/.changelog/adfe779a-b5b4-44bd-8e23-5cdb4db5aef8.json
@@ -1,7 +1,7 @@
 {
   "id": "adfe779a-b5b4-44bd-8e23-5cdb4db5aef8",
   "type": "bugfix",
-  "collapse": true,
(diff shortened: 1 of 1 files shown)


lucix-aws (approved):
LGTM, JFYI - integration tests will fail in forks, don't worry about those.

.changelog/adfe779a-b5b4-44bd-8e23-5cdb4db5aef8.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "adfe779a-b5b4-44bd-8e23-5cdb4db5aef8",
+  "type": "bugfix",
+  "collapse": true,
```
lucix-aws, line 4: We only need to collapse for sweeping changes that touched many modules.

```suggestion
  "collapse": false,
```

PR #3008, opened 2025-02-13
Track credential providers via User-Agent Feature ids

Madrigal:
We already have a way to track feature ids via the user agent middleware. This PR adds a way to track which credential provider(s) have been used for getting the credentials on a certain service.

These features have been defined on a SEP, and a single provider can have multiple features depending on how the credentials were resolved

This is implemented by having the `config` object track which credential providers have been used, since it's the main point of entry where credentials are being loaded via `config.LoadDefaultConfig`.  

## Why not use the existing ProviderName field on aws.Credentials?
Because of 2 reasons

1. The same provider can be used for multiple features we want to track, so we can't tell just by the provider which feature to add
2. Adding this to the credentials object would make it more difficult to emit these metrics, since they'd have to be added only when credentials are resolved.

## Why not extend the interface of credential provider to add this new method?
Because that would be a breaking change, since any credential provider implemented by a 3rd party would no longer satisfy the interface.

## Why some providers have credential chain passed on the options struct instead of being part of the main provider?
Mainly due to backwards compatibility. The ones where options are used have a `New()` method on the package, and this is the main way to create a new provider. Adding a new field would change the signature.

## How do we handle credential refresh?
Other SDKs have the concept of credential chain, where a provider is tried, and if it fails, there is a chain of credential providers that gets retried until valid credentials are found. In Go, we resolve credentials when we create the config object, but after that no changes are made to the credentials

## Are there any known gaps?
Main thing is capturing intermediate credentials. We'd like to track the calls made to SSO to see which provider chain has been used. However, SSO doesn't use the `aws.Credentials` object and instead uses an access token that is set directly into it, so this whole process fails and we don't set any feature on the user agent.

### DIFF
--- a/service/medialive/api_op_StartMonitorDeployment.go
+++ b/service/medialive/api_op_StartMonitorDeployment.go
@@ -176,9 +176,6 @@ func (c *Client) addOperationStartMonitorDeploymentMiddlewares(stack *middleware
 	if err = addCredentialSource(stack, options); err != nil {
 		return err
 	}
-	if err = addIdempotencyToken_opStartMonitorDeploymentMiddleware(stack, options); err != nil {
-		return err
-	}
 	if err = addOpStartMonitorDeploymentValidationMiddleware(stack); err != nil {
 		return err
 	}


aws/credentials.go
```diff
@@ -70,6 +70,56 @@ func (AnonymousCredentials) Retrieve(context.Context) (Credentials, error) {
...
+	// CredentialsStsSessionToken credentials resolved from STS using a session token 	S
+	CredentialsStsSessionToken
+	// CredentialsProfile  credentials resolved from a config file(s) profile with static credentials
+	CredentialsProfile
+	// CredentialsProfileSourceProfile credentials resolved from a source profile in a config file(s) profile
+	CredentialsProfileSourceProfile
+	// CredentialsProfileNamedProvider credentials resolved from a named provider in a config file(s) profile (like EcsContainer)
+	CredentialsProfileNamedProvider
+	// CredentialsProfileStsWebIDToken  credentials resolved from configuration for assuming a role with STS using web identity token in a config file(s) profile
+	CredentialsProfileStsWebIDToken
```
lucix-aws, line 104: STS

aws/credentials.go
```diff
@@ -70,6 +70,56 @@ func (AnonymousCredentials) Retrieve(context.Context) (Credentials, error) {
...
+	// CredentialsProfileSso credentials resolved from an SSO session in a config file(s) profile
+	CredentialsProfileSso
```
lucix-aws, line 106: SSO

aws/credentials.go
```diff
@@ -125,6 +175,13 @@ type CredentialsProvider interface {
 	Retrieve(ctx context.Context) (Credentials, error)
 }
 
+// CredentialProviderChain allows any credential provider to track
+// the chain where they were sourced. For example, if the credentials came from a
+// call to a role specified in the profile, this method will give the whole precedence chain
+type CredentialProviderChain interface {
```
lucix-aws, line 181: I don't think we can name it like this - `ProviderChain` really implies that this is a provider _composed_ of multiple providers, i.e. a chain that tries providers until one resolves.

aws/config.go
```diff
@@ -31,6 +31,9 @@ type Config struct {
 	// sources, and resolve credentials using the SDK's default credential chain.
 	Credentials CredentialsProvider
 
+	// Stores which source(s) were used to get the credentials.
+	CredentialChain []CredentialSource
```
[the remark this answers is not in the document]
lucix-aws: Looking more at the code I have a better idea of how this is used, but this is really internal state we should be keeping track off only when resolving credentials, I'm not seeing why it needs to be an exported field on config.

credentials/endpointcreds/provider.go
```diff
@@ -191,3 +194,11 @@ func (p *Provider) resolveAuthToken() (string, error) {
 
 	return authToken, nil
 }
+
+// CredentialChain returns the credential chain that was used to construct this provider
+func (p *Provider) CredentialChain() []aws.CredentialSource {
+	if p.options.CredentialChain == nil {
+		return []aws.CredentialSource{aws.CredentialsProcess}
```
lucix-aws, line 201: Double-check these, I don't think this is the right enum value.

wty-Bryant (changes requested):

aws/credentials.go
Madrigal: Changed, also changed all other instances. Kind of surprised this wasn't picked up by vet, but I guess it doesn't know that STS is a noun

aws/credentials.go
Madrigal: Changed

aws/credentials.go
Madrigal: Yeah agree. Discussed offline and changed

aws/config.go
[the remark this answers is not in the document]
Madrigal: moved to the context object inside `resolve_credentials`. Note that some internal method had to be modified to return `context`, since sometimes you need to pass information about a credentials resolver up the stack

codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/CredentialSourceFeatureTrackerGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    public void writeAdditionalFiles(GoCodegenContext ctx) {
+        if (!AwsSignatureVersion4.hasSigV4X(ctx.model(), ctx.settings().getService(ctx.model()))) {
+            return;
+        }
+
+        ctx.writerDelegator().useFileWriter("api_client.go", ctx.settings().getModuleName(), goTemplate("""
+                $aws:D $awsMiddleware:D
+
+                type SetCredentialSourceMiddleware struct {
```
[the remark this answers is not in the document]
Madrigal: changed to be unexported

credentials/stscreds/web_identity_provider.go
```diff
@@ -64,6 +64,9 @@ type WebIdentityRoleOptions struct {
 	// want to use as managed session policies.  The policies must exist in the
 	// same account as the role.
 	PolicyARNs []types.PolicyDescriptorType
+
+	// The chain of providers that was used to create this provider
+	CredentialChain []aws.CredentialSource
```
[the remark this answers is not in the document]
Madrigal: Changed docs on all providers to indicate that they shouldn't be modifying this value directly

aws/credentials.go
```diff
@@ -70,6 +70,56 @@ func (AnonymousCredentials) Retrieve(context.Context) (Credentials, error) {
...
 }
 
+// CredentialSource is the source of the credential provider.
+// A provider can have multiple credential sources: For example, a provider that reads a profile, calls ECS to
+// get credentials and then assumes a role using STS will have all these as part of its provider chain.
+type CredentialSource int
+
+const (
+	// CredentialsUndefined is the sentinel zero value
+	CredentialsUndefined CredentialSource = iota
```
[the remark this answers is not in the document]
Madrigal: I like it more as well. Renaming

credentials/static_provider.go
```diff
@@ -21,7 +21,16 @@ func (*StaticCredentialsEmptyError) Error() string {
 // A StaticCredentialsProvider is a set of credentials which are set, and will
 // never expire.
 type StaticCredentialsProvider struct {
-	Value aws.Credentials
+	Value  aws.Credentials
+	Source []aws.CredentialSource
```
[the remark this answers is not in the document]
Madrigal: This has been changed on the most recent version

lucix-aws (approved):

wty-Bryant (approved):

PR #3025, opened 2025-03-03
add test that verifies SRA order of operations

lucix-aws:
Middleware snapshot tests, on their own, are not sufficient to prevent regressions, because a snapshot test only displays and tracks changes to a state, it does not catch when changes to that state are incorrect. To address this gap, we are implementing a new new test that verifies correct operation ordering.

[internal: addresses AI 2349272]


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/customization/SraOperationOrderTest.java
```diff
@@ -0,0 +1,91 @@
...
+                		*stack = *inner
+                		return errTestReturnEarly
+                	}
+                }
+                """, SmithyGoDependency.ERRORS, SmithyGoDependency.SMITHY_MIDDLEWARE);
+    }
+
+    private GoWriter.Writable renderTest(String operationName) {
+        return goTemplate("""
+                $1D $2D $3D $4D $5D $6D
```
Madrigal, line 45: man I love that we can just import things and then let Go fmt sort things later.

Madrigal (approved):
Took a peek an protocol tests and a service team. Generated tests look good

PR #3029, opened 2025-03-04
Add credential refresh retry test

lucix-aws:
Adds a "credential refresh" test in the kitchensink service, that explicitly verifies that credentials are re-retrieved when an operation retries.

### DIFF
--- a/internal/kitchensinktest/credential_refresh_retry_test.go
+++ b/internal/kitchensinktest/credential_refresh_retry_test.go
@@ -1,0 +1,60 @@
+package kitchensinktest
+
+import (
+	"context"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/internal/sdk"
+)
+
+func TestCredentialRefreshRetry(t *testing.T) {
+	sdk.NowTime = func() time.Time { return time.Unix(0, 0) }
+	defer func() {
+		sdk.NowTime = time.Now
+	}()
+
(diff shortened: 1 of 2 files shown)


internal/kitchensinktest/credential_refresh_retry_test.go
```diff
@@ -0,0 +1,60 @@
...
+		sdk.NowTime = time.Now
+	}()
+
+	credvalues := []aws.Credentials{
+		{AccessKeyID: "foo", SecretAccessKey: "bar"},
+		{AccessKeyID: "baz", SecretAccessKey: "qux"},
+	}
+	mhttp := &mockHTTP{
+		resps: []*http.Response{
+			{StatusCode: 500, Body: http.NoBody},
```
Madrigal, line 26: it's interesting to me that we retry by default on 500. I guess this is because of high availability in services?

internal/kitchensinktest/options.go
```diff
@@ -56,6 +62,9 @@ type Options struct {
 	// used over the deprecated EndpointResolver.
 	EndpointResolverV2 EndpointResolverV2
 
+	// Signature Version 4 (SigV4) Signer
+	HTTPSignerV4 HTTPSignerV4
```
Madrigal, line 66: Do we actually need to have a full auth scheme and v4 signer to test credentials refresh? Would this still work if we had static credentials?

lucix-aws: There's no credentials provider in options without sigv4 or sigv4a (note how the diff shows it was added here)

Madrigal (approved):

PR #3050, opened 2025-04-03
Feat Getobject Version Control

wty-Bryant:
This PR adds an ETag check from the initial GET response to confirm object parts are downloaded from the same version. If the object is modified during the multi-part download and no VersionId is provided in the request, the download will fail. Alternatively, users can still multipart download a specific version of a large object by passing VersionId in the request input. 
This fixes issue aws/aws-sdk-go#4986

### DIFF
--- a/feature/s3/manager/download.go
+++ b/feature/s3/manager/download.go
@@ -359,8 +359,8 @@ func (d *downloader) downloadChunk(chunk dlchunk) error {
 
 	// Get the next byte range of data
 	params.Range = aws.String(chunk.ByteRange())
-	if params.VersionId == nil && d.getETag() != "" {
-		params.IfMatch = aws.String(d.getETag())
+	if params.VersionId == nil && d.etag != "" {
+		params.IfMatch = aws.String(d.etag)
 	}
 
 	var n int64
@@ -469,13 +469,6 @@ func (d *downloader) setTotalBytes(resp *s3.GetObjectOutput) {
 
 		d.totalBytes = total
 	}
-}
-
-func (d *downloader) getETag() string {
-	d.m.Lock()
-	defer d.m.Unlock()
-
-	return d.etag
 }
 
 func (d *downloader) incrWritten(n int64) {


feature/s3/manager/download.go
```diff
@@ -401,6 +404,7 @@ func (d *downloader) tryDownloadChunk(params *s3.GetObjectInput, w io.Writer) (i
 		return 0, err
 	}
 	d.setTotalBytes(resp) // Set total if not yet set.
+	d.setETag(resp)
```
Madrigal, line 407: Shouldn't this be set only once instead of at each chunk?

wty-Bryant: yea, the func would only set it in first GET when etag is empty

Madrigal: I see, this is because of the mutex we added to setETag. Unfortunately this means all threads will have to wait for the lock since we use the same mutex as the downloader and they all try to call setETag when getting a result. Can we use something like [sync.OnceValue](https://pkg.go.dev/sync#OnceValue) to set this instead, so we don't have all go routines waiting on the lock?

feature/s3/manager/download.go
```diff
@@ -401,6 +405,9 @@ func (d *downloader) tryDownloadChunk(params *s3.GetObjectInput, w io.Writer) (i
 		return 0, err
 	}
 	d.setTotalBytes(resp) // Set total if not yet set.
+	d.once.Do(func() {
+		d.etag = aws.ToString(resp.ETag)
```
Madrigal, line 409: is eTag guaranteed to appear on the response? otherwise we will panic when doing `aws.ToString`

feature/s3/manager/download.go
```diff
@@ -464,6 +471,13 @@ func (d *downloader) setTotalBytes(resp *s3.GetObjectOutput) {
 	}
 }
 
+func (d *downloader) getETag() string {
+	d.m.Lock()
```
Madrigal, line 475: If we set this with `sync.Once`, we don't need to grab the lock to read the value

feature/s3/manager/download_test.go
```diff
@@ -522,6 +550,73 @@ func TestDownload_WithRange(t *testing.T) {
...
+	if e, a := 3, *invocations; e != a {
+		t.Errorf("expect %v API calls, got %v", e, a)
+	}
+
+	expectVersions := []string{"vid", "vid", "vid"}
+	if e, a := expectVersions, *versions; !reflect.DeepEqual(e, a) {
+		t.Errorf("expect %v version ids, got %v", e, a)
+	}
+
+	expectETags := []string{"", "", ""}
```
Madrigal, line 581: Not sure if the expected response is empty eTag, but this will likely still hold true: if a version id is set, the object should always have the same eTag

feature/s3/manager/download_test.go
```diff
@@ -577,6 +672,63 @@ func TestDownload_WithFailure(t *testing.T) {
 	}
 }
 
+func TestDownload_WithMismatch(t *testing.T) {
```
Madrigal, line 675: took me a bit to read this, but when you make a call to S3 with an `IfMatch` that doesn't match, you get back a `PreconditionFailed` error.

wty-Bryant: yea, this case just ensures downloader could react to the specific error

feature/s3/manager/download.go
wty-Bryant: Manually get object with/without version, and etag will always be returned even if the object is empty. `aws.ToString` would nil check and return empty string for nil input

feature/s3/manager/download_test.go
wty-Bryant: `expectEtags` is list of `IfMatch` captured from requests, in this case versionId is passed by user so no ETag would be set to `IfMatch`

Madrigal (approved):
LGTM as soon as CI passes

Issue #2900, opened 2024-11-15
feature/dynamodb/attributevalue needs a new major version to remove all of the opt-in flags that should be default


lucix-aws:
### Describe the feature

todo

### Use Case

todo

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

todo

### Go version used

todo

PR #3054, opened 2025-04-08
remove all of the opt-in flags that should be default

raduendv:
addresses #2900 - removes all mentioned opt-in flags and turns them on by default


lucix-aws:
We can't just break the API like this. The linked issue mentions that this would need to be a new major version of the module(s).

lucix-aws:
It also mentions that we can't do that right now due to issues it would bring with the internal release.

PR #3065, opened 2025-04-16
implement progress listener events for PutObject

lucix-aws:
Implements new `ProgressListener` event posting for `PutObject`.

This is mostly a drop-in add to the existing codebase, with the exception that intermediate part sizes now need a bit more plumbing throughout.

The new `singleObjectProgressEmitter` should be blindly reusable for `GetObject` / `DownloadObject`.

### DIFF
--- a/feature/s3/transfermanager/putobject_test.go
+++ b/feature/s3/transfermanager/putobject_test.go
@@ -553,6 +553,99 @@ func TestUploadZeroLenObject(t *testing.T) {
 	}
 }
 
+func TestProgressListener_SingleUpload_SeekableBody(t *testing.T) {
+	c, _, _ := s3testing.NewUploadLoggingClient(nil)
+
+	listener := &mockListener{}
+
+	var opts Options
+	opts.ProgressListeners.Register(listener)
+
+	mgr := New(c, opts)
+
+	body := "foobarbaz"
+	in := &PutObjectInput{
+		Bucket: "Bucket",
+		Key:    "Key",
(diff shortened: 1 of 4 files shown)


feature/s3/transfermanager/putobject_test.go
```diff
@@ -553,6 +553,99 @@ func TestUploadZeroLenObject(t *testing.T) {
 	}
 }
 
+func TestProgressListener_SingleUpload_SeekableBody(t *testing.T) {
```
Madrigal, line 556: we're missing a test for when the progress listener fails

feature/s3/transfermanager/putobject_test.go
```diff
@@ -553,6 +553,99 @@ func TestUploadZeroLenObject(t *testing.T) {
...
+	// 40MB / 8 (default part size) = 5 expected events
+	// we keep the input size a multiple of the part size so the intermediate
+	// byte counts are predictable
+	const eightMB = 1024 * 1024 * 8
+	listener.expectByteTransfers(t,
+		eightMB,
+		eightMB*2,
+		eightMB*3,
+		eightMB*4,
+		eightMB*5)
```
Madrigal, line 646: [nit] It'll be good to have a test for a size that has a different length trailing body, although this should be enough since we're just iterating the events received

Madrigal (approved):
minor comments

wty-Bryant (approved):

PR #3067, opened 2025-04-17
[do not merge] feat: support account id in imds / new profile configs

lucix-aws:
DO NOT MERGE: the new `-extended` metadata endpoint is not yet available.

* Support the new `*-extended/` credentials endpoint for IMDS, with automatic fallback to the old one
* Add support for `AWS_EC2_INSTANCE_PROFILE_NAME` and friends
* Retrieve account ID in IMDS calls (on both the new and legacy endpoint per SEP).
* Add implicit profile caching. This has actually been specified in older IMDS credential SEPs, but we never did it for some reason.
* Add new `disable_ec2_metadata` shared config setting.

### DIFF
--- a/config/shared_config.go
+++ b/config/shared_config.go
@@ -80,6 +80,7 @@ const (
 
 	ec2MetadataServiceEndpointKey = "ec2_metadata_service_endpoint"
 
+	ec2MetadataDisabledKey   = "disable_ec2_metadata"
 	ec2MetadataV1DisabledKey = "ec2_metadata_v1_disabled"
 
 	ec2InstanceProfileNameKey = "ec2_instance_profile_name"
@@ -265,6 +266,11 @@ type SharedConfig struct {
 	//
 	// ec2_metadata_service_endpoint=http://fd00:ec2::254
 	EC2IMDSEndpoint string
+
+	// Specifies if the EC2 IMDS service client is enabled.
+	//
+	// disable_ec2_metadata=true
+	EC2IMDSClientEnableState imds.ClientEnableState
 
 	// Specifies that IMDS clients should not fallback to IMDSv1 if token
 	// requests fail.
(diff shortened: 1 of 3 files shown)


config/resolve_credentials_test.go
```diff
@@ -750,6 +757,103 @@ func TestResolveCredentialsEcsContainer(t *testing.T) {
...
+			defer cleanupFn()
+
+			// setupCredentialsEndpoints sets this above and then we hold onto
+			// it for this test
+			ec2MetadataURL := os.Getenv("AWS_EC2_METADATA_SERVICE_ENDPOINT")
+
+			restoreEnv := awstesting.StashEnv()
+			defer awstesting.PopEnv(restoreEnv)
+
+			os.Setenv("AWS_EC2_METADATA_SERVICE_ENDPOINT", ec2MetadataURL)
```
Madrigal, line 816: nit: use `t.Setenv` instead

config/env_config.go
```diff
@@ -916,3 +923,11 @@ func (c EnvConfig) GetS3DisableExpressAuth() (value, ok bool) {
 
 	return *c.S3DisableExpressAuth, true
 }
+
+func (c EnvConfig) getEC2InstanceProfileName() (string, bool, error) {
+	if len(c.EC2InstanceProfileName) == 0 {
+		return "", false, nil
```
Madrigal, line 929: It was discussed to err on this rather than leaving it empty. My main concern would be that someone would want to set this to something on their shell, like `export AWS_EC2_INSTANCE_PROFILE_NAME=$(some_process_that_returns_a_value)`, and that something returns an empty string, then they would make a request to the base URL rather than to the profile they were trying to use

Madrigal (commented):
Minor c

Madrigal (approved):

wty-Bryant (approved):

Issue #3032, opened 2025-03-09
WARN Response has no supported checksum. Not validating response payload. for simple S3 get


couling:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

This looks like a regression of #1606

Using `github.com/aws/aws-sdk-go-v2/service/s3 v1.78.1` a simple getObject request causes a warning:

```go
        result, err := client.GetObject(ctx, &s3.GetObjectInput{
		Bucket: &bucket,
		Key:    &key,
	})
```

```
SDK 2025/03/09 14:23:48 WARN Response has no supported checksum. Not validating response payload.
```

I understand that there is now a setting to turn of these warnings generally, but that's rather like cutting the wire to the warning light. Either this warning is a false positive and needs fixing, or else better instructions need to be given [here](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html) whatever developer error causes the error.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

Be able to perform a simple get request without warnings by following the [documentation](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html)

### Current Behavior

Generates a warning
```
SDK 2025/03/09 14:23:48 WARN Response has no supported checksum. Not validating response payload.
```

### Reproduction Steps

Follow instructions for a GetObject request in [documentation](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html).

### AWS Go SDK V2 Module Versions Used

```
	github.com/aws/aws-sdk-go-v2/config v1.29.9
	github.com/aws/aws-sdk-go-v2/service/s3 v1.78.1
```

### Compiler and Version used

go version go1.24.0 darwin/amd64

### Operating System and version

MacOS 15

jonathan-innis:
This warning appeared to start just firing randomly after we upgraded to the latest version of aws-sdk-go-v2 and called GetObject as well -- this strikes me as a regression since we didn't see this before -- agree that it's not entirely clear what this error means from the docs.

entigo-mart-erlenheim:
[...]

PR #3072, opened 2025-04-23
do not write WARN log for missing checksum on partial content

swtch1:
Partial fix for https://github.com/aws/aws-sdk-go-v2/issues/3032.  This suppresses WARN logs when the response status is `206 Partial Content`.  A 206 response will never have a checksum.


lucix-aws:
closing in favor of #3075

PR #3081, opened 2025-05-03
Remove credential config of GetBucketRegion by default

### LINKED ISSUE #3077: MIGRATION ISSUE: `GetBucketRegion` in v2 requires valid credentials
### Pre-Migration Checklist

- [x] I've read the [Migration Guide](https://aws.github.io/aws-sdk-go-v2/docs/migrating/).
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-go) for similar migration issues.

### Go Version Used

Go 1.24

### Describe the Migration Issue

Originally posted in https://github.com/trufflesecurity/trufflehog/pull/4069#issuecomment-2816678647.

---

I discovered an undocumented behavioral difference between V1’s `GetBucketRegionWithClient` and V2's `GetBucketRegion`.

Specifically, V1’s `GetBucketRegionWithClient` does not return an error when credentials are invalid, whereas V2’s `GetBucketRegion` fails under the same conditions.

In the following code examples, I used a fake access key, fake secret key, and a fake role ARN.

## `GetBucketRegionWithClient` in V1

```go
package main

import (
	"context"
	"log"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/credentials"
	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/s3"
	"github.com/aws/aws-sdk-go/service/s3/s3manager"
	"github.com/aws/aws-sdk-go/service/sts"
)

func main() {
	ctx := context.Background()

	cfg := &aws.Config{
		Region:      aws.String("us-east-1"),
		Credentials: credentials.NewStaticCredentials("fakeDummy", "fakeDummy", ""),
	}
	sess, err := session.NewSession(cfg)
	if err != nil {
		log.Fatal(err)
	}

	stsClient := sts.New(sess)
	cfg.Credentials = stscreds.NewCredentialsWithClient(stsClient, "000000000:role/service-role/fake-non-existent-role")

	s3Client := s3.New(sess)

	region, err := s3manager.GetBucketRegionWithClient(ctx, s3Client, "truffletestbucket-s3-role-assumption")
	log.Println("Region: ", region)
	log.Println("Err: ", err)
}
```

## `GetBucketRegion` in V2

```go
package main

import (
	"context"
	"log"

	"github.com/aws/aws-sdk-go-v2/credentials"
	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
	s3manager "github.com/aws/aws-sdk-go-v2/feature/s3/manager"
)

func main() {
	ctx := context.Background()

	cfg, err := config.LoadDefaultConfig(ctx,
		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider("fakeDummy", "fakeDummy", "")),
	)
	if err != nil {
		log.Fatal(err)
	}

	stsClient := sts.NewFromConfig(cfg)
	provider := stscreds.NewAssumeRoleProvider(stsClient, "arn:aws:iam::000000000:role/service-role/fake-non-existent-role")
	cfg.Credentials = aws.NewCredentialsCache(provider)

	s3Client := s3.NewFromConfig(cfg)

	region, err := s3manager.GetBucketRegion(ctx, s3Client, "truffletestbucket-s3-role-assumption")
	log.Println("Region: ", region)
	log.Println("Err: ", err)
}
```

### Observed Differences/Errors

# V1 Code Snippet Output

Running the V1 code snippet above produces the following output:

```
2025/04/19 19:04:01 Region:  us-east-2
2025/04/19 19:04:01 Err:  <nil>
```

Upon inspecting the [source code of V1's `GetBucketRegionWithClient`](https://github.com/aws/aws-sdk-go/blob/v1.55.6/service/s3/s3manager/bucket_region.go#L104-L110), we can see that it bypasses the configured credentials and issues the `HeadBucket` request using `credentials.AnonymousCredentials`:

```go
func GetBucketRegionWithClient(ctx aws.Context, svc s3iface.S3API, bucket string, opts ...request.Option) (string, error) {
	req, _ := svc.HeadBucketRequest(&s3.HeadBucketInput{
		Bucket: aws.String(bucket),
	})
	req.Config.S3ForcePathStyle = aws.Bool(true)

	req.Config.Credentials = credentials.AnonymousCredentials
```

# V2 Code Snippet Output

Running the V2 code snippet above code results in the following output:

```
2025/04/19 19:03:17 Region:
2025/04/19 19:03:17 Err:  operation error S3: HeadBucket, get identity: get credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 331c40dd-d8c4-42ca-9105-fcff18f9a8fb, api error InvalidClientTokenId: The security token included in the request is invalid.
```

The error originates from the following call chain:

1. https://github.com/aws/aws-sdk-go-v2/blob/service/s3/v1.79.2/service/s3/auth.go#L274-L283
2. https://github.com/aws/aws-sdk-go-v2/blob/v1.36.3/internal/auth/smithy/credentials_adapter.go#L40-L43

### Additional Context

It turns out that valid credentials are not required to retrieve a bucket's region using the [HeadBucket API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html).

A simple `HEAD` request returns the `x-amz-bucket-region` header, even when the response status is `403 Forbidden`:

```shell
curl --head truffletestbucket-s3-role-assumption.s3.amazonaws.com

HTTP/1.1 403 Forbidden
x-amz-bucket-region: us-east-2
x-amz-request-id: 7E2S67HYJWFER9XD
x-amz-id-2: ZRbFyUcIJDb7y15lJvnG9M2vGqF55h7NQLaYJvQFh2CQTsMsCZDpwHa1VIQGHIHd52sOwaJwJ3vYDkxULXKp+tlDH1l5lD2Tr5dbpV+CS/E=
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Sat, 19 Apr 2025 11:47:05 GMT
Server: AmazonS3
```

This matches the test case for `TestGetBucketRegion_Exists`, where we expect the bucket region is included in the header when the response is `403 Forbidden`.

https://github.com/aws/aws-sdk-go-v2/blob/cfef7b8bb84e6f11f5294901ebbb45eccc702b96/feature/s3/manager/bucket_region_test.go#L50-L53

wty-Bryant:
The s3 transfer manager helper func `GetBucketRegion` is ported from go v1 but missed default anonymous credential config, this PR patches the config back so user could get a public bucket's region without credentials.
Fixes #3077

### DIFF
--- a/.changelog/096161d8acea4d53afe332ec60aac3db.json
+++ b/.changelog/096161d8acea4d53afe332ec60aac3db.json
@@ -1,0 +1,8 @@
+{
+    "id": "096161d8-acea-4d53-afe3-32ec60aac3db",
+    "type": "bugfix",
+    "description": "Enable user to call GetBucketRegion without credential config",
+    "modules": [
+        "feature/s3/manager"
+    ]
+}


feature/s3/manager/bucket_region.go
```diff
@@ -63,20 +63,21 @@ const bucketRegionHeader = "X-Amz-Bucket-Region"
...
-	clientOptionFns := make([]func(*s3.Options), len(optFns)+1)
+	clientOptionFns := make([]func(*s3.Options), len(optFns)+2)
 	clientOptionFns[0] = func(options *s3.Options) {
 		options.APIOptions = append(options.APIOptions, captureBucketRegion.RegisterMiddleware)
 	}
-	copy(clientOptionFns[1:], optFns)
+	clientOptionFns[1] = func(options *s3.Options) {
+		options.Credentials = nil
+	}
+	copy(clientOptionFns[2:], optFns)
```
Juneezee, line 80: Thanks for fixing the bug!

Out of curiosity, is there a reason why we don't just set `options.Credentials = nil `directly in `clientOptionFns[0]`?

```suggestion
	clientOptionFns[0] = func(options *s3.Options) {
		options.APIOptions = append(options.APIOptions, captureBucketRegion.RegisterMiddleware)
		options.Credentials = nil
	}
	copy(clientOptionFns[1:], optFns)
```

lucix-aws (approved):

feature/s3/manager/bucket_region.go
wty-Bryant: It doesn't matter whether you put it at index 0 or 1 as long as it is executed ahead other functional config (e.g. user want to reset their credentials using a option fn)

cartermckinnon:
@lucix-aws this broke things for private buckets in `cn-northwest-1`. The `HeadBucket` behavior is different there, you must provide credentials or you receive a `401` (and there is no `x-amz-bucket-region` header in the response).

It works as expected in `cn-north-1` (even for buckets that reside in `cn-northwest-1`). I reached out to the S3 team as well, not clear if this is expected behavior.

I can work around this by explicitly using the client's credentials for the call:
```
cfg, _ := config.LoadDefaultConfig(context.TODO())
region, err := manager.GetBucketRegion(context.TODO(), s3.NewFromConfig(cfg), bucketName, func(o *s3.Options) {
	o.Credentials = cfg.Credentials
})
```

But this worked with defaults prior to `feature/s3/manager@v1.17.75`.

Issue #941, opened 2020-12-01
Add support for S3 transfer manager callbacks


bunjiboys:
**Is your feature request related to a problem? Please describe.**
With the non-Golang SDK's the S3 manager package supports providing a callback function to get information about data being sent to the S3 API, making it a lot easier to reliably track transfer speeds. Currently with the Go SDK, the only option is to wrap the writer/reader API's to do calculation based purely on local speeds.

**Describe the solution you'd like**
Similarly to the Python and Node/JavaScript API's, add support for either a callback function or provide a channel that can receive the number of bytes transferred as each part is uploaded.

**Describe alternatives you've considered**
As described above, I currently have a workaround wrapping the reader API for files to roughly track speeds, using the io.Writer* and io.Reader interfaces. However, this only tells you the speed of your local disks, and not the actual network transfers.

lucix-aws:
This is referred to as "progress listeners" in the modern transfer manager specification, it's definitely something we intend to support in the next iteration of the feature.

PR #3083, opened 2025-05-05
add DownloadObject progress listeners

lucix-aws:
Adds progress listeners to the new transfer manager's DownloadObject implementation.

Part of #941


feature/s3/transfermanager/downloadobject_test.go
```diff
@@ -19,23 +19,26 @@ import (
...
+		expectRanges         []string
+		partNumber           int32
+		versionID            string
+		partsCount           int32
+		expectParts          []int32
+		expectVersions       []string
+		expectETags          []string
+		expectErr            string
+		dataValidationFn     func(*testing.T, *types.WriteAtBuffer)
+		listenerValidationFn func(*testing.T, *mockListener, any, any, error)
```
wty-Bryant, line 41: Good use case of `any` compatible with later listeners

wty-Bryant (approved):

feature/s3/transfermanager/api_op_DownloadObject.go
```diff
@@ -776,14 +790,6 @@ func (d *downloader) incrWritten(n int64) {
 	d.written += n
 }
 
-// getTotalBytes is a thread-safe getter for retrieving the total byte status.
-func (d *downloader) getTotalBytes() int64 {
```
Madrigal, line 780: [nit] was this just unused all around? I suspect we're removing this since the actual solution is "use a progress listener", right?

Madrigal (approved):

feature/s3/transfermanager/api_op_DownloadObject.go
lucix-aws: Yeah, this was unused. I assume it was ported over from v1 but refactoring made it unnecessary.

PR #3085, opened 2025-05-14
Add echo-service model for optional codegen testing of cbor/json services

Madrigal:
Add a new `service/echo` codegen model, without generating a client for it. 

This won't actually be plugged in, but this is useful for internal testing.

Followed the same steps as #3023

### DIFF
--- a/codegen/echo-service/model/main.smithy
+++ b/codegen/echo-service/model/main.smithy
@@ -1,0 +1,54 @@
+$version: "2.0"
+
+namespace smithy.sdk.example
+
+use aws.api#service
+use aws.auth#sigv4
+use smithy.protocols#rpcv2Cbor
+ 
+@title("Echo Service")
+@service(sdkId: "Echo")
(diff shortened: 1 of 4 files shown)


lucix-aws (approved):

codegen/echo-service/model/main.smithy
```diff
@@ -0,0 +1,54 @@
...
+// comment cbor and uncomment this for enabling awsJson
+// @aws.protocols#awsJson1_0
+service EchoService {
+    version: "2020-07-02"
+    operations: [EchoOperation]
+}
+
+operation EchoOperation {
+    input := with [ComplexStructureMixin] {}
+    output := with [ComplexStructureMixin] {}
```
wty-Bryant, line 22: Just curious: why we use `ComplexStructureMixin` for both input and output

Madrigal: This is an Echo service, so it returns the same shape for input and output

PR #3109, opened 2025-06-07
Use `crypto/subtle.ConstantTimeCompare` in sigv4a signer.

teo-tsirpanis:


lucix-aws:
This isn't right, the stdlib version of constant compare you're using returns 0/1 for equality. We need to actually **compare** the two operands and check for ordered comparison. If you look at the implementation for our constant compare you will notice that it returns -1/0/1 and the key derivation routine depends on this. As you can see above from the test failures this actually breaks the signer.

teo-tsirpanis:
Oops, didn't notice. 😳

PR #3121, opened 2025-06-25
Join errors when retry token is unavailable

NetcraftPm:
If we fail to acquire a retry token, we are currently discarding the original error. According to the [Retry Behaviour reference guide](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html), we are expected to return the actual response (error) in this situation.

With these changes, we now return the joined error, which allows callers to inspect and match on the error that caused the request to fail.

### DIFF
--- a/.changelog/3723ba75-d027-4e70-93f3-6202b10d6edd.json
+++ b/.changelog/3723ba75-d027-4e70-93f3-6202b10d6edd.json
@@ -1,0 +1,9 @@
+{
+    "id": "3723ba75-d027-4e70-93f3-6202b10d6edd",
+    "type": "bugfix",
+    "description": "Include original error when retry token unavailable",
+    "collapse": false,
+    "modules": [
+        "aws"
+    ]
+}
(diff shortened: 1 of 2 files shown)


lucix-aws:
This seems ok but shouldn't the retry error wrap the original error instead?

NetcraftPm:
> This seems ok but shouldn't the retry error wrap the original error instead?

I'm happy to make it wrap the error in `GetRetryToken` instead if you prefer. My logic was that it seemed that given the method name is `GetRetryToken` it doesn't make much sense for it to wrap the error it's given, as that's conceptually to determine what (size of) token to reserve. We could also use `fmt.Errorf` to join the errors, but it feels like that would make the actual error string quite hard to read.

NetcraftPm:
@lucix-aws Hi, just wondering if you would mind taking another look at this in light of my comment last month?

lucix-aws:
Integration tests fail in forks, don't worry about that.

NetcraftPm:
@lucix-aws Great, thanks! Anything else you need from me on this?

lucix-aws:
It's good to go. I'm planning to merge this after today's release and then it'll be available in the Tuesday release.

PR #3127, opened 2025-07-02
MOD: include zero content-length header if present

vjoke:
# In Brief
This change addresses an issue in the SigV4 signing process where an explicit Content-Length: 0 header would be omitted from the set of signed headers.
# Description of Problem
The v4 signer is responsible for building a canonical request, which includes a list of signed headers. The current implementation has two relevant behaviors:
It automatically adds the Content-Length header to the signed headers only if the request body's length is greater than zero (length > 0).
It iterates through the request's existing headers but contains a condition to always skip a header named Content-Length, to prevent it from being added twice.
The combination of these two behaviors means that if a request has a zero-length body and an explicit Content-Length: 0 header, the header is never included in the signature. This can lead to signature mismatch errors for AWS services that require Content-Length to be signed, even for empty bodies (e.g., S3 PutObject).
# The Fix
This PR modifies the condition that skips existing Content-Length headers. Instead of unconditionally skipping it, the condition is now strings.EqualFold(k, contentLengthHeader) && length > 0.
This ensures that the pre-existing Content-Length header is only skipped if the signer has already handled it (because length > 0). If length is 0, the signer will now correctly process and sign an explicit Content-Length: 0 header found on the request.
# How to Test
This can be verified with a unit test that creates a request with a zero-byte body, adds an explicit Content-Length: 0 header, and asserts that content-length is present in the canonical headers string generated by the signer.

### DIFF
--- a/aws/signer/v4/v4.go
+++ b/aws/signer/v4/v4.go
@@ -434,7 +434,7 @@ func (s *httpSigner) buildCanonicalHeaders(host string, rule v4Internal.Rule, he
 		if !rule.IsValid(k) {
 			continue // ignored header
 		}
-		if strings.EqualFold(k, contentLengthHeader) {
+		if strings.EqualFold(k, contentLengthHeader) && length > 0 {
 			// prevent signing already handled content-length header.
 			continue
 		}


lucix-aws:
Are you using the signer directly or is this fixing a failure to sign a request in this SDK against an official AWS service?

If the former -- we recently released a standalone signing module `github.com/smithy-go/aws-http-auth`. At this point, if this is for a general sigv4 issue, this contribution should probably be made there (OR it may be that that version is sufficiently configurable enough such that this isn't an issue). While it's always been possible to use this signer for whatever, it's really more meant to be used by the SDK and it has some behavioral quirks.

If the latter -- please elaborate.

PR #3145, opened 2025-07-20
feat: add support for per service options to Config

### LINKED ISSUE #3141: add support for per-service options to aws.Config
### Describe the feature

Allow for `aws.Config` to hold "options" to be applied only when constructing the client for a particular service.

### Use Case

In a monorepo, it is useful to have a single framework for configuring the SDK. Today, while we can configure the "base" `aws.Config` itself, it is not feasible to for example apply a consistent configuration to all DynamoDB service clients.

While we can somewhat achieve this today by writing our own wrapper functions around every service client constructor, this requires all developers to remember to do this.

### Proposed Solution

Add a field to `aws.Config` that maps from a service ID (e.g., `dynamodb.ServiceID`) to a list of callback functions (`[]func(any)`).

When calling for example `dynamodb.NewFromConfig`, it will check if the `aws.Config` contains an entry for the service ID in question. If it does, it will pass the service-specific options struct (e.g., `dynamodb.Options`) to each callback.

Alternatively, add a field like `func(string) []func(any)` to `aws.Config`, which is a function that will accept the service ID and return the list of options callback, if any. The advantage of this approach is that the set of options can be determined more dynamically (e.g., mapping the service ID to an environment variable).

### Other Information

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request

n/a

### Go version used

n/a

Aakkash-Suresh:
## Summary
Implements the feature requested in #3141 to allow `aws.Config` to hold service specific options that are automatically applied during client construction.

## Problem
Currently, service specific options must be manually passed to each client creation:
```go
ddbClient1 := dynamodb.NewFromConfig(cfg, func(o *dynamodb.Options) {
    o.DisableValidateResponseChecksum = true
})
ddbClient2 := dynamodb.NewFromConfig(cfg, func(o *dynamodb.Options) {
    o.DisableValidateResponseChecksum = true
})
```

## Solution
Add service-specific options to `aws.Config` that are automatically applied during client construction:
```go
cfg := aws.NewConfig().
    WithRegion("us-west-2").
    WithServiceOptions("DynamoDB", func(opts any) {
        if ddbOpts, ok := opts.(*dynamodb.Options); ok {
            ddbOpts.DisableValidateResponseChecksum = true
        }
    })

// Options are automatically applied
ddbClient1 := dynamodb.NewFromConfig(cfg)
ddbClient2 := dynamodb.NewFromConfig(cfg)
```

## Related Issue
Closes #3141


aws/config.go
```diff
@@ -192,6 +192,45 @@ type Config struct {
 	// This variable is sourced from environment variable AWS_RESPONSE_CHECKSUM_VALIDATION or
 	// the shared config profile attribute "response_checksum_validation".
 	ResponseChecksumValidation ResponseChecksumValidation
+
+	// ServiceOptions provides service specific configuration options that will be applied
+	// when constructing clients for specific services. The key is the service ID (e.g., "DynamoDB"),
+	// and the value is a slice of callback functions that will be applied to the service's Options struct.
+	ServiceOptions map[string][]func(any)
```
rittneje, line 199: @Aakkash-Suresh I'm wondering if this should instead be `[]func(string, any)`. That would enable use cases in which the framework that is setting these options does not have to directly know about all possible services.

For example:
```go
WithServiceOptions(func(serviceID string, opts any) {
    field := reflect.ValueOf(opts).Elem().FieldByName("RetryMaxAttempts")
    if field.IsZero() {
        return
    }

    maxAttempts, err := strconv.Atoi(os.Getenv("AWS_SDK_" + serviceID + "_MAX_ATTEMPTS"))
    if err == nil && maxAttempts >= 0 {
        field.SetValue(reflect.ValueOf(maxAttempts))
    }
})
```

lucix-aws:
You're going to need to regenerate the service clients since this includes a codegen change.

aws/config.go
Aakkash-Suresh: Hmm, Great idea. I've implemented the field as you suggested. This makes it much easier for frameworks to configure services without having to know about every single AWS service.

Aakkash-Suresh:
> You're going to need to regenerate the service clients since this includes a codegen change.

Service clients have been regenerated. All api_client.go files now include the ApplyServiceOptions call. 

@lucix-aws Let me know if i'm missing something here.

aws/config.go
```diff
@@ -192,6 +192,40 @@ type Config struct {
...
+	for _, callback := range c.ServiceOptions {
+		callback(serviceID, options)
+	}
+}
+
+// WithServiceOptions adds service specific options to the config.
+// This function can be chained with other config builder methods.
+func (c *Config) WithServiceOptions(callbacks ...func(string, any)) *Config {
+	if c.ServiceOptions == nil {
+		c.ServiceOptions = make([]func(string, any), 0)
```
rittneje, line 218: This line is unnecessary. append will take care of it.

aws/config.go
```diff
@@ -192,6 +192,33 @@ type Config struct {
...
+// This function is intended to be used by service clients in their NewFromConfig functions.
+func (c Config) ApplyServiceOptions(serviceID string, options any) {
+	for _, callback := range c.ServiceOptions {
+		callback(serviceID, options)
+	}
+}
+
+// WithServiceOptions adds service specific options to the config.
+// This function can be chained with other config builder methods.
+func (c *Config) WithServiceOptions(callbacks ...func(string, any)) *Config {
```
lucix-aws, line 212: These `With...()` APIs go on `LoadOptions`, not `Config` (notice how there weren't any before)

aws/config.go
```diff
@@ -192,6 +192,33 @@ type Config struct {
...
+	c.ServiceOptions = append(c.ServiceOptions, callbacks...)
+	return c
+}
+
+// WithRegion sets the region for the config.
+// This function can be chained with other config builder methods.
+func (c *Config) WithRegion(region string) *Config {
```
lucix-aws, line 219: Not sure why you added this, these go on `LoadOptions` and there's already a `WithRegion` there.

aws/config.go
```diff
@@ -192,6 +192,33 @@ type Config struct {
...
+
+	// ServiceOptions provides service specific configuration options that will be applied
+	// when constructing clients for specific services. Each callback function receives the service ID
+	// and the service's Options struct, allowing for dynamic configuration based on the service.
+	ServiceOptions []func(string, any)
+}
+
+// ApplyServiceOptions applies service specific options from the config to the given options struct.
+// This function is intended to be used by service clients in their NewFromConfig functions.
+func (c Config) ApplyServiceOptions(serviceID string, options any) {
```
lucix-aws, line 204: There's no real reason to have this "public but you shouldn't call it"-type API -- just generate the logic to do this in `NewFromConfig` etc.

service/accessanalyzer/api_client.go
```diff
@@ -440,6 +440,7 @@ func NewFromConfig(cfg aws.Config, optFns ...func(*Options)) *Client {
 	resolveUseDualStackEndpoint(cfg, &opts)
 	resolveUseFIPSEndpoint(cfg, &opts)
 	resolveBaseEndpoint(cfg, &opts)
+	cfg.ApplyServiceOptions(ServiceID, &opts)
```
lucix-aws, line 443: We _probably_ have to actually pass these new functional options down into `New()` and call them there, where the other functional options are called.

lucix-aws (changes requested):

service/accessanalyzer/api_client.go
rittneje: They would not work without extra wrapping as the type signatures are incompatible.

rittneje: or just:
```go
return New(opts, func(o *Options) {
    for _, opt := range cfg.ServiceOptions {
        opt(ServiceID, o)
    }
    for _, opt := range optFns {
        opt(o)
    }
})
```

aws/config.go
Aakkash-Suresh: Makes sense! I'll remove the `ApplyServiceOptions` method and generate the logic directly in the service client's `NewFromConfig` function instead. This is much cleaner.

service/accessanalyzer/api_client.go
Aakkash-Suresh: Got it! I'll modify the codegen template to pass the ServiceOptions as functional options to `New()` where they can be called alongside the other options. 

Also, I'll keep the ServiceID parameter in the callback since it enables use cases where the framework doesn't need to know about all possible services upfront, like setting environment based options dynamically.

aws/config.go
lucix-aws: Are you using an LLM to respond to these comments?

Aakkash-Suresh: Not really, I'm using them to rephrase my comment at some places 😄

rittneje:
@lucix-aws Is there anything still pending for this change?

lucix-aws:
I'll try to find time to look at it again this week.

lucix-aws:
Ignore failing integration tests and codegen tests.

PR #3159, opened 2025-08-05
add bedrock env token provider

lucix-aws:
Adds support for `AWS_BEARER_TOKEN_BEDROCK`, which when configured will switch the bedrock and bedrockruntime services to using HTTP bearer auth.

### DIFF
--- a/service/bedrock/env_token_provider_test.go
+++ b/service/bedrock/env_token_provider_test.go
@@ -2,7 +2,6 @@ package bedrock
 
 import (
 	"context"
-	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


service/bedrock/env_token_provider_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"context"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/smithy-go/auth/bearer"
+)
+
+func patchEnv(k, v string) func() {
```
Madrigal, line 13: you can use [t.Setenv](https://pkg.go.dev/testing#T.Setenv) instead

service/bedrock/env_token_provider_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Token configured with auth scheme preference also set in env. The Bedrock
+// token is more specific which is why it takes precedence
+func TestEnvTokenProviderSEP2(t *testing.T) {
+	defer patchEnv("AWS_BEARER_TOKEN_BEDROCK", "bedrock-token")()
+
+	cfg := aws.Config{
+		// simulate loaded from env/shared config
+		AuthSchemePreference: []string{"sigv4", "httpBearerAuth"},
```
Madrigal, line 62: this would also work even if httpBearerAuth wasn't on the preference, right?

service/bedrock/env_token_provider_test.go
```diff
@@ -0,0 +1,107 @@
...
+		os.Unsetenv(k)
+	}
+}
+
+// Valid service-specific token configured
+func TestEnvTokenProviderSEP0(t *testing.T) {
+	defer patchEnv("AWS_BEARER_TOKEN_BEDROCK", "bedrock-token")()
+
+	svc := New(Options{})
+	token, err := svc.Options().BearerAuthTokenProvider.RetrieveBearerToken(context.Background())
```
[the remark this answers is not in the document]
lucix-aws: That would be a failure for this test as well which would be what we'd want to see here.

Madrigal (approved):

PR #3163, opened 2025-08-08
Feat S3 Transfer Manager v2 DownloadDirectory

wty-Bryant:
Implement v2 s3 transfer manager's DownloadDirectory api bound to single union client which mimics normal service client's initialization and api call. User could now download a bucket's objects to local directory with customized filtering and file naming in a single operation call.

Test: This PR passed unit tests for DownloadDirectory which covers use cases with different config and objects naming to confirm this key-prefix downloader could find valid objects, generate correct file path and pass download requests to S3 concurrently. It also passed integration test which uploads objects with different sizes to S3 via S3 client, then downloads objects back using new directory downloader and checks their content consistency

### DIFF
--- a/feature/s3/transfermanager/api_op_DownloadDirectory.go
+++ b/feature/s3/transfermanager/api_op_DownloadDirectory.go
@@ -236,7 +236,7 @@ func (d *directoryDownloader) downloadObject(ctx context.Context, ch chan object
 			d.emitter.Start(ctx, d.in)
 		})
 
-		err = os.MkdirAll(filepath.Dir(data.path), os.ModePerm)
+		err = os.MkdirAll(filepath.Dir(data.path), 0755)
 		if err != nil {
 			d.setErr(fmt.Errorf("error when creating directory for file %s: %v", data.path, err))
 			continue


feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -0,0 +1,252 @@
...
+	// is different from local file separator
+	S3Delimiter string
+
+	// A callback func to allow users to fileter out unwanted objects
+	// according to bool returned from the function
+	Filter ObjectFilter
+
+	// A callback function to allow customers to update individual
+	// GetObjectInput that the S3 Transfer Manager generates
+	Callback GetRequestCallback
```
lucix-aws, line 41: You seem to be missing a few inputs from the SEP
* recursive
* failurePolicy

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -0,0 +1,252 @@
...
+
+	i := directoryDownloader{c: c, in: input, options: c.options.Copy()}
+	for _, opt := range opts {
+		opt(&i.options)
+	}
+
+	return i.downloadDirectory(ctx)
+}
+
+type directoryDownloader struct {
```
lucix-aws, line 96: Progress listener support?

feature/s3/transfermanager/api_op_DownloadDirectory.go
lucix-aws: Talked offline, design for these seems to be somewhat in-flux so we can defer for now.

PR #3173, opened 2025-08-27
Update to smithy-go v1.23.0

lucix-aws:


Madrigal (approved):
Approved, side note: haven't occurred to me to add two changelogs for the same change to emphasize how one particular module would be impacted. Nice.

wty-Bryant (approved):

PR #3185, opened 2025-09-09
Bump smithy go version to allow unused required parameter in endpoint rule set

Madrigal:
Adding changes from aws/smithy-go#601


lucix-aws (approved):

lucix-aws:
Don't you have to regenerate everything?

Madrigal:
> > Don't you have to regenerate everything?

ugh, yeah was thinking this was not impacting the current models but yeah, adding the regen models

PR #3199, opened 2025-10-01
Feat checksum reuse

wty-Bryant:
Current checksum middleware calculates checksum of request payload for every re-attempt, which will be time-consuming for large request. This PR caches the first checksum calculated and re-use it in later retry, which avoids repeated checksum calculation and guarantees payload consistency check in case request body is updated between reattempts.

### DIFF
--- a/service/internal/checksum/middleware_compute_input_checksum.go
+++ b/service/internal/checksum/middleware_compute_input_checksum.go
@@ -368,7 +368,7 @@ func (m *AddInputChecksumTrailer) HandleFinalize(
 		if checksum == "" {
 			checksum, e = checksumReader.Base64Checksum()
 			if e != nil {
-				return out, metadata, fmt.Errorf("failed to get computed checksum, %w", err)
+				return out, metadata, fmt.Errorf("failed to get computed checksum, %w", e)
 			}
 		}
 		// Record the checksum and algorithm that was computed


lucix-aws:
Tests failing?

service/internal/checksum/middleware_compute_input_checksum.go
```diff
@@ -65,6 +65,10 @@ type ComputeInputPayloadChecksum struct {
 	// when used with trailing checksums, and aws-chunked content-encoding.
 	EnableDecodedContentLengthHeader bool
 
+	checksum string
+
+	sha256Checksum string
```
Madrigal, line 70: just for my education, why is sha256 treated differently than other algorithms for checksum?

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -968,6 +968,246 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
...
+			},
+			updatedStreamBody: "world hello",
+			expectHeader: http.Header{
+				"Content-Encoding": []string{"aws-chunked"},
+				"X-Amz-Trailer":    []string{"x-amz-checksum-crc32"},
+			},
+			expectPayload:         []byte("b\r\nworld hello\r\n0\r\nx-amz-checksum-crc32:DUoRhQ==\r\n\r\n"),
+			expectDeferToFinalize: true,
+			expectChecksumMetadata: map[string]string{
+				"CRC32": "DUoRhQ==",
```
Madrigal, line 1073: Why would the updated stream body have the same checksum as the original request?

service/internal/checksum/middleware_compute_input_checksum_test.go
```diff
@@ -968,6 +968,246 @@ func TestComputeInputPayloadChecksum(t *testing.T) {
...
+					r := smithyhttp.NewStackRequest().(*smithyhttp.Request)
+					r.URL, _ = url.Parse("http://example.aws")
+					r.ContentLength = 11
+					r = requestMust(r.SetStream(bytes.NewReader([]byte("hello world"))))
+					return r
+				}(),
+			},
+			updatedStreamBody: "world hello",
+			expectHeader: http.Header{
+				"X-Amz-Checksum-Crc32": []string{"DUoRhQ=="},
```
Madrigal, line 1020: > guarantees payload consistency check in case request body is updated between reattempts.

is that right? it seems like we're keeping the same checksum even with updated payload

service/internal/checksum/middleware_compute_input_checksum.go
Madrigal, line 70: more of a "for my understanding" question, but why is sha256 treated differently than the rest of the algorithms for checksum?

wty-Bryant: Yes, sha256 value is cached in ctx [here](https://github.com/aws/aws-sdk-go-v2/blob/fcd6c667064dfd76542078bcd5b435e9df401aef/service/internal/checksum/middleware_compute_input_checksum.go#L211) for later sigv4

service/internal/checksum/middleware_compute_input_checksum_test.go
wty-Bryant: The SEP didn't say which value should we set in metadata, but imo we should always set the original value here, if the payload is updated between retry, error will be returned and picked up so metadata value doesn't matter in this case

service/internal/checksum/middleware_compute_input_checksum_test.go
wty-Bryant: The logic here is "if there is a checksum value available from previous failed request, just use it", so the old checksum should still be set as header

service/internal/checksum/middleware_compute_input_checksum.go
wty-Bryant: Same answer listed above, just for payload signing

Madrigal (approved):

PR #3202, opened 2025-10-07
[merge 10/27/25] validate individual part ranges/size in upload/download

lucix-aws:
https://github.com/aws/aws-sdk-go-v2/issues/3213

Adds new SEP-specified durability checks to the existing transfer manager:
* Uploads will now verify that we actually send the number of parts we expected.
* Downloads will verify that S3 returns the correct range per part.

I have also elected to add a disable switch for these, since this is technically a behavioral break and there is a possibility that folks out there might actually be hitting these failure points without realizing. I'll add a pinned announcement issue to highlight this change when we ship it.

These changes are effectively verified by the existing unit/integration tests passing.

### DIFF
--- a/feature/s3/manager/validate_parts_test.go
+++ b/feature/s3/manager/validate_parts_test.go
@@ -32,3 +32,21 @@ func TestDownload_RangeMismatch(t *testing.T) {
 		t.Errorf("error mismatch:\n%v !=\n%v", err, "invalid content range")
 	}
 }
+
+func TestDownload_RangeMismatchDisabled(t *testing.T) {
+	c, _, _ := newDownloadBadRangeClient(buf12MB)
+
+	d := manager.NewDownloader(c, func(d *manager.Downloader) {
+		d.Concurrency = 1
+		d.DisableValidateParts = true
+	})
+
+	w := manager.NewWriteAtBuffer(make([]byte, len(buf12MB)))
+	_, err := d.Download(context.Background(), w, &s3.GetObjectInput{
+		Bucket: aws.String("bucket"),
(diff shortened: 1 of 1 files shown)


Madrigal:
change seems fine, but should we add some tests for it? 

who tests the validators? who watches the watchmen?

lucix-aws:
> change seems fine, but should we add some tests for it? 
> 
> 
> 
> who tests the validators? who watches the watchmen?

It basically seemed to me like the existing unit tests all do cover it, since the new logic is now running for all of them.

Madrigal:
> > change seems fine, but should we add some tests for it?
> > who tests the validators? who watches the watchmen?
> 
> It basically seemed to me like the existing unit tests all do cover it, since the new logic is now running for all of them.

I'm not sure how hard it is to mock a case where we get a different number of parts and I'll be fine if that's a medium/hard thing to do, but we're adding the `TrustMeBroThePartsAreCorrect` flag that we were definitely not testing before

lucix-aws:
> > > change seems fine, but should we add some tests for it?
> 
> > > who tests the validators? who watches the watchmen?
> 
> > 
> 
> > It basically seemed to me like the existing unit tests all do cover it, since the new logic is now running for all of them.
> 
> 
> 
> I'm not sure how hard it is to mock a case where we get a different number of parts and I'll be fine if that's a medium/hard thing to do, but we're adding the `TrustMeBroThePartsAreCorrect` flag that we were definitely not testing before

Well evidently based on Tianyi's last comment, we do absolutely need a positive test for this, so I'll need to figure it out.

Issue #3208, opened 2025-10-15
cannot use credential_process on windows when it has a quoted path containing space e.g. appstream


mrgum:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

I'm using a tool built on aws-sdk-go-v2 (rclone.org) and it fails to work inside aws appstream with the supplied default .aws/config file of

```
[profile appstream_machine_role]
credential_process = "C:\Program Files\Amazon\Photon\PhotonRoleCredentialProvider\PhotonRoleCredentialProvider.exe" --role=Machine
```

it gives an error

```
'\"C:\Program Files\Amazon\Photon\PhotonRoleCredentialProvider\PhotonRoleCredentialProvider.exe\"' is not recognized as an internal or external command,
```

changing that to avoid the " and spaces 

```
[profile appstream_machine_role]
credential_process = C:\Progra~1\Amazon\Photon\PhotonRoleCredentialProvider\PhotonRoleCredentialProvider.exe --role=Machine
```

avoids this issue but it would be good if it could work out of the box

I don't think this happened in as-sdk-go (v1)

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

sdk works using appstream's credential_process

### Current Behavior

```
'\"C:\Program Files\Amazon\Photon\PhotonRoleCredentialProvider\PhotonRoleCredentialProvider.exe\"' is not recognized as an internal or external command,
```

### Reproduction Steps

thanks to https://github.com/ncw

```
package main

import (
	"context"
	"log"
	"os"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/s3"
)

func main() {
	if len(os.Args) < 2 {
		log.Fatalf("Need bucket name as parameter")
	}
	bucket := os.Args[1]

	// Load the Shared AWS Configuration (~/.aws/config)
	cfg, err := config.LoadDefaultConfig(context.TODO())
	if err != nil {
		log.Fatal(err)
	}

	// Create an Amazon S3 service client
	client := s3.NewFromConfig(cfg)

[...]

PR #3209, opened 2025-10-15
windows: split up arguments keeping quoted strings with spaces

mrgum:
refences issue #3208

deals with credentials_process containing "c:\program files\"... as used in aws appstream


mrgum:
p.s. I'm a golang novice so please rewrite it :)

wty-Bryant:
Closed the PR since the change here may break current user code relying on that behavior and you have workaround which is more like a profile mis-configuration

PR #3216, opened 2025-10-20
Speed up unit tests by removing duplicate work

Madrigal:
Currently, the tests we run on CI have the following structure

```
	cd ./internal/repotools/cmd/eachmodule \
		&& go run . -p $(subst _,/,$(subst unit-race-modules-,,$@)) ${EACHMODULE_FLAGS} \
		"go vet ${BUILD_TAGS} --all ./..." \
		"go test ${BUILD_TAGS} ${RUN_NONE} ./..." \
		"go test -timeout=2m ${UNIT_TEST_TAGS} -race -cpu=4 ./..."
```

`eachmodule` command finds all Go modules in the directory (usually root) and runs the commands you pass as arguments concurrently, so if you have for example modules `A`, `B` and `C`, you'd run 9 commands on this example, since you have `go vet`, `go test` and another `go test`.

## Problem
We are doing duplicate work with this. We run `go test` twice for each module. The core issue is that the first command expands to 

```
go test -tags example,codegen,integration,ec2env,perftest -run NONE
```

Which is a fancy way of saying "Compile test that have these tags, which we don't normally include when compiling, but don't run any of the tests there".
This is however executed on *all* modules, where there are only a handful of modules that have these tags, so this is unnecessary work on each module. 

Another minor issues is that we run `go vet` as a separate target. `vet` can be run as part of `go test` by passing `go test -vet=all` flag, so Go can take advantage of the already parsed AST.

## Solution

This PR adds a new `findtaggedmodules` target, which finds all modules that contain those tags, and adds a new target `build-tagged-modules`, to only run the first `go test -tags example,codegen,integration,ec2env,perftest -run NONE` on those modules.
Then, for the rest of the modules, it adds the `-vet=all` to run `vet` when running testing.

## Results

Last run took 1h 7m [ref](https://github.com/aws/aws-sdk-go-v2/actions/runs/18567912402), where the test that took the longest was 

```
 Unix SDK tests (macos-latest, 1.23)
succeeded in 1h 7m 24s 
```

The fastest test was 

```
 Unix SDK tests (ubuntu-latest, 1.25)
succeeded in 42m 32s 
```

On this run, execution took 43 m [ref](https://github.com/aws/aws-sdk-go-v2/actions/runs/18693907255?pr=3216), where the test that took the longest was

```
 Windows SDK Tests (windows-latest, 1.25)
succeeded in 43m 29s 
```

and the fastest test was 

```
 Unix SDK tests (ubuntu-latest, 1.25)
succeeded in 25m 24s 
```

### DIFF
--- a/.github/workflows/closed-issue-message.yml
+++ b/.github/workflows/closed-issue-message.yml
@@ -2,6 +2,10 @@ name: Closed Issue Message
 on:
     issues:
        types: [closed]
+
+permissions:
+  issues: write
+
 jobs:
     auto_comment:
         runs-on: ubuntu-latest
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ main ]
 
+
+permissions:
+  contents: read
+
 env:
   # get owner of the repository. used by forks.
   SMITHY_GO_REPOSITORY: ${{ github.event.pull_request.head.repo.owner.login }}/smithy-go
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,6 +9,10 @@ on:
       - main
       - 'feat-**'
 
+
+permissions:
+  contents: read
+
 env:
   EACHMODULE_CONCURRENCY: 2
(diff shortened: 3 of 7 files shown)


lucix-aws (approved):
11/10, naturally let's wait for CI

wty-Bryant (approved):

PR #3261, opened 2026-01-01
all: Replace ioutil.TempFile with os.CreateTemp

12ya:
Updated all usages of the deprecated ioutil.TempFile to use os.CreateTemp across test and utility files. This change improves compatibility with newer Go versions and follows current best practices for temporary file creation.


lucix-aws:
https://github.com/aws/aws-sdk-go-v2/pull/3265#issuecomment-3785286768

PR #3263, opened 2026-01-01
all: Replace ioutil.NopCloser with io.NopCloser in tests

12ya:
Updated all test files to use io.NopCloser instead of the deprecated ioutil.NopCloser for wrapping readers. This change improves compatibility with newer Go versions and removes unnecessary imports of ioutil.


lucix-aws:
https://github.com/aws/aws-sdk-go-v2/pull/3265#issuecomment-3785286768

PR #3267, opened 2026-01-01
all: Refactor stripPort to use strings.Cut for parsing

12ya:
Updated the stripPort function in both host.go files to use strings.Cut for improved clarity and efficiency when parsing host and port components, replacing manual index lookups.


lucix-aws:

PR #3268, opened 2026-01-01
all: Refactor portOnly to use strings.Cut for parsing

12ya:
Updated the portOnly function in both host.go files to use strings.Cut for improved clarity and correctness when extracting the port from host strings, including better handling of IPv6 addresses and malformed input.


lucix-aws:

PR #3272, opened 2026-01-02
internal/auth/scheme, config/provider:

12ya:
Remove unused servicesObjectProvider and minor cleanup

Deleted the unused servicesObjectProvider interface and related function from provider.go. Also removed unnecessary fmt import and simplified error string returns in scheme.go.


lucix-aws:

Issue #2726, opened 2024-07-25
(short issue description)


kmacmcfarlane:
### Acknowledgements

- [X] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [X] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

When working with dynamodb using attributevalue.MarshalMap() on a struct with `omitempty` on a field, I want to be able to return a `types.AttributeValueMemberNULL` in order to null out a field in dynamodb.

To achieve this, I want to use a double pointer on the field such that I can assign a nil pointer to my primative type to the double pointer. When encoding to JSON, for instance, this will encode the value `null` for that field in the output. The `MarshalMap()` appears to mistakenly think that the value double pointed to is nil, that it should not output anything when `omitempty` is set on the struct.

### Expected Behavior

I would expect `types.AttributeValueMemberNULL` to be returned for the key in the result av map.

### Current Behavior

The key for the struct field is omited from the map.

### Reproduction Steps

```
package store

import (
	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

var _ = Describe("endorsements store", func() {

	It("set Foo to attributevalue NULL", func() {

		// given
		var nilVal *int
		s := struct {
			Foo **int `dynamodbav:"foo,omitempty"`
		}{
			Foo: &nilVal,
		}

		av, err := attributevalue.MarshalMap(s)
		Expect(err).ShouldNot(HaveOccurred())
		Expect(av["Foo"]).Should(BeAssignableToTypeOf(types.AttributeValueMemberNULL{}))
	})
})
```

### Possible Solution

Check if a value of a pointer is a non-nil pointer to another pointer, in that case, return a `types.AttributeValueMemberNULL` instead of omitting the field from the output.

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2 v1.30.3
github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue v1.14.10
github.com/aws/aws-sdk-go-v2/feature/dynamodb/expression v1.7.32
[...]

PR #3290, opened 2026-01-22
attributevalue marshal to AttributeValueMemberNULL with double pointer when outer pointer is not nil

raduendv:
**Fixes**: attributevalue.MarshalMap() to AttributeValueMemberNULL field with double pointer [#2726](https://github.com/aws/aws-sdk-go-v2/issues/2726)

# Description

```go
var nilVal *int
s := struct {
	Foo **int `dynamodbav:"foo,omitempty"`
}{
	Foo: &nilVal,
}
```

The property `foo` should be serialized to `&types.AttributeValueMemberNULL{}` instead of falling back to `omitempty` since the first level pointer is not nil.


feature/dynamodb/attributevalue/encode.go
```diff
@@ -549,9 +561,19 @@ func (e *Encoder) encodeMap(v reflect.Value, fieldTag tag) (types.AttributeValue
...
+		// at all levels of nested maps or slices. This is necessary for correct DynamoDB marshaling
+		// when omitemptyelem is specified on the parent map or slice field.
+		//
+		// Note: For pointer types (including double pointers), omitempty only omits the field if the outer pointer is nil.
+		// If the outer pointer is non-nil but the inner pointer is nil (e.g., **int where *int is nil),
+		// omitempty does not omit the field; instead, the encoder emits a DynamoDB NULL attribute for that field.
 		elem, err := e.encode(elemVal, tag{
-			OmitEmpty: fieldTag.OmitEmptyElem,
-			NullEmpty: fieldTag.NullEmptyElem,
+			OmitEmpty:     fieldTag.OmitEmptyElem,
```
lucix-aws, line 574: This seems like a separate fix from the double-pointer issue? Additionally wouldn't we also need to change this for lists?

I would think this gets fixed in a separate PR though.

lucix-aws (commented):
The main part of this seems fine. While we typically would require this get gated behind an opt-in, in this case I believe it should be fixed directly because of the following
* this does not OMIT anything from the data, rather it adds something that was arguably missing (unlike OmitNullAttributeValues, OmitEmptyTime)
* there's no possibility of unexpected switching to a totally separate, dormant code path (UseEncodingMarshalers)
* the decode half of this already works this way, as your tests demonstrate

See my comment elsewhere about the encodeMap change though

feature/dynamodb/attributevalue/encode.go
raduendv: Initially addressed the reported bug with the direct pointer handler. I've now extended the fix to `encodeMap` and `encodeListElems` to match the same functionality. These weren't in the original bug path, but they need the same logic to stay consistent.

The new pointer handler checks both `OmitEmpty` and `OmitEmptyElem` to decide whether to omit an element or serialize it as `NULL`. Propagating both flags recursively through maps and lists ensures a non-nil outer pointer with a nil inner pointer gets omitted consistently under `omitemptyelem` across all collection types. Direct struct fields tagged with `omitempty` still serialize that state as `NULL`.

**List with `**int`:**

```go
var inner *int
outer := &inner

input := struct {
	Values []**int `dynamodbav:",omitemptyelem"`
}{
	Values: []**int{outer},
}
```

The element gets omitted:

```json
{
	"Values": {
		"L": []
	}
}
```

**Map with `**int`:**

```go
var inner *int
outer := &inner

input := struct {
	Values map[string]**int `dynamodbav:",omitemptyelem"`
}{
	Values: map[string]**int{"value": outer},
}
```

The entry gets omitted:

```json
{
	"Values": {
		"M": {}
	}
}
```

Issue #3294, opened 2026-01-28
Proposal: Data Mapper


raduendv:
### Describe the feature

**Proposal:** Add an "enhanced DynamoDB client" to the AWS SDK for Go v2, providing a strongly-typed, extensible, and ergonomic API for working with DynamoDB tables. This would include support for batch operations, schema mapping, extension hooks, and utilities for encoding/decoding Go structs to DynamoDB items.

### Use Case

Many of us working with DynamoDB need a higher-level, type-safe API that reduces boilerplate and improves maintainability. A built-in enhanced client would simplify CRUD operations, batch reads/writes, and schema management, and would be especially valuable for applications requiring custom extensions, efficient batch processing, and strong compile-time guarantees.

### Proposed Solution

Introduce a generic Table abstraction, batch read/write operations, schema builders, and extension registries as part of the SDK. These would wrap the low-level DynamoDB API with higher-level, idiomatic Go constructs. Batch operations should include retry logic and error thresholds. The design should allow for custom extension points and support for Go generics.

#### Example Usage

```go
type Order struct {
	OrderID   string
	Customer  string
	Total     float64
}

schema := enhancedclient.NewSchema[Order]().WithName("orders")

// Create a strongly-typed table
table, err := enhancedclient.NewTable[Order]
    ddbClient,
    enhancedclient.WithSchema(schema), // optional
)
if err != nil {
	// handle error
}

// Put an item
order := &Order{OrderID: "123", Customer: "Alice", Total: 42.0}
_, err = table.PutItem(ctx, order)

// Get an item
result, err := table.GetItem(ctx, enhancedclient.Map{"order_id": &types.AttributeValueMemberS{Value: "123"}})
// result is instance of Order{}
```

### Other Information

- Batch get/put/delete with iterators
- Schema-based encoding/decoding for Go structs
- Extension registry for custom behaviors
- Utilities for field mapping and attribute value conversion
- Comprehensive test coverage for table and batch operations

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

```
	github.com/aws/aws-sdk-go-v2 v1.39.4
	github.com/aws/aws-sdk-go-v2/config v1.31.15
	github.com/aws/aws-sdk-go-v2/feature/dynamodb/expression v1.8.19
	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.52.2
[...]

PR #3295, opened 2026-01-28
Data Mapper

raduendv:
Data Mapper implementation #3294


OrHayat:
im not aws worker or something but i kind of wanted to propuse similar thing. and i have done similar work at my company(resulted in different api because of some of the folowing things ill mention)

there are some really good ideas in this pr like table lifecycle hooks on the request to update/put and so on..

my main issuess with that pr is 
1) u mixed new serializer code with tables abstraction so u did alot of duplicate work above existing dynamodb.attributevalue codebase. - you also kept alot of the inefficienes that it does
when i say that i mean to compare to JSON v2 package - you check once what the type of the serialzied item is
and then u get/build serialzer/de-seiarlizer for that type  and call into it
instead of dynamiclly checking each and every type that the value can be

2) you made tag  struct even more bloated then previous version of serializer
 i would personally move some of those fields into schema defintion in the table constructor
 that because it make it wierd to denormlize stuff easily(you need to duplicate code)
 for example by moving  non seralization stuff to the table defintion you can do
 
```go                                                                                                                
  type User struct {                           
      Name  string `dynamodbav:"name"`                                                                                 
      Email string `dynamodbav:"email"`                                                                                
  }                                                                                                                    
                                                                                                                       
  type Order struct {
      Total  int    `dynamodbav:"total"`
      Status string `dynamodbav:"status"`
  }

  // Clean composition — no duplicated fields
  type UserOrders struct {
      User
      Order
  }
  ```
 and store struct  each one at its own table - table struct define the schema /autoversion/any other stuff - not the  serializer itself
 
3a) lack of transaction is pretty huge u can build generic wrapper on transactions(read/write ones)
if u want fully generic api each table append "step" to the transaction and at the end you have non generic  execute call that does the transaction
3b)same to 3a but lack of batchGet/Batch write on multiple table instances

4) missing opportunity have way to validate the tables pk/sk
aginst real table in some nice way(you have dynamodb api can compare schema to the existing table schema - this will allows checking of bugs at CI/start time of app instead of runtime and having it failed silently

PR #3328, opened 2026-02-19
credentials/ssocreds: fix dropped error

alrs:
This fixes a dropped `err` variable in `credentials/ssocreds`. Unit tests continue to pass after this change.


Madrigal:
I agree that this was likely a miss when it was first implemented. 

However, this can have more important consequences, since if an invalid token was received we would have previously ignored the issue and returned a zero-d time and we'd now throw an error on unmarshal.

Have you encountered this in your workflow?

alrs:
I haven't run in to this. I go after swallowed errors because of the difficulties they present when trying to debug.

If this is too big a change, I can just turn it into a warning and be rid of `os.Exit(1)`. If that is still a bridge too far, I'm happy to close this PR.

PR #3329, opened 2026-02-19
parameterize generate-dev

lucix-aws:
see source comment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,8 @@ REPOTOOLS_CALCULATE_RELEASE_VERBOSE_FLAG=-v=${REPOTOOLS_CALCULATE_RELEASE_VERBOS
 
 REPOTOOLS_CALCULATE_RELEASE_ADDITIONAL_ARGS ?=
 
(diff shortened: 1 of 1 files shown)


Madrigal:
Love the idea (I'm running into this right now), ideally there would be a way to do this to also just generate the service you're interested in, since that's what takes the most time.

My workflow right now is something like

```
// make changes to smithy-go
cd ../smithy-go && make && cd ../aws-sdk-go-v2 && SMITHY_GO_BUILD_API=com.amazonaws.s3#AmazonS3 make smithy-build && gofmt -w -s service/s3
```

and then manually check changes

Madrigal (approved):

lucix-aws:
Not so coincidentally I have almost the exact same thing but with aliases

```smg && smgb && go2 && SMITHY_GO_BUILD_API=com.amazonaws.sqs SMITHY_GO_EXPERIMENTAL_SERDE=com.amazonaws.sqs make generate-dev && cd service/sqs && go build ./...```

Issue #3330, opened 2026-02-23
config: tilde (~) not expanded in AWS_CONFIG_FILE and AWS_SHARED_CREDENTIALS_FILE environment variables


LinguineCode:
## Describe the bug

When `AWS_CONFIG_FILE` or `AWS_SHARED_CREDENTIALS_FILE` is set to a path starting with `~` (e.g. `~/.aws/config`), the Go SDK does not expand the tilde to the user's home directory. The literal string `~/.aws/config` is passed to file open calls, which fails because no such path exists.

This is inconsistent with:

1. **botocore (Python SDK)**, which calls `os.path.expanduser()` on these paths before opening them
2. **The AWS SDKs and Tools Reference Guide**, which [specifies](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html) that `~` followed by `/` (or the platform path separator) at the start of a shared config/credentials file path should resolve to the user's home directory

## Expected Behavior

`AWS_CONFIG_FILE=~/.aws/config` should resolve to `$HOME/.aws/config` (e.g. `/home/user/.aws/config`) before the SDK attempts to open the file, consistent with other AWS SDKs and the shared configuration specification.

## Current Behavior

The SDK reads the raw environment variable value via `os.Getenv()` in `config/env_config.go` and stores it without any path expansion. When it later attempts to open `~/.aws/config` as a literal file path, it fails.

## Reproduction Steps

```bash
export AWS_CONFIG_FILE=~/.aws/config
export AWS_PROFILE=my-profile
# Any Go program using aws-sdk-go-v2 to load config will fail
```

This commonly affects users of [direnv](https://direnv.net/) who set `AWS_CONFIG_FILE` in `.envrc` files, and any environment where these variables are set with `~` rather than the expanded `$HOME`.

## Solution

The SDK's `internal/shareddefaults` package already has a `UserHomeDir()` function used to construct default paths. A small `ExpandHomePath()` helper can be added alongside it and applied to the two `os.Getenv()` calls in `NewEnvConfig()`:

- `config/env_config.go` line 343: `cfg.SharedCredentialsFile`
- `config/env_config.go` line 344: `cfg.SharedConfigFile`

I have a PR ready with the fix and tests.

## References

- [AWS SDKs and Tools Reference Guide - Shared config and credentials files: Location](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html)
[...]

PR #3331, opened 2026-02-23
config: expand tilde (~) in AWS_CONFIG_FILE and AWS_SHARED_CREDENTIALS_FILE env vars

LinguineCode:
## Description

Fixes #3330

When `AWS_CONFIG_FILE` or `AWS_SHARED_CREDENTIALS_FILE` is set to a path starting with `~` (e.g. `~/.aws/config`), the SDK passes the literal string to file open calls without expanding the tilde to the user's home directory. This is inconsistent with botocore and with the [AWS SDKs and Tools Reference Guide](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html), which specifies that `~` followed by `/` at the start of a file path should resolve to the home directory.

## Changes

- **`internal/shareddefaults/shared_config.go`**: Added `ExpandHomePath()` helper that expands a leading `~/` to `UserHomeDir()`. This function is placed alongside the existing `UserHomeDir()` it depends on.
- **`config/env_config.go`**: Wrapped the two `os.Getenv()` calls for `AWS_CONFIG_FILE` and `AWS_SHARED_CREDENTIALS_FILE` with `shareddefaults.ExpandHomePath()`.

## Testing

- Added `internal/shareddefaults/expand_home_test.go` with 8 test cases covering: empty string, absolute paths, relative paths, bare `~`, `~/path`, nested paths, tilde mid-path (no expansion), and `~username` (no expansion).
- Added test case 55 to `config/env_config_test.go` verifying end-to-end tilde expansion when `AWS_CONFIG_FILE` and `AWS_SHARED_CREDENTIALS_FILE` are set with `~/` prefixes.

All existing tests continue to pass.

### DIFF
--- a/config/env_config.go
+++ b/config/env_config.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	"github.com/aws/aws-sdk-go-v2/internal/shareddefaults"
 	smithyrequestcompression "github.com/aws/smithy-go/private/requestcompression"
 )
 
@@ -340,8 +341,8 @@ func NewEnvConfig() (EnvConfig, error) {
 	setStringFromEnvVal(&cfg.Region, regionEnvKeys)
 	setStringFromEnvVal(&cfg.SharedConfigProfile, profileEnvKeys)
 
-	cfg.SharedCredentialsFile = os.Getenv(awsSharedCredentialsFileEnv)
-	cfg.SharedConfigFile = os.Getenv(awsConfigFileEnv)
+	cfg.SharedCredentialsFile = shareddefaults.ExpandHomePath(os.Getenv(awsSharedCredentialsFileEnv))
+	cfg.SharedConfigFile = shareddefaults.ExpandHomePath(os.Getenv(awsConfigFileEnv))
 
 	cfg.CustomCABundle = os.Getenv(awsCABundleEnv)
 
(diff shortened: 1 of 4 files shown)


LinguineCode:
Hey folks -- quick note of context on this PR.

I've been a systems architect for 30 years and an exclusive AWS shop for the last 15. Not "multi-cloud with a preference" -- I mean I have never taken a gig on another cloud provider. AWS is my platform. I've been a proponent and unashamed fanboy since the days when people still asked "but what if Amazon shuts it down?"

In all that time, I've been a consumer of open source -- building my career and my clients' businesses on the shoulders of work that people contributed for free. This is my first PR to an open source project, ever. And I'm glad it's to an AWS repo.

I ran into this bug the hard way: hours of debugging why `direnv exec . terraform plan` worked fine with the AWS CLI but silently failed with Terraform. I used [Claude Code](https://docs.anthropic.com/en/docs/claude-code) to help me hunt down the root cause -- tracing the discrepancy from my `.envrc` through direnv, into the Terraform provider, and finally to the two `os.Getenv()` calls in this SDK that don't expand `~` the way botocore does. Once we understood the problem, Claude Code helped me navigate this codebase, write the fix and tests, and put this PR together.

I've spent 30 years benefiting from open source without ever contributing back. The fact that I can now go from "this is broken and I'm frustrated" to "here's a tested fix with an issue and a PR" in a single sitting -- that genuinely moves me. The barrier to entry just got a whole lot lower, and I hope this is the first of many.

Looking forward to feedback and happy to iterate on anything.

PR #3338, opened 2026-03-03
read the correct signing name for sigv4a

lucix-aws:
As you can see in the diff we were always reading the sigv4 property. In practice this never mattered because all of the existing AWS endpoint rules set both (to the same thing) for v4 and v4a, but this will future proof us in that regard.


Madrigal (approved):

Madrigal:
Tests are failing with GH shenanigans

> fatal: unable to access 'https://github.com/aws/smithy-go/': The requested URL returned error: 500

PR #3339, opened 2026-03-04
track protocol benchmarks

lucix-aws:
track https://github.com/aws/smithy-go/pull/634

e.g.

```bash
cd internal/protocoltest/jsonrpc10
go test -bench ./...
```


Madrigal:
Approved since it's just the gen code from the smithy PR

PR #3353, opened 2026-03-18
add sagemakerruntimehttp bidirectional stream API

lucix-aws:


Madrigal:
Was finally able to look into this. We have an error on how we handle Event Stream header values

We write these values as StringValue https://github.com/aws/aws-sdk-go-v2/blob/main/aws/protocol/eventstream/header.go#L142

But we try to read them as *StringValue https://github.com/aws/aws-sdk-go-v2/blob/main/internal/protocoltest/awsrestjson/deserializers.go#L13733

This operation fails because of this, so we would need to fix that before releasing this

lucix-aws:
closing in favor of #3418

Issue #3354, opened 2026-03-20
Lambda Recursion Detection will not work with Lambda Managed Instances


bmoffatt:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

[Lambda recursion detection](https://docs.aws.amazon.com/lambda/latest/dg/invocation-recursion.html), added to `aws-sdk-go-v2` by #2105 in 2023, took a dependency on the environment variable `_X_AMZN_TRACE_ID`. For customers using [Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances.html), for which support was added in [github.com/aws/aws-lambda-go@v1.51.0](https://github.com/aws/aws-lambda-go/releases/tag/v1.51.0), the `aws-sdk-go-v2` recursion detection will no longer work, as the environment variable will not be present. The environment variable cannot be used to pass the trace id, as Lambda Managed Instances environments handle concurrent requests within the same process.

Good news, the `context.Context` passed to handlers by `github.com/aws/aws-lambda-go`, includes the trace id. This is how the integration was done with the (recently deprecated) https://github.com/aws/aws-xray-sdk-go. 

```go
traceID, ok := ctx.Value("x-amzn-trace-id").(string)
```

Here's where it is set today: https://github.com/aws/aws-lambda-go/blob/v1.53.0/lambda/invoke_loop.go#L72

And where it was set in 2017: https://github.com/aws/aws-lambda-go/blob/v1.0.0/lambda/function.go#L53

That `_X_AMZN_TRACE_ID` is also set, is a more 'recent' (2020) addition, done to resolve a documentation inconsistency https://github.com/aws/aws-lambda-go/pull/269

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

X-Amzn-Trace-Id to be added to request headers when the sdk is used with Lambda Managed Instances

### Current Behavior

Header will not be present

### Reproduction Steps

N/A

### Possible Solution

I suggest switching `aws-sdk-go-v2` to looking for the context value, and falling back to the environment variable for back-compat with customers who're passing `context.Background()` rather than the lambda provided `context.Context`. 

### AWS Go SDK V2 Module Versions Used

[...]

PR #3357, opened 2026-03-23
In Lambda recursion detection middleware, fetch the trace id from context before the environment variable

bmoffatt:
# **PLEASE READ BEFORE CONTINUING**

**DO NOT** submit pull requests that directly modify generated source files, e.g. `/service/s3/api_client.go`. Generated source files will always include an identifying header:

```
// Code generated by smithy-go-codegen DO NOT EDIT.
```

Manual changes to these files will be overwritten by code generation that occurs as part of the daily SDK release process.

**DO NOT** submit pull requests that directly modify files in the `/codegen/sdk-codegen/aws-models` folder. These are API model files, owned by each AWS service team, that are updated automatically as part of the daily SDK release process. Local changes to these files will not persist.

If you believe the contents of any of these files need to be changed, please [open an issue](https://github.com/aws/aws-sdk-go-v2/issues/new/choose).

#

See issue: https://github.com/aws/aws-sdk-go-v2/issues/3354

# Changelog

Please **DO NOT** include a changelog entry in your pull request (you may see
what these look like in other PRs submitted directly by SDK team members). We
will take care of this for you.

* Updates `aws/middleware/recursion_detection.go` to grab the trace ID from `ctx.Value("x-amzn-trace-id")`, falls back to `os.LookupEnv("_X_AMZN_TRACE_ID")` for back compat.
* Updates test cases to `aws/middleware/recursion_detection_test.go`


lucix-aws:
CI is failing, you can ignore integ/codegen tests but the go tests need to pass

lucix-aws:
this is fine but we should really be using the _context_ as a fallback, not the env. i don't want the values being swapped out from under anyone on the off chance that they are somehow different today for environments where both would be set. this is also sort of a special implementation detail of go (all the SDKs have this recursion detection and they'd all be looking at the env) so i'd rather have the code do the blessed thing(TM) first and then check the language-specific thing as a followup

bmoffatt:
Yeah I can flip the order. I suppose that'd be less foot-gun-y for sdk users not using Lambda Managed Instances that are plumbing like `context.TODO()` instead of the lambda context.

I'm not sure however that this would be *more consitent* with the other SDKs though. The java sdk equivalent of this change falls back to the env (well, [actually a system property](https://github.com/aws/aws-sdk-java-v2/blob/512738eeb842bb4cce78d03c53c8117edb0b7309/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/interceptor/TracingSystemSetting.java#L24-L27) now... [it's a long story...](https://github.com/aws/aws-lambda-java-libs/issues/261#issuecomment-1213635113))

ref: https://github.com/aws/aws-sdk-java-v2/pull/6433/changes

Ditto with the node.js SDK https://github.com/aws/aws-sdk-js-v3/blob/a84cde64ab1985fbc1058f3188f82dd1b34428ce/packages-internal/middleware-recursion-detection/src/recursionDetectionMiddleware.ts#L38-L40

Issue #3187, opened 2025-09-10
Clock-skew offset is not updated after time re-sync, first attempts always use wrong timestamps


eattker:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Overview

This library automatically adjusts the signing timestamp when the server clock differs from the client clock and responds with "RequestTimeTooSkewed" error. However when the clocks became synchronized again, the library does not update the offset back to the newly calculated skew (close to zero).

When we shift the client time to the future, we can see that the signing timestamps of the first attempts are in the future, then it adjust the timestamp (using negative offset) to be in the present in the retries. This works when the offset is not close to zero.
However it can't adjust the timestamp for the second try, the proper timestamp is sent only in the third attempt, see later. All subsequent requests will use the correct timestamp at the first attempt. Requests without a "retryer" will be succeeded.

When we shift back the client time to present, it always sends the first request with timestamp in the past (due to negative offset), the second attempt will use the correct timestamp. But all subsequent requests still uses timestamp in past (due to negative timestamp) at the first attempt and adjust it in the second attempt. Requests without "retryer" will fail because of the wrong timestamp.

### Observer timeline:

Time on both side: **2025**0910050607Z  
Shift the client time forward by 1 year to **2026**0910050607Z

Request 1
-> Attempt 1 : **2026**0910050607Z
<- Response 1 : RequestTimeTooSkewed, **2025**0910050607Z
-> Attempt 2 : **2026**0910050607Z
<- Response 2 : RequestTimeTooSkewed, **2025**0910050607Z
-> Attempt 3 : **2025**0910050607Z
<- Response 3 : OK

Request 2
-> Attempt 1 : **2026**0910050607Z
<- Response 1 : OK

Shifting the client time back to present  (**2025**0910050607Z)

Request 3
-> Attempt 1 : **2024**0910050607Z
<- Response 1 : RequestTimeTooSkewed, **2025**0910050607Z
-> Attempt 2 : **2025**0910050607Z
<- Response 2 : OK

Request 4 ...
-> Attempt 1 : **2024**0910050607Z
<- Response 1 : RequestTimeTooSkewed, **2025**0910050607Z
-> Attempt 2 : **2025**0910050607Z
<- Response 2 : OK

[...]

PR #3359, opened 2026-03-25
fix stale skew and delayed skew healing

lucix-aws:
closes #3187

fixes two separate issues that prevented a corrected skew from making it back to the client right away, and that delayed a skew from being reused for an attempt.

unfortunately i forgot to not rebase so the diff is large

### DIFF
--- a/internal/protocoltest/awsrestjson/deserializers.go
+++ b/internal/protocoltest/awsrestjson/deserializers.go
@@ -13715,6 +13715,19 @@ func awsRestjson1_deserializeEventMessageHeadersEvent(v *types.HeadersEvent, msg
 	}
 
 	{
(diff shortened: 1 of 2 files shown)


lucix-aws:
FYI to the reviewer, the only thing you really need to look at is aws/retry/middleware.go. The stale context and skew persistence logic are both there.

The old middleware that handled the offset to context and storing back to the client atomic was deprecated and unregistered from middleware, that's now handled in the above.

There's also a new kitchen sink test that tests skew with a real client.

lucix-aws:
Tests passed, I'm going to regenerate for https://github.com/aws/aws-sdk-go-v2/pull/3361 and then merge

PR #3361, opened 2026-03-26
fix order of generated event header handlers

lucix-aws:
in smithy 1.64 it looks like some new headers were added to protocol tests and we never properly sorted those, so that codegenned output has been shuffling around for months now. this fixes it.


Madrigal:
finally!

PR #3363, opened 2026-03-26
remove staling workflows

lucix-aws:
We feel that these are overly hostile to users, and they can sometimes suppress issues that should stay open, so we are removing them.


Madrigal:
#nostale 

(I do wonder if eventually we'll bring something similar back if the number of issues grows huge, but I'm fine on removing them altogether and adjust later on)

PR #3364, opened 2026-03-26
reinstate transfer manager Location output on upload

### LINKED ISSUE #3323: feature/s3/transfermanager: v1 parity: location output
see #3317 

> UploadObjectOutput.Location is always nil

lucix-aws:
closes #3323 

**There is a slight behavior change here.** The source comment calls it out but basically v1 used this recordLocationClient thing for both single- and multi-upload, rather than copying the complete MPU location value for the latter. The API-borne location %-encodes the path. I imagine it's better to keep that behavior for both paths here (so basically ignore Location for MPU) but lmk if you disagree.


wty-Bryant (approved):

magnetikonline:
Hello @lucix-aws, are we any closer to getting this merged into mainline?

Madrigal (approved):

PR #3368, opened 2026-03-27
drop service/internal/benchmark

lucix-aws:
These have never been deployed in any kind of CI and they're just taking up repo space, so we're removing them. We're putting a placeholder first because completely removing it will break some internal references (that were also literally just removing it in internal builds lol). Will follow up with a full removal once that's addressed.


wty-Bryant (approved):

lucix-aws:
close/reopen to rerun checks, there was a network timeout

PR #3369, opened 2026-03-30
Revert "drop service/internal/benchmark"

wty-Bryant:
Reverts aws/aws-sdk-go-v2#3368, our repo tool will mark a module without go.mod as dropped module and fail if it contains any .go file


lucix-aws (approved):

Madrigal (approved):
Previous changed cause some issues with our release system. This is still dead weight, so we'll likely still remove it later

PR #3377, opened 2026-04-10
fix(service/s3/validators): validate empty bucket to prevent panic in dnsCompatibleBucketName method

hanagantig:
## Description  
Fixes a panic in `dnsCompatibleBucketName` when an empty bucket name is provided.

The function assumes a non-empty string and directly accesses `bucket[0]`, which causes a runtime panic when `bucket == ""`.

**Before (bug):**  
- Empty bucket name (`""`) is not handled  
- Code accesses `bucket[0]` → out-of-bounds panic  

**After (fix):**  
- Empty bucket name is explicitly handled  
- Function safely returns `false` before any indexing  

---

## Root Cause  
`dnsCompatibleBucketName` directly indexes the first character:

```go
if !((bucket[0] > 96 && bucket[0] < 123) || (bucket[0] > 47 && bucket[0] < 58)) {
    return false
}

### DIFF
--- a/service/s3/internal/customizations/update_endpoint.go
+++ b/service/s3/internal/customizations/update_endpoint.go
@@ -278,7 +278,7 @@ func hostCompatibleBucketName(u *url.URL, bucket string) bool {
 // dnsCompatibleBucketName returns true if the bucket name is DNS compatible.
 // Buckets created outside of the classic region MUST be DNS compatible.
 func dnsCompatibleBucketName(bucket string) bool {
-	if strings.Contains(bucket, "..") {
+	if bucket == "" || strings.Contains(bucket, "..") {
 		return false
 	}
 
--- a/service/s3/internal/customizations/update_endpoint_internal_test.go
+++ b/service/s3/internal/customizations/update_endpoint_internal_test.go
@@ -74,3 +74,49 @@ func TestRemoveBucketFromPath(t *testing.T) {
 		})
 	}
 }
+
+func TestDNSCompatibleBucketName(t *testing.T) {
+	cases := []struct {
+		bucket   string
+		expected bool
+	}{
+		// valid bucket names
+		{"abc", true},
+		{"a", true},
+		{"ab", true},
+		{"my-bucket", true},
+		{"my.bucket", true},
+		{"my-bucket-123", true},
+		{"123bucket", true},
+		{"a1b2c3", true},
+		{"bucket.with.dots", true},
+		// empty bucket
+		{"", false},
+		// consecutive dots
+		{"a..bc", false},
+		// invalid characters
+		{"a$b$c", false},
+		{"bucket_name", false},
+		{"UPPERCASE", false},
+		{"Bucket", false},
+		{"bucket name", false},
+		{"bucket@name", false},
+		// starts with invalid character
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Madrigal** [comment]
Unfortunately, you are making code changes to a generated file. See the [first line of that file](https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/validators.go#L1C1-L1C52)

```
// Code generated by smithy-go-codegen DO NOT EDIT.
```

The service model is the source of truth for this, and it's owned by S3 directly, not by the SDK, so we can't accept this PR. You can create an issue and we can own reaching out to S3 about adding this validation on their model.

THAT SAID, it seems like the issue you are facing is with one of our customizations on `dnsCompatibleBucketName`. If instead of assuming a non-empty bucket we instead returned `false` on empty bucket (since empty names are not DNS compatible), would that solve your issue?

hanagantig:
hello @Madrigal 
Thanks for the clarification — you’re absolutely right, I missed that this is a generated file. Sorry about that.

My initial thought was to handle this at the validation level to prevent similar issues in other parts of the code that might process the bucket name. But given that the validation is model-driven, I understand why this approach doesn’t fit here.

I’ve instead updated dnsCompatibleBucketName to safely handle empty input and added a few unit tests to cover this case.

To be honest, I’m not deeply familiar with all the context here — I just ran into this issue while using the library in my project and wanted to help address it.

Let me know if this approach looks good to you 👍

hanagantig:
Other than that, I’ll also open an issue to address this at the model level so that proper validation is enforced upstream as well.

lucix-aws:
This code path isn't actually hit in the modern endpoints 2.0 route is it?

lucix-aws:
Can you show us your client config? Specifically how you're configuring the endpoint resolver

hanagantig:
after digging into my setup, I found that this code path is actually hit when using `WithEndpointResolverWithOptions`.

I noticed it’s marked as deprecated, but it still works when I provide my custom endpoint URL. While testing this configuration, I ran into the panic caused by an empty bucket name.

So it seems this path is still reachable in certain setups.

lucix-aws:
If you don't use that and instead set BaseEndpoint on the client to your custom endpoint does that not work?

lucix-aws:
That's the correct way to set a custom endpoint for a service, essentially.

hanagantig:
yes, I tried using `BaseEndpoint` and it works as expected — no issues there.
The panic only occurs with the deprecated configuration `WithEndpointResolverWithOptions`.

lucix-aws:
Thanks for the PR but given the extreme sensitivity of endpoint resolution (especially this deprecated legacy endpoint resolution which unfortunately probably has many callers out in the wild still) I would prefer to not change this code unless absolutely necessary. Since the v2 path (BaseEndpoint) works I think we should just be recommending that to users instead.

hanagantig:
My only concern is that since the deprecated configuration is still usable and likely present in existing setups, hitting a runtime panic for something like an empty bucket might be a bit harsh. Ideally, even in legacy paths, it would fail gracefully with an error rather than panic.

But I see your points. Thanks @lucix-aws for the clarification and your time.

PR #3381, opened 2026-04-14
fold internal/ini into config

lucix-aws:
internal/ini is only used by config. having an internal/ module is effectively both pointless and puts strain on the public module proxy, since the proxy hosts the tags but they're impossible to actually consume outside of the SDK.

I'm looking into doing this for a handful of our other "internal" modules (e.g. internal/v4a), we'd basically just fold them into the core aws-sdk-go-v2 runtime.


Madrigal:
How do we know this change is safe to release for existing customer? Can you elaborate how would it work for someone that today is depending on both the main and the internal module?

lucix-aws:
They couldn't have both, the only thing that depends on internal/ini _is_ config itself, so they're either on a version of config that has it folded in or they're on an older one that doesn't. When they upgrade from not folded to folded, they'd lose the "transitive" version.

Madrigal:
OK so to clarify, if you have a `go.mod` like

```
require (
	github.com/aws/aws-sdk-go-v2 v1.41.4
	github.com/aws/aws-sdk-go-v2/config v1.32.12
	github.com/aws/aws-sdk-go-v2/service/uxc v1.0.0
)

require (
    // ...
	github.com/aws/aws-sdk-go-v2/internal/ini v1.8.6 // indirect
)
```

It will be replaced with

```
require (
	github.com/aws/aws-sdk-go-v2 v1.41.4
	github.com/aws/aws-sdk-go-v2/config v1.32.13 // version bump
	github.com/aws/aws-sdk-go-v2/service/uxc v1.0.0
)

require (
    // ...
	// no internal/ini indirect
)
```

The `internal` package means that no external module will be importing this, so there's 0 chance this breaks someone importing it directly https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages

Madrigal (approved):

lucix-aws:
So yes, but let's consider another one of these random internal modules, like presigned-url. Suppose you did have two services that used it, one upgraded w/ the folded version and the other not. You'd just have

```
require (
   github.com/aws/aws-sdk-go-v2/service/presignable v1.0.0 // old, depends on internal/presigned-url
   github.com/aws/aws-sdk-go-v2/service/presignable2 v1.0.1 // new, depends on aws-sdk-go-v2 v1.2.0 which has internal/presigned-url folded in
)

require (
   github.com/aws/aws-sdk-go-v2 v1.2.0 // required by presignable2, has the code of internal/presigned-url
  github.com/aws/aws-sdk-go-v2/internal/presigned-url v1.0.0 // the final version of this before we folded it in and stopped publishing new ones.
)
```

The two things would exist independently and wouldn't interact at all. If you upgraded presignable2 then internal/presigned-url would just go away.

PR #3383, opened 2026-04-15
fold internal/ into main runtime module

lucix-aws:
same reason as #3381


lucix-aws:
This PR appears to be completely broken due to some sort of GitHub outage, as the diff isn't showing. Closing in favor of #3384

PR #3388, opened 2026-04-16
s3/transfermanager: fix panic in concurrentReader when len(p) is smaller than cap(p)

islishude:
## Summary

This fixes a panic in `concurrentReader.read` when the caller passes a slice where `len(p) < cap(p)`.

The reader was using `cap(p)` to:
- detect empty input
- calculate the max part index to process
- decide whether `p[index:]` was safe to slice

That is not safe for an [`io.Reader`](https://pkg.go.dev/io#Reader), because reads must be bounded by `len(p)`, not `cap(p)`. 

> Read reads up to len(p) bytes into p. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered. 

In this case, the old logic could treat an offset as valid and then panic with a slice-bounds error when evaluating `p[index:]`.

## Changes

- replace `cap(p)` with `len(p)` in `concurrentReader.read`
- keep part selection and direct chunk writes aligned with the actual readable window
- add regression coverage for both paths that previously panicked:
  - reading from a buffered chunk
  - reading from a newly received chunk

## Testing

- added `TestConcurrentReaderReadUsesSliceLenForBounds`
- ran the concurrent reader test file successfully

### DIFF
--- a/feature/s3/transfermanager/concurrent_reader_test.go
+++ b/feature/s3/transfermanager/concurrent_reader_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+
 	"io"
 	"math"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


feature/s3/transfermanager/concurrent_reader_test.go
```diff
@@ -3,16 +3,68 @@ package transfermanager
...
+			r.buf[1] = &outChunk{
+				body:   bytes.NewReader([]byte("chunk")),
+				index:  1,
+				length: int64(len("chunk")),
+			}
+		},
+		"received chunk": func(r *concurrentReader) {
+			r.capacity = 1
+			r.ch <- outChunk{
+				body:   bytes.NewReader([]byte("chunk")),
```
wty-Bryant, line 34: So this case just triggers the index 1 chunk can be received by `read()` but out of scope of len(p) right? I asked since theoretically index 1 chunk will only be received when capacity is at least 2

PR #3389, opened 2026-04-16
s3/transfermanager: avoid double-closing concurrentReader channel after read errors

islishude:
## Summary

Fix a panic in transfermanager concurrentReader when Read is called again after a non-EOF error.

## Problem

When concurrentReader.Read encounters a non-EOF error, it closes r.ch to unblock the internal reader goroutine. If the caller invokes Read again, the method previously continued through the same cleanup path and attempted to close r.ch a second time, causing this panic: `panic: close of closed channel`

## Changes

Return the stored non-EOF error immediately on subsequent Read calls
Guard r.ch close with sync.Once so channel shutdown is idempotent
Add a regression test covering repeated Read calls after a download error

## Test

`go test ./... in feature/s3/transfermanager`

### DIFF
--- a/feature/s3/transfermanager/concurrent_reader.go
+++ b/feature/s3/transfermanager/concurrent_reader.go
@@ -42,7 +42,6 @@ type concurrentReader struct {
 	ctx context.Context
 	m   sync.Mutex
 	wg  sync.WaitGroup
-	co  sync.Once
 
 	err error
 }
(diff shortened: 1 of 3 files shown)


islishude:
Hi there, could you team please take a look? and please also #3388

wty-Bryant:
sorry, dealing with some other feature parity this week, will look into those 2 PRs recently

feature/s3/transfermanager/concurrent_reader.go
```diff
@@ -100,7 +106,9 @@ func (r *concurrentReader) Read(p []byte) (int, error) {
 	r.wg.Wait()
 
 	if e := r.getErr(); e != nil && e != io.EOF {
-		close(r.ch)
+		r.co.Do(func() {
```
wty-Bryant: nit: this once.Do might not be necessary based on your change in line 55:
1. if this non EOF error comes from this Read(), then close the channel as it is
2. if this error comes is stored by earlier call of Read(), then line55 check will just return

wty-Bryant (approved):

PR #3392, opened 2026-04-20
Auto-discover protocoltest instead of manually adding them to build

Madrigal:
Currently, protocol tests are created by manually updating [codegen/protocol-test-codegen/smithy-build.json](https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/protocol-test-codegen/smithy-build.json), where we manually input which service to use according to a template.

This means when a new protocol test gets added, we need to read the release notes and remember to manually add them to this file.

With this PR, protocol tests are instead read from the runtime class path. We find all Smithy services available and generate the `smithy-build.json` file manually. This is done on as a new step `generate-smithy-build` on protocol test gradle file which is attached to the `build` step, which now generates the whole content of `internal/protocoltest`.

This PR leaves existing tests exactly as-is, name and everything. The diff of `smithy-build.json` only indicates changes on position, but all entries are the same. 

Removing services in this script instead of [ProtocolUtils in smithy-go](https://github.com/aws/smithy-go/blob/main/codegen/smithy-go-codegen/src/main/java/software/amazon/smithy/go/codegen/ProtocolUtils.java) since we're removing whole services instead of just individual tests.

Added a step to add manual files to `internal/protocoltest` since there are some tests that we wrote manually and would be lost.


internal/protocoltest/README.md
```diff
@@ -0,0 +1,43 @@
+# Protocol Test Codegen – Manual Files
```
wty-Bryant, line 1: nit: why we need 2 same README here?

codegen/protocol-test-codegen/build.gradle.kts
```diff
@@ -51,15 +55,117 @@ tasks.create<SmithyBuild>("buildSdk") {
...
+
+            projectionsBuilder.withMember(projName, Node.objectNodeBuilder()
+                .withMember("transforms", Node.fromNodes(
+                    Node.objectNodeBuilder()
+                        .withMember("name", "includeServices")
+                        .withMember("args", Node.objectNode()
+                            .withMember("services", Node.fromStrings(shapeId)))
+                        .build(),
+                    Node.objectNodeBuilder()
+                        .withMember("name", "removeUnusedShapes")
```
wty-Bryant, line 123: nit: how later build gradle tasks use this removedUnusedShapes node?

internal/protocoltest/awsrestjson/snapshot/api_op_TestNoInputNoPayload.go.snap
```diff
@@ -1,40 +0,0 @@
-TestNoInputNoPayload
```
wty-Bryant, line 1: nit: not familiar with test cases in protocol test, but why we need to delete those 2 snaps?

codegen/protocol-test-codegen/smithy-build.json
```diff
@@ -1,240 +1,281 @@
 {
-  "version": "1.0",
```
lucix-aws, line 2: The sdk-codegen project does the same programmatic generation but we don't check in the file at all, we should do the same here, mainly so someone doesn't edit this one by mistake (but also we just don't need it).

codegen/protocol-test-codegen/manual/README.md
```diff
@@ -0,0 +1,43 @@
+# Protocol Test Codegen – Manual Files
```
lucix-aws, line 1: We might want to just have a GoIntegration that generates this stuff instead of having to have this new mechanism but it's nbd. I imagine it could just gate on "is service in protocoltests namespace" etc.

codegen/protocol-test-codegen/smithy-build.json
Madrigal: Ah, this sounds good to me, will remove it and just keep it as part of `generate`

internal/protocoltest/awsrestjson/snapshot/api_op_TestNoInputNoPayload.go.snap
Madrigal: Snapshot tests point to existing tests. These tests were renamed from `api_op_TestNoInputNoPayload` to `api_op_TestGet...` on [this commit](https://github.com/aws/aws-sdk-go-v2/commit/253d7b135c8d2b547dc15bce26f9a6715498928d#diff-9916533b4f242bdea33221f72bc87b1a2abafe8a13194b396cf3631a79e2d315). However, these snapshot tests were not removed when they were renamed.

codegen/protocol-test-codegen/build.gradle.kts
Madrigal: This is used by `smithy`, not by Gradle. The transformation removes shapes that are not associated to a service or operation, see [official docs](https://smithy.io/2.0/guides/smithy-build-json.html#removeunusedshapes)

This is replicating existing behavior, since we have this transformation on all services, see https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/protocol-test-codegen/smithy-build.json#L13

internal/protocoltest/README.md
Madrigal: It's an artifact of manually copying all files from `manual` to here. There's some discussion about just not checking them in, which may be the way we go instead.

codegen/protocol-test-codegen/manual/README.md
Madrigal: This sounds good, but if we have a GoIntegration writing manual files will mean writing a Java file that creates a Go template, right?

codegen/protocol-test-codegen/smithy-build.json
Madrigal: Removed it from checking it in, but it is kind of annoying that every time I run `generate` the file is there preventing me from switching to a different branch if it exists on that branch.

I could add it to `.gitignore`, but I'm not sure if that would cause any issue with workflows that expect the file to exist.

codegen/protocol-test-codegen/manual/README.md
lucix-aws: correct yes. up to you

lucix-aws (approved):

codegen/protocol-test-codegen/smithy-build.json
Madrigal: I added it go .gitignore

lucix-aws (approved):

PR #3400, opened 2026-04-24
Add preview of changes for standard retry mode behind flag

Madrigal:
Add preview of new standard retry behavior behind `AWS_NEW_RETRIES_2026` flag.

All changes are gated behind the `AWS_NEW_RETRIES_2026` environment variable and have no impact on existing behavior unless the flag is explicitly set to "true". This feature is expected to be enabled by default later in the year.

This is part of a cross-SDK change to standardize how we handle retries, and will be implemented by all AWS SDKs (Java, Python, Rust, etc.)

### Changes

Retry token bucket costs
- Non-throttling retry cost increases from 5 to 14 tokens, reducing retry amplification during service outages.
- Throttling errors now use a discounted cost of 5 tokens (previously, timeouts had a special cost of 10). This
allows more retries when a service explicitly signals "try again later."
- Timeouts are no longer treated differently from other transient errors.

Backoff timing
- Base backoff for non-throttling errors reduced from 1,000ms to 50ms, significantly reducing latency for
transient failures.
- Throttling errors retain the 1,000ms base backoff.
- MAX_BACKOFF is now applied before jitter rather than after.
- DynamoDB and DynamoDB Streams use a 25ms base backoff and 4 max attempts (up from 3).

x-amz-retry-after header
- The SDK now honors the x-amz-retry-after response header, which specifies a server-recommended backoff in
milliseconds.
- The value is clamped between the computed backoff and 5 seconds above it.
- Invalid header values are silently ignored.

Long-polling operations
- Operations marked as long-polling (currently manually adding  SQS `ReceiveMessage`, SFN `GetActivityTask`, SWF `PollForActivityTask/PollForDecisionTask`, but will be implemented as a trait on the service) now back off even when the retry quota is exhausted, preventing request amplification on polling endpoints.
- Codegen support added via LongPollingRetryIntegration, which checks for the LongPollTrait Smithy trait with a
hardcoded fallback for known operations until the trait is widely applied.

Attempt token bookkeeping
- NoRetryIncrement is now only applied on the first attempt's success, not after retries. This prevents double-
counting when the retry token is also refunded on success.

### Backwards compatibility

- All behavioral changes are gated behind AWS_NEW_RETRIES_2026=true. Without the flag, the SDK behaves
identically to before this change.
- Public constants (DefaultRetryCost, DefaultRetryTimeoutCost) retain their original values.
- New public API surface (ThrottlingRetryCost, Throttles, LongPolling on StandardOptions; AddWithLongPolling
wrapper; NewExponentialJitterBackoffWithOptions) is additive only.
- Existing tests explicitly set the flag to false to ensure they validate legacy behavior.

### Testing
- All standard mode test cases from the cross-SDK specification are implemented in retry2026_test.go, covering:
basic retries, max attempts, quota exhaustion, quota recovery, exponential backoff, max backoff, throttling
costs, DynamoDB-specific config, long-polling backoff, x-amz-retry-after header (valid, min-clamped, max-clamped,
invalid), long-polling with throttle errors, and non-retryable errors.
- Concurrent shared retry quota test validates correct token bucket behavior across goroutines.


aws/retry/middleware.go
```diff
@@ -285,10 +305,13 @@ func (r *Attempt) handleAttempt(
 	// Get the retry delay before another attempt can be made, and sleep for
 	// that time. Potentially early exist if the sleep is canceled via the
 	// context.
-	retryDelay, reqErr := r.retryer.RetryDelay(attemptNum, err)
+	retryDelay, reqErr := r.retryer.RetryDelay(attemptNum-1, err)
```
lucix-aws: -1 here why?

aws/retry/middleware.go
```diff
@@ -276,6 +278,24 @@ func (r *Attempt) handleAttempt(
 	// Get a retry token that will be released after the
 	releaseRetryToken, retryTokenErr := r.retryer.GetRetryToken(ctx, err)
 	if retryTokenErr != nil {
+		// Long-polling operations must still back off when quota is exceeded.
+		if newRetries2026() {
+			longPolling := false
+			if std, ok := r.retryer.(*Standard); ok {
+				longPolling = std.IsLongPolling()
```
lucix-aws: do we need both the option _and_ the wrap? it seems like this is just a fact of certain operations so i'm not sure why we'd make it configurable

aws/retry/jitter_backoff.go
```diff
@@ -25,13 +37,53 @@ func NewExponentialJitterBackoff(maxBackoff time.Duration) *ExponentialJitterBac
...
 	}
 }
 
+// ExponentialJitterBackoffOption is a functional option for ExponentialJitterBackoff.
+type ExponentialJitterBackoffOption func(*ExponentialJitterBackoff)
+
+// WithBaseDelay sets the base delay for non-throttle errors.
+func WithBaseDelay(d time.Duration) ExponentialJitterBackoffOption {
+	return func(j *ExponentialJitterBackoff) {
+		j.baseDelay = d
```
lucix-aws, line 51: * am i reading correctly that this and WithThrottleCheck won't be used unless they're in the 2026 path
* this is only used as a vehicle to customize ddb's values, right? could it be internalized? i don't want to add to the retry API surface unless we absolutely have to

Madrigal: > am i reading correctly that this and WithThrottleCheck won't be used unless they're in the 2026 path

Correct

> this is only used as a vehicle to customize ddb's values, right? could it be internalized? i don't want to add to the retry API surface unless we absolutely have to

That's correct. When you talk about it being internalized, did you have anything specific in mind? We could expose `BaseDelay` as a customizable field on the retryer set to the default of 50 ms instead, but unsure if you had something else in mind that would allow us to keep the same external API and still be able to use it from DynamoDB

Madrigal: Added `BaseDelay` as a customizable field instead, I think it's easier to reason about and may allow someone in the future to change how they want to retry

aws/retry/middleware.go
Madrigal: Removed the option

aws/retry/middleware.go
Madrigal: TL;DR this is correct but should be gated, changed and added a test

Our new retries strategy specifies that "When backing off after the first failure, i MUST be 0", so doing this offset to start at `attemptNum-1`. However, you're right that this isn't backwards compatible and missed it on the review. I added a new test that would fail with this behavior and enable this only when the flag is enabled.

lucix-aws (approved):

Issue #3358, opened 2026-03-23
Reading EventStream header values fails with unexpected event header ... with type ...


Madrigal:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

While testing #3353 I noticed that we failed reading event stream headers

2026/03/23 15:58:07 Failed to end session: unexpected event header DataType with type eventstream.StringValue:

This is because we write these values as `StringValue` (or really any `XXXValue`) https://github.com/aws/aws-sdk-go-v2/blob/main/aws/protocol/eventstream/header.go#L142

But we try to read them as `*StringValue` https://github.com/aws/aws-sdk-go-v2/blob/main/internal/protocoltest/awsrestjson/deserializers.go#L13733

At the time of writing, no services live on the SDK (besides sagemakerruntimehttp2) reads from eventstream headers

```
$ rg 'hv, ok := headerValue' service
(no results)
```

Since we're in the middle of a refactor of schema deserializing, I'd rather hold off making changes until we roll this out and have more extensive event stream testing

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Expected to be able to read this value

### Current Behavior

Error is returned

### Reproduction Steps

On PR https://github.com/aws/aws-sdk-go-v2/pull/3353

### Possible Solution

This is already addressed on the new schema serialization https://github.com/aws/smithy-go/pull/641/

If we had to roll the change before migration, we would need to change the values here to be non-pointable https://github.com/aws/aws-sdk-go-v2/blob/main/codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AwsEventStreamUtils.java#L2124

### AWS Go SDK V2 Module Versions Used

latest

### Compiler and Version used

x

### Operating System and version

x

lucix-aws:
OBE #3412

PR #3418, opened 2026-05-20
reinstate sagemaker bidi api

lucix-aws:
#3358 was fixed with the introduction of event stream protocol tests so we can release this API now


service/sagemakerruntimehttp2/deserializers.go
```diff
@@ -1,3 +1,871 @@
...
+
+func awsRestjson1_deserializeEventMessageResponsePayloadPart(v *types.ResponsePayloadPart, msg *eventstream.Message) error {
+	if v == nil {
+		return fmt.Errorf("unexpected serialization of nil %T", v)
+	}
+
+	{
+		headerValue := msg.Headers.Get("CompletionState")
+		if headerValue != nil {
+			hv, ok := headerValue.(eventstream.StringValue)
```
Madrigal, line 226: This is the actual thing that had to change for future reference, on #3353 it was `*eventstream.StringValue` which was incorrect

Madrigal (approved):
Discussed offline. I did some testing in the past for this operation, and while I can't find the exact testing with the live service, I do remember that after fixing that eventHeader I was able to make a call to the service, so given that we now also have eventstream test (#3412) we're comfortable pushing this

PR #3408, opened 2026-05-09
fix(s3/transfermanager): respect MultipartUploadThreshold in upload decision

nghiack7:
## Problem

`Options.MultipartUploadThreshold` is documented and defaults to 16 MB, but the upload decision logic in `nextReader()` never reads it. The threshold used is `PartSizeBytes` (default 8 MB), so:

- Files between 8 MB and 16 MB are uploaded via **multipart** even though the caller expects a single `PutObject` for files below their configured threshold.
- Setting a custom `MultipartUploadThreshold` has no effect.

Reported in #3333.

## Root Cause

```go
// Before: threshold is PartSizeBytes, MultipartUploadThreshold ignored
r := io.LimitReader(u.in.Body, u.options.PartSizeBytes)
firstPart, _ := io.ReadAll(r)
if int64(len(firstPart)) < u.options.PartSizeBytes {
    return ..., io.EOF  // → PutObject
}
return ..., nil  // → multipart
```

## Fix

Read up to `max(MultipartUploadThreshold, PartSizeBytes)` bytes. If all data fits within the threshold (EOF), use `PutObject`. Otherwise use multipart and re-inject any excess lookahead bytes back into the body via `io.MultiReader` so subsequent part reads remain at the normal `PartSizeBytes` granularity.

```go
// After: use MultipartUploadThreshold as the decision boundary
readLimit := u.options.MultipartUploadThreshold
if u.options.PartSizeBytes > readLimit {
    readLimit = u.options.PartSizeBytes
}
lookahead, _ := io.ReadAll(io.LimitReader(u.in.Body, readLimit))
if int64(len(lookahead)) < u.options.MultipartUploadThreshold {
    return ..., io.EOF  // → PutObject
}
// re-inject overflow so part sizes stay at PartSizeBytes
if int64(len(lookahead)) > u.options.PartSizeBytes {
    overflow := lookahead[u.options.PartSizeBytes:]
    u.in.Body = io.MultiReader(bytes.NewReader(overflow), u.in.Body)
    lookahead = lookahead[:u.options.PartSizeBytes]
}
return ..., nil  // → multipart
```

## Behaviour Changes

| File size | Before | After |
|-----------|--------|-------|
| < PartSizeBytes (8 MB) | PutObject | PutObject |
| PartSizeBytes ≤ size < MultipartUploadThreshold (8–16 MB) | **Multipart** (bug) | **PutObject** (fixed) |
| ≥ MultipartUploadThreshold (16 MB) | Multipart | Multipart |
[...]

Issue #3407, opened 2026-05-05
Configurable MaxUploadParts for transfermanager UploadObject


Nouuu:
### Describe the feature

`feature/s3/transfermanager`’s `Client.UploadObject` enforces the S3 multipart hard limit of 10,000 parts via an internal constant (`defaultMaxUploadParts = 10000`), but does not expose any user-configurable setting to cap the maximum number of parts to a lower value (e.g., 1000).

The deprecated `feature/s3/manager` uploader exposed `Uploader.MaxUploadParts`, which allowed callers to intentionally trade fewer part requests for larger part sizes.

I’m requesting an equivalent configuration option in `transfermanager.Options` (or similar) so users can set a maximum part count like **1000**.

### Use Case

We upload very large objects and want to limit the number of S3 `UploadPart` requests for operational reasons (request volume, latency, throttling, cost/accounting, and easier reasoning about behavior).

Example: we want multipart uploads to use **at most 1000 parts**. Today, the only lever is to manually calculate `PartSizeBytes`, and this is awkward, especially when the upload size is known only via `ContentLength`/`io.Seeker` and you’d prefer the library to do the part-size math (as the old uploader did).

Also, for unknown-size streaming uploads, having a configured maximum part count provides a predictable failure mode earlier than hitting the hard S3 10,000 limit.

### Proposed Solution

Add a configurable max-part option to `feature/s3/transfermanager.Options`, e.g.:

```go
type Options struct {
    // ...
    // MaxUploadParts caps the number of multipart upload parts used by UploadObject.
    // Default should preserve current behavior (10000).
    MaxUploadParts int32
}
```

Suggested semantics:
- Default `MaxUploadParts` = 10000 (preserve current behavior).
- Validate `1 <= MaxUploadParts <= 10000`.
- When object size is known (via `io.Seeker` or `UploadObjectInput.ContentLength`), automatically increase `PartSizeBytes` so:
  - `ceil(objectSize / PartSizeBytes) <= MaxUploadParts` (e.g., 1000).
- When size is unknown (streaming), error once the part count would exceed `MaxUploadParts`, with a message similar to the old uploader (“exceeded configured MaxUploadParts … adjust PartSizeBytes …”).

This would restore feature parity with `feature/s3/manager.Uploader.MaxUploadParts` while keeping the S3 hard limit enforced.

### Other Information

Relevant code:
- `defaultMaxUploadParts` is internal and fixed:
[...]

Issue #3341, opened 2026-03-07
transfermanager: AbortMultipartUpload fails when context is done.


proost:
### Describe the feature

When a multipart upload fails due to context done, multiUploader.fail() calls AbortMultipartUpload using the same context. The abort request never reaches S3, leaving orphaned parts that incur storage costs.
                                                                                                                                                                                                                                 
  Add a MultipartCleanupTimeout option to transfermanager.Options that creates a fresh context with a timeout for cleanup calls, so abort succeeds even when the original context is done.

### Use Case

 When using UploadObject or UploadDirectory with a cancellable context (e.g. user-initiated cancellation, deadline), the SDK should still clean up multipart uploads. Currently, users must either:
  - Catch MultipartUploadError, extract the upload ID, and manually call AbortMultipartUpload with a new context
  - Configure S3 lifecycle rules to eventually clean up orphaned parts.

Both are workarounds for what should work out of the box. It can be more convenient for user. because no additional code is needed.


### Proposed Solution

1. Add a new option to transfermanager.Options:
```
  // MultipartCleanupTimeout is the timeout for cleanup operations
  // (e.g. AbortMultipartUpload) when a multipart upload fails.
  // A fresh context with this timeout is used so cleanup succeeds
  // even when the original context is cancelled.
  /// Defaults to 0 (uses the original context).
  MultipartCleanupTimeout time.Duration
```
  2. Update multiUploader.fail() to use a detached context:
```
  func (u *multiUploader) fail(ctx context.Context, ...) {
      if u.options.MultipartCleanupTimeout > 0 {
          var cancel context.CancelFunc
          ctx, cancel = context.WithTimeout(context.Background(), u.options.MultipartCleanupTimeout)
          defer cancel()
      }
      u.options.S3.AbortMultipartUpload(ctx, params, ...)
  }
```


### Other Information

- Default is 0 for backward compatibility — existing behavior is unchanged.



### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

1.41.2

### Go version used

1.25

wty-Bryant:
This is a corner case omitted during the initial implementation, put it in the queue and will fix that once we have bandwidth

proost:
@wty-Bryant 
Thanks!

If you are ok in this feature, i can implement.

But if you plan to rewrite whole handling fail scenario logic, i leave in the queue. 

How do you think?

[...]

Issue #3326, opened 2026-02-17
Support for ChecksumType in the TransferManager upload functionality


rohithkumar31:
### Describe the feature

CRC algorithms support FULL_OBJECT checksum type for multipart uploads. Currently, the transfer manager doesn't provide an option to specify the Checksum type.

### Use Case

This feature helps with the data integrity when migration objects from on-premise to S3 at large scale. 

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Go SDK V2 Module Versions Used

github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.17.61
github.com/aws/aws-sdk-go-v2 v1.36.1

### Go version used

1.23.0

wty-Bryant:
I saw the checksum type in `CreateMultipartUploadInput`, this request is put into our TODO list

wty-Bryant:
#3424 should fix the issue

PR #3424, opened 2026-05-26
Feat tmv2 parity

wty-Bryant:
Updated feature/s3/transfermanager to add following features missed from old tm v1 or complained by users:

- Add back range download for single object #3322 
- Support checksum type config #3326 
- Fix error when running following listener/cleanup of failed transferl, which is caused by context cancellation #3341 
- Honor mpu threshold option which was ignored before #3408 
- Add back max upload parts option for single object #3407

### DIFF
--- a/feature/s3/transfermanager/api_op_DownloadObject.go
+++ b/feature/s3/transfermanager/api_op_DownloadObject.go
@@ -580,7 +580,9 @@ func (d *downloader) download(ctx context.Context) (*DownloadObjectOutput, error
 	if d.options.GetObjectType == types.GetObjectParts {
 		output = d.getChunk(ctx, 1, "", clientOptions...)
 		if d.err != nil {
-			d.emitter.Failed(ctx, d.err)
+			freshCtx, cancel := d.freshContext(ctx)
+			defer cancel()
+			d.emitter.Failed(freshCtx, d.err)
 			return output, d.err
 		}
 
@@ -608,7 +610,9 @@ func (d *downloader) download(ctx context.Context) (*DownloadObjectOutput, error
 		if rng := aws.ToString(d.in.Range); rng != "" {
 			rangeStart, rangeEnd, err := getReqRange(rng)
 			if err != nil {
-				d.emitter.Failed(ctx, d.err)
+				freshCtx, cancel := d.freshContext(ctx)
+				defer cancel()
+				d.emitter.Failed(freshCtx, d.err)
 				return nil, err
 			}
 			d.offset = rangeStart
@@ -631,7 +635,9 @@ func (d *downloader) download(ctx context.Context) (*DownloadObjectOutput, error
 					return out, nil
 				}
 			}
-			d.emitter.Failed(ctx, d.err)
+			freshCtx, cancel := d.freshContext(ctx)
+			defer cancel()
+			d.emitter.Failed(freshCtx, d.err)
 			return nil, d.err
 		}
 		total := d.totalBytes
@@ -659,7 +665,9 @@ func (d *downloader) download(ctx context.Context) (*DownloadObjectOutput, error
 	}
 
 	if d.err != nil {
-		d.emitter.Failed(ctx, d.err)
+		freshCtx, cancel := d.freshContext(ctx)
+		defer cancel()
+		d.emitter.Failed(freshCtx, d.err)
 		return nil, d.err
 	}
 
@@ -833,6 +841,13 @@ func (d *downloader) setTotalBytes(resp *s3.GetObjectOutput) {
 
 		d.totalBytes = total
 	}
+}
+
+func (d *downloader) freshContext(ctx context.Context) (context.Context, context.CancelFunc) {
+	if d.options.FailTimeout <= 0 {
+		return ctx, func() {}
+	}
+	return context.WithTimeout(context.Background(), d.options.FailTimeout)
 }
 
 func (d *downloader) setOutput(resp *DownloadObjectOutput) {
--- a/feature/s3/transfermanager/api_op_DownloadObject.go
+++ b/feature/s3/transfermanager/api_op_DownloadObject.go
@@ -100,6 +100,9 @@ type DownloadObjectInput struct {
 	// [RFC 7232]: https://tools.ietf.org/html/rfc7232
 	IfUnmodifiedSince *time.Time
 
+	// Downloads the specified byte range of an object. This field only applies when GetObjectType is GetObjectRanges
+	Range *string
+
 	// Confirms that the requester knows that they will be charged for the request.
 	// Bucket owners need not specify this parameter in their requests. If either the
 	// source or destination S3 bucket has Requester Pays enabled, the requester will
@@ -602,6 +605,16 @@ func (d *downloader) download(ctx context.Context) (*DownloadObjectOutput, error
 			d.wg.Wait()
 		}
 	} else {
+		if rng := aws.ToString(d.in.Range); rng != "" {
+			rangeStart, rangeEnd, err := getReqRange(rng)
+			if err != nil {
+				d.emitter.Failed(ctx, d.err)
+				return nil, err
+			}
+			d.offset = rangeStart
+			d.totalBytes = rangeEnd + 1
+			d.pos = rangeStart
+		}
 		d.getChunk(ctx, 0, d.byteRange(), clientOptions...)
 		if d.err != nil {
 			// early check to see if error is caused by range download a zero object
@@ -780,7 +793,7 @@ func (d *downloader) tryDownloadChunk(ctx context.Context, params *s3.GetObjectI
 
 	d.totalBytesOnce.Do(func() {
 		d.setTotalBytes(out)
-		d.emitter.Start(ctx, d.in, d.totalBytes)
+		d.emitter.Start(ctx, d.in, d.totalBytes-d.offset)
 	}) // Set total in first GET
 
 	var n int64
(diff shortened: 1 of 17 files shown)


feature/s3/transfermanager/api_op_DownloadObject.go
```diff
@@ -843,14 +871,22 @@ func (d *downloader) byteRange() string {
...
+		return -1, -1, fmt.Errorf("invalid range format %s, should be bytes=start-end format", rng)
+	}
+
+	start, err := strconv.ParseInt(sub2[0], 10, 64)
 	if err != nil {
-		return 0, 0, fmt.Errorf("error when parsing request start: %v", err)
+		return -1, -1, fmt.Errorf("invalid range start %v", err)
 	}
-	end, err := strconv.ParseInt(ranges[1], 10, 64)
+	end, err := strconv.ParseInt(sub2[1], 10, 64)
```
Madrigal: nit: I think this would be a bit clearer

```suggestion
	rangeFmt := strings.Split(rng, "=")
	if len(rangeFmt) != 2 {
		return -1, -1, fmt.Errorf("invalid range format %s, should be bytes=start-end format", rng)
	}
	startEnd := strings.Split(rangeFmt[1], "-")
	if len(startEnd) != 2 {
		return -1, -1, fmt.Errorf("invalid range format %s, should be bytes=start-end format", rng)
	}

	start, err := strconv.ParseInt(startEnd[0], 10, 64)
if err != nil {
		return -1, -1, fmt.Errorf("invalid range start %v", err)
}
	end, err := strconv.ParseInt(startEnd[1], 10, 64)
```

feature/s3/transfermanager/api_op_DownloadObject_integ_test.go
```diff
@@ -41,6 +42,16 @@ func TestInteg_DownloadObject(t *testing.T) {
 				},
 			},
 		},
+		"range get large object body with range input": {
+			Body:       bytes.NewReader(largeObjectBuf),
+			ExpectBody: largeObjectBuf[1:10485761],
+			Range:      "bytes=1-10485760",
```
Madrigal, line 48: why is the expect `[x-y]` but the range is `[x-(y-1)]`

feature/s3/transfermanager/api_op_PutObject_integ_test.go
```diff
@@ -6,13 +6,21 @@ import (
...
 
 func TestInteg_PutObject(t *testing.T) {
 	cases := map[string]putObjectTestData{
 		"seekable body":         {Body: strings.NewReader("hello world"), ExpectBody: []byte("hello world")},
 		"empty string body":     {Body: strings.NewReader(""), ExpectBody: []byte("")},
 		"multipart upload body": {Body: bytes.NewReader(largeObjectBuf), ExpectBody: largeObjectBuf},
+		"multipart upload body with full object checksum type": {
+			Body:              bytes.NewReader(largeObjectBuf),
+			ExpectBody:        largeObjectBuf,
+			ChecksumAlgorithm: types.ChecksumAlgorithmCrc32c, // only CRC algorithms support full object checksum
```
Madrigal: This may no longer be true after #3422

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -945,6 +1050,37 @@ func TestUploadWithContextCanceled(t *testing.T) {
 	}
 }
 
+func TestUploadWithContextCanceledWhenComplete(t *testing.T) {
```
Madrigal, line 1120: Wouldn't calling "completeMultiPart" be the last thing we do, and hence the whole upload could be now considered complete? why do we want to abort when this happens?

feature/s3/transfermanager/downloadobject_test.go
```diff
@@ -88,6 +89,50 @@ func TestDownloadObject(t *testing.T) {
...
 		},
+		"single range download with specified range input": {
+			data:        buf20MB,
+			getObjectFn: s3testing.RangeGetObjectFn,
+			optFn: func(o *Options) {
+				o.GetObjectType = types.GetObjectRanges
+			},
+			rng:               "bytes=2-8388609",
+			expectInvocations: 1,
+			expectRanges:      []string{"bytes=2-8388609"},
```
Madrigal, line 100: These numbers seem very magic, from where are they coming from?

feature/s3/transfermanager/options.go
```diff
@@ -17,6 +19,15 @@ type Options struct {
 	// The threshold bytes to decide when the file should be multi-uploaded
 	MultipartUploadThreshold int64
 
+	// FailTimeout is the timeout for transfer failure handling when a transfer fails.
+	// A fresh context with this timeout is used so failure followup (AbortMPU for upload or possible progress listener work)
+	// succeed even when the original context is canceled.
+	// Defaults to 0 (uses the original context) for fail case not caused by ctx cancellation.
+	FailTimeout time.Duration
+
+	// The max parts count for a multi part upload, which could not exceed defaultMaxUploadParts defined by s3 side
```
Madrigal: `defaultMaxUploadParts` is not exported on the package, so users don't know what we're referring to if they are just looking at the docs. 

We can point them to the docs or just say "10,000" https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html although I worry if this number is ever changed

feature/s3/transfermanager/api_op_UploadObject_test.go
Madrigal: More than this, I'd like a change where we abort something that times out before it has completed, which is probably what the original requester was asking for

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -279,10 +275,73 @@ func TestUploadOrderMultiDifferentPartSize(t *testing.T) {
 
 	// Part lengths
 	if len := getReaderLength((*args)[1].(*s3.UploadPartInput).Body); 1024*1024*11 != len {
-		t.Errorf("expect %d, got %d", 1024*1024*7, len)
+		t.Errorf("expect %d, got %d", 1024*1024*11, len)
```
Madrigal, line 278: why did this changed to 11?

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -127,65 +147,41 @@ func TestUploadOrderMultiTriggerredBySinglePartSize(t *testing.T) {
...
-	if "UPLOAD-ID" != aws.ToString(resp.UploadID) {
-		t.Errorf("expect %q, got %q", "UPLOAD-ID", aws.ToString(resp.UploadID))
-	}
-
 	if "VERSION-ID" != aws.ToString(resp.VersionID) {
 		t.Errorf("expect %q, got %q", "VERSION-ID", aws.ToString(resp.VersionID))
 	}
 
-	// Validate input values
-	v := aws.ToString((*args)[1].(*s3.UploadPartInput).UploadId)
```
Madrigal, line 143: did you intend to remove these validations? if so, why?

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -127,65 +147,41 @@ func TestUploadOrderMultiTriggerredBySinglePartSize(t *testing.T) {
 		t.Errorf("Expected no error but received %v", err)
 	}
 
-	if diff := cmpDiff([]string{"CreateMultipartUpload", "UploadPart", "CompleteMultipartUpload"}, *invocations); len(diff) > 0 {
+	if diff := cmpDiff([]string{"PutObject"}, *invocations); len(diff) > 0 {
```
Madrigal, line 150: So now we're changing an upload that is less than MPU treshold to just call `putObject`, correct? is this the same across other SDKs?

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -279,10 +275,73 @@ func TestUploadOrderMultiDifferentPartSize(t *testing.T) {
...
+	// Part lengths
+	if e, a := 1024*1024*16, getReaderLength((*args)[1].(*s3.UploadPartInput).Body); int64(e) != a {
+		t.Errorf("expect %d, got %d", e, a)
+	}
+	if e, a := 1024*1024*4, getReaderLength((*args)[2].(*s3.UploadPartInput).Body); int64(e) != a {
+		t.Errorf("expect %d, got %d", e, a)
+	}
+}
+
+func TestUploadOrderMultiWithMaxParts(t *testing.T) {
```
Madrigal, line 316: we should add a test when we can't fulfill `MaxUploadParts`

> When size is unknown (streaming), error once the part count would exceed MaxUploadParts, with a message similar to the old uploader (“exceeded configured MaxUploadParts … adjust PartSizeBytes …”).

feature/s3/transfermanager/api_op_DownloadObject_integ_test.go
wty-Bryant: range's start and end index is inclusive, but slicing's last byte is exclusive (e.g. [x-y))

feature/s3/transfermanager/api_op_PutObject_integ_test.go
wty-Bryant: I just referred to s3 doc for full object checksum: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity-upload.html#ChecksumTypes-Uploads . I can delete this comment for future change

feature/s3/transfermanager/downloadobject_test.go
wty-Bryant: they are just random number ensuring range get works when not starting from a part's first byte

feature/s3/transfermanager/options.go
wty-Bryant: that max might change recently by s3, I prefer to just add the reference link

feature/s3/transfermanager/api_op_UploadObject_test.go
wty-Bryant: Yes, that's how the SEP defines the threshold, it decides when to use mpu

feature/s3/transfermanager/api_op_UploadObject_test.go
wty-Bryant: the old value is wrong (we set part size here to 11MB)

feature/s3/transfermanager/api_op_UploadObject_test.go
wty-Bryant: The abort is inherited from tm v1 and required by SEP, it happens when either UploadPart or CompleteMPU fails. In original issue, the user's ctx cancellation led to mpu fail, and the follow-up abort failed to clean up the whole process since that dead ctx was passed to the AbortMPU. That's why we need a new ctx for abort and listener.Fail call

Madrigal: I understand the need for a new context, I'm not questioning that. 

I misunderstood this test. You are simulating a FAIL on CompleteMPU, so if we can't verify that this completed we want to call Abort. This makes sense.

I'd still like to see a case where we fail before calling Complete, which I suspect will be the most common scenario, timing out before you call Complete

feature/s3/transfermanager/api_op_UploadObject_test.go
wty-Bryant: because this case now validates single upload, so previous multipart prop check should be removed

feature/s3/transfermanager/api_op_UploadObject_test.go
wty-Bryant: added for unseekable body

feature/s3/transfermanager/options.go
```diff
@@ -17,6 +19,16 @@ type Options struct {
...
```
Madrigal:
```suggestion
	// The max parts count for a multi part upload, which may not exceed "Maximum number of parts per upload" defined by S3 
```

Madrigal (approved):

feature/s3/transfermanager/options.go
wty-Bryant: I will update this to "must not exceed" since s3 limit is the hard limit, any max upload parts value greater than default (10000 for now) will be error back

PR #3425, opened 2026-05-27
credentials/logincreds: create new cache files with 0600

lucix-aws:
* NEW cache files are created with 0600. existing files are untouched
* add support for `AWS_RESTRICT_FILE_PERMISSIONS`
  * the SEP did not specify a shared config and internal discussion apparently reached consensus that such a config was overkill, so that's why that's missing


wty-Bryant (approved):
There's no shared config option needed based on changelog description, everything looks good

Madrigal (approved):

PR #3426, opened 2026-05-28
update to smithy-go v1.26.0

lucix-aws:
check the commit list, there is a minor fix in the endpoint auth scheme resolver for s3 to ensure that still works (and i added a test for it)


wty-Bryant (approved):
double confirm the SMITHY_GO_CODEGEN_VERSION matches the latest commit id of smithy go main branch

PR #3427, opened 2026-05-29
fix flaky s3 manager test

lucix-aws:
caught in an internal build, the concurrency in this test means that one of a few orders of operations could happen, so just expect either or

### DIFF
--- a/feature/s3/transfermanager/api_op_UploadObject_test.go
+++ b/feature/s3/transfermanager/api_op_UploadObject_test.go
@@ -364,12 +364,15 @@ func TestUploadOrderMultiExceedMaxParts(t *testing.T) {
 		t.Errorf("expect %q to be contained in %q", e, a)
 	}
 
-	vals := []string{"CreateMultipartUpload", "UploadPart", "UploadPart", "AbortMultipartUpload"}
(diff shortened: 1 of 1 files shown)


wty-Bryant (approved):

feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -364,12 +364,15 @@ func TestUploadOrderMultiExceedMaxParts(t *testing.T) {
 		t.Errorf("expect %q to be contained in %q", e, a)
 	}
 
-	vals := []string{"CreateMultipartUpload", "UploadPart", "UploadPart", "AbortMultipartUpload"}
-	if !reflect.DeepEqual(vals, *ops) {
-		t.Errorf("expect %v, got %v", vals, *ops)
+	// The worker may or may not send part 2 before observing the error set
+	// by the main goroutine, so both sequences are valid.
+	withTwo := []string{"CreateMultipartUpload", "UploadPart", "UploadPart", "AbortMultipartUpload"}
```
wty-Bryant, line 369: I forgot the `readChunk` might not be quick enough to send the second UploadPart before next main iteration set error

PR #3437, opened 2026-06-04
Fix tmv2 flaky test to avoid double closing test channel

wty-Bryant:


feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -1099,6 +1099,7 @@ func TestUploadWithContextCanceledWhenUploadPart(t *testing.T) {
 	}
 	u := New(c, func(o *Options) {
 		o.FailTimeout = 5 * time.Second
+		o.Concurrency = 1
```
lucix-aws: i would prefer we don't pin all these to 1 concurrency because we actually want to test the concurrency. can this be solved in another way?

wty-Bryant: yep I can use some once to do that instead

lucix-aws (approved):

Issue #3321, opened 2026-02-12
refactor invokeOperation to handle common middlewares


lucix-aws:
This is a screenshot of `gsa` on a sample program I wrote that calls all SQS operations. (note this is on an in-progress branch for our serialization refactor but the problem is the same in main).

<img width="792" height="705" alt="Image" src="https://github.com/user-attachments/assets/710d046e-3b4f-4428-8890-27e3cffac2e1" />

Looking into the breakdown, the individually-generated `addOperation*Middlewares` functions account for ~7kb of code. That's _per-operation_. What sticks out to me here is that their code (so, the middlewares they add) is 99% identical across each one. The only things that will differ will be the two serde middlewares which are unique per operation (which actually won't be the case anymore in the upcoming serde rewrite), and then any customization middlewares that would have had a specific OperationPredicate in codegen.

We should be able to refactor this so invokeOperation adds all of the common middlewares, and then the individual funcs only add per-operation ones. That should save 6kb of compiled code per function, or ~120kb in SQS.

We have the existing middleware snapshot tests so it will be incredibly easy to verify that we don't break anything when we do this.

PR #3443, opened 2026-06-05
feat: mark common plugins and fix per-op ordering for invokeOperation…

Tongs2000:
Issue: #3321

  Companion to aws/smithy-go#674. Marks AWS-specific common middlewares and fixes per-op ordering to support the invokeOperation refactor.

  Changes:
  - Mark common plugins (addClientRequestID, addRetry, addRawResponseToMetadata, addSpanRetryLoop, addClientUserAgent, addUserAgentRetryMode, addRecursionDetection,
  addSetLegacyContextSigningOptionsMiddleware, addInterceptBeforeRetryLoop, addInterceptAttempt) with `.isCommon(true)`
  - Refactor RegisterServiceMetadataMiddleware to emit a single shared `newServiceMetadataMiddleware(region, operation)` function per service
  - Fix addComputeContentLength to use Insert relative to ClientRequestID
  - Fix S3 Add100Continue to use Insert relative to RecursionDetection
  - Remove unused awsmiddleware import in AwsHttpPresignURLClientGenerator (pre-existing bug)

  Pending after smithy-go#674 merges:
  - Bump SMITHY_GO_CODEGEN_VERSION
  - Run make generate (will produce the regenerated output)
  - Add changelog

  Snapshot tests pass locally across all ~300 services. SQS benchmark shows ~47% reduction in per-operation compiled function size (~2,550 bytes saved per operation).

  By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


service/internal/s3shared/s3100continue.go
```diff
@@ -13,9 +13,9 @@ const default100ContinueThresholdBytes int64 = 1024 * 1024 * 2
 // Add100Continue add middleware, which adds {Expect: 100-continue} header for s3 client HTTP PUT request larger than 2MB
 // or with unknown size streaming bodies, during operation builder step
 func Add100Continue(stack *middleware.Stack, continueHeaderThresholdBytes int64) error {
-	return stack.Build.Add(&s3100Continue{
+	return stack.Build.Insert(&s3100Continue{
 		continueHeaderThresholdBytes: continueHeaderThresholdBytes,
-	}, middleware.After)
+	}, "RecursionDetection", middleware.Before)
```
wty-Bryant, line 18: is this change to make sure the 100continue middleware doesn't break snapshot order after the common plug in change?

Tongs2000: Yes. `RecursionDetection` moved to `addCommonMiddlewares` (runs first), so it now registers in the Build step before per-op middlewares. The original `Add`(After) here would place `S3100Continue` after `RecursionDetection`, but the snapshot expects it before. Changing to `Insert(..., "RecursionDetection", Before)` preserves the original ordering.

wty-Bryant (approved):
lgtm, wait for folLuc and Luis's comment

PR #3445, opened 2026-06-06
fix(feature/s3/transfermanager): allow multiple symlinks to share the same target file in UploadDirectory

### LINKED ISSUE #3439: feature/s3/transfermanager: UploadDirectory fails on multiple files symlinking to the same target due to incorrect duplicate path checking
### Acknowledgements

### Describe the bug

When using transfermanager.UploadDirectory to upload a local directory, if the directory contains multiple different files that happen to be symbolic links pointing to the exact same underlying target file, the upload fails with a traversed duplicate path error.

This introduces a severe issue when trying to upload standard structures like the Hugging Face cache directory, where multiple distinct files (e.g., in a snapshot folder) are symlinks sharing the same data blob under the hood.

### Regression Issue

### Expected Behavior

Even if two different symlink files (e.g., link1.model and link2.model) point to the same target file, they represent two distinct objects that need to be uploaded to S3 under their respective keys. The manager should allow this and not treat it as a duplicate path error.

### Current Behavior

The traverseSymlink function incorrectly uses the fully resolved target file path (path) for deduplication checks:

`error when getting abs path of file /root/.cache/huggingface/hub/models--internlm--Intern-S1-mini/snapshots/0f25c8f60f4190a18dd6620c94bdccedbebc208d/tokenizer_IUPAC.model: traversed duplicate path: /root/.cache/huggingface/hub/models--internlm--Intern-S1-mini/blobs/a6b5c578315e2d6aef8af99243d23549e06a08d6c5f319199b17a373433993d4`
Because a previous symlink already resolved to that same blob and marked it in u.traversed, the second unique symlink hits the global registry collision and crashes the entire upload session.

### Reproduction Steps

1. Create a directory where two unique files are symlinked to the same target:

```Bash
mkdir test_dir
echo "shared data" > test_dir/real_target.txt
ln -s real_target.txt test_dir/file_a.txt
ln -s real_target.txt test_dir/file_b.txt
```
2. Run UploadDirectory targeting test_dir.

3. The process will abort upon reaching file_b.txt because its resolved target path (real_target.txt) was already registered when processing file_a.txt.

### Possible Solution

In github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/api_op_UploadDirectory.go, the duplicate check inside traverseSymlink should be performed against the originPath (the path of the file inside the upload directory) to ensure we are deduplicating the actual source files, not their resolved targets.

```Go
// getAbsPath resolves a path's destination absolute path with deduplication
// in case any symlink causes traverse loop or repeated upload
func (u *directoryUploader) getAbsPath(path string) (string, error) {
  
    originPath := path


    if u.traversed[originPath] != nil {
       return "", fmt.Errorf("traversed duplicate path %s", originPath)
    }
    u.traversed[originPath] = struct{}{}

    fileInfo, err := os.Lstat(path)
    if err != nil {
       return "", fmt.Errorf("error when stating info of file %s: %v", path, err)
    }

    if fileInfo.Mode()&os.ModeSymlink == os.ModeSymlink {
       if !aws.ToBool(u.in.FollowSymbolicLinks) {
          return "", nil
       }
       path, err = u.traverseSymlink(path)
       if err != nil {
          return "", err
       }
    }

    return path, nil
}

func (u *directoryUploader) traverseSymlink(path string) (string, error) {
	originPath := path
	for {
		dst, err := os.Readlink(path)
		if err != nil {
			return "", fmt.Errorf("error when reading symlink of %s: %v", originPath, err)
		}
		if filepath.IsAbs(dst) {
			path = dst
		} else {
			path = filepath.Join(filepath.Dir(path), dst)
		}
        
		// FIX: Deduplicate based on the file's path within the upload directory, not its resolved target
		if u.traversed[originPath] != nil {
			return "", fmt.Errorf("traversed duplicate path: %s", originPath)
		}
        
		fileInfo, err := os.Lstat(path)
		if err != nil {
			return "", fmt.Errorf("error when stating linked path %s: %v", path, err)
		}
		if fileInfo.Mode()&os.ModeSymlink != os.ModeSymlink {
			return path, nil
		}
        
		// FIX: Register the original file path as processed
		u.traversed[originPath] = struct{}{}
	}
}
```

_No response_

- file: github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/api_op_UploadDirectory.go

The issue resides in the traverseSymlink implementation within feature/s3/transfermanager.

Currently, the u.traversed map is used as a global registry across the entire directory walk session to detect infinite symlink loops:

```
if u.traversed[path] != nil {
    return "", fmt.Errorf("traversed duplicate path: %s", path)
}
...
u.traversed[path] = struct{}{}
```

### Compiler and Version used

1.25.5

Alpine Linux v3.23

pedrorfdez:
## Description

Fixes #3439

`UploadDirectory` with `FollowSymbolicLinks: true` fails with `traversed duplicate path` when two or more distinct symlinks inside the source directory resolve to the same underlying file or directory (e.g. a Hugging Face model cache where multiple snapshot entries point to the same blob).

Each symlink represents a distinct S3 key subtree and all uploads are valid and expected. The error was a false positive.

## Root Cause

`u.traversed` was used as a global session-wide registry for resolved paths. When any symlink resolved to a path already in the map, whether a file or a directory, the upload failed. This caused false positives whenever two distinct symlinks pointed to the same underlying file or directory, both of which are legitimate and should produce separate S3 objects under different keys.

## Fix

- Removed `u.traversed` entirely.
- Directory loop detection now uses an `ancestors` set passed through recursive `traverse` calls. A directory is added to `ancestors` before recursing into its children and removed afterward (DFS backtracking). This means only directories **currently active on the call stack** are checked for cycles, which correctly allows sibling symlinks to the same directory while still detecting actual infinite loops.
- `traverseSymlink` retains its existing **local `visited` set** per resolution chain to detect circular symlink chains (e.g. `a → b → a`), without any cross-contamination between independent symlink resolutions.

## Testing

- Added three new test cases to `TestUploadDirectory`:
  - `"symlink pointing to another file already in source is uploaded"`: a symlink resolving to a file that exists elsewhere in the source tree should succeed, not error.
  - `"multiple symlinks pointing to the same target file are each uploaded"`: directly reproduces the scenario in #3439.
  - `"two symlinks referring to the same directory are each uploaded"`: covers the edge case where two sibling symlinks resolve to the same directory; both subtrees must be uploaded independently.
- The previously incorrect test case `"error when a symlink refers to another file under source"` has been replaced, as that behaviour was wrong.
- All existing tests continue to pass, including the legitimate loop-detection case `"error when a symlink refers to its upper dir"`.

### DIFF
--- a/feature/s3/transfermanager/api_op_UploadDirectory.go
+++ b/feature/s3/transfermanager/api_op_UploadDirectory.go
@@ -130,7 +130,6 @@ type directoryUploader struct {
 
 	filesUploaded atomic.Int64
 	filesFailed   atomic.Int64
-	traversed     map[string]any
 
 	err error
 
@@ -151,7 +150,7 @@ func (u *directoryUploader) uploadDirectory(ctx context.Context) (*UploadDirecto
 	}
 
 	if aws.ToBool(u.in.Recursive) {
-		u.traverse(aws.ToString(u.in.Source), aws.ToString(u.in.KeyPrefix), ch)
+		u.traverse(aws.ToString(u.in.Source), aws.ToString(u.in.KeyPrefix), ch, map[string]struct{}{})
 	} else {
 		files, err := u.traverseFolder(aws.ToString(u.in.Source))
 		if err != nil {
@@ -210,8 +209,6 @@ func (u *directoryUploader) uploadDirectory(ctx context.Context) (*UploadDirecto
 }
 
 func (u *directoryUploader) init() {
-	u.traversed = make(map[string]any)
-
 	u.failurePolicy = TerminateUploadPolicy{}
 	if u.in.FailurePolicy != nil {
 		u.failurePolicy = u.in.FailurePolicy
@@ -228,8 +225,15 @@ type fileEntry struct {
 }
 
 // traverse recursively visits each folder and sends each
-// valid file's request to worker goroutines
-func (u *directoryUploader) traverse(path, keyPrefix string, ch chan fileEntry) {
+// valid file's request to worker goroutines.
+//
+// ancestors tracks the real directory paths currently active on the call
+// stack. It is used to detect symlinks that would create an infinite loop by
+// pointing back to a directory that is already being recursed into. Entries
+// are added before recursing into a directory and removed afterward (DFS
+// backtracking), so two sibling symlinks that resolve to the same directory
+// are each traversed independently without triggering a false positive.
+func (u *directoryUploader) traverse(path, keyPrefix string, ch chan fileEntry, ancestors map[string]struct{}) {
 	if u.getErr() != nil {
 		return
 	}
@@ -256,14 +260,24 @@ func (u *directoryUploader) traverse(path, keyPrefix string, ch chan fileEntry) 
 		return
 	}
 	if fileInfo.IsDir() {
+		// Detect symlink-induced directory loops: if absPath is already an
+		// ancestor on the current recursion stack we would loop forever.
+		if _, loop := ancestors[absPath]; loop {
+			u.setErr(fmt.Errorf("traversed duplicate path %s", absPath))
+			return
+		}
 		subFiles, err := u.traverseFolder(absPath)
 		if err != nil {
 			u.setErr(fmt.Errorf("error when traversing folder %s: %v", absPath, err))
 			return
 		}
+		// Mark this directory as active for all children, then unmark it
+		// once done so sibling subtrees can legitimately enter the same dir.
+		ancestors[absPath] = struct{}{}
 		for _, f := range subFiles {
-			u.traverse(filepath.Join(path, f), key, ch)
+			u.traverse(filepath.Join(path, f), key, ch, ancestors)
 		}
+		delete(ancestors, absPath)
 	} else {
 		if u.in.Filter != nil && !u.in.Filter.FilterFile(path) {
 			return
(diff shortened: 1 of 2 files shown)


pedrorfdez:
Ready for review

feature/s3/transfermanager/api_op_UploadDirectory.go
```diff
@@ -288,11 +293,20 @@ func (u *directoryUploader) getAbsPath(path string) (string, error) {
...
+		if err != nil {
+			return "", fmt.Errorf("error when stating resolved path %s: %v", path, err)
+		}
 	}
-	if u.traversed[path] != nil {
-		return "", fmt.Errorf("traversed duplicate path %s", path)
+
+	// Only deduplicate directory paths to prevent traversal loops.
+	if fileInfo.IsDir() {
+		if u.traversed[path] != nil {
```
wty-Bryant: This duplicate folder check should not be added since a folder contains shared files could also be referred by different symlinks under root source.

pedrorfdez: You are completely right, I'll implement a solution

pedrorfdez (commented):
missclicked review

pedrorfdez:
The new approach passes an ancestors set through the recursive traverse calls: a directory is added to the set before recursing into it and removed after (DFS backtracking), so only directories that are currently active on the call stack are checked.

This correctly allows sibling symlinks to the same directory while still catching actual traversal loops. A new test case covers this scenario.

Ready for review, I updated the PR description too, it is a good guidance.

wty-Bryant (commented):
lgtm, waiting for other teammate's view and approval, i think it's the only solution in our loop detection case considering:

- a loop in a uploaded directory might only happen when there's a symlink that can make a loop routing back to itself
- since `traverseSymlink` jumps out once a non symlink is referred (e.g. a folder that could contain symlink previously traversed in current iteration), the deduplication in `traverseSymlink` itself is not enough for complete loop detection

therefore, the only way to check loop here is tracing every dfs path being traversed now with nodes map refreshed at each layer

lucix-aws (approved):

PR #3454, opened 2026-06-24
aws/transport/http: set TLS curve preferences

Ankit026-code:
## Summary

This change updates the default HTTP transport configuration to explicitly set TLS curve preferences.

## Changes

- Added TLS curve preferences to `tls.Config` in `defaultHTTPTransport`:
  - CurveP256
  - CurveP384
  - CurveP521

- Ensures consistent elliptic curve selection for TLS connections.

- Added unit test to validate:
  - TLS minimum version is TLS 1.2
  - TLS configuration is properly initialized

## Motivation

Improves TLS configuration clarity and ensures predictable elliptic curve usage in secure connections.

## Testing

- Ran `go test ./aws/transport/http`
- All tests passing locally

## Notes

No changes to generated code or service models.

### DIFF
--- a/aws/transport/http/client.go
+++ b/aws/transport/http/client.go
@@ -193,9 +193,13 @@ func defaultHTTPTransport() *http.Transport {
 		ForceAttemptHTTP2:     true,
 		TLSClientConfig: &tls.Config{
 			MinVersion: DefaultHTTPTransportTLSMinVersion,
+			CurvePreferences: []tls.CurveID{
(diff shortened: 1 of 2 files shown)


lucix-aws:
Is there motivation for this beyond clarity? How were these values chosen? What potential impact does this have across the 400+ AWS services supported by this SDK today?

I am concerned about changing this without more context, even with that I would have pretty significant backwards-compatibility concerns.

Issue #3469, opened 2026-07-03
feature/s3/transfermanager always enables CRC32 checksums with no way to disable them


mxgnus-de:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

The new feature/s3/transfermanager always enables CRC32 request checksums for uploads, even when the application does not explicitly configure a checksum algorithm.

As a result, upload requests always contain checksum-related headers such as:

```
X-Amz-Sdk-Checksum-Algorithm: CRC32
X-Amz-Content-Sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER
X-Amz-Trailer: x-amz-checksum-crc32
Content-Encoding: aws-chunked
```

This causes uploads to fail against some S3-compatible object storage implementations that do not support this behavior.

This appears to be a regression compared to the legacy feature/s3/manager, where the same issue was previously fixed:

Issue: https://github.com/aws/aws-sdk-go-v2/issues/3007
Fix: https://github.com/aws/aws-sdk-go-v2/pull/3151

From investigating the implementation, the Transfer Manager appears to always set a default checksum algorithm internally instead of allowing it to remain unset:

https://github.com/aws/aws-sdk-go-v2/blob/a4cd1c9dcbc1492cd27b72f2a21682d43dab7f40/feature/s3/transfermanager/options.go#L83-L87

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

If no checksum algorithm is explicitly configured by the application, the Transfer Manager should not automatically enable CRC32 request checksums.

Alternatively, it should expose a configuration option allowing applications to disable automatic checksum calculation, matching the flexibility that was added to the legacy feature/s3/manager.

The environment variables
```
AWS_REQUEST_CHECKSUM_CALCULATION=when_required
AWS_RESPONSE_CHECKSUM_VALIDATION=when_required
```
should also prevent unnecessary checksum headers from being added, but currently they do not.

### Current Behavior

Every upload performed by feature/s3/transfermanager includes CRC32 checksum headers regardless of configuration.

The following request headers are always present:
```
X-Amz-Sdk-Checksum-Algorithm: CRC32
[...]

PR #3470, opened 2026-07-07
Add missing options.RequestChecksumCalculation

lucix-aws:
adds checksum control, mirroring the SEP and old v1 behavior, but v1 actually had an oversight which ignored the setting for single-part fallback so this is more complete


lucix-aws:
closes #3469

lucix-aws:
the serde snapshot tests are going to fail for the moment because we haven't updated our release process to update them. ignore that for now

wty-Bryant (approved):

PR #3472, opened 2026-07-10
Fix Clock Skew according to internal specification

Madrigal:
Update our implementation of Clock Skew to match our internal specification. 

## Update Skew computation 
- Replaced the current `serverTime - responseAt` formula with the internal spec midpoint calculation: `serverTime - midpoint(requestAt, responseAt)`, which bounds measurement error to half the round-trip time.
- Added a `maxTrustedRequestDuration` (15 min) guard — responses slower than this discard the skew candidate.
- Added an `Age` header check — cached responses are not used for skew measurement.

## Add disable flag
- Added `DisableClockSkewCorrection` flag to disable clock skew computation
- Wired it through the config resolution chain: env var (`AWS_DISABLE_CLOCK_SKEW_CORRECTION`), shared config profile (`disable_clock_skew_correction`), and programmatic.

## Update codes
- Unified `definiteSkewCodes` and p`ossibleSkewCodes` into a single `clockSkewCodes` set. 
- Added `RequestTimeTooSkewed` and `AccessDeniedException` to the set; removed RequestExpired and RequestInTheFuture.


aws/middleware/clockskew_test.go
```diff
@@ -0,0 +1,158 @@
...
+	if skew, ok := GetAttemptSkew(mdDisabled); ok {
+		t.Fatalf("disabled: expected no recorded skew, got %v", skew)
+	}
+
+	_, mdEnabled, err := RecordResponseTiming{}.
+		HandleDeserialize(context.Background(), smithymiddleware.DeserializeInput{}, next)
+	if err != nil {
+		t.Fatalf("enabled: unexpected error: %v", err)
+	}
+	if skew, ok := GetAttemptSkew(mdEnabled); !ok || skew != 5*time.Minute {
```
wty-Bryant, line 155: nit: from first glance there might be transient case where calculated  skew is slight diff from exact 5 min since we use midpoint of req/resp time for calculation now, but that should not impact too much

lucix-aws (approved):

lucix-aws:
needs a changelog btw, fixandship

PR #3482, opened 2026-07-21
test: add kitchen-sink response body close behavior tests

Tongs2000:
## Description                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                            
  Adds kitchen-sink tests that lock in the current response-body close behavior,                                                                                                                                                                                                                                            
  so the upcoming middleware-stack refactor can be verified not to change it.                                                                                                                                                                                                                                               
  Per review feedback on smithy-go#683, these tests go into main *first* (before                                                                                                                                                                                                                                            
  that change merges) to establish the baseline.                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                            
  Response-body close behavior is implemented independently across codegen paths:                                                                                                                                                                                                                                           
  awsJson (HttpRpc) vs restJson1 (HttpBinding), and serde2 vs legacy serde. Since                                                                                                                                                                                                                                           
  most services still generate with legacy serde, each protocol is covered in                                                                                                                                                                                                                                               
  both serde variants, and each is tested for the output categories it can                                                                                                                                                                                                                                                  
  express:                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                            
  | Category | awsJson serde2 | awsJson legacy | restJson1 serde2 | restJson1 legacy |                                                                                                                                                                                                                                      
  |----------|:---:|:---:|:---:|:---:|                                                                                                                                                                                                                                                                                      
  | non-streaming (close on success **and** error) | ✅ | ✅ | ✅ | ✅ |                                                                                                                                                                                                                                                    
  | streaming blob / caller-owned (open on success, close on error) | n/a¹ | n/a¹ | ✅ | ✅ |                                                                                                                                                                                                                               
  | event stream (body owned by stream reader) | ✅ | ✅ | ✅ | ✅ |                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                            
  ¹ awsJson is an RPC protocol and cannot express an `@httpPayload` streaming                                                                                                                                                                                                                                               
  blob, so that category only exists on the REST protocol.                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                            
  ## Changes                                                                                                                                                                                                                                                                                                                
  - Add a restJson1 kitchen-sink service (streaming blob can't be modeled under                                                                                                                                                                                                                                             
    the existing awsJson service), and add a `SubscribeEvents` event stream                                                                                                                                                                                                                                                 
    operation to the awsJson service.                                                                                                                                                                                                                                                                                       
  - Generate four kitchen-sink clients via `smithy-build` projections: awsJson and                                                                                                                                                                                                                                          
    restJson1, each in serde2 and legacy-serde variants.                                                                                                                                                                                                                                                                    
  - Add response-body close tests for every protocol/serde/category combination;                                                                                                                                                                                                                                            
    event stream close tests live in per-client `internal/testing` modules.                                                                                                                                                                                                                                                 
  - Register the new modules in `modman.toml`.                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  The change is split into two commits: hand-written source (models, smithy-build                                                                                                                                                                                                                                           
  config, tests) and the generated clients (`make generate` output).                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                            
  ## Testing                                                                                                                                                                                                                                                                                                                
  All new tests pass on this branch (main behavior). They were also run against                                                                                                                                                                                                                                             
  the middleware-refactor branch, where they caught a real regression (streaming                                                                                                                                                                                                                                            
  operations no longer closed the body on the error path); that has been fixed on                                                                                                                                                                                                                                           
  the refactor branch. This confirms the tests actually guard the behavior.


lucix-aws:
These tests look good but I think we want another service that is generated with legacy serde since most services are still on that, probably a restjson one

Tongs2000:
> These tests look good but I think we want another service that is generated with legacy serde since most services are still on that, probably a restjson one

Good call. Will add a legacy-serde restjson service (via useLegacySerde: true in smithy-build) alongside the serde2 one. Confirmed it generates the legacy serializers/deserializers, and the close-body tests pass against it. So both serde paths are now covered.

lucix-aws (approved):

PR #3489, opened 2026-07-22
Fix transfer manager v2 uploader flaky test

wty-Bryant:
Existing transfer manager v2 upload retry test uses a mock http server locally to upload a 24MB object, where most of time is consumed on transport/socket layer. Under a concurrent multipart upload case, 11(attemps) * 8MB(part size) * 3(parts) = 264MB are copied through the loopback TCP stack (write/read syscalls, netpoll, bufio, HTTP framing), plus server-side io.Copy of every body, connection churn, and the Expect: 100-continue handshakes. All those wire workflow causes overloads of CI's poor CPU and leads to flaky timeout. This PR replaces the mock server with just a mock http client so no wire roundtrip will be counted

Local starving test is shown here though it may not 100% mock CI cpu core config:
go to `feature/s3/transfermanager` package then run
```
go test -race -c -o /tmp/tm.test ./ && \
  for i in $(seq $(sysctl -n hw.ncpu)); do yes >/dev/null & done; HOGS=$(jobs -p); \
  GOMAXPROCS=2 /tmp/tm.test -test.run '^TestUploadRetry$' -test.cpu=2 -test.count=20 -test.timeout=30s; \
  echo "exit=$?"; kill $HOGS 2>/dev/null
```
It uses max process 2 and different hogs to mock CI cpu's workload, `exit=0` should be output showing the low cpu core like CI still passed the unit test under timeout

### DIFF
--- a/feature/s3/transfermanager/api_op_UploadObject_test.go
+++ b/feature/s3/transfermanager/api_op_UploadObject_test.go
@@ -1318,7 +1318,7 @@ func (c *retryHTTPClient) Do(r *http.Request) (*http.Response, error) {
 	case r.Method == "PUT":
 		defer func() {
 			if err := r.Body.Close(); err != nil {
-				log.Printf("failed to close response body: %q", err)
+				log.Printf("failed to close request body: %q", err)
 			}
 		}()
 		partStr := r.URL.Query().Get("partNumber")
--- a/feature/s3/transfermanager/api_op_UploadObject_test.go
+++ b/feature/s3/transfermanager/api_op_UploadObject_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"reflect"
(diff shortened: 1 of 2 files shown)


feature/s3/transfermanager/api_op_UploadObject_test.go
```diff
@@ -1308,76 +1301,70 @@ func TestUploadRetry(t *testing.T) {
...
-			return
+		if part <= 0 || part > int64(len(c.failsLeft)) {
+			return &http.Response{StatusCode: 400, Status: "BadRequest"}, fmt.Errorf("invalid partNumber %v", part)
 		}
-		s.partHandlers[part-1].ServeHTTP(w, r)
+
+		n, _ := io.Copy(io.Discard, r.Body)
+		if c.failsLeft[part-1] == 0 {
+			if e, a := r.ContentLength, n; e != a {
+				c.t.Errorf("expect content length to be %d, got %d", e, a)
```
Madrigal: we take a member `t.Testing` only to make this call and a call to error if we can't close the request body (or at least this is the only call I could find). I see that before we were returning an error in the S3 shape before, and we still do that here when we return a `500 InternalException`, can we do that instead of taking a `t.Testing` object?

wty-Bryant: a union client behavior makes sense, will get rid of the testing field

wty-Bryant: I directly return 400 response here, for the request body close error since it happens finally in a defer func, I just log the possible close error (the body has been drained so failed close won't impact test)

Madrigal (approved):

PR #3491, opened 2026-07-24
Add SSE-C, versionId and other fields to the first HeadObject call on Transfer Manager

### LINKED ISSUE #3490: feature/s3/transfermanager: GetObject HeadObject calls don't forward SSE-C params
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk-go-v2/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date

### Describe the bug

`transfermanager.Client.GetObject()` calls `HeadObject` internally to determine object size before issuing parallel ranged GETs. These HeadObject calls only pass `Bucket` and `Key` — they do not forward `SSECustomerAlgorithm`, `SSECustomerKey`, or `SSECustomerKeyMD5` from the `GetObjectInput`.

For SSE-C encrypted objects, S3 requires the encryption key on every request including HeadObject. Without it, S3 returns `403 AccessDenied`, making the transfer manager unusable for SSE-C objects.

The ranged GET calls work correctly — `mapGetObjectInput()` properly forwards all SSE-C fields. It's only the HeadObject calls that are missing them.

### Affected code

In `feature/s3/transfermanager/api_op_GetObject.go`, the `getter.get()` method:

**GetObjectParts path (~line 629):**
```go
out, err := g.options.S3.HeadObject(ctx, &s3.HeadObjectInput{
    Bucket:     g.in.Bucket,
    Key:        g.in.Key,
    PartNumber: aws.Int32(1),
}, clientOptions...)
```

**GetObjectRanges path (~line 651):**
```go
out, err := g.options.S3.HeadObject(ctx, &s3.HeadObjectInput{
    Bucket: g.in.Bucket,
    Key:    g.in.Key,
}, clientOptions...)
```

Both should additionally include:
```go
    SSECustomerAlgorithm: g.in.SSECustomerAlgorithm,
    SSECustomerKey:       g.in.SSECustomerKey,
    SSECustomerKeyMD5:    g.in.SSECustomerKeyMD5,
```

### Regression Issue

### Expected Behavior

`transfermanager.GetObject()` should work with SSE-C encrypted objects when `SSECustomerAlgorithm`, `SSECustomerKey`, and `SSECustomerKeyMD5` are provided in `GetObjectInput`.

### Current Behavior

HeadObject returns `403 AccessDenied` because the SSE-C params are not forwarded, making parallel downloads impossible for SSE-C objects.

### Reproduction Steps

```go
tmClient := transfermanager.New(s3Client, func(o *transfermanager.Options) {
    o.GetObjectType = types.GetObjectRanges
})

// This fails with 403 on the internal HeadObject call
out, err := tmClient.GetObject(ctx, &transfermanager.GetObjectInput{
    Bucket:               aws.String("my-bucket"),
    Key:                  aws.String("my-sse-c-object"),
    SSECustomerAlgorithm: aws.String("AES256"),
    SSECustomerKey:       aws.String(base64Key),
    SSECustomerKeyMD5:    aws.String(keyMD5),
})
```

### Possible Solution

Add the three SSE-C fields to both `HeadObjectInput` structs in `getter.get()`. The fields are already available on `g.in` (the `GetObjectInput`).

### Workaround

Wrap the `S3APIClient` passed to `transfermanager.New()` with an implementation that intercepts `HeadObject` and injects the SSE-C params:

```go
type sseCClient struct {
    transfermanager.S3APIClient
    alg, key, keyMD5 *string
}

func (c *sseCClient) HeadObject(ctx context.Context, input *s3.HeadObjectInput, optFns ...func(*s3.Options)) (*s3.HeadObjectOutput, error) {
    input.SSECustomerAlgorithm = c.alg
    input.SSECustomerKey = c.key
    input.SSECustomerKeyMD5 = c.keyMD5
    return c.S3APIClient.HeadObject(ctx, input, optFns...)
}
```

- Note: `DownloadObject()` is NOT affected because it doesn't call HeadObject — it discovers size from the first ranged GET's `Content-Range` header.
- SSE-S3 (AES256) and SSE-KMS (aws:kms) are also NOT affected since they don't require client-side params on every request.

### SDK version used

`github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager v0.1.21` (also verified in v0.3.5)

### Environment details

Go 1.24, linux/amd64

Madrigal:
Fixes #3490

## Description

Fix `transfermanager.GetObject()` failing for SSE-C encrypted objects and other scenarios where request-level fields are required on every S3 API call.

### Problem

The transfer manager calls `HeadObject` internally to discover object size before issuing parallel ranged/part GETs. These `HeadObject` calls only passed `Bucket` and `Key` — they didn't forward fields from the user's input that S3 requires on every request to the object.

This caused failures for:
- **SSE-C encrypted objects** — S3 requires the encryption key on every request including HeadObject, returning 400 without it (reported at #3490)

Additionally, I took a look at the other fields we added when calling `GetObject` and added the following parameters

- **Requester-pays buckets** — S3 requires the `x-amz-request-payer` header, returning 403 without it
- **Version-specific downloads** — HeadObject could check a different version than what gets downloaded
- **Conditional requests** — `IfModifiedSince`/`IfUnmodifiedSince` weren't applied consistently
- **Cross-account safety** — `ExpectedBucketOwner` wasn't enforced on the HeadObject call

The ranged/part GET calls themselves were fine — `mapGetObjectInput()` already forwards all fields correctly. It was only the manually-constructed `HeadObject` calls that were missing them.

### Fix

Add the missing fields to both `HeadObjectInput` structs in `getter.get()` (the `GetObjectParts` path and the `GetObjectRanges` path):

- `SSECustomerAlgorithm`, `SSECustomerKey`, `SSECustomerKeyMD5`
- `RequestPayer`
- `ExpectedBucketOwner`
- `VersionId`
- `IfModifiedSince`, `IfUnmodifiedSince`

### Testing
Full repro at https://gist.github.com/Madrigal/27ccd834d4623bf5179f789199a90807
* Created a bucket with SSE-C configuration
* Uploaded a large object, verified it fails
* Apply the fix in this PR
* Ensure it suceeds

### DIFF
--- a/feature/s3/transfermanager/api_op_GetObject.go
+++ b/feature/s3/transfermanager/api_op_GetObject.go
@@ -630,9 +630,12 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 	if g.options.GetObjectType == types.GetObjectParts {
 		// must know the part size before creating stream reader
 		out, err := g.options.S3.HeadObject(ctx, &s3.HeadObjectInput{
-			Bucket:     g.in.Bucket,
-			Key:        g.in.Key,
-			PartNumber: aws.Int32(1),
+			Bucket:               g.in.Bucket,
+			Key:                  g.in.Key,
+			PartNumber:           aws.Int32(1),
+			SSECustomerAlgorithm: g.in.SSECustomerAlgorithm,
+			SSECustomerKey:       g.in.SSECustomerKey,
+			SSECustomerKeyMD5:    g.in.SSECustomerKeyMD5,
 		}, clientOptions...)
 		if err != nil {
 			return nil, err
@@ -652,8 +655,11 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 		r.partsCount = partsCount
 	} else {
 		out, err := g.options.S3.HeadObject(ctx, &s3.HeadObjectInput{
-			Bucket: g.in.Bucket,
-			Key:    g.in.Key,
+			Bucket:               g.in.Bucket,
+			Key:                  g.in.Key,
+			SSECustomerAlgorithm: g.in.SSECustomerAlgorithm,
+			SSECustomerKey:       g.in.SSECustomerKey,
+			SSECustomerKeyMD5:    g.in.SSECustomerKeyMD5,
 		}, clientOptions...)
 		if err != nil {
 			return nil, err
--- a/feature/s3/transfermanager/api_op_GetObject.go
+++ b/feature/s3/transfermanager/api_op_GetObject.go
@@ -636,6 +636,11 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 			SSECustomerAlgorithm: g.in.SSECustomerAlgorithm,
 			SSECustomerKey:       g.in.SSECustomerKey,
 			SSECustomerKeyMD5:    g.in.SSECustomerKeyMD5,
+			ExpectedBucketOwner:  g.in.ExpectedBucketOwner,
+			RequestPayer:         s3types.RequestPayer(g.in.RequestPayer),
+			VersionId:            g.in.VersionID,
+			IfModifiedSince:      g.in.IfModifiedSince,
+			IfUnmodifiedSince:    g.in.IfUnmodifiedSince,
 		}, clientOptions...)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


feature/s3/transfermanager/api_op_GetObject.go
```diff
@@ -630,9 +630,17 @@ func (g *getter) get(ctx context.Context) (out *GetObjectOutput, err error) {
 	if g.options.GetObjectType == types.GetObjectParts {
 		// must know the part size before creating stream reader
 		out, err := g.options.S3.HeadObject(ctx, &s3.HeadObjectInput{
-			Bucket:     g.in.Bucket,
-			Key:        g.in.Key,
-			PartNumber: aws.Int32(1),
+			Bucket:               g.in.Bucket,
```
wty-Bryant, line 633: maybe we can create a map func for head input, but not a big deal

wty-Bryant (approved):

feature/s3/transfermanager/api_op_GetObject.go
Madrigal: I've thought about this, on the conversion between GetObject and what the Transfer manager does we we read the mapping from `tm-mapping-reference.json` and from there we do the mapping. I'm not sure how widely used HeadObject is across SDKs, but we could define on that JSON how we want to translate between input and HeadObject, at the risk of making the reference file non standard. 

What do you think, is this what you had in mind?

wty-Bryant: from mapping reference perspective we don't encourage other sdks to rely on head object to download, so it makes sense to not create mapper func like normal i/o here, we can keep it as it is

PR #3498, opened 2026-07-28
Migrate serde2 kitchensink eventstream tests to smithy-go/eventstream

Tongs2000:
## Summary
  The `eventstream_test.go` files in the serde2 `internal/kitchensinktest` and `internal/kitchensinktestrestjson` modules imported the legacy `github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream` (+ `/eventstreamapi`) package, while the co-located generated `eventstream.go` already uses `github.com/aws/smithy-go/eventstream`.
  Because the generated code no longer references the legacy package, the dependency was only pulled in by the hand-written test, so it had a `require` line but no in-repo `replace` directive. When the smithy-go upgrade re-resolved every module's versions, `go mod tidy` in the clean preview build bumped the unpinned require to an unpublished pseudo-version whose `go.sum` entry was missing, breaking the build:
  ```
  missing go.sum entry for module providing package
  github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream
  ```
  ## Fix
  Migrate the tests to `smithy-go/eventstream` so they match the generated code. The APIs are identical; only the message-type constants moved from the `eventstreamapi` subpackage into the `eventstream` package itself. `go mod tidy` then drops the stale require line from `go.mod`/`go.sum`, removing the dependency that could be bumped. Legacy kitchensink modules are left unchanged: their generated code still uses the legacy package and their `go.mod` carries a local `replace` directive, so they remain self-consistent.
  ## Testing
  - `go test ./...` passes in both `internal/kitchensinktest` and `internal/kitchensinktestrestjson`.
  - Reproduced against the actual failed preview build artifact: on the original snapshot `go mod tidy` fails with `unknown revision aws/protocol/eventstream/v1.7.15-zeta...` (the smithy-go bump had re-resolved the unpinned require to an unpublished pseudo-version). After applying this change, `aws/protocol/eventstream` drops out of `go.mod`/`go.sum` entirely and `go mod verify` reports `all modules verified`.

### DIFF
--- a/internal/kitchensinktest/eventstream_test.go
+++ b/internal/kitchensinktest/eventstream_test.go
@@ -9,8 +9,7 @@ import (
 	"sync/atomic"
 	"testing"
 
-	"github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream"
-	"github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream/eventstreamapi"
+	"github.com/aws/smithy-go/eventstream"
 	smithyendpoints "github.com/aws/smithy-go/endpoints"
 	"github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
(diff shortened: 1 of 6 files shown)


lucix-aws (approved):
merge on ci pass

wty-Bryant (approved):
waiting for ci

PR #3499, opened 2026-07-29
Set FIPS-approved TLS curve preferences when the FIPS module is active

### LINKED ISSUE #3345: `defaultHTTPTransport` does not set `CurvePreferences`, breaking `GODEBUG=fips140=only` on Go 1.25+
### Acknowledgements

### Describe the bug

All AWS SDK API calls fail under Go 1.25+ with `GODEBUG=fips140=only`:

```
crypto/ecdh: use of X25519 is not allowed in FIPS 140-only mode
```

The SDK's `defaultHTTPTransport()` does not set `CurvePreferences` on its `tls.Config`, so Go's defaults apply -- which include X25519. Go 1.25 enforces FIPS rejection of X25519 during TLS handshake (Go 1.24 introduced `fips140=only` but did not enforce X25519 rejection due to [golang/go@10cef81](https://github.com/golang/go/commit/10cef816aa9769345016c04032090ae7f5851f5c)).

| Go | `fips140=only` | Client | Result |
|----|----------------|--------|--------|
| 1.24 | yes | SDK default | OK -- X25519 not enforced |
| 1.25 | yes | SDK default | **FAIL** -- X25519 rejected |
| 1.25 | yes | custom (P-256/P-384/P-521 only) | OK |

### Regression Issue

### Expected Behavior

Both tests should succeed -- the SDK default client should work under `GODEBUG=fips140=only` without requiring a custom HTTP client:

```
--- Test 1: SDK default client (no workaround) ---
  OK

--- Test 2: custom client with FIPS-safe curves ---
  OK
```

### Current Behavior

Test 1 fails at the TLS level because the SDK default transport negotiates X25519:

```
--- Test 1: SDK default client (no workaround) ---
  FAIL: operation error STS: GetCallerIdentity, exceeded maximum number of attempts, 3,
    https response error StatusCode: 0, RequestID: , request send failed,
    Post "https://sts-fips.us-east-1.amazonaws.com/":
    crypto/ecdh: use of X25519 is not allowed in FIPS 140-only mode

--- Test 2: custom client with FIPS-safe curves ---
  OK
```

Test 2 succeeds only because it explicitly sets FIPS-safe `CurvePreferences`.

### Reproduction Steps

Save and run the following with Go 1.25+ built with `GOFIPS140=latest`:

```go
package main

import (
	"context"
	"crypto/tls"
	"fmt"
	"net/http"
	"os"

	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
)

func main() {
	fmt.Printf("GODEBUG=%s\n\n", os.Getenv("GODEBUG"))

	printSDKTransportDefaults()

	fmt.Println("\n--- Test 1: SDK default client (no workaround) ---")
	callSTS(nil)

	fmt.Println("\n--- Test 2: custom client with FIPS-safe curves ---")
	callSTS(fipsSafeHTTPClient())
}

func printSDKTransportDefaults() {
	tr := awshttp.NewBuildableClient().GetTransport()
	tc := tr.TLSClientConfig
	fmt.Println("SDK default transport TLS config:")
	fmt.Printf("  MinVersion: %s\n", tlsVersionName(tc.MinVersion))
	if len(tc.CurvePreferences) == 0 {
		fmt.Println("  CurvePreferences: [] (Go defaults, includes X25519)")
	} else {
		fmt.Printf("  CurvePreferences: %v\n", tc.CurvePreferences)
	}
}

func callSTS(httpClient *http.Client) {
	ctx := context.Background()
	opts := []func(*config.LoadOptions) error{
	}
	if httpClient != nil {
		opts = append(opts, config.WithHTTPClient(httpClient))
	}

	cfg, err := config.LoadDefaultConfig(ctx, opts...)
	if err != nil {
		fmt.Printf("  config error: %v\n", err)
		return
	}

	_, err = sts.NewFromConfig(cfg).GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
	if err != nil {
		fmt.Printf("  FAIL: %v\n", err)
		return
	}
	fmt.Println("  OK")
}

func fipsSafeHTTPClient() *http.Client {
	return &http.Client{
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{
				MinVersion: tls.VersionTLS12,
				CurvePreferences: []tls.CurveID{
					tls.CurveP256,
					tls.CurveP384,
					tls.CurveP521,
				},
			},
		},
	}
}

func tlsVersionName(v uint16) string {
	switch v {
	case tls.VersionTLS10:
		return "TLS 1.0"
	case tls.VersionTLS11:
		return "TLS 1.1"
	case tls.VersionTLS12:
		return "TLS 1.2"
	case tls.VersionTLS13:
		return "TLS 1.3"
	default:
		return fmt.Sprintf("0x%04x", v)
	}
}
```

#### Command to Execute
```bash 
GODEBUG=fips140=only GOFIPS140=latest AWS_USE_FIPS_ENDPOINT=true go run main.go
```

### Possible Solution

In [`aws/transport/http/client.go`](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/transport/http/client.go), `defaultHTTPTransport()` only sets `MinVersion`:

Set FIPS-safe `CurvePreferences` in `defaultHTTPTransport()`, or detect `fips140=only` at runtime and exclude non-FIPS curves.

_No response_

- aws-sdk-go-v2 v1.41.3
- aws-sdk-go-v2/config v1.32.11
- aws-sdk-go-v2/service/sts v1.41.8

### Compiler and Version used

go version go1.25.0 darwin/arm64

macOS 15.7.3 (Sequoia), build 24G419

ATKasem:
Fixes #3345

### Problem

Every SDK request fails on Go 1.25+ under `GODEBUG=fips140=only`:

```
crypto/ecdh: use of X25519 is not allowed in FIPS 140-only mode
```

`defaultHTTPTransport()` built its `tls.Config` with only `MinVersion` set, leaving `CurvePreferences` empty so Go's defaults applied. Those lead with X25519, which `crypto/ecdh` rejects outright in FIPS 140-only mode, so the handshake dies before any AWS call is made. Go 1.24 introduced `fips140=only` but did not enforce the X25519 rejection; 1.25 does, which is why this surfaced now.

### Approach

The issue suggested either setting FIPS-safe `CurvePreferences` outright or detecting FIPS at runtime. I went with the second: setting them unconditionally would strip X25519 *and* the Go 1.24+ post-quantum `X25519MLKEM768` hybrid from every SDK user, the overwhelming majority of whom are not in FIPS mode. That is a real downgrade to pay for a FIPS-only bug.

Instead, `defaultTLSCurvePreferences` returns `nil` normally — Go's defaults, completely unchanged — and the NIST curves only when `crypto/fips140.Enabled()` reports the module is active. The module targets `go 1.24`, so `crypto/fips140` is available without build tags.

This also covers `fips140=on`, where X25519 is permitted but not approved; restricting to approved curves there matches FIPS intent rather than leaving a second inconsistent mode.

### Verification

Behavior of the default transport, using a small program that reads `NewBuildableClient().GetTransport().TLSClientConfig` against this branch:

**Before (`origin/main`)**

```
--- normal mode ---
fips140.Enabled() = false
CurvePreferences: [] (Go defaults, includes X25519)
--- GODEBUG=fips140=only ---
fips140.Enabled() = true
CurvePreferences: [] (Go defaults, includes X25519)   <-- handshake fails here
```

**After (this branch)**

```
--- normal mode ---
fips140.Enabled() = false
CurvePreferences: [] (Go defaults, includes X25519)   <-- unchanged
--- GODEBUG=fips140=only ---
fips140.Enabled() = true
CurvePreferences: [P-256 P-384 P-521]
```

The normal-mode line is identical before and after, which is the point: non-FIPS users keep X25519 and the post-quantum hybrid.

Unit tests cover both branches of the selection, assert no non-approved curve can appear under FIPS, and check the transport's own TLS config. They pass in both modes:

```
$ GOFIPS140=latest GODEBUG=fips140=only go test -v -run 'TestDefaultTLSCurvePreferences|TestDefaultHTTPTransport_TLSConfig' ./aws/transport/http/
--- PASS: TestDefaultTLSCurvePreferences (0.00s)
    --- PASS: TestDefaultTLSCurvePreferences/fips_enabled_restricts_to_approved_curves (0.00s)
    --- PASS: TestDefaultTLSCurvePreferences/fips_disabled_defers_to_go_defaults (0.00s)
--- PASS: TestDefaultTLSCurvePreferences_NoX25519UnderFIPS (0.00s)
--- PASS: TestDefaultHTTPTransport_TLSConfig (0.00s)
ok  	github.com/aws/aws-sdk-go-v2/aws/transport/http	0.360s
```

Also clean: `go build ./...`, `go vet ./aws/transport/http/`, `gofmt -l` (no output), and the full `./aws/transport/http/...` package tests.

I did not run the issue's end-to-end STS reproduction, since it needs live AWS credentials against a FIPS endpoint. The transport-level output above isolates the same root cause.

Per CONTRIBUTING, no `.changelog` entry is included.

### DIFF
--- a/aws/transport/http/client.go
+++ b/aws/transport/http/client.go
@@ -2,6 +2,7 @@ package http
 
 import (
 	"context"
+	"crypto/fips140"
 	"crypto/tls"
 	"net"
 	"net/http"
@@ -27,6 +28,19 @@ var (
 
 	// Default to TLS 1.2 for all HTTPS requests.
 	DefaultHTTPTransportTLSMinVersion uint16 = tls.VersionTLS12
+
+	// DefaultHTTPTransportTLSCurvePreferencesFIPS is the elliptic curve preference
+	// list applied to the default transport when the FIPS 140-3 module is active.
+	//
+	// Go's default preferences lead with X25519, which crypto/ecdh rejects under
+	// GODEBUG=fips140=only, failing every TLS handshake the SDK attempts. Only the
+	// NIST curves are FIPS-approved, so restricting to them keeps the default
+	// client usable in FIPS deployments.
+	DefaultHTTPTransportTLSCurvePreferencesFIPS = []tls.CurveID{
+		tls.CurveP256,
+		tls.CurveP384,
+		tls.CurveP521,
+	}
 )
 
 // Timeouts for net.Dialer's network connection.
@@ -178,6 +192,16 @@ func defaultDialer() *net.Dialer {
 	}
 }
 
+// defaultTLSCurvePreferences returns the curve preferences for the default
+// transport. Outside FIPS mode it returns nil so Go's own defaults apply,
+// preserving X25519 and the post-quantum X25519MLKEM768 hybrid.
+func defaultTLSCurvePreferences(fipsEnabled bool) []tls.CurveID {
+	if !fipsEnabled {
+		return nil
+	}
+	return DefaultHTTPTransportTLSCurvePreferencesFIPS
+}
+
 func defaultHTTPTransport() *http.Transport {
 	dialer := defaultDialer()
 
@@ -192,7 +216,8 @@ func defaultHTTPTransport() *http.Transport {
 		ExpectContinueTimeout: DefaultHTTPTransportExpectContinueTimeout,
 		ForceAttemptHTTP2:     true,
 		TLSClientConfig: &tls.Config{
-			MinVersion: DefaultHTTPTransportTLSMinVersion,
+			MinVersion:       DefaultHTTPTransportTLSMinVersion,
+			CurvePreferences: defaultTLSCurvePreferences(fips140.Enabled()),
 		},
 	}
 
--- a/aws/transport/http/client_test.go
+++ b/aws/transport/http/client_test.go
@@ -1,6 +1,8 @@
 package http
 
 import (
+	"crypto/fips140"
+	"crypto/tls"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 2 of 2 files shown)


lucix-aws (approved):

wty-Bryant (commented):
Codegen and integ tests failed because pr is from fork repo branch, merged since other go tests passed

PR #3505, opened 2026-07-31
upgrade to smithy-go v1.27.6

### LINKED ISSUE #3504: service/pricingplanmanager: all subscription operations fail to deserialize — httpbinding leaks header-bindings mode into struct httpPayload members
### Acknowledgements
- [x] I have searched for duplicate or closed issues
- [x] I have read the [Contributing Guidelines](https://github.com/aws/aws-sdk-go-v2/blob/main/CONTRIBUTING.md)

### Describe the bug

`service/pricingplanmanager` v1.0.0 (new schema-based smithy-go codegen) cannot deserialize the response of **any** subscription operation (`GetSubscription`, `CreateSubscription`, `UpdateSubscription`, `CancelSubscription`, `AssociateResourcesToSubscription`, ...). Every call against the real service fails with:

```
operation error Pricing Plan Manager: GetSubscription, https response error StatusCode: 200, RequestID: ...,
deserialization failed, unable to parse time string, parse errors:
 * "2006-01-02T15:04:05.999999999Z": parsing time "" as "2006-01-02T15:04:05.999999999Z": cannot parse "" as "2006"
 ...
```

Note the HTTP status is 200 and the body is well-formed JSON — the failure is client-side.

### Root cause (smithy-go `transport/http/protocol/internal/httpbinding`)

These operations' outputs have **both** a header binding (`eTag` ← `ETag` header) and a **structure-typed `httpPayload` member** (`subscription`). In smithy-go v1.27.5 `httpbinding.ShapeDeserializer`:

1. `isBindingSet` returns `true` for the `httpPayload` member when the payload is non-empty, so the payload member is appended to `d.bindings`.
2. `ReadStructMember` yields each binding member with `d.inBindings = true`, and nothing resets it before the generated code descends into the payload struct's `Deserialize`.
3. Every nested member of `Subscription` is then read through the bindings path: `ReadString` calls `readHeaderString` — the nested members have no `httpHeader` trait, so `isHTTPHeader` yields an empty name and `Header.Get("")` returns `""` (silent data corruption); the first timestamp member (`createdAt`) goes through `readHeaderTime`/`parseTimestamp("")` and fails hard with the error above.

Instrumenting `ReadTime` confirms it: it is invoked for `Subscription.createdAt` with `inBindings=true, inBody=false, depth=2`.

`ListSubscriptions` works because its output has no header bindings.

### Regression Issue

New service module; broken since its initial release.

### Expected Behavior

The struct-typed `httpPayload` member is deserialized from the response body; header bindings only apply to the members that carry HTTP binding traits. `GetSubscription` returns the parsed `Subscription`.

### Current Behavior

All operations returning `ETag` header + `subscription` payload fail with the deserialization error above; had there been no timestamp members, nested strings would silently deserialize as `""`.

### Reproduction Steps

Fails identically against the live service, but reproducible offline:

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"io"
	"net/http"

	"github.com/aws/aws-sdk-go-v2/service/pricingplanmanager"
)

type fakeRT struct{}

func (fakeRT) Do(*http.Request) (*http.Response, error) {
	h := http.Header{}
	h.Set("Content-Type", "application/json")
	h.Set("ETag", "2")
	body := `{"arn":"arn:aws:pricingplanmanager::111122223333:subscription:sub_X","createdAt":"2026-06-18T14:42:24.679859965Z","planFamily":"CloudFront","planTier":"FREE","resourceArns":["a","b"],"scheduledChange":null,"status":"ACTIVE","statusReason":null,"updatedAt":"2026-06-18T14:42:46.090267004Z","usageLevel":null}`
	return &http.Response{StatusCode: 200, Header: h, Body: io.NopCloser(bytes.NewReader([]byte(body)))}, nil
}

func main() {
	client := pricingplanmanager.New(pricingplanmanager.Options{
		Region:     "us-east-1",
		HTTPClient: fakeRT{},
		Credentials: aws.CredentialsProviderFunc(func(context.Context) (aws.Credentials, error) {
			return aws.Credentials{AccessKeyID: "AKID", SecretAccessKey: "SECRET"}, nil
		}),
	})
	_, err := client.GetSubscription(context.Background(), &pricingplanmanager.GetSubscriptionInput{Arn: aws.String("arn:x")})
	fmt.Println(err)
}
```

### Possible Solution

In `httpbinding.ShapeDeserializer.ReadStructMember`, clear `d.inBindings` before yielding a member that carries the `httpPayload` trait (nested struct reads then correctly delegate to the body deserializer via the `depth > 1` path in `ReadStruct`). Alternatively, exclude structure-typed `httpPayload` members from `d.bindings` and dispatch them after the header bindings with `inBindings=false`.

- github.com/aws/aws-sdk-go-v2 v1.43.2
- github.com/aws/aws-sdk-go-v2/service/pricingplanmanager v1.0.0
- github.com/aws/smithy-go v1.27.5

### Compiler and Version used

go version go1.26.5 darwin/arm64

macOS 26 (Darwin 25.5.0), also reproduced on Linux

lucix-aws:
closes #3504 

our brand new snapshot tests passing verifies this, since they caught the deserializer bug as well


wty-Bryant (approved):

lucix-aws:
(ci needs to pass)

service/pricingplanmanager/response_snapshot/AssociateResourcesToSubscription.response.snap
```diff
@@ -2,4 +2,4 @@
 Content-Type: application/json
 Etag: __ETag__
 
-{"arn":"__Arn__","createdAt":"2000-01-01T00:00:00Z","planFamily":"__PlanFamily__","planTier":"__PlanTier__","resourceArns":["__Member__","__Member__"],"scheduledChange":{"changeType":"DOWNGRADE","effectiveDate":"2000-01-01T00:00:00Z","planTier":"__PlanTier__","usageLevel":"__UsageLevel__","status":"PENDING_APPROVAL","statusReason":"__StatusReason__","updatedAt":"2000-01-01T00:00:00Z","usageLevel":"__UsageLevel__"}}
```
Madrigal, line 5: (non blocking) this was invalid JSON

Madrigal (approved):

PR #3514, opened 2026-08-12
Close each file after its transfer finished during directory upload/download

### LINKED ISSUE #3512: transfermanager.UploadDirectory does not close files that it opens
### Acknowledgements

### Describe the bug

I have noticed that the S3 `transfermanager` is not closing files that it opens when calling `UploadDirectory`. This has caused problems in my app which calls `UploadDirectory` and then immediately tries to remove the uploaded directory afterwards with `os.RemoveAll`. Because there are still open file handles, this results in a "directory not empty" error (note this will only happen for certain filesystems, like in an NFS volume where there will be a bunch of `.nfs...` files for the open file handles).

### Regression Issue

### Expected Behavior

`transfermanager.UploadDirectory` should close all file handles it opens before it returns.

### Current Behavior

`transfermanager.UploadDirectory` leaves files open after returning.

### Reproduction Steps

Here is a simple app that will reproduce the issue. Create a directory with some files to upload, then run the app while specifying the `dir`, `bucket`, and `keyPrefix` flags, and it will print out the open file handles at the end of its run:

<details>
<summary>Reproduction code</summary>  

```go
package main

import (
	"context"
	"flag"
	"fmt"
	"os"
	"sync"

	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
)

func main() {
	dir := flag.String("dir", "", "directory to upload")
	bucket := flag.String("bucket", "", "bucket to upload to")
	keyPrefix := flag.String("keyPrefix", "", "keyPrefix to upload to")
	flag.Parse()

	if err := run(dir, bucket, keyPrefix); err != nil {
		fmt.Fprintf(os.Stderr, "ERROR: %v\n", err)
		os.Exit(1)
	}
}

func run(dir, bucket, keyPrefix *string) error {
	ctx := context.Background()
	captured := &fileCapture{}

	awsCfg, err := config.LoadDefaultConfig(ctx)
	if err != nil {
		return fmt.Errorf("config.LoadDefaultConfig: %w", err)
	}

	s3client := s3.NewFromConfig(awsCfg)
	client := transfermanager.New(s3client)

	output, err := client.UploadDirectory(ctx, &transfermanager.UploadDirectoryInput{
		Source:    dir,
		Bucket:    bucket,
		KeyPrefix: keyPrefix,
		Callback:  captured,
	})
	if err != nil {
		return fmt.Errorf("UploadDirectory: %w", err)
	}

	openFiles := captured.openFiles()

	fmt.Printf("objects uploaded: %d\n", output.ObjectsUploaded)
	fmt.Printf("file handles captured: %d\n", captured.len())
	fmt.Printf("file handles still open after UploadDirectory returned: %d\n", len(openFiles))
	for _, file := range openFiles {
		fmt.Printf("  OPEN: %s (fd=%d)\n", file.Name(), file.Fd())
	}

	return nil
}

// fileCapture receives the original *os.File bodies opened by UploadDirectory.
type fileCapture struct {
	mu    sync.Mutex
	files []*os.File
}

func (capture *fileCapture) UpdateRequest(input *transfermanager.UploadObjectInput) {
	file, ok := input.Body.(*os.File)
	if !ok {
		return
	}

	capture.mu.Lock()
	defer capture.mu.Unlock()
	capture.files = append(capture.files, file)
}

func (capture *fileCapture) len() int {
	capture.mu.Lock()
	defer capture.mu.Unlock()
	return len(capture.files)
}

func (capture *fileCapture) openFiles() []*os.File {
	capture.mu.Lock()
	defer capture.mu.Unlock()

	var open []*os.File
	for _, file := range capture.files {
		if _, err := file.Stat(); err == nil { // Stat should fail if the file is closed
			open = append(open, file)
		}
	}
	return open
}
```

</details>

Example run:

```
> ls assets
bar.txt  baz.txt  foo.txt
> go build -o repro
> ./repro -dir assets -bucket my.bucket -keyPrefix example
objects uploaded: 3
file handles captured: 3
file handles still open after UploadDirectory returned: 3
  OPEN: assets/bar.txt (fd=5)
  OPEN: assets/foo.txt (fd=7)
  OPEN: assets/baz.txt (fd=6)
```

### Possible Solution

Close any opened file handles before UploadDirectory returns

_No response_

```go
module uploaddirectoryrepro

go 1.24

require (
	github.com/aws/aws-sdk-go-v2/config v1.32.36
	github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager v0.3.12
	github.com/aws/aws-sdk-go-v2/service/s3 v1.107.1
)

require (
	github.com/aws/aws-sdk-go-v2 v1.43.5 // indirect
	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.7.17 // indirect
	github.com/aws/aws-sdk-go-v2/credentials v1.19.35 // indirect
	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.18.36 // indirect
	github.com/aws/aws-sdk-go-v2/internal/configsources v1.4.36 // indirect
	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.7.36 // indirect
	github.com/aws/aws-sdk-go-v2/internal/v4a v1.4.37 // indirect
	github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.16 // indirect
	github.com/aws/aws-sdk-go-v2/service/internal/checksum v1.9.29 // indirect
	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.36 // indirect
	github.com/aws/aws-sdk-go-v2/service/internal/s3shared v1.19.37 // indirect
	github.com/aws/aws-sdk-go-v2/service/signin v1.5.5 // indirect
	github.com/aws/aws-sdk-go-v2/service/sso v1.33.5 // indirect
	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.38.5 // indirect
	github.com/aws/aws-sdk-go-v2/service/sts v1.45.5 // indirect
	github.com/aws/smithy-go v1.27.7 // indirect
)
```

### Compiler and Version used

go version go1.26.4 darwin/arm64

macOS 26.6.1

wty-Bryant:
Fixes #3512 , this pr fixes the open file leak for both directory upload/download

### DIFF
--- a/feature/s3/transfermanager/api_op_DownloadDirectory.go
+++ b/feature/s3/transfermanager/api_op_DownloadDirectory.go
@@ -288,10 +288,9 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 		err = d.failurePolicy.OnDownloadFailed(d.in, input, err)
 		if err != nil {
 			return fmt.Errorf("error when heading info of object %s: %v", data.key, err)
-		} else {
-			d.objectsFailed.Add(1)
-			return nil
 		}
+		d.objectsFailed.Add(1)
+		return nil
 	}
 
 	d.progressOnce.Do(func() {
@@ -323,10 +322,9 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 		err = d.failurePolicy.OnDownloadFailed(d.in, input, err)
 		if err != nil {
 			return fmt.Errorf("error when getting object and writing to local file %s: %v", data.path, err)
-		} else {
-			d.objectsFailed.Add(1)
-			return nil
 		}
+		d.objectsFailed.Add(1)
+		return nil
 	}
 
 	d.objectsDownloaded.Add(1)
--- a/feature/s3/transfermanager/api_op_DownloadDirectory.go
+++ b/feature/s3/transfermanager/api_op_DownloadDirectory.go
@@ -269,13 +269,11 @@ func (d *directoryDownloader) downloadObject(ctx context.Context, ch chan object
 			continue
 		}
 
-		if err := d.downloadSingleObject(ctx, data); err != nil {
-			d.setErr(err)
-		}
+		d.downloadSingleObject(ctx, data)
 	}
 }
 
-func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data objectEntry) error {
+func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data objectEntry) {
 	input := &GetObjectInput{
 		Bucket: d.in.Bucket,
 		Key:    aws.String(data.key),
@@ -287,10 +285,11 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 	if err != nil {
 		err = d.failurePolicy.OnDownloadFailed(d.in, input, err)
 		if err != nil {
-			return fmt.Errorf("error when heading info of object %s: %v", data.key, err)
+			d.setErr(fmt.Errorf("error when heading info of object %s: %v", data.key, err))
+			return
 		}
 		d.objectsFailed.Add(1)
-		return nil
+		return
 	}
 
 	d.progressOnce.Do(func() {
@@ -299,12 +298,14 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 
 	err = os.MkdirAll(filepath.Dir(data.path), 0755)
 	if err != nil {
-		return fmt.Errorf("error when creating directory for file %s: %v", data.path, err)
+		d.setErr(fmt.Errorf("error when creating directory for file %s: %v", data.path, err))
+		return
 	}
 
 	file, err := createFileFn(data.path)
 	if err != nil {
-		return fmt.Errorf("error when creating file %s: %v", data.path, err)
+		d.setErr(fmt.Errorf("error when creating file %s: %v", data.path, err))
+		return
 	}
 	var fileCopyFail bool
 	defer func() {
@@ -321,15 +322,15 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 		// where s3.GetObject is really called, must be handled by failure policy
 		err = d.failurePolicy.OnDownloadFailed(d.in, input, err)
 		if err != nil {
-			return fmt.Errorf("error when getting object and writing to local file %s: %v", data.path, err)
+			d.setErr(fmt.Errorf("error when getting object and writing to local file %s: %v", data.path, err))
+			return
 		}
 		d.objectsFailed.Add(1)
-		return nil
+		return
 	}
 
 	d.objectsDownloaded.Add(1)
 	d.emitter.ObjectsTransferred(ctx, n)
-	return nil
 }
 
 func (d *directoryDownloader) freshContext(ctx context.Context) (context.Context, context.CancelFunc) {
--- a/feature/s3/transfermanager/api_op_DownloadDirectory.go
+++ b/feature/s3/transfermanager/api_op_DownloadDirectory.go
@@ -17,6 +17,13 @@ import (
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 )
 
+// createFileFn is a thin indirection over os.Create used when writing
+// downloaded objects to disk. It exists so tests can verify that every file
+// handle DownloadDirectory creates is subsequently closed, guarding against a
+// file-handle leak analogous to the one fixed for aws/aws-sdk-go-v2#3512. In
+// prod env it is exactly os.Create.
+var createFileFn = os.Create
+
 // DownloadDirectoryInput represents a request to the DownloadDirectory() call
 type DownloadDirectoryInput struct {
 	// Bucket where objects are downloaded from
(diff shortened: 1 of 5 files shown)


feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -262,54 +269,67 @@ func (d *directoryDownloader) downloadObject(ctx context.Context, ch chan object
...
-		file, err := os.Create(data.path)
-		if err != nil {
-			d.setErr(fmt.Errorf("error when creating file %s: %v", data.path, err))
-			continue
+		if fileCopyFail {
+			os.Remove(data.path) // only remove the file if the copy failed
 		}
-		n, err := io.Copy(file, out.Body)
+	}()
+	n, err := io.Copy(file, out.Body)
```
Tongs2000, line 319: Overall this looks great. One small corner case I wanted to confirm. In `downloadSingleObject` Since the `defer` runs before the caller's `setErr`, and the new `setErr` keeps the first non-cancellation error, if `io.Copy` fails and `Close()` also fails on the same object, the surfaced error will be the close error rather than the copy error. Which do we actually want to report? Same shape exists in `uploadSingleFile`.

wty-Bryant: actually that's a subtle case where the close error will be surfaced first, I think we can just set error within single object func here, good catch

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -700,3 +701,168 @@ func TestDownloadDirectoryWithContextCanceled(t *testing.T) {
...
+					ContentLength: aws.Int64(int64(len(c.Data))),
+					PartsCount:    aws.Int32(c.PartsCount),
+					ETag:          aws.String(etag),
+				}, nil
+			},
+			failurePolicy: IgnoreDownloadFailurePolicy{},
+			// GetObject only performs a HeadObject up front and streams the
+			// body lazily, so os.Create runs for all four objects. foo/zoo/bar
+			// and baz then fail during io.Copy (body read), so their files are
+			// created, closed, and removed; the other two download normally.
```
Madrigal, line 809: Do we have a test somewhere else of this behavior, where if a file fails to download we actually remove the empty file we created before reading the body?

feature/s3/transfermanager/upload_directory_test.go
```diff
@@ -645,3 +646,156 @@ func TestUploadDirectoryWithContextCanceled(t *testing.T) {
...
+func (c *openFileCapture) count() int {
+	return len(c.files)
+}
+
+// stillOpen returns the names of captured files that are still open. A closed
+// *os.File returns an error from Stat, so a nil error means the handle leaked.
+func (c *openFileCapture) stillOpen() []string {
+	var open []string
+	for _, f := range c.files {
+		if _, err := f.Stat(); err == nil {
```
Madrigal: nit: Ideally I'd like to check we actually get `fs.ErrClosed` instead of a generic error, but one of your tests is doing Stat on a file you deleted, so you'd get a different error

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -700,3 +701,168 @@ func TestDownloadDirectoryWithContextCanceled(t *testing.T) {
...
+	sort.Strings(open)
+	return open
+}
+
+// TestDownloadDirectoryClosesCreatedFiles verifies that DownloadDirectory
+// closes every file handle it creates before returning - on the success path
+// and when individual object downloads fail and are ignored. Regression test
+// for a file-handle leak analogous to aws/aws-sdk-go-v2#3512.
+func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
+	_, filename, _, _ := runtime.Caller(0)
```
Madrigal: Do you need `runtime.Caller(0)`? I think you can just use `filepath.Join` like we do here https://github.com/aws/aws-sdk-go-v2/blob/fe3cb6a41e2183899e83c5a35e3fde5402b18d85/credentials/ssocreds/sso_cached_token_test.go#L71

lucix-aws (approved):

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -262,54 +269,68 @@ func (d *directoryDownloader) downloadObject(ctx context.Context, ch chan object
...
+
+	file, err := createFileFn(data.path)
+	if err != nil {
+		d.setErr(fmt.Errorf("error when creating file %s: %v", data.path, err))
+		return
+	}
+	var fileCopyFail bool
+	defer func() {
+		if err := file.Close(); err != nil {
+			d.setErr(fmt.Errorf("error when closing file %s: %v", data.path, err))
```
Madrigal, line 313: Wouldn't this override an existing issue if we had already call `setErr`? I suspect that may be more important than reporting we failed to close the file.

feature/s3/transfermanager/download_directory_test.go
wty-Bryant: that's a simplification

feature/s3/transfermanager/api_op_DownloadDirectory.go
wty-Bryant: that's why I change setErr here to align with ur earlier change: it will only override a nil error or cancellation error, and setErr is called right after an err is found as tao suggested

feature/s3/transfermanager/upload_directory_test.go
wty-Bryant: yea errorClosed is more straightforward showing "we get an error from closed file"

feature/s3/transfermanager/api_op_DownloadDirectory.go
Madrigal: I was thinking more of actually checking `d.err` somehow, but if you're confident this will only override a nil or cancellation error I'm OK with it

feature/s3/transfermanager/download_directory_test.go
wty-Bryant: okay I replace all download directory destination path to relative path, for upload directory, since there's some symlink relative path test config, I prefer to leave that as it was. But I did find a real bug in upload directory's symlink traverse where we should resolve a relative symlink from the initial path itself instead of its parent folder.

feature/s3/transfermanager/upload_directory_test.go
wty-Bryant: update the logic to directly check os.ErrClosed

Madrigal (approved):

PR #3518, opened 2026-08-14
Fix route53 error close body

Tongs2000:
## Description
  
  The Route53 custom error deserialization middleware (`custom_error_deser.go`) leaked the HTTP connection on the `InvalidChangeBatch` error path.

  Two issues:

  1. **Dropped Closer**: the middleware rewinds the response body with `io.NopCloser(io.MultiReader(...))`, whose `Close` is a no-op — so closing the response body never reached the underlying transport body.

  2. **Skipped close (Route53-specific)**: when the root element is `InvalidChangeBatch`, this middleware deserializes it into an error and returns it. When a deserialize middleware returns a non-nil error, the caller forwards it up without closing the response body — the convention being that an error path either has no reusable body (e.g. a transport failure) or the returning layer has already cleaned it up. Route53 returned an error here without closing the body, so on this path the connection was neither drained nor released.

  This is the same class of bug fixed for S3 in #3517, but Route53 additionally requires the explicit close because it returns an error (S3 only rewrites the status code and returns `nil`).

  ## Fix

  - Forward the original body's `Closer` via a small `replayReadCloser` (same pattern as the S3 fix) so closing the response body reaches the transport body.

  - Explicitly call `CloseResponseBody` on the `InvalidChangeBatch` path, after custom error deserialization has read the body, so the connection is drained and released.

  ## Testing

  Added `TestResponseBodyClosed`, which asserts the original response body is closed on both the error and success paths. Existing `TestCustomErrorDeserialization` continues to pass.

### DIFF
--- a/service/route53/internal/customizations/custom_error_deser.go
+++ b/service/route53/internal/customizations/custom_error_deser.go
@@ -60,12 +60,20 @@ func (m *processResponse) HandleDeserialize(
 		return out, metadata, nil
 	}
 
-	// rewind response body
-	response.Body = io.NopCloser(io.MultiReader(&readBuff, response.Body))
+	// rewind response body, forwarding the original body's Closer so closing it
+	// still closes the underlying transport body (io.NopCloser would drop it).
+	originalBody := response.Body
+	response.Body = &replayReadCloser{
+		Reader: io.MultiReader(&readBuff, originalBody),
+		Closer: originalBody,
+	}
 
 	// if start tag is "InvalidChangeBatch", the error response needs custom unmarshaling.
 	if strings.EqualFold(t.Name.Local, "InvalidChangeBatch") {
-		return out, metadata, route53CustomErrorDeser(&metadata, response)
+		// The caller won't close the response body when we return an error, so close it here.
+		err = route53CustomErrorDeser(&metadata, response)
+		smithyhttp.CloseResponseBody(ctx, response, false, err)
+		return out, metadata, err
 	}
 
 	return out, metadata, err
@@ -90,4 +98,10 @@ func route53CustomErrorDeser(metadata *middleware.Metadata, response *smithyhttp
 		Message:  ptr.String("ChangeBatch errors occurred"),
 		Messages: err.Messages,
 	}
+}
+
+// replayReadCloser pairs a Reader with an independent Closer.
+type replayReadCloser struct {
+	io.Reader
+	io.Closer
 }
--- a/service/route53/internal/customizations/custom_error_deser_test.go
+++ b/service/route53/internal/customizations/custom_error_deser_test.go
@@ -3,6 +3,7 @@ package customizations_test
 import (
 	"context"
 	"errors"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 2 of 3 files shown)


service/route53/internal/customizations/custom_error_deser.go
```diff
@@ -60,12 +60,20 @@ func (m *processResponse) HandleDeserialize(
 		return out, metadata, nil
 	}
 
-	// rewind response body
-	response.Body = io.NopCloser(io.MultiReader(&readBuff, response.Body))
+	// rewind response body, forwarding the original body's Closer so closing it
+	// still closes the underlying transport body (io.NopCloser would drop it).
+	originalBody := response.Body
+	response.Body = &replayReadCloser{
```
wty-Bryant, line 66: nit: go assignment for new response.Body happens after replayReadCloser definition on right hand side so we can just replace originalBody with response.Body here, non-block

service/route53/internal/customizations/custom_error_deser_test.go
```diff
@@ -130,3 +131,56 @@ type fakeCredentials struct{}
...
+func TestResponseBodyClosed(t *testing.T) {
+	cases := map[string]string{
+		"error":   `<InvalidChangeBatch><Messages><Message>oops</Message></Messages></InvalidChangeBatch>`,
+		"success": `<ChangeResourceRecordSetsResponse><ChangeInfo><Id>mockID</Id></ChangeInfo></ChangeResourceRecordSetsResponse>`,
+	}
+
+	for name, body := range cases {
+		t.Run(name, func(t *testing.T) {
+			status := 200
+			if name == "error" {
```
wty-Bryant: nit: just from personal preference i would directly set each case as a type (in case we want to add more test data) containing status code as field

service/route53/internal/customizations/custom_error_deser_test.go
```diff
@@ -130,3 +131,56 @@ type fakeCredentials struct{}
...
+			}
+			tracked := &bodyCloseTracker{Reader: strings.NewReader(body)}
+			svc := route53.New(route53.Options{
+				Region:      "us-east-1",
+				Credentials: &fakeCredentials{},
+				Retryer:     aws.NopRetryer{},
+				HTTPClient:  &mockHTTPClient{&http.Response{StatusCode: status, Body: tracked}},
+			})
+
+			_, _ = svc.ChangeResourceRecordSets(context.Background(), &route53.ChangeResourceRecordSetsInput{
```
wty-Bryant: we may want to check error based on status code and body returned from mock client

service/route53/internal/customizations/custom_error_deser.go
Tongs2000: I think we can keep `originalBody` for readability. Also it keeps this symmetric with the same fix already merged for S3 in #3517.

service/route53/internal/customizations/custom_error_deser_test.go
Tongs2000: Done. Switched to a struct per case with an `explicit` status field.

service/route53/internal/customizations/custom_error_deser_test.go
Tongs2000: Done. The test now also asserts an error is returned on the error path and none on the success path.

wty-Bryant (approved):

PR #3520, opened 2026-08-17
Fix tmv1 uploadretry flaky test

wty-Bryant:
Same to previous transfer manager v2, tm v1 uses mock http server to run the upload retry test. Consequently concurrent 5mb request body upload causes flaky transport layer overload, which leads to CI failure. This pr fixes it in the same way of #3489 to replace unnecessary http server with just mock http client


Tongs2000 (approved):
LGTM, mirrors the merged #3489.

Issue #3503, opened 2026-07-30
Expanding S3 200-error handling to all applicable operations


Madrigal:
# Expanding S3 200-error handling to all applicable operations

## What's changing

The AWS SDK for Go v2 is expanding its S3 200-error detection middleware to cover all S3 operations that return structured XML responses. Previously, this middleware only applied to `CopyObject`, `UploadPartCopy`, and `CompleteMultipartUpload`.

## Why

S3 can occasionally return an error response with a 200 HTTP status code. Without this middleware, the SDK treats the response as successful and attempts to deserialize it normally — which can result in empty or partially populated output structs with no error returned to the caller.

With this change, the SDK will correctly detect these responses as errors and return them to the caller, allowing retry logic to handle them appropriately.

## What this means for you

- If your application was previously receiving successful responses with empty or unexpected data from S3 operations, you may now see errors returned instead. **This is the correct behavior** — the previous silent failure was a bug.
- The errors surfaced are transient (similar to 5xx) and are automatically retried by the SDK's default retry logic.
- No action is required unless you have code that specifically expects a successful response in cases where S3 is actually returning a 200-error.

## Rollout plan

This change is being rolled out in waves, starting with low-traffic operations and progressively enabling higher-traffic ones. Each wave soaks before the next is enabled so any regressions have time to surface.

## Affected operations

All S3 operations **except** those with streaming or non-XML response bodies:

- GetObject
- GetObjectAnnotation
- GetObjectTorrent
- SelectObjectContent
- GetBucketPolicy

These are excluded and unaffected by this change.

## Reporting issues

If you believe this change is causing unexpected behavior in your application, please [open an issue](https://github.com/aws/aws-sdk-go-v2/issues/new) with:

- The S3 operation affected
- The error message you're receiving
- Your SDK version

PR #3523, opened 2026-08-18
Expand S3 operations that check for an error inside an HTTP 200 response (wave 2/4)

Madrigal:
Related to #3503


Tongs2000 (commented):
LGTM. Minor: `WAVE_2` lists **18** operations but only **17** files were regenerated. It seems that `GetPublicAccessBlock` appears in both `WAVE_1` and `WAVE_2`. Harmless for correctness since the set dedupes, but worth removing the duplicate so the per-wave volume buckets stay accurate.

Tongs2000 (approved):
LGTM. Feel free to merge once the branch is updated and CI is green.

PR #3532, opened 2026-08-25
support connection read timeouts

lucix-aws:
framework to support connection read timeouts. we'll do a phased rollout internally but customers can opt in for now with the new env


lucix-aws:
you can filter to only the 1st commit if you want to avoid the codegen diff

aws/transport/http/timeouts_test.go
```diff
@@ -0,0 +1,367 @@
...
+
+			var timeout *ResponseTimeoutError
+			if !errors.As(err, &timeout) {
+				t.Fatalf("expect *ResponseTimeoutError, got %T: %v", err, err)
+			}
+		})
+	}
+}
+
+func TestReadTimeoutSlowButProgressingSucceeds(t *testing.T) {
```
Madrigal, line 210: love this test, reading the implementation I wasn't sure if this was covered but this test makes it clear it does

Madrigal (commented):
there are some merging conflicts, but besides that it LGTM. This would be enabled in a separate PR at a later time and this is just the mechanism to support it

internal/timeouts/timeouts.go
```diff
@@ -0,0 +1,46 @@
...
+const enableReadTimeoutEnvVar = "AWS_ENABLE_DEFAULT_READ_TIMEOUT_2026"
+
+var enableFromEnv = sync.OnceValue(func() bool {
+	return os.Getenv(enableReadTimeoutEnvVar) == "true"
+})
+
+// GetServiceReadTimeout reports the SDK's default read timeout for a service,
+// and whether one applies.
+func GetServiceReadTimeout(serviceID string) (time.Duration, bool) {
+	if enableReadTimeout2026 {
```
wty-Bryant: so this if-else controls internal/external rollout of a service's read timeout right?

internal/timeouts/timeouts.go
```diff
@@ -0,0 +1,46 @@
...
+		if len(readTimeout2026Rollout) > 0 && !readTimeout2026Rollout[serviceID] {
```
wty-Bryant: do we need the empty check of `readTimeout2026Rollout` here? my understanding is a nil rollout map means all services don't have a default read timeout

internal/timeouts/timeouts.go
lucix-aws: this is _internal-only_, if you use the external env it'll just turn on everything

internal/timeouts/timeouts.go
lucix-aws: we do not, this was poorly written and i've just fixed it

aws/transport/http/timeouts.go
```diff
@@ -0,0 +1,59 @@
...
+	if errors.Is(err, os.ErrDeadlineExceeded) {
+		return n, &ResponseTimeoutError{TimeoutDur: c.timeout}
+	}
+
+	return n, err
+}
+
+// installReadTimeout wraps the transport's dialer so every connection it opens
+// carries this client's read timeout.
+func (b *BuildableClient) installReadTimeout(tr *http.Transport) {
```
wty-Bryant, line 35: nit: prefer to put this func under client.go where `BuildableClient` is defined, but non-blocking

wty-Bryant (approved):

PR #3541, opened 2026-09-01
fix(s3): guard dnsCompatibleBucketName against empty bucket names

### LINKED ISSUE #3378: S3 model does not validate empty bucket names (leads to panic in SDK)
### Acknowledgements

### Describe the bug

The S3 service model does not validate empty bucket names, which allows invalid input (`""`) to pass through SDK validation layers.

This can lead to runtime issues in SDK logic that assumes a non-empty bucket name. For example, in the Go SDK, this results in a panic in `dnsCompatibleBucketName` due to direct string indexing (`bucket[0]`).

### Regression Issue

### Expected Behavior

- Empty string bucket is rejected early with clear errors  
- No reliance on downstream defensive checks
- Invalid input should result in a proper validation error instead of panic

### Current Behavior

Currently, the generated validation only checks whether the `Bucket` field is `nil`, but does not reject empty string values:

- `Bucket == nil` → validation error  
- `Bucket == ""` → passes validation (unexpected)

This allows invalid input to propagate into downstream logic. This results in a panic in `dnsCompatibleBucketName` due to direct string indexing (`bucket[0]`).
```golang
func dnsCompatibleBucketName(bucket string) bool {
	if strings.Contains(bucket, "..") {
		return false
	}

	// checks for `^[a-z0-9][a-z0-9\.\-]{1,61}[a-z0-9]$` domain mapping
	if !((bucket[0] > 96 && bucket[0] < 123) || (bucket[0] > 47 && bucket[0] < 58)) {
		return false
	}
```

### Reproduction Steps

Calling `PutObject` with an empty bucket name:

```go
_, err := client.PutObject(ctx, &s3.PutObjectInput{
    Bucket: aws.String(""),
    Key:    aws.String("test.txt"),
    Body:   strings.NewReader("data"),
})
```
---
### Possible Solution

I raised a [PR](https://github.com/aws/aws-sdk-go-v2/pull/3377) to address this issue in `dnsCompatibleBucketName` by safely handling empty input and preventing the panic. However, I guess this still needs to be addressed at the S3 service model level to ensure proper validation across all SDKs.

_No response_

```
github.com/aws/aws-sdk-go-v2 v1.41.5
github.com/aws/aws-sdk-go-v2/config v1.32.14
github.com/aws/aws-sdk-go-v2/service/s3 v1.99.0
```

### Compiler and Version used

go1.24.0 darwin/arm64

macOS Tahoe Version 26.3

warren830:
## Description

`dnsCompatibleBucketName` indexed `bucket[0]` without a length check, so an empty bucket name panicked instead of being reported as not DNS compatible:

```
panic: runtime error: index out of range [0] with length 0
  .../s3/internal/customizations.dnsCompatibleBucketName({0x0, 0x0})
      service/s3/internal/customizations/update_endpoint.go:289
```

`hostCompatibleBucketName` delegates to it, so endpoint-style resolution panicked for the same input rather than falling back to path-style addressing.

This returns `false` for an empty name, which is the correct answer for that predicate: an empty name is not DNS compatible, so the caller uses path-style and the request then fails through normal error handling instead of taking down the process.

## Motivation and Context

Fixes #3378

## Testing

Added `TestDNSCompatibleBucketName` and `TestHostCompatibleBucketName` to `update_endpoint_internal_test.go`. Both are offline table-driven tests and need no credentials.

Verified against `main` at `5a5ee7a73`, Go 1.27.0 on linux/arm64:

- The empty-name case panics before the change and passes after it (deterministic red -> green).
- `go test ./internal/customizations/` — pass.
- `go test ./...` in `service/s3` — 4 packages, 0 failures.
- `go vet ./internal/customizations/`, `gofmt`, and `git diff --check` — clean.

## Scope

Two files, +62 lines, no public API change, no generated code touched:

- `service/s3/internal/customizations/update_endpoint.go` (+4)
- `service/s3/internal/customizations/update_endpoint_internal_test.go` (+58, tests only)

Note on placement: the issue also raises validating empty `Bucket` in the S3 model's generated validation layer. I kept this PR to the panic itself so it is small and safe to review; happy to follow up on the model-level validation separately if you'd like it, since that lives in codegen.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/aws/aws-sdk-go-v2/blob/main/CONTRIBUTING.md) document
- [x] Local run of `go test` passes
- [x] Branch is based on the latest `main`

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


warren830:
Hi maintainers — this PR is now marked ready for review and the branch has been updated to current `main`, but no workflows have run yet: fork PRs from a first-time contributor need a one-time approval before CI can start.

Could someone approve the workflow run when convenient? The change is small and self-contained — a guard that rejects an empty bucket name before it is indexed, plus a test that reproduces the panic — and the relevant tests pass locally.

wty-Bryant (commented):
Thx for contributing the code, but based on previous [comment](https://github.com/aws/aws-sdk-go-v2/pull/3377#issuecomment-4237979111), due to the sensitivity of such deprecated legacy endpoint resolution which might still be used by many users in wild, we prefer to not update its behavior for now. Additionally, have u ever tried replacing legacy ep resolver with ep2.0's base endpoint pointing to your customized endpoint, as suggested in the old issue? That is the default way we config custom endpoint in endpoint resolver v2.

PR #3543, opened 2026-09-01
Remove span retry loop middleware

Tongs2000:
## Description

`spanRetryLoop` occupied a Finalize slot on every operation of every service for the sole purpose of opening a tracing span around the retry loop. `retry.Attempt` is the retry loop, so the span now starts at the top of its `HandleFinalize` and ends when it returns. The middleware and its registrar are gone from codegen.

This also drops a large per-request allocation: `spanRetryLoop` held a full copy of `Options` (784 bytes on sqs) to read a single `TracerProvider` that `invokeOperation` had already placed in the context, where `tracing.StartSpan` picks it up on its own.

## Behavior change

These changes only affect tracing behavior around the retry loop. Request bytes, retry decisions, and request execution are otherwise unchanged. With the default `NopTracerProvider` there is no observable difference.

**Presigned requests no longer emit a `RetryLoop` span.** Presign converters remove `retry.Attempt`, so the retry-loop span logic no longer runs on that path. Presigned requests never executed a retry loop, so this aligns the emitted tracing with actual behavior.

**`InterceptBeforeRetryLoop` now runs outside the `RetryLoop` span.** Previously it was only covered because `spanRetryLoop` happened to be registered first before `"Retry"`. That contradicted the hook's contract that `BeforeRetryLoopInterceptor` runs immediately before entering the retry loop. As a result, any time spent in that hook, and any tracing work that uses the current span from context there, is now associated with the operation span instead of `RetryLoop`.


codegen/smithy-aws-go-codegen/src/main/java/software/amazon/smithy/aws/go/codegen/AssembleMiddlewareStack.java
```diff
@@ -240,28 +225,6 @@ public void processFinalizedModel(GoSettings settings, Model model) {
...
-                return "spanRetryLoop";
-            }
-
-            public Map<String, Symbol> getFields() {
-                return Map.of("options", buildPackageSymbol("Options"));
-            }
-
-            public Writable getFuncBody() {
-                return goTemplate("""
-                        tracer := operationTracer(m.options.TracerProvider)
```
lucix-aws, line 255: This line went away in the migration to the runtime? what is `operationTracer` doing here?

lucix-aws: nvm i can't read

lucix-aws (approved):

PR #3546, opened 2026-09-03
Remove retry metrics header middleware

Tongs2000:
## Description

`retry.MetricsHeader` held a Finalize slot inside the retry loop on every operation of every service in order to write one header, `Amz-Sdk-Request`. It built that header entirely from the retry metadata that `Attempt` records on the context before each attempt, and `Attempt` already reads the same metadata, so `Attempt` now sets the header itself and the middleware is no longer registered.

The exported `MetricsHeader` type stays and its body is unchanged, marked `Deprecated` in the manner of `SignHTTPRequestMiddleware`. Registering it by hand still works.

Added `TestAttemptMiddleware_setsRetryMetricsHeader`. Nothing covered the header on this path: the existing `MetricsHeader` test exercises the middleware directly, the middleware-order snapshots record names rather than headers, and `Amz-Sdk-Request` is in the request snapshots' `skipHeaders` because its value varies per attempt. Removing the new call site leaves every other test in the repo green.

## Behavior change

No wire-format change for stacks built via `retry.AddRetryMiddlewares`. `Amz-Sdk-Request` is still written once per attempt with the same value before the signed request is sent.

`retry.AddRetryMiddlewares` also no longer registers `MetricsHeader` as a separate Finalize middleware, so clients using that helper drop one Finalize slot while continuing to send the same header because `Attempt` now writes it directly.

Custom stacks that register `Attempt` without also registering `MetricsHeader` will now emit `Amz-Sdk-Request`.


aws/retry/middleware_test.go
```diff
@@ -90,6 +90,38 @@ func TestMetricsHeaderMiddleware(t *testing.T) {
 	}
 }
 
+func TestAttemptMiddleware_setsRetryMetricsHeader(t *testing.T) {
```
lucix-aws: This is a good candidate for a kitchensink test I think (doesn't need to be in every protocol since it's not protocol-specific)

Basically call an operation that you rig to fail x times and make sure the header is still there each time. Bonus points if you test it against the old impl.

aws/retry/middleware_test.go
```diff
@@ -90,6 +90,38 @@ func TestMetricsHeaderMiddleware(t *testing.T) {
...
+			if len(sent) == 1 {
+				return out, metadata, mockRetryableError{b: true}
+			}
+			return out, metadata, nil
+		}))
+	if err != nil {
+		t.Fatalf("expect no error, got %v", err)
+	}
+
+	expect := []string{"attempt=1; max=3", "attempt=2; max=3"}
```
lucix-aws: test the ttl part as well?

aws/retry/middleware_test.go
Tongs2000: Added `internal/kitchensinktest/retry_metrics_header_test.go`: rigs the HTTP client to fail **twice** and asserts `Amz-Sdk-Request` on all three attempts. It also passes against the old implementation, where `MetricsHeader` is still on the stack.

aws/retry/middleware_test.go
Tongs2000: Covered in the `kitchensink` test. `ttl` needs a deadline and a positive clock skew, and the skew only exists once a response has been seen, so one run with two failures hits both branches: attempt 1 has a deadline but no skew yet and gets no `ttl`, while attempts 2 and 3 do. `TestMetricsHeaderMiddleware` in `aws/retry/middleware_test.go` already pins the formatting of the `ttl` value itself.

lucix-aws (approved):

Issue #3552, opened 2026-09-09
feature/s3/transfermanager v0.4.1+: GetObject Body.Read deadlocks permanently on large multipart objects read in small chunks


moitias:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

Hi!
  We recently updated transfermanager to v0.4.3 along with other changes and immediately after deploying noticed that all S3 ingestion stopped with no errors, panics or other signals explaining why.

A pprof profile of goroutines was taken and quickly identified the source as concurrentReader.Read():
```
goroutine 32108 [sync.WaitGroup.Wait, 64 minutes]:
sync.(*WaitGroup).Wait(...)
github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager.(*concurrentReader).Read(...)
        transfermanager@v0.4.3/concurrent_reader.go:109

goroutine 37193 [chan send, 64 minutes]:
github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager.(*concurrentReader).downloadChunk(...)
        transfermanager@v0.4.3/concurrent_reader.go:178
created by (*concurrentReader).Read in goroutine 32108
```
Namely, two goroutines launched by transfermanager (Read and downloadChunk) being blocked.

After investigating, the culprit was deemed to be how Read handles and waits for the download workers (see context section below for details of the investigation)

***Workaround***

Before the bug is fixed, forcing rangeRead by doing:
```
transfermanager.New(s3client, func(o *transfermanager.Options) {
  o.GetObjectType = transfermanagertypes.GetObjectRanges 
})
```
fixes the issue.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### Expected Behavior

- Read not hanging
- Canceling context cancels the read

### Current Behavior

- Read hangs forever
- Canceling the context does not cancel it

### Reproduction Steps

```
package main

import (
	"bufio"
	"bytes"
	"context"
	"errors"
	"io"
	"log"
	"os"
	"runtime/pprof"
	"sync/atomic"
	"time"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
[...]

PR #3554, opened 2026-09-09
fix GetObject deadlock

lucix-aws:
`concurrentReader.partRead` had a "quick exit" that returned as soon as the caller's read buffer (`p`) was filled from previously buffered parts **without** falling through to the receive loop that drains the channel, so fall through to that instead


lucix-aws:
closes #3552

Madrigal (approved):
Merge after CI pass

wty-Bryant (approved):

Issue #3526, opened 2026-08-21
s3/transfermanager: DownloadObject corrupts multipart objects with unequal part sizes


acj:
### Acknowledgements

- [x] I have searched (https://github.com/aws/aws-sdk/issues?q=is%3Aissue) for past instances of this issue
- [x] I have verified all of my SDK modules are up-to-date (you can perform a bulk update with `go get -u github.com/aws/aws-sdk-go-v2/...`)

### Describe the bug

If I do a multipart upload to S3 and use parts of varying sizes, then a `transfermanager.Client` with default options (`GetObjectType` in particular) can corrupt the object during download. `DownloadObject` reads the size of the first part and assumes that all other parts have the same size, which is not guaranteed.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### Expected Behavior

Calling `DownloadObject` on a vanilla `transfermanager.Client` returns exactly the object content that was uploaded, regardless of the object's part layout.

### Current Behavior

The downloaded objects are corrupted because parts are written to incorrect offsets, but `DownloadObject` does not return an error.

### Reproduction Steps

With the code below in `main.go`, run
```
go mod init transfermanager-repro
go get github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager@v0.3.15
go run .
```

Output:
```
$ go run .
err=<nil>
received=12582912
want=12582912
got=13631488
equal=false
```

```go
package main

import (
	"bytes"
	"context"
	"fmt"
	"io"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
	tmtypes "github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/types"
	"github.com/aws/aws-sdk-go-v2/service/s3"
)

const mib = 1 << 20

var parts = [][]byte{
	bytes.Repeat([]byte{'A'}, 6*mib),
	bytes.Repeat([]byte{'B'}, 5*mib),
	bytes.Repeat([]byte{'C'}, 1*mib),
}

type mockS3 struct{ transfermanager.S3APIClient }

func (*mockS3) GetObject(
	_ context.Context,
	in *s3.GetObjectInput,
	_ ...func(*s3.Options),
) (*s3.GetObjectOutput, error) {
	i := int(aws.ToInt32(in.PartNumber)) - 1
[...]

PR #3560, opened 2026-09-16
Download directory objects through DownloadObject instead of GetObject

RahulJasapara:
## Summary

`DownloadDirectory` downloaded each object with `GetObject` and `io.Copy`ed the returned stream into the destination file. This replaces that with a single `DownloadObject` call writing into the file through its `io.WriterAt`.

On a 64 GiB directory (1024 × 64 MiB objects, 8 MiB parts), the rewiring produces **+71% throughput, 49× lower peak memory, 3.8× better CPU efficiency.** Against the CRT S3 client on the same box in the same session, the CPU-per-byte gap closes from **5.1× to 2.0×** at **99× less memory**.

## What changed

5 files, +335/-37, all under `feature/s3/transfermanager`.

- `downloadSingleObject` creates the destination directory and file first, then calls `DownloadObject` with that file as the `io.WriterAt`. The `GetObject` call and the `io.Copy` into the file are gone.
- New unexported `mapDownloadObjectInput` copies every `GetObjectInput` field onto a `DownloadObjectInput`.
- The cleanup flag went from `fileCopyFail` to `transferFailed`, so the file is removed on any transfer failure instead of only a copy failure.
- Transferred bytes for the progress emitter come from `out.ContentLength`, which `DownloadObject` sets to the number of bytes it wrote, instead of `io.Copy`'s return value.
- Three unit tests added in `download_directory_test.go`, and one integration case in `api_op_DownloadDirectory_integ_test.go`. Two assertions tightened in `setup_integ_test.go`, described at the end. Their shared `describeBytesDiff` helper sits in `shared_test.go` so both the tagged and untagged test files can use it.

## Why it is faster

`GetObject` returns an `io.Reader`, and an ordered reader forces in-order part delivery. Its `concurrentReader` buffers completed parts in memory until the parts before them arrive, bounded by `GetObjectBufferSize` (50 MiB) per in-flight object. At 44 object workers that is the multi-gigabyte resident set, and head-of-line blocking is what capped throughput.

`DownloadObject` takes an `io.WriterAt` and writes each part at its computed offset. Parts never wait on each other and nothing is buffered.

The removed `HeadObject` round trip is a secondary win, not the main one.

## Benchmarks

`c7gn.16xlarge`, us-east-1, 1024 × 64 MiB objects (64 GiB), 8 MiB parts, 44 object workers, 8 part workers, `ranges` mode, both arms on `f0c4cca160b`.

Destination on tmpfs, a RAM-backed filesystem, isolating the code change from storage:

| | before | after | |
|---|---|---|---|
| Throughput | 46.849 Gbps | **80.125 Gbps** | +71% |
| Peak RSS | 4.60 GiB | **95.95 MiB** | 49× lower |
| Avg CPU | 48% | **16%** | |
| CPU-s/GiB | 7.42 | **1.97** | 3.8× better |
| NIC utilization | 23% | **40%** | |

Real disk (xfs on the EBS root volume) with the CRT S3 client as a reference point. All three cells ran back to back in one session, each wrapped in `/usr/bin/time -v` so CPU seconds and peak RSS come from a single instrument. 4 passes = 256 GiB per arm, destination `/home/ec2-user/dl-disk`:

| | before | after | `crt-c` (target 200 Gbps) |
|---|---|---|---|
| Throughput | 19.448 Gbps | 19.905 Gbps | 19.000 Gbps |
| Peak RSS | 4,696 MiB | **91.77 MiB** | 9,073 MiB |
| CPU-s / GiB | 6.540 | **2.584** | 1.271 |

- after vs before: +2.3% throughput, **51× less memory**, **2.5× better CPU**.
- after vs `crt-c`: **99× less memory**, 2.0× worse CPU, down from 5.1× worse.

aws-c-s3 has no directory operation, so the CRT cell is a flat 1024-task workload dispatched at once against a 44-object-worker pool here.

## Behavior changes

1. **The destination file is created before the first GET**, since `DownloadObject` writes through an `io.WriterAt`. An object that fails outright (403, 404, failed precondition) would leave a zero-length file where the old code left nothing, so cleanup was widened from copy failures to any transfer failure, including the failure-policy-ignored path where the overall call still succeeds. If a file already existed at that path it is truncated here and then removed by that cleanup, where the old code left it untouched until `GetObject` had succeeded.

2. **`Start` fires before `MkdirAll`** rather than after the first object's `GetObject` returned, since no intermediate call is left to hang it off. If every object fails, `Start` now fires where before it did not.

3. **A failure mid-transfer can leave holes rather than a valid prefix**, since `WriteAt` writes at computed offsets. Any returned error triggers cleanup, so this is only observable if the process is killed outright.

4. **Registered `ObjectProgressListeners` now fire during a directory download.** `GetObject` emits no progress events, so the object level was silent regardless of what the customer registered. This matches `UploadDirectory`, which passes no options to `UploadObject` (`api_op_UploadDirectory.go:407`). Directory-level events are unaffected.

## Concurrency

Unchanged. The per-call `Concurrency` still sizes the object worker pool, and part concurrency still comes from the client options, since the inner call is made on `d.c` exactly as `d.c.GetObject` was. No per-call functional options are forwarded to it, which the previous code did not do either, and which `UploadDirectory` also does not do. There is a comment at the call site (`api_op_DownloadDirectory.go:309-310`), because the absence is invisible: reusing `d.options.Concurrency` for parts would turn 44 object workers into 44² = 1,936 in-flight requests.

## Public API

No exported symbol is added, removed or changed shape.

`Callback.UpdateRequest(*GetObjectInput)` still receives a `GetObjectInput`, and `FailurePolicy.OnDownloadFailed` still receives the original input. `mapDownloadObjectInput` projects one input type onto the other. Every `GetObjectInput` field carries across.

A customer-supplied `Range` behaves identically: the write offset is `start + cur` where `start = pos - offset`, so ranged bytes land at file offset 0 as `io.Copy` placed them.

## Testing

Three unit tests were added to `download_directory_test.go`. All pass.

- **`TestDownloadDirectoryWriteOffsets`** downloads a 3-part object whose parts are deliberately unequal (1000, 700 and 300 bytes) and compares the resulting file byte for byte. Regression coverage for #3526.
- **`TestMapDownloadObjectInputIsTotal`** fills every field of `GetObjectInput` with a distinct value and requires each one to arrive on the `DownloadObjectInput`. A field added to both types later but forgotten in the mapper fails this test instead of silently dropping customer input.
- **`TestDownloadDirectoryNoHeadObject`** asserts the directory path issues no `HeadObject`, and separately asserts exactly 2 `GetObject` calls, one per single-part object, so the test cannot pass by downloading nothing at all.

To confirm those tests fail when the code is wrong, the source was broken on purpose once per test and then restored:

| Break introduced | Test that caught it | Failure reported |
|---|---|---|
| remove the #3536 offset correction (`api_op_DownloadObject.go:802-813`) | `TestDownloadDirectoryWriteOffsets` | expect 2000 bytes, got 2300 bytes |
| remove `VersionID` from `mapDownloadObjectInput` | `TestMapDownloadObjectInputIsTotal` | `VersionID is not carried across` |
| change `transferFailed = true` to `false` | existing file-cleanup test | stray file left behind |

Everything else that was run:

- `go test -tags integration -count=1 -run TestInteg_Download .` green against real S3 in 68.6s, covering `TestInteg_DownloadObject`, `TestInteg_DownloadObject_UnequalSize`, which exercises 6/5/1 MiB parts in both `parts` and `ranges` mode, and `TestInteg_DownloadDirectory`, which compares every downloaded file byte for byte.
- `go vet ./...` clean, with and without the `integration` tag.
- `go test -race -count=1 ./...` passes.

One integration case was added, `TestInteg_DownloadDirectory/unequal part sizes`. It uploads a 12 MiB object as real 6/5/1 MiB parts through `CreateMultipartUpload`, plus a 1 MiB single-part object under the same prefix, then compares every downloaded file byte for byte. The directory harness only did single `PutObject` uploads before, so it could not produce real part boundaries, which is what #3526 corrupted. That added a `PartSizes` field and a `putMultipartObject` helper to `setup_integ_test.go`, called only from `testDownloadDirectory` so both existing `WithChangingPartSize` functions are untouched.

## Also in this PR

`setup_integ_test.go` compared downloaded bytes against uploaded bytes with `bytes.EqualFold`. Case folding treats `0x41` and `0x61` as equal, so those assertions could pass on binary corruption.

Fixed in `testDownloadObject` and `testDownloadDirectory`, the two assertions guarding the paths this PR changes. `bytes.EqualFold` remains in `testPutObject`, `testGetObject` and `testUploadDirectory`, which cover upload and `GetObject` paths this PR does not touch. Happy to fix those here or in a follow-up.

Failure messages changed too. Dumping megabytes of binary with `%s` is not diagnostic, and printing only lengths misses the likeliest failure: a misplaced `WriteAt` produces a file of exactly the right length with bytes in the wrong place. `describeBytesDiff` reports the offset of the first difference when lengths match, and the lengths when they do not.


feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -281,23 +281,12 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
 	if d.in.Callback != nil {
 		d.in.Callback.UpdateRequest(input)
 	}
-	out, err := d.c.GetObject(ctx, input)
```
RahulJasapara, line 284: The failure-policy handling moved to the `DownloadObject` call below. The old code invoked `OnDownloadFailed` at two points, at the GET and at the copy. There is one transfer call now, so there is one place that handles it.

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -281,23 +281,12 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
...
-		err = d.failurePolicy.OnDownloadFailed(d.in, input, err)
-		if err != nil {
-			d.setErr(fmt.Errorf("error when heading info of object %s: %v", data.key, err))
-			return
-		}
-		d.objectsFailed.Add(1)
-		return
-	}
 
 	d.progressOnce.Do(func() {
```
RahulJasapara, line 285: This code is unchanged, but worth noting: `Start` now fires before `MkdirAll` instead of after a successful `GetObject`. That's behavior change 2 in the description

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -307,30 +296,60 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
...
 	}
 
 	d.objectsDownloaded.Add(1)
-	d.emitter.ObjectsTransferred(ctx, n)
+	// DownloadObject sets ContentLength to the number of bytes it wrote.
+	d.emitter.ObjectsTransferred(ctx, aws.ToInt64(out.ContentLength))
+}
+
+// mapDownloadObjectInput returns the DownloadObjectInput equivalent of in, writing to w.
+func mapDownloadObjectInput(in *GetObjectInput, w io.WriterAt) *DownloadObjectInput {
```
RahulJasapara, line 329: All 20 `GetObjectInput` fields are copied. `DownloadObjectInput` adds only `WriterAt`. `TestMapDownloadObjectInputIsTotal` fails if a field is later added to both types and missed here.

feature/s3/transfermanager/setup_integ_test.go
```diff
@@ -510,8 +510,8 @@ func testDownloadObject(t *testing.T, bucket string, testData downloadObjectTest
 		return
 	}
 
-	if e, a := testData.ExpectBody, w.Bytes(); !bytes.EqualFold(e, a) {
-		t.Errorf("expect %s, got %s", e, a)
+	if e, a := testData.ExpectBody, w.Bytes(); !bytes.Equal(e, a) {
```
RahulJasapara, line 513: This file had five `bytes.EqualFold` comparisons of downloaded bytes against uploaded bytes. Case folding treats `0x41` and `0x61` as equal, so those assertions could pass on corrupted binary data. I changed the two that guard the paths this PR touches, here and in `testDownloadDirectory`. The three in `testPutObject`, `testGetObject` and `testUploadDirectory` cover paths this PR does not change, so I left them unchanged.

feature/s3/transfermanager/setup_integ_test.go
```diff
@@ -782,10 +782,26 @@ func testDownloadDirectory(t *testing.T, bucket string, testData downloadDirecto
...
+		if e, a := expectData, b; !bytes.Equal(e, a) {
+			t.Errorf("downloaded file %s does not match the uploaded object: %s", f, describeBytesDiff(e, a))
+		}
+	}
+}
+
+// describeBytesDiff summarizes how two byte slices differ, for assertions on
+// payloads too large to print. Misplaced writes preserve length, so the offset
+// of the first difference is reported as well as the lengths.
+func describeBytesDiff(expect, actual []byte) string {
```
RahulJasapara: The old message dumped megabytes of binary with %s. Printing lengths instead would miss the failure this change risks most, since a misplaced WriteAt gives a file of exactly the right length with bytes in the wrong place. This reports the offset of the first difference, and only runs on an already-failing assertion.

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,192 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+	}
+	if e, a := int64(2), out.ObjectsDownloaded; e != a {
+		t.Fatalf("expect %d objects downloaded, got %d", e, a)
+	}
+
+	if n := len(s3Client.HeadObjectInputs); n != 0 {
+		t.Errorf("expect no HeadObject calls on the directory path, got %d", n)
+	}
+	// An empty HeadObjectInputs is also consistent with no download happening.
+	if s3Client.GetObjectInvocations == 0 {
```
wty-Bryant: we should still check the invocation count matches the number of GetObject calls (2 in this case)

RahulJasapara: Good catch, done in commit 7dce264. Asserts exactly 2 now, one GET per single-part object.

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,192 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+		}
+		src.Field(i).Set(v)
+	}
+
+	var w nopWriterAt
+	got := reflect.ValueOf(mapDownloadObjectInput(in, w)).Elem()
+
+	for i := 0; i < src.NumField(); i++ {
+		f := src.Type().Field(i)
+		if !f.IsExported() {
```
wty-Bryant: nit: why we care about public/private field here? All GetObject/DownloadObject input fields are public. But this is non-blocker

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,192 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("expect to read downloaded file, got %v", err)
+	}
+	if e, a := len(want), len(got); e != a {
+		t.Fatalf("expect downloaded file to be %d bytes, got %d", e, a)
+	}
+	if !bytes.Equal(want, got) {
+		for i := range want {
+			if want[i] != got[i] {
+				t.Fatalf("downloaded file differs at offset %d: expect %q, got %q", i, want[i], got[i])
```
wty-Bryant: i like this first diff index error fmt

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,192 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+	}
+	if e, a := int64(1), out.ObjectsDownloaded; e != a {
+		t.Fatalf("expect %d objects downloaded, got %d", e, a)
+	}
+
+	got, err := os.ReadFile(filepath.Join(dstPath, "multipart", "object"))
+	if err != nil {
+		t.Fatalf("expect to read downloaded file, got %v", err)
+	}
+	if e, a := len(want), len(got); e != a {
```
wty-Bryant: nit: seems like `describeBytesDiff` can be used here as well?

wty-Bryant (commented):
Since GetObject callback interface is a public var, we can not simply replace GetObjectInput in directory downloader with DownloadObject (that is an initial design gap ignoring performance). So the Get to Download request conversion is correct here. Waiting for other reviewer's feedback

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: Removed

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: Done. `describeBytesDiff` was in `setup_integ_test.go`, which is `//go:build integration`, so this untagged file couldn't call it. Moved it to `shared_test.go` so both use it. Also switched it to `%q` to keep the format you liked. Adds a 4th file to the PR, but I can revert if you'd rather not.

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,179 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+			return p, nil
+		}
+		switch t.Elem().Kind() {
+		case reflect.String:
+			p := reflect.New(t.Elem())
+			p.Elem().Set(reflect.ValueOf(fmt.Sprintf("value-%d", seed)).Convert(t.Elem()))
+			return p, nil
+		case reflect.Bool:
+			p := reflect.New(t.Elem())
+			p.Elem().SetBool(true)
```
Madrigal: We're always setting `bool` to true regardless of seed

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -307,30 +296,60 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
...
 			return
 		}
 		d.objectsFailed.Add(1)
 		return
 	}
 
 	d.objectsDownloaded.Add(1)
-	d.emitter.ObjectsTransferred(ctx, n)
+	// DownloadObject sets ContentLength to the number of bytes it wrote.
+	d.emitter.ObjectsTransferred(ctx, aws.ToInt64(out.ContentLength))
```
Madrigal, line 325: Is `ObjectsTransferred` not supposed to be the number of objects transferred rather than bytes? since we also have `bytesTransferred`?

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,179 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+
+	s3Client, _ := s3testing.NewDownloadDirectoryClient()
+	s3Client.ListObjectsData = [][]s3types.Object{{
+		{Key: aws.String("multipart/object")},
+	}}
+	s3Client.GetObjectFn = s3testing.UnequalPartGetObjectFn
+	s3Client.PartsData = [][]byte{partA, partB, partC}
+	s3Client.PartsCount = 3
+
+	dstPath := filepath.Join("testdata", "write-offsets")
```
Madrigal: `testdata` is where we store our test suite, this seems more like where you are writing output data to. I'd rather use `os.CreateTemp(t.TempDir(), "whatever-destination-name")`

feature/s3/transfermanager/download_directory_test.go
```diff
@@ -874,3 +874,179 @@ func TestDownloadDirectoryClosesCreatedFiles(t *testing.T) {
...
+	s3Client.GetObjectFn = s3testing.PartGetObjectFn
+	s3Client.Data = []byte("hello world")
+	s3Client.PartsCount = 1
+
+	dstPath := filepath.Join("testdata", "no-head-object")
+	defer os.RemoveAll(dstPath)
+
+	out, err := New(s3Client).DownloadDirectory(context.Background(), &DownloadDirectoryInput{
+		Bucket:      aws.String("mock-bucket"),
+		Destination: aws.String(dstPath),
```
Madrigal, line 894: you may be able to use `os.DevNull` since you don't really care about what's written here

feature/s3/transfermanager/api_op_DownloadDirectory.go
RahulJasapara: The parameter is bytes by design. `progress_listener.go:343` takes a byte delta, adds it to `bytesTransferred`, and increments`objectsTransferred` by 1 itself, then emits an event carrying both. `api_op_UploadDirectory:423` makes the identical call, and the code I replaced passed `io.Copy`'s byte count to the same method, so this PR doesn't change the contract, only where the byte count comes from. i agree that the name is misleading. Renaming touches `UploadDirectory` and the shared emitter, so I'd rather follow up separately.

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: I just switched to `t.TempDir()` in 1d3f6ab per your other comment, which gets the output out of the repo anyway.

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: Fixed in 1d3f6ab with `t.TempDir()`, which also drops the `defer os.RemoveAll`. Used it directly rather than `os.CreateTemp` since that returns a file and `Destination` needs a directory. 

btw four older tests in this file do the same thing (488, 651, 674, 844). should I convert them here or follow up later?

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: yeah I deleted rather than fixed in 1d3f6ab. neither input type has a bool field and a bool can't be seed-distinct anyway since `false` is the zero value. so only `true` proves it was carried. A bool added later now hits the existing `t.Fatalf` instead of silently passing.

feature/s3/transfermanager/api_op_DownloadDirectory.go
```diff
@@ -281,23 +281,12 @@ func (d *directoryDownloader) downloadSingleObject(ctx context.Context, data obj
...
-		return
-	}
 
 	d.progressOnce.Do(func() {
 		d.emitter.Start(ctx, d.in)
 	})
 
-	err = os.MkdirAll(filepath.Dir(data.path), 0755)
-	if err != nil {
+	if err := os.MkdirAll(filepath.Dir(data.path), 0755); err != nil {
```
lucix-aws, line 289: So your PR description mentioned this, but what happens if there was an existing file here already? cc @wty-Bryant

feature/s3/transfermanager/download_directory_test.go
lucix-aws: We generally don't want to touch the filesystem in unit tests unless we absolutely need to.

feature/s3/transfermanager/api_op_DownloadDirectory.go
wty-Bryant: the `MkdirAll` does nothing and returns nil if the directory already exists  https://pkg.go.dev/os#MkdirAll , if u mean the destination file already exists, the result depends on the `createFileFn` below (default to `os.Create` truncating old one)

lucix-aws: > default to os.Create truncating old one

yeah, do we care about this basically, that we're now premptively deleting whatever file was there before attempting to get it

wty-Bryant: normally we don't, except the file system is case-insensitive (foo remove Foo), i will look at it

feature/s3/transfermanager/download_directory_test.go
RahulJasapara: Neither of these two tests actually needs the filesystem but `createFileFn` returns a concrete `*os.File` (`api_op_DownloadDirectory.go:25`), so there's no way to swap in `types.WriteAtBuffer`, and `os.MkdirAll` (:289) and `os.Remove` (:305) have no seam either. Adding that seam also means rewriting `TestDownloadDirectoryClosesCreatedFiles`. Do you want me to do that in this PR or file it as a follow-up?

lucix-aws: if we've already got tests in this using the filesystem then dw about it for now

PR #3563, opened 2026-09-21
feat: Add aws-chunked response decoding reader

Madrigal:
Add an internal service/internal/awschunked package providing an
io.ReadCloser that decodes an inbound `aws-chunked` encoded stream, yielding
only the decoded payload bytes to the caller.

Trailer fields carried after the terminating zero-length chunk are
captured and exposed once the stream reaches EOF, via Trailers() and a
fire-once OnTrailers callback, so a later stage can consume them.

The reader parses but ignores chunk extensions for forward compatibility,
and optionally verifies the total decoded length against a caller-supplied
value to guard against truncation.


service/internal/awschunked/go.sum
lucix-aws: does this need to be a module? i'd like to get away from random internal/ modules because they're not really _necessary_ (there is not a ton of code here, or any random dependency we want to isolate) and it's also taxing public infrastructure with modules that nobody can actually use but us

lucix-aws (approved):

service/internal/awschunked/go.sum
Madrigal: fair call, I removed the module components

PR #3564, opened 2026-09-23
Checksum non-200 HTTP responses

Madrigal:
Relax the constraint we added on #3075


service/internal/checksum/middleware_validate_output.go
```diff
@@ -100,10 +100,13 @@ func (m *validateOutputPayloadChecksum) HandleDeserialize(
 
 	// Skip validation if no checksum algorithm or checksum is available.
 	if len(expectedChecksum) == 0 || len(algorithmToUse) == 0 {
-		if response.Body != http.NoBody && m.LogValidationSkipped {
+		// Only log for successful responses. Error responses (4xx/5xx) carry an
+		// error document rather than an object payload and legitimately have no
+		// checksum, so logging there is just noise.
+		if response.StatusCode < 400 && response.Body != http.NoBody && m.LogValidationSkipped {
```
lucix-aws: should this be <300 instead? idk

Madrigal: I put it mostly so we don't log on 404 (the original complaint) and 5xx. 3xx are redirects which I suspect we'll rarely see on this middleware

lucix-aws (approved):

